PR #66, opened 2022-04-22
improvement: add name to enum

dsinghvi:


packages/api/src/fern.yml
```diff
@@ -101,6 +101,9 @@ types:
   EnumValue:
     extends: WithDocs
     fields:
+      name:
+        type: string
+        docs: Name is an alphanumeric way to reference this enum. Useful for code generation.
```
zachkirsch, line 106: nbd it's more than alphanumeric (has to start with a letter). This will be cool once we have validation we can add that right here: `regex: /[a-z][a-z0-9]*/i`

zachkirsch (approved):

packages/api/src/fern.yml
dsinghvi: good point and underscores are allowed too

PR #120, opened 2022-05-02
fix: add auth block to IR

dsinghvi:


packages/compiler/ir-generation/src/converters/services/convertHttpService.ts
```diff
@@ -91,3 +92,12 @@ function convertHttpMethod(method: RawSchemas.HttpEndpointSchema["method"]): Htt
             return HttpMethod.Delete;
     }
 }
+
+function convertHttpAuth(auth: "bearer" | "basic"): HttpAuth {
+    switch (auth) {
+        case "basic":
+            return HttpAuth.Basic;
+        case "bearer":
+            return HttpAuth.Bearer;
```
zachkirsch, line 101: Good opportunity for 
```ts
default:
  assertNever(auth);
```

packages/compiler/ir-generation/src/converters/services/convertHttpService.ts
```diff
@@ -20,6 +20,7 @@ export function convertHttpService({
 
     return {
         docs: serviceDefinition.docs,
+        auth: serviceDefinition.auth !== undefined ? convertHttpAuth(serviceDefinition.auth) : null,
```
zachkirsch, line 23: `!= null` not `!== undefined`

PR #125, opened 2022-05-05
improvement: utility to invoke plugin docker

dsinghvi:
**PR adds**: 
- `code-generation` package in the `compiler` folder. Currently this package only has a `pluginInvoker.ts` file but I could see us adding more things like plugin error handling, plugin logs, & invoking other parts of the plugin framework here. 

**Questionable things**: 
- I created a test that downloads the [java plugin](https://hub.docker.com/repository/docker/fernapi/fern-java-client) and invokes it. The test passes in a hardocded `ir.json` and `config.json`. Yet another place we are hardcoding IR, so open to changing if you have a better idea.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,11 +7,6 @@ jobs:
       image: ubuntu-2004:current
     steps:
       - checkout
-      # - run:
-      #     name: Install Node
-      #     command: |
-      #       sudo apt update
-      #       sudo apt install nodejs npm yarn
       - run:
           name: Run code-generation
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -43,5 +43,8 @@
     "react": "^18.0.0",
     "stylelint": "^14.6.1"
   },
-  "packageManager": "yarn@3.2.0"
+  "packageManager": "yarn@3.2.0",
+  "dependencies": {
```
zachkirsch, line 47: can you delete this?

PR #145, opened 2022-05-07
Revert "Revert "Delete fern.sh (#137)""

zachkirsch:
This reverts commit 90a6beaa2143871767a3c7780a6063b874a22260.

### DIFF
--- a/packages/compiler/plugins/plugin-runner/src/runPlugin.ts
+++ b/packages/compiler/plugins/plugin-runner/src/runPlugin.ts
@@ -26,28 +26,28 @@ export async function runPlugin({
 }: runPlugin.Args): Promise<void> {
     const config: PluginConfig = {
         irFilepath: DOCKER_PATH_TO_IR,
-        output:
-            pluginInvocation.absolutePathToOutput != null
-                ? {
-                      path: DOCKER_CODEGEN_OUTPUT_DIRECTORY,
-                      pathRelativeToRootOnHost:
-                          absolutePathToProjectConfig != null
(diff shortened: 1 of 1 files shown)


packages/compiler/plugins/plugin-runner/src/runPlugin.ts
```diff
@@ -9,44 +11,39 @@ const DOCKER_PATH_TO_IR = path.join(DOCKER_FERN_DIRECTORY, "ir.json");
...
-    pluginConfig,
     pathToWriteConfigJson,
-    pluginOutputDirectory,
-    outputPathRelativeToRootOnHost,
+    relativeWorkspacePath,
 }: runPlugin.Args): Promise<void> {
-    const configJson = {
-        outputPathRelativeToRootOnHost,
+    const config: PluginConfig = {
+        relativeWorkspacePathOnHost: relativeWorkspacePath,
```
dsinghvi, line 28: we need this to be `outputPathRelativeToRootOnHost`

zachkirsch: Oops thanks

zachkirsch: wait those are different things

zachkirsch: oh i'm tracking I think

dsinghvi: yup they are different things -- was incorrect on what i needed at first

PR #153, opened 2022-05-10
improvement: fern init

dsinghvi:
`fern init` is run at the root of users directory and generates a `fern.config.json`, `fern-api` directory, with a `sample api` and a sample `.fernrc.yml`


dsinghvi:
@zachkirsch things we need to figure out are: 
1) Should we have a form that asks users what languages they want to generate in initially? 
2) What should we name folder & api in .fernrc.yml (considering that the sample apis are food delivery apis)?

packages/init/src/init.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export async function initialize(): Promise<void> {
+    await writeProjectConfigIfNotExists();
+    await writeApiSubDirectoryIfNotExists();
+}
+
+const API_DIRECTORY = "fern-api";
+const SRC_DIRECTORY = "src";
+const API_SRC_DIRECTORY = `${API_DIRECTORY}/${SRC_DIRECTORY}`;
+const API_WORKSPACE_DEFINITION_FILE = `${API_DIRECTORY}/src/${WORKSPACE_DEFINITION_FILENAME}`;
```
zachkirsch, line 19: can also use path.join

packages/init/src/init.ts
```diff
@@ -0,0 +1,56 @@
...
+        await writeSampleApiToDirectory(API_WORKSPACE_DEFINITION_FILE);
+    }
+}
+
+async function writeWorkspaceDefinitionFile(dir: string): Promise<void> {
+    const workspaceDefinition: WorkspaceDefinitionSchema = {
+        name: "Food Delivery API",
+        input: SRC_DIRECTORY,
+        plugins: [],
+    };
```
zachkirsch, line 35: you can pull this out if you want into a global like you did with DEFAULT_PROJECT_CONFIG

packages/init/src/sampleApi.ts
```diff
@@ -0,0 +1,120 @@
+import { RawSchemas } from "@fern-api/syntax-analysis";
+import { writeFile } from "fs/promises";
+import yaml from "js-yaml";
+
+export async function writeSampleApiToDirectory(dir: string): Promise<void> {
+    await writeFile(`${dir}/${ORDER_API_FILENAME}`, yaml.dump(ORDER_API));
+    await writeFile(`${dir}/${MENU_API_FILENAME}`, yaml.dump(MENU_API));
```
zachkirsch, line 7: path.join

packages/init/src/init.ts
zachkirsch: it's better cross platform, e.g. on windows when the separator is `\`

packages/init/src/init.ts
```diff
@@ -0,0 +1,56 @@
...
+    await writeFile(dir, yaml.dump(workspaceDefinition));
+}
+
+const DEFAULT_PROJECT_CONFIG: ProjectConfigSchema = {
+    workspaces: ["**"],
+};
+
+async function writeProjectConfigIfNotExists(): Promise<void> {
+    if (!doesPathExist(PROJECT_CONFIG_FILENAME)) {
+        await writeFile(PROJECT_CONFIG_FILENAME, JSON.stringify(DEFAULT_PROJECT_CONFIG));
```
zachkirsch, line 45: `JSON.stringify(DEFAULT_PROJECT_CONFIG, undefined, 4)` will pretty-print it

packages/init/src/init.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+export async function doesPathExist(filepath: string): Promise<boolean> {
```
zachkirsch, line 49: does this need to be exported?

packages/init/src/sampleApi.ts
```diff
@@ -0,0 +1,120 @@
...
+                        parameters: {
+                            restaurantId: "RestaurantId",
+                        },
+                        response: "list<MenuItem>",
+                    },
+                },
+            },
+        },
+    },
+};
```
zachkirsch, line 120: not super into to `init` dumping a big yaml, let's chat when you come in

packages/init/src/init.ts
dsinghvi: thanks

packages/init/src/init.ts
dsinghvi: nope

PR #161, opened 2022-05-11
improvement: pipe through workspace name in ir

dsinghvi:


packages/cli/src/__test__/generate/__snapshots__/generate.test.ts.snap
```diff
@@ -28,17 +28,13 @@ export class Client implements Client {
...
             method: \\"DELETE\\",
             headers: {},
             token: this.token,
-            body: {
-                content: JSON.stringify(request),
-                contentType: \\"application/json\\",
-            },
+            body: request,
         });
-        const response = JSON.parse(encodedResponse.toString());
```
dsinghvi, line 41: @zachkirsch was this just not generated in a previous test

PR #220, opened 2022-06-15
improvement: endpoint errors must correspond to defined errors

dsinghvi:


packages/compiler/plugins/docker-utils/src/__test__/resources/basic-writer/host-output/my-file.txt
```diff
@@ -0,0 +1 @@
+Test!
```
zachkirsch, line 1: prob don't wanna check this in

PR #244, opened 2022-06-20
improvement: define generator publish config 

dsinghvi:


packages/fern-typescript/cli/src/generator/generator-config/schemas/GeneratorPublishConfigSchema.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from "zod";
+
+// IMPORTANT: don't use strictObject so we're forward compat if new keys are added
+
+export const GeneratorPublishConfigSchema = z.object({
+    version: z.string(),
```
dsinghvi, line 6: @zachkirsch how do we generate versions for `fern generate` again?

PR #277, opened 2022-06-25
fix: npm publish logs + works

dsinghvi:


packages/fern-typescript/cli/src/commands/runCommand.ts
```diff
@@ -37,20 +37,31 @@ export async function runCommand({
...
     }
 
     if (config.publish != null) {
+        const { registryUrl, token } = config.publish.registries.npm;
         await runNpmCommandInOutputDirectory(
             "config",
             "set",
-            `${scopeWithAtSign}:${config.publish.registries.npm.registryUrl}`,
+            `${scopeWithAtSign}:registry`,
             config.publish.registries.npm.registryUrl
```
zachkirsch, line 54: registryUrl

PR #305, opened 2022-06-29
improvement: org name in project config

dsinghvi:
Adding a mandatory `org` key to `fern.config.json`. The org should get propagated to any nested `.fernrc.yml` files. This means that `fern.config.json` is required, but maybe we wind that back.


dsinghvi:
One other thing to think about is how to handle open source repos (i.e. someone makes an open source contribution to fern which changes the ir, how do they run `fern generate`) That's why i think specifying orgName in repo and maybe configuring access to the org can be done in a web app.

dsinghvi:
Screen Shot 2022-06-28 at 9 31 28 PM

packages/ete-tests/src/__test__/generate/fixtures/fern.config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "workspaces": ["**"],
+  "org": "fern"
```
dannysheridan, line 3: cool that this leaves room for future configurations:

- authorization
- billing
- analytics

Issue #379, opened 2022-07-14
Empty extends causes fern check to fail


dsinghvi:
```yaml
types: 
  Base: 
    properties: 
      dummy: string
  Derived: 
    extends: Base
```
fails. Instead the yaml needs to be: 
```yaml
types: 
  Base: 
    properties: 
      dummy: string
  Derived: 
    extends: Base
    properties: {}
```

dsinghvi:
Can we default the former in the yaml to assume properties are `{}` ? @zachkirsch wdyt

zachkirsch:
Fixed in `0.0.148`

dsinghvi:
The CLI throws an error 
```
fern check [workspaces...]

Validates your Fern API Definition

Positionals:
  workspaces  If omitted, every workspace specified in the project-level configu
              ration (fern.config.json) will be processed. [array] [default: []]

Options:
      --help     Show help                                             [boolean]
  -v, --version  Show version number                                   [boolean]

Error: Unexpected value: {"extends":"BaseValidation"}
    at assertNever (/usr/local/lib/node_modules/fern-api/bundle.cjs:334096:9)
    at visitRawTypeDeclaration (/usr/local/lib/node_modules/fern-api/bundle.cjs:339887:3)
    at visitTypeDeclaration (/usr/local/lib/node_modules/fern-api/bundle.cjs:339916:9)
    at visitTypeDeclarations (/usr/local/lib/node_modules/fern-api/bundle.cjs:339912:11)
    at async Object.types (/usr/local/lib/node_modules/fern-api/bundle.cjs:340107:7)
    at async visitObject (/usr/local/lib/node_modules/fern-api/bundle.cjs:339691:5)
    at async visitFernYamlAst (/usr/local/lib/node_modules/fern-api/bundle.cjs:340099:3)
    at async getTypesByFilepath (/usr/local/lib/node_modules/fern-api/bundle.cjs:346790:5)
    at async Object.create (/usr/local/lib/node_modules/fern-api/bundle.cjs:346756:29)
    at async Promise.all (index 0)``` instead now

dsinghvi:
Closing because this hasn't been requested by the community.

PR #383, opened 2022-07-14
Allow no properties on object (if extends is present)

zachkirsch:
Fixes https://github.com/fern-api/fern/issues/379


dsinghvi:
yoooo thank you

PR #388, opened 2022-07-15
improvement: validate path parameters are defined

dsinghvi:
Fixes https://github.com/fern-api/fern/issues/386

### DIFF
--- a/packages/ete-tests/src/__test__/ir/fixtures/simple/api/src/api.yml
+++ b/packages/ete-tests/src/__test__/ir/fixtures/simple/api/src/api.yml
@@ -29,7 +29,7 @@ services:
           method: GET
           path: /{movieId}
           path-parameters:
-            postId: MovieId
+            movieId: MovieId
           response: Movie
           errors:
             - NotFoundError
--- a/packages/ete-tests/src/__test__/ir/__snapshots__/ir.test.ts.snap
+++ b/packages/ete-tests/src/__test__/ir/__snapshots__/ir.test.ts.snap
@@ -153,7 +153,7 @@ exports[`fern-ir-tests simple 1`] = `
                         },
                         \\"pathParameters\\": [
                             {
-                                \\"key\\": \\"postId\\",
+                                \\"key\\": \\"movieId\\",
                                 \\"valueType\\": {
                                     \\"name\\": \\"MovieId\\",
                                     \\"fernFilepath\\": [],


packages/ete-tests/src/__test__/ir/fixtures/simple/api/src/api.yml
```diff
@@ -29,7 +29,7 @@ services:
           method: GET
           path: /{movieId}
           path-parameters:
-            postId: MovieId
+            movieId: MovieId
```
zachkirsch, line 32: lmao nice

packages/validator/src/rules/no-undefined-path-parameters/__test__/no-undefined-path-parameters.test.ts
```diff
@@ -0,0 +1,14 @@
...
+import { NoUndefinedPathParametersRule } from "../no-undefined-path-parameters";
+
+describe("no-undefined-path-parameters", () => {
+    it("simple", async () => {
+        const violations = await getViolationsForRule({
+            rule: NoUndefinedPathParametersRule,
+            absolutePathToDefinition: path.join(__dirname, "fixtures", "simple"),
+        });
+
+        expect(violations).toMatchSnapshot();
```
zachkirsch, line 12: the function is `toMatchInlineSnapshot()` and then run the test, but that wasn't that helpful actually. You can just do `expect(violations).toEqual([])`, and run the test. You'll get an error that the test failed and it'll print the actual value, and then you can just copy and paste into the test:
```
// good to type the `actual` so we get compile breaks if we change what a rule violation looks like
const actual: RuleViolation[] = [ ... ]

expect(violations).toEqual(actual);
```

packages/validator/src/rules/no-undefined-path-parameters/no-undefined-path-parameters.ts
```diff
@@ -0,0 +1,56 @@
+import { constructHttpPath } from "@fern-api/ir-generator";
+import chalk from "chalk";
+import { Rule, RuleViolation } from "../../Rule";
+
+export const NoUndefinedPathParametersRule: Rule = {
+    name: "no-duplicate-enum-values",
```
zachkirsch, line 6: wrong name - we can get rid of this field tbh. actually the benefit of keeping it is if the rule throws, we can catch and be like `rule <xyz> threw an error: <error>`

packages/validator/src/rules/no-undefined-path-parameters/no-undefined-path-parameters.ts
```diff
@@ -0,0 +1,56 @@
...
+    create: () => {
+        return {
+            httpEndpoint: ({ endpoint }) => {
+                const errors: RuleViolation[] = [];
+
+                const urlPathParameters = new Set();
```
zachkirsch, line 12: this is probably getting typed as `Set<any>`. This should be
```ts
const urlPathParameters = new Set<PathParameter>();
```

packages/validator/src/rules/no-undefined-path-parameters/no-undefined-path-parameters.ts
```diff
@@ -0,0 +1,56 @@
...
+                                message: `Endpoint path has duplicate path parameter: ${chalk.bold(
+                                    part.pathParameter
+                                )}`,
+                            });
+                        }
+                        urlPathParameters.add(part.pathParameter);
+                    });
+                }
+
+                const definedPathParameters = new Set(Object.keys(endpoint["path-parameters"] ?? {}));
```
zachkirsch, line 28: nbd, but the only thing about using sets everywhere is that you lose the order of how things appear in the yaml.

packages/validator/src/rules/no-undefined-path-parameters/__test__/no-undefined-path-parameters.test.ts
dsinghvi: Ah gotcha -- i tried that it was just annoying to get bolded and colored values copy + pasted

packages/validator/src/rules/no-undefined-path-parameters/no-undefined-path-parameters.ts
dsinghvi: gotcha so within the same endpoint things can get jumbled?

PR #389, opened 2022-07-15
improvement: add basic auth

dsinghvi:


packages/fern-typescript/client/src/__test__/fixtures/auth/src/auth.yml
```diff
@@ -0,0 +1,46 @@
+types:
+  SomeType: string
+services:
```
dsinghvi, line 3: this will be useful when i write the yaml of all the possible endpoints/schemas

packages/fern-typescript/commons/src/dependencies/getReferenceToFernServiceUtils.ts
```diff
@@ -91,3 +91,21 @@ function addFernServiceUtilsImport({
     });
     dependencyManager.addDependency(PACKAGE_NAME, VERSION);
 }
+
+export function invokeMaybeGetter(maybeGetter: ts.Expression): ts.Expression {
```
dsinghvi, line 95: lol there's literally no way someone knows what this does -- my workflow was modify, run ete tests, check updated snapshots

packages/fern-typescript/client/src/http/utils.ts
```diff
@@ -4,30 +4,35 @@ export function doesServiceHaveHeaders(service: HttpService): boolean {
...
 
-export function doesServiceHaveAuth(
+export function doesServiceHaveBearerAuth(
     service: HttpService
-): { hasAuth: false } | { hasAuth: true; isOptional: boolean; authType: HttpAuth.Bearer | HttpAuth.Basic } {
+): { hasAuth: false } | { hasAuth: true; isOptional: boolean } {
+    return doesServiceHaveAuth(service, HttpAuth.Bearer);
+}
+
+export function doesServiceHaveBasicAuth(
```
dsinghvi, line 13: having the 2 functions `doesServiceHaveBasicAuth` and `doesServiceHaveBearerAuth` is def cleaner cause it gets rid of the weird error case -- didn't do it originally cause then it would mean another iteration over the endpoints

PR #394, opened 2022-07-17
improvement: validation rule to check if imports actually exist

dsinghvi:
trying to implement a validation rule to check if imported files actually exist, had to do a couple of interesting things: (1) pass down absoluteDefinitionPath all the way down as RuleRunnerArgs , (2) once i had the absolute path, i could append the imported file path and then check if the file exists but this introduces async into the rules which messes with a lot of the types. curious how you’d approach this instead

Fixes https://github.com/fern-api/fern/issues/387

### DIFF
--- a/packages/ete-tests/src/__test__/validate/__snapshots__/validate.test.ts.snap
+++ b/packages/ete-tests/src/__test__/validate/__snapshots__/validate.test.ts.snap
@@ -1,3 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`fern-ir-tests simple 1`] = `""`;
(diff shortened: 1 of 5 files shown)


zachkirsch:
Pushed up! Mostly just added a bunch of awaits everywhere now that the rule is async. Also put the `absolutePath` inside the workspace rather than passing it to the rule.

Terminology is getting kinda confusing (e.g. what's even is a a `Workspace`?). Put some thoughts here: https://www.notion.so/usebirch/Fern-Workspace-Terminology-dae519a16ce1421fac15f377c0c82550

dsinghvi:
ahh i didn't catch the workspace concept

PR #396, opened 2022-07-17
improvement: init has lang server comment

dsinghvi:
@zachkirsch i have just hardcoded the url to the next release and we'll have to remember to bump this whenever the json schema changes. 
The better option i think is to inject a `REF` constant into the bundle through webpack's `DefinePlugin`. `Ref` should either be the commit hash or the tag which we pass to the init command. What do you think of that approach? If good, can take a stab at it.

### DIFF
--- a/packages/ete-tests/src/__test__/init/__snapshots__/init.test.ts.snap
+++ b/packages/ete-tests/src/__test__/init/__snapshots__/init.test.ts.snap
@@ -17,7 +17,7 @@ generators: []
 `;
 
 exports[`fern init 3`] = `
-"# yaml-language-server: $schema=https://raw.githubusercontent.com/fern-api/fern/0.0.152/workspace-schema.json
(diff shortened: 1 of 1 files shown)


zachkirsch:
If we can dump the json schema to s3 in circle releases, then we just keep the URL constant. Thoughts?

dsinghvi:
thought about that too -- its pretty much close to defaulting the url to "main". downside is that if you are an older cli version and ir has been changed then all the type hints are broken

zachkirsch:
This is trueee but same problem if you upgrade your cli and the schema is pinned to a particular version. And in general if your CLI is behind, your type hints will be right but you can get errors when you hit generate if fiddle is on a later version of the IR.  

prob the ideal approach is to put a version in the yaml and then we never really “break”. 

For now I’m down to just put “main”. If someone is on an older version of the CLI we’d probably want them to upgrade anyway. Wyt

PR #413, opened 2022-07-20
improvement: pre-release testing

dsinghvi:


dsinghvi:
Just replied in Slack -- plan is to just have this run right before you tag just so ik all the generator versions in `fern add` work fine

dsinghvi:
Do you know how i can make the `fern generate` output less "interactive" and more just update based

zachkirsch:
Nothing cheap - right now we just wipe and repaint every x milliseconds. It’s stateless, it just paints the latest reply from the BE. To only log updated we’ll need to change it to be stateful so it can tell what changed between polls

.circleci/config.yml
```diff
@@ -57,6 +57,22 @@ jobs:
...
+      - run: |
+          npm install -g fern-api@next
+          mkdir pre-release-test
+          cd pre-release-test
+          fern init --organization fern
+          fern add java
+          fern add typescript
+          fern add postman
+          fern generate
+          rm -rf pre-release-test
```
zachkirsch, line 74: i think there should be a `cd ..` first. and maybe `rm -r`, the `-f` is probably swallowing that error

.circleci/config.yml
```diff
@@ -57,6 +57,22 @@ jobs:
           name: Ensure no changes to git-tracked files
           command: git --no-pager diff
 
+  pre-release-test:
+    machine:
+      image: ubuntu-2004:current
```
zachkirsch, line 62: no idea if machine is more expensive, but if it is we don't need docker for this one

.circleci/config.yml
```diff
@@ -111,6 +127,10 @@ workflows:
               only: /.*/
           context:
             - auth0-dev
+      - pre-release-test:
+          filters:
+            tags:
+              only: /.*/
```
zachkirsch, line 133: youre prob tracking but rn this will run on all commits

dsinghvi: yup will restrict to main branch

zachkirsch: do we want it to run pre publishing snapshots or just releases

.circleci/config.yml
zachkirsch: @dsinghvi

dsinghvi: ah okay

.circleci/config.yml
dsinghvi: just releases

dsinghvi: there's an if case that terminates early if there is no git tag

.circleci/config.yml
```diff
@@ -57,6 +57,27 @@ jobs:
...
+      - image: cimg/node:current
+    steps:
+      - checkout
+      - node/install-packages:
+          pkg-manager: yarn-berry
+      - run:
+          name: ETE Test
+          command: |
+            yarn workspace @fern-api/cli dist
+            cli_path = yarn workspace @fern-api/cli exec pwd
```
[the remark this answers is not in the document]
dsinghvi: thanks

PR #416, opened 2022-07-22
improvement: substitute env variables in .fernrc

dsinghvi:
```
config: 
   key: ${ENV_VAR} 
```
Now `ENV_VAR` is substituted in the `.fernrc.yml`


packages/commons/src/config-types/workspace-definition/__test__/substituteEnvVariables.test.ts
```diff
@@ -0,0 +1,18 @@
...
+        process.env.ENV_VAR = "test";
+        const rawYaml = `
+        foo: bar
+        baz: 
+          qux: 
+            thud: \${ENV_VAR}
+        plugh: \${ENV_VAR}`;
+        const parsedYaml = yaml.load(rawYaml);
+        const substitutedYaml = substituteEnvVariables(parsedYaml);
+        console.log(substitutedYaml);
```
zachkirsch, line 15: Prob shouldn’t log in the test

packages/commons/src/config-types/workspace-definition/__test__/substituteEnvVariables.test.ts
```diff
@@ -0,0 +1,18 @@
+import yaml from "js-yaml";
+import { substituteEnvVariables } from "../substituteEnvVariables";
+
+describe("substituteEnvVariables", () => {
+    it("basic", () => {
+        process.env.ENV_VAR = "test";
```
zachkirsch, line 6: Nbd, but this will pollute the actual environment, you can use jest.mock in general for this kind of thing

packages/commons/src/config-types/workspace-definition/substituteEnvVariables.ts
```diff
@@ -0,0 +1,24 @@
...
+        }
+    } else if (key !== undefined) {
+        set(target, key, typeof value === "string" ? replaceStringWithEnvVariable(value) : value);
+    }
+    return target;
+}
+
+function replaceStringWithEnvVariable(val: string): string {
+    return val.replace(/\$\{(\w+)\}/g, (_substring, envVarName) => {
+        return process.env[envVarName] ?? "";
```
zachkirsch, line 22: We should probably throw if it’s not defined rather than swallow

packages/commons/src/config-types/workspace-definition/substituteEnvVariables.ts
```diff
@@ -0,0 +1,24 @@
+import { isPlainObject, set } from "lodash";
+
+export function substituteEnvVariables(yaml: unknown): unknown {
```
zachkirsch, line 3: By the time we’re in this file, the yaml has been parsed into a JSON object (by yaml.load), so I wouldn’t expect to see the word yaml anywhere. (Having a bit of trouble following what yaml/processedYaml/target/value are)

zachkirsch:
This is epic!!

packages/commons/src/config-types/workspace-definition/__test__/substituteEnvVariables.test.ts
dsinghvi: got it, i googled on stack overflow and this was the most upvoted answer

packages/commons/src/config-types/workspace-definition/__test__/substituteEnvVariables.test.ts
dsinghvi: ah just did that to get a hold of the object for assertion -- will take out

packages/commons/src/config-types/workspace-definition/substituteEnvVariables.ts
dsinghvi: errors thrown in the cli are pretty shitty so if you add some way to `alert error` and `exit` then i can make that change?

packages/commons/src/config-types/workspace-definition/substituteEnvVariables.ts
dsinghvi: that makes sense -- stole parts of this code from some other library so can fix it up and make it more sensible

packages/commons/src/config-types/workspace-definition/substituteEnvVariables.ts
zachkirsch: Yeah agree, we need to fix that. Still better to have a shitty stacktrace than to swallow imo, esp since it’s probably us using it

PR #492, opened 2022-07-29
Updating docs to remove npm-dev link

dannysheridan:


dannysheridan:
@zachkirsch no dice on the test running successfully in CircleCI

zachkirsch:
actually, different failure! on Circle you're failing ETE tests (those are run with `yarn test:ete` instead of `yarn test`). The error is that you're specifying `path-param:` instead of `path-parameters:`

dannysheridan:
hmmm 39 errors that look like they're caused by `Cannot find module '@fern-api/config-management-commons' or its corresponding type declarations.`

Screen Shot 2022-07-29 at 10 52 20 PM

PR #530, opened 2022-08-04
improvement: fern upgrade

dsinghvi:


packages/cli/cli/src/commands/upgrade-generator/upgradeGeneratorInWorkspace.ts
```diff
@@ -0,0 +1,22 @@
+import { upgradeGeneratorsIfPresent } from "@fern-api/manage-generator";
+import { loadRawWorkspaceConfiguration } from "@fern-api/workspace-configuration";
+import { writeFile } from "fs/promises";
+import yaml from "js-yaml";
+import { loadProject } from "../utils/load-project/loadProject";
+
+export async function upgradeGeneratorsInWorkspace(commandLineWorkspaces: readonly string[]): Promise<void> {
```
zachkirsch, line 7: `readonly`!!

packages/cli/cli/src/commands/upgrade-generator/upgradeGeneratorInWorkspace.ts
```diff
@@ -0,0 +1,22 @@
...
+    const { workspaceConfigurations } = await loadProject({ commandLineWorkspaces });
+
+    for (const workspaceConfigurationFilePath of workspaceConfigurations) {
+        const workspaceConfiguration = await loadRawWorkspaceConfiguration(workspaceConfigurationFilePath);
+        const updatedWorkspaceConfiguration = upgradeGeneratorsIfPresent({ workspaceConfiguration });
+        if (updatedWorkspaceConfiguration.updatedWorkspaceConfiguration !== workspaceConfiguration) {
```
zachkirsch, line 13: This is testing reference equality, you'd need to do a deep equals. But I think we can just skip this check

dsinghvi: ah makes sense, i just stole it from addGenerator

packages/cli/cli/src/commands/upgrade-generator/upgradeGeneratorInWorkspace.ts
```diff
@@ -0,0 +1,22 @@
...
+            await writeFile(workspaceConfigurationFilePath, yaml.dump(updatedWorkspaceConfiguration));
+        }
+        for (const upgradeInfo of updatedWorkspaceConfiguration.upgrades) {
+            console.log(
```
zachkirsch, line 17: nit ``console.log(chalk.green(`my message`))``

packages/config-management/manage-generator/src/generatorInvocations.ts
```diff
@@ -0,0 +1,50 @@
...
+    generate: {
+        enabled: true,
+        output: "./generated-openapi.yml",
+    },
+    config: {
+        format: "yaml",
+    },
+};
+
+export function getGeneratorInvocationsByName(): Record<string, GeneratorInvocationSchema> {
```
zachkirsch, line 43: nbd but generally a function like this can just be an object:

```ts
export const GENERATOR_INVOCATIONS: Record<string, GeneratorInvocationSchema> = {
  [JAVA_GENERATOR_INVOCATION.name]: JAVA_GENERATOR_INVOCATION,
  ...
}

// no need to export these anymore, all consumers should use GENERATOR_INVOCATIONS
const JAVA_GENERATOR_INVOCATION: GeneratorInvocationSchema = {
...
```

Also, I would make the map from something stronger than `string` so the lookups are typesafe, e.g.
```ts
export enum Generator {
  Java,
  TypeScript,
  ...
}

export const GENERATOR_INVOCATIONS: Record<Generator, GeneratorInvocationSchema> = {
```

packages/cli/cli/src/commands/upgrade-generator/upgradeGeneratorInWorkspace.ts
```diff
@@ -0,0 +1,23 @@
...
+import chalk from "chalk";
+import { loadProject } from "../utils/load-project/loadProject";
+
+export async function upgradeGeneratorsInWorkspace(commandLineWorkspaces: readonly string[]): Promise<void> {
+    const { workspaceConfigurations } = await loadProject({ commandLineWorkspaces });
+
+    for (const workspaceConfigurationFilePath of workspaceConfigurations) {
+        const workspaceConfiguration = await loadRawWorkspaceConfiguration(workspaceConfigurationFilePath);
+        const updatedWorkspaceConfiguration = upgradeGeneratorsIfPresent({ workspaceConfiguration });
+        await writeFile(workspaceConfigurationFilePath, yaml.dump(updatedWorkspaceConfiguration));
```
zachkirsch, line 14: Or even better,
```ts
if (updatedWorkspaceConfiguration.upgrades.length > 0)
```

packages/config-management/manage-generator/src/generatorInvocations.ts
dsinghvi: ah thanks -- i couldnt figure out how to statically construct a record with dynamic keys

packages/config-management/manage-generator/src/upgradeGenerator.ts
```diff
@@ -0,0 +1,45 @@
...
+export function upgradeGeneratorsIfPresent({
+    workspaceConfiguration,
+}: {
+    readonly workspaceConfiguration: WorkspaceConfigurationSchema;
+}): GeneratorUpgradeResult {
+    const generatorInvocationsByName = getGeneratorInvocationsByName();
+    const upgrades: GeneratorUpgradeInfo[] = [];
+    const updatedGenerators = workspaceConfiguration.generators.map((generatorConfig) => {
+        const updatedInvocation = generatorInvocationsByName[generatorConfig.name];
+        if (updatedInvocation != null && updatedInvocation.version !== generatorConfig.version) {
```
zachkirsch, line 24: nice thing about making the `Record` key stronger than `string` is that you can scrap this null check, and you'll get a compile error if you forget a generator in that map

zachkirsch: `updatedInvocation.version !== generatorConfig.version` should we make sure that the new version is later than the old version?

packages/config-management/manage-generator/src/generatorInvocations.ts
dsinghvi: the name still needs to be `"fernapi/fern-java"`

zachkirsch: You can do `Java = “name”` in the enum

packages/cli/cli/src/commands/upgrade-generator/upgradeGeneratorInWorkspace.ts
zachkirsch: Ah cool, we should fix that too at some point

PR #550, opened 2022-08-09
Use older versions in cli upgrade test

zachkirsch:


zachkirsch:
@dsinghvi fysa - noticed an issue where the .fernrc used in the `upgrade` test would get edited in place (so after the first run, the .fernrc had all the new versions)

dsinghvi:
> @dsinghvi fysa - noticed an issue where the .fernrc used in the `upgrade` test would get edited in place (so after the first run, the .fernrc had all the new versions)

ah good point

PR #568, opened 2022-08-12
Improve names in IR

zachkirsch:


packages/generation/ir/ir-types/src/types.yml
```diff
@@ -7,13 +7,12 @@ types:
     docs: "A type, which is a name and a shape"
     extends: commons.WithDocs
     properties:
-      name: DeclaredTypeName
+      name: TypeName
       shape: Type
-  DeclaredTypeName:
+  TypeName:
```
dsinghvi, line 12: lets not do this -- it conflicts with java poet type name

zachkirsch: cool, thoughts on me changing errorname and servicename to match

packages/generation/ir/ir-types/src/auth.yml
```diff
@@ -2,9 +2,10 @@
 
 imports:
   types: types.yml
-  commons: ./commons.yml
+  commons: commons.yml
+  http: services/http.yml
 types:
-  AuthSchemeName: string
+  AuthSchemeName: commons.StringWithAllCasings
```
dsinghvi, line 8: why does the auth scheme name need to be `camel cased/ snake cased` ?

packages/generation/ir/ir-types/src/types.yml
dsinghvi: like `DeclaredErrorName` ?

dsinghvi: and `DeclaredServiceName` ?

packages/generation/ir/ir-types/src/auth.yml
zachkirsch: yeah good point, we can just use string here and enforce a good name in the linter

packages/generation/ir/ir-types/src/types.yml
zachkirsch: yeah

packages/generation/ir/ir-types/src/auth.yml
dsinghvi: cool and the name there would be UpperCamelCase just like types or error names

PR #590, opened 2022-08-18
improvement: add generator works with new layout

dsinghvi:
also provide feedback to user when a generator is added


zachkirsch:
Merging via https://github.com/fern-api/fern/pull/591

PR #654, opened 2022-08-29
Warn when receiving unknown status type from fiddle

zachkirsch:
Fixes https://github.com/fern-api/fern/issues/653


zachkirsch:
Fixes https://github.com/fern-api/fern/issues/653

PR #683, opened 2022-09-11
improvement: allow defining path parameters at http service level

dsinghvi:


packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
```diff
@@ -30,6 +30,7 @@ export function convertHttpService({
             fernFilepath,
         },
         basePath: serviceDefinition["base-path"],
```
zachkirsch, line 32: Isn't this effectively a break anyway? Since this will be `basePath: /{appId}/` which will break old generators. All to say do we even need to v2?

fern/ir-types/definition/services/http.yml
```diff
@@ -9,7 +9,10 @@ types:
     extends: commons.WithDocs
     properties:
       name: serviceCommons.DeclaredServiceName
-      basePath: optional<string>
+      basePath:
+        type: optional<string>
+        docs: This is deprecated.
+      basePathV2: optional<HttpPath>
```
zachkirsch, line 15: I think if the service is gonna have the HttpPath in the IR, should the path params go here too?

packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
dsinghvi: if we don't v2 then we guarantee that any runs of `fern generate` with non upgraded generators will fail as well (which will likely be the common case)

dsinghvi: by keeping around the old field we wont be breaking usage of non-upgraded generators

fern/ir-types/definition/services/http.yml
dsinghvi: down to do that

packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
zachkirsch: Makes sense- and if you want to use the new path params thing you have to make sure the generators you’re using support the v2?

dsinghvi: exactly

PR #785, opened 2022-10-03
improvement: consume fiddle local code-gen apis

dsinghvi:


packages/cli/migrations/src/migrations/0.0.207/index.ts
```diff
@@ -0,0 +1,9 @@
+import { VersionMigrations } from "../../types/VersionMigrations";
+import UnionSinglePropertyKeyMigration from "./add-mode-to-draft-generators";
```
zachkirsch, line 2: AddModeToDraftGeneratorsMigration

packages/cli/migrations/src/migrations/0.0.207/add-mode-to-draft-generators/migration.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!YAML.isSeq(draftGenerators)) {
+        context.fail(`draft generators are not a list in ${filepath}`);
+        return;
+    }
+    draftGenerators.items.forEach((draftGenerator) => {
+        if (!YAML.isMap(draftGenerator)) {
+            context.fail(`draft generator is not an object in ${filepath}`);
+            return;
+        }
+        const name = draftGenerator.get("name");
```
zachkirsch, line 42: def not worth changing, but for context I just remembered there's a boolean config somewhere that if it's true it'll return you the `Scalar` node with comments vs just the string

packages/cli/migrations/src/migrations/0.0.207/index.ts
dsinghvi: ah thanks

packages/cli/migrations/src/migrations/0.0.207/add-mode-to-draft-generators/migration.ts
dsinghvi: ah you're right

dsinghvi: if i pass in true ill get the node as opposed to the `string`

PR #846, opened 2022-10-22
improvement: support `origin` 

dsinghvi:


fern/ir-types/definition/ir.yml
```diff
@@ -25,6 +26,7 @@ types:
         type: FernConstants
         docs: this is deprecated. use constantsV2
       constantsV2: constants.Constants
+      environments: environment.Environments
```
dsinghvi, line 29: should we make this optional so that its easier for generators to check the case `are environments empty` ? Currently they have to check `environments.defaultEnvironment != null && environments.otherEnvironments.length == 0`

dsinghvi: gonna go ahead and do it

PR #850, opened 2022-10-24
Use visitObject for api files too

zachkirsch:


packages/cli/yaml/validator/src/createAstVisitorForRules.ts
```diff
@@ -44,16 +44,16 @@ export function createAstVisitorForRules({
...
+            ...createAstNodeVisitor("docs"),
+            ...createAstNodeVisitor("import"),
+            ...createAstNodeVisitor("typeReference"),
+            ...createAstNodeVisitor("typeDeclaration"),
+            ...createAstNodeVisitor("typeName"),
+            ...createAstNodeVisitor("httpService"),
+            ...createAstNodeVisitor("httpEndpoint"),
+            ...createAstNodeVisitor("queryParameter"),
+            ...createAstNodeVisitor("errorDeclaration"),
+            ...createAstNodeVisitor("errorReference"),
```
dsinghvi, line 56: why not noop? i.e. root api file doesnt contain types?

PR #854, opened 2022-10-24
Replace 'in' with 'includes' in default-env validation rule

zachkirsch:


zachkirsch:
@dsinghvi fysa you can't use "in" for lists

PR #861, opened 2022-10-28
Try/catch generator invocations

zachkirsch:


packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForWorkspace.ts
```diff
@@ -21,7 +21,7 @@ export async function runRemoteGenerationForWorkspace({
         return;
     }
 
-    await Promise.all(
+    const results = await Promise.allSettled(
```
dsinghvi, line 24: so is `allSettled` the big change here? basically wait until each task is actually done?

zachkirsch: yes exactly

before the generator invocations were supposed to _not_ throw (and there was one place we were accidentally throwing which was causing danny's issue)

now, the generators invocations can throw as much as they want, and we catch them here with allSettled

PR #868, opened 2022-10-28
Fix upgrading when generators.yml is empty

zachkirsch:


zachkirsch:
the root cause was that the CLI was throwing on unknown generators

PR #871, opened 2022-10-31
Always ignore disabled rules

zachkirsch:


dsinghvi (approved):
We followed a really similar pattern for "disabled" migrations on other backends

PR #950, opened 2022-12-02
fix: add more logging during remote generation task

dsinghvi:


packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForWorkspace.ts
```diff
@@ -23,20 +23,25 @@ export async function runRemoteGenerationForWorkspace({
...
+                } catch (e) {
+                    interactiveTaskContext.logger.error("Failed");
+                    interactiveTaskContext.logger.debug(`${e}`);
+                }
+            })
         )
     );
 
-    if (results.some((result) => result.status === "rejected")) {
+    if (results.some((result) => result.status === "rejected" || result.value === false)) {
```
zachkirsch, line 44: nit `|| !result.value`

packages/cli/cli/src/cli.ts
```diff
@@ -63,7 +63,8 @@ async function runCli() {
     } catch (error) {
         if (error instanceof FernCliError) {
             if (cliContext.getLogLevel() === LogLevel.Debug) {
-                cliContext.logger.error(`${error.name}: ${error.message}\n${error.stack}`);
+                cliContext.logger.error(`${error.name}: ${error.message}`);
+                cliContext.logger.error(`${error.stack}`);
```
zachkirsch, line 67: Not sure if this was an intentional change but the reason for the `\n` was so the prefix only appears one:
```
[fern-typescript]: error message
                   at bundle.js:3213 ...
```

vs

```
[fern-typescript]: error message
[fern-typescript]: at bundle.js:3213 ...
```

packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForWorkspace.ts
```diff
@@ -23,20 +23,25 @@ export async function runRemoteGenerationForWorkspace({
...
+                        organization,
+                        workspace,
+                        interactiveTaskContext,
+                        generatorInvocation,
+                        version,
+                        audiences: generatorGroup.audiences,
+                    });
+                } catch (e) {
+                    interactiveTaskContext.logger.error("Failed");
+                    interactiveTaskContext.logger.debug(`${e}`);
```
zachkirsch, line 38: This will probably log out `Error`, you should do `interactiveTaskContext.logger.debug(e);`. We have logic here to stringify it correctly https://github.com/fern-api/fern/blob/main/packages/cli/cli/src/cli-context/logErrorMessage.ts

packages/cli/cli/src/cli.ts
dsinghvi: gotcha i made that change it just looked weird

dsinghvi: Screen Shot 2022-12-02 at 5 45 04 PM

zachkirsch: Lmao looks like we can just scrap the first log then and just log the stack

zachkirsch: Oh got it there’s no message

zachkirsch: Cool yeah what you have looks perfect

zachkirsch: And maybe we only log the first line if there’s a non empty message?

dsinghvi: perfect will change that

packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForWorkspace.ts
dsinghvi: Screen Shot 2022-12-02 at 6 04 09 PM

PR #954, opened 2022-12-05
add error discrimination config

dsinghvi:
Adds `error-discrimination` as a configuration in the root `api.yaml`. Errors can either be discriminated by status-code or properties as seen below. 

```yaml
error-discrimination:
  strategy: status-code
```

```yaml
error-discrimination:
  strategy: property
  property-name: errorName
```

Other things in the PR are
- *validation*: if there are declared errors, then `error-discrimination` must be present
- *validation*: if the strategy is status-code, then make sure that status codes don't conflict
- *migration*: the `error-discriminant` config is removed and auto-migrated to property based error-discrimination

### DIFF
--- a/packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
+++ b/packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
@@ -96,15 +96,17 @@ export async function generateIntermediateRepresentation({
                 if (errors == null) {
                     return;
                 }
-                const isStatusCodeDiscriminated =
-                    workspace.rootApiFile["error-discrimination"]?.strategy === "status-code";
+                const errorDiscriminationSchema = workspace.rootApiFile["error-discrimination"];
+                if (errorDiscriminationSchema == null) {
+                    throw new Error("Please specify error-discrimination in api.yml.");
+                }
                 for (const [errorName, errorDeclaration] of Object.entries(errors)) {
                     const convertedErrorDeclaration = convertErrorDeclaration({
                         errorName,
                         errorDeclaration,
                         file,
                         typeResolver,
-                        isStatusCodeDiscriminated,
+                        errorDiscriminationSchema,
                     });
                     audienceIrGraph?.addError(convertedErrorDeclaration);
                     intermediateRepresentation.errors.push(convertedErrorDeclaration);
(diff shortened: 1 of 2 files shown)


fern/ir-types/definition/errors.yml
```diff
@@ -21,6 +24,10 @@ types:
       typeV3: optional<types.TypeReference>
       http: optional<HttpErrorConfiguration>
       statusCode: integer
+  ErrorDeclarationDiscriminant:
```
zachkirsch, line 27: Can we call this `ErrorDeclarationDiscriminant` to distinguish between discriminant ('errorName') and discriminant value ('PostNotFoundError')

packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
```diff
@@ -95,12 +96,15 @@ export async function generateIntermediateRepresentation({
                 if (errors == null) {
                     return;
                 }
+                const isStatusCodeDiscriminated =
```
zachkirsch, line 99: not a huge deal, but if we add a new strategy we won't get any compile breaks in this package. Alternative way is to pass in the `RawSchemas.ErrorDiscriminationConfigSchema` into convertErrorDeclaration, and have a helper function there (e.g. `convertDiscriminantValue`) that switches on the `strategy` and assertsNever in the default case.

packages/cli/migrations/src/migrations/0.0.248/add-error-discrimination-config/migration.ts
```diff
@@ -0,0 +1,36 @@
...
+                context.failWithoutThrowing(`Failed to migrate ${filepath}`, error);
+            }
+        }
+    },
+};
+
+async function migrateRootApiFile(filepath: AbsoluteFilePath, _context: TaskContext): Promise<void> {
+    const contents = await readFile(filepath);
+    const parsedDocument = YAML.parseDocument(contents.toString());
+    const errorDiscriminant = parsedDocument.get("error-discriminant", true);
```
zachkirsch, line 26: sneaky `true` param here i remember

zachkirsch (approved):
epic

packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
dsinghvi: makes sense -- i didn't want to pass it in to the converter and do it outside since its going to be the same for every error. but maybe i can convert outside

packages/cli/migrations/src/migrations/0.0.248/add-error-discrimination-config/migration.ts
dsinghvi: yup -- im not changing this key so don't need the comments

PR #980, opened 2022-12-10
Upgrade generators to latest version, not min version

zachkirsch:


zachkirsch:
Closing as sometimes people want to just update the CLI and not the generators. We should have another command to upgrade generators. Tracking at https://github.com/fern-api/fern/issues/981

PR #1005, opened 2022-12-15
support `fern register`

dsinghvi:


packages/cli/cli/src/cli.ts
```diff
@@ -268,6 +269,53 @@ function addIrCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
...
+            }
+            const workspace = workspaces[0];
+            if (workspace == null) {
+                cliContext.fail(`Failed to load ${argv.api}`);
+                return;
+            }
+            if (argv.token == null) {
+                cliContext.fail("Token must be specified");
+                return;
+            }
```
dsinghvi, line 308: @zachkirsch had to do a bunch of checks (want a single `Workspace` and a required `token`) is there a way around doing these checks manually?

Issue #1040, opened 2022-12-24
I want to add a description to my API


dannysheridan:
I want to add a high-level overview / description about my API. 

Suggestion: this should live in `api.yml` as `docs`.

dannysheridan:
Example: https://developers.gigs.com/docs/api/72b08d51d7429-gigs-core-api

https://i.ibb.co/vqN2THJ/Screenshot-2022-12-23-at-5-06-59-PM.png

PR #1041, opened 2022-12-24
add top level api docs

dsinghvi:
fixes https://github.com/fern-api/fern/issues/1040


dannysheridan:
Thanks for making this one available today. It'll be used in every API I model.

PR #1082, opened 2022-12-30
fern `upgrade` optionally takes version

dsinghvi:


packages/cli/cli/src/cli.ts
```diff
@@ -343,15 +343,21 @@ function addUpgradeCommand({
...
-            }),
+            yargs
+                .option("rc", {
+                    boolean: true,
+                    hidden: true,
+                    default: false,
+                })
+                .option("version", {
+                    string: true,
+                    description: "The version to upgrade to. If unspecified, will upgrade to latest.",
```
zachkirsch, line 354: can we change to `"The version to upgrade to. Defaults to the latest release."`

packages/cli/cli/src/commands/upgrade/upgrade.ts
```diff
@@ -38,12 +38,15 @@ const PREVIOUS_VERSION_ENV_VAR = "FERN_PRE_UPGRADE_VERSION";
 export async function upgrade({
     cliContext,
     includePreReleases,
+    upgradeVersion,
 }: {
     cliContext: CliContext;
     includePreReleases: boolean;
+    upgradeVersion?: string;
```
zachkirsch, line 45: `upgradeVersion: string | undefined` (reasoning: if we forget to pass it, that's probably a mistake on our end, so we'd want the compiler to warn)

zachkirsch: also nitty but can we call it `toVersion` or `targetVersion` or something like that

PR #1272, opened 2023-02-03
Add fdr UI

zachkirsch:


fern/api/definition/api/__package__.yml
```diff
@@ -0,0 +1,89 @@
...
+      key: string
+      value: TypeReference
+
+  PrimitiveType:
+    enum:
+      - integer
+      - double
+      - long
+      - string
+      - boolean
```
dsinghvi, line 70: can we make these unions -- i wanna make it easy to support `regex` in string for example ?

PR #1335, opened 2023-02-25
[openapi validator] log all breadcrumbs when running `fern check`

TeisJayaswal:


zachkirsch:
Closing because we deleted the openapi validator

PR #1353, opened 2023-03-04
add `base-properties` to union

dsinghvi:


packages/cli/generation/ir-migrations/src/migrations/v9-to-v8/migrateFromV9ToV8.ts
```diff
@@ -23,6 +23,12 @@ export const V9_TO_V8_MIGRATION: IrMigration<
         [GeneratorName.POSTMAN]: AlwaysRunMigration,
     },
     migrateBackwards: (v9): IrVersions.V8.ir.IntermediateRepresentation => {
+        for (const [_, type] of Object.entries(v9.types)) {
+            if (type.shape._type === "union" && type.shape.baseProperties.length > 0) {
+                throw new Error("Failed to migrate IR because you're using base-properties on union.");
```
zachkirsch, line 28: nit can you do

```
throw new Error(`Failed to migrate IR because ${type.name.name.originalName} uses "base-properties".`);
```

PR #1367, opened 2023-03-07
fix: introduce undiscrminated unions

dsinghvi:


packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
```diff
@@ -109,6 +109,9 @@ export function convertTypeExample({
                 wireValue: example,
             });
         },
+        undiscriminatedUnion: () => {
+            throw new Error("Unsupported!");
```
zachkirsch, line 113: can you say what's going on here in case this actually gets thrown
`Examples are not supported for undiscriminated unions.`

Ideally this is a fern check rule as well

dsinghvi: just put it as a place holder for now -- my thought was i'm going to need to fill this out once i support examples? wdyt

packages/cli/generation/ir-migrations/src/migrations/v12-to-v11/migrateFromV12ToV11.ts
```diff
@@ -24,14 +24,114 @@ export const V12_TO_V11_MIGRATION: IrMigration<
...
+            types: v11Types,
+            services: mapValues(v12.services, (service) => convertService(service, { taskContext, targetGenerator })),
         };
     },
 };
 
+function getUndiscriminatedUnionsErrorMessage(context: IrMigrationContext): string {
+    return (
+        `Generator ${context.targetGenerator.name}@${context.targetGenerator.version}` +
+        " does not support undiscriminated unions" +
```
zachkirsch, line 129: missing period at end of this line

packages/cli/yaml/yaml-schema/src/ast/visitors/visitTypeDeclarations.ts
```diff
@@ -136,6 +136,28 @@ export async function visitTypeDeclaration({
                 examples: visitExamples,
             });
         },
+        undiscriminatedUnion: async (union) => {
+            await visitObject(union, {
+                docs: createDocsVisitor(visitor, nodePathForType),
+                discriminated: noop,
+                union: async (unionMembers) => {
+                    for (const [index, unionMember] of unionMembers.entries()) {
+                        const nodePathForUnionType = [...nodePathForType, `union[${index}]`];
```
zachkirsch, line 145:
```ts
const nodePathForUnionType = [...nodePathForType, { key: "union", arrayIndex: index }];
```

dsinghvi: thats sick

packages/cli/yaml/yaml-schema/src/schemas/UndiscriminatedUnionSchema.ts
```diff
@@ -0,0 +1,17 @@
...
+export const UndiscriminatedUnionSchema = BaseTypeDeclarationSchema.extend({
+    union: z.array(
+        z.union([
+            z.string(),
+            WithDocsSchema.extend({
+                type: z.string(),
+            }),
+        ])
+    ),
+    discriminated: z.literal(false),
```
zachkirsch, line 14: on the other union could see you doing `discriminated: z.optional(z.literal(true))`

I think it's more common to have optional booleans default to false, but in this case it's defaulting to true. any ideas for a key that defaults to false? nbd if not.

packages/cli/yaml/yaml-schema/src/utils/visitRawTypeDeclaration.ts
```diff
@@ -40,9 +45,23 @@ export function isRawObjectDefinition(rawTypeDeclaration: TypeDeclarationSchema)
...
+    );
+}
+
+export function isRawUndiscriminatedUnionDefinition(
+    rawTypeDeclaration: TypeDeclarationSchema
+): rawTypeDeclaration is UndiscriminatedUnionSchema {
+    const undiscriminatedUnionSchema = rawTypeDeclaration as UndiscriminatedUnionSchema;
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    return (rawTypeDeclaration as UnionSchema).union != null;
+    return undiscriminatedUnionSchema.union != null && undiscriminatedUnionSchema.discriminated != null;
```
zachkirsch, line 64: If you add the `optional(literal(true))` you'll have to change this logic

packages/cli/yaml/yaml-schema/src/schemas/UndiscriminatedUnionSchema.ts
dsinghvi: `undiscriminated` ?  feels more complex than now

packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
zachkirsch: Oh dope. Gonna be tricky to support I think because you’ll have to do the Jackson thing of figuring out if the example is valid. (IMO better to disallow then allow invalid example shapes)

PR #1368, opened 2023-03-07
Simplify v12->v11 migration

zachkirsch:


packages/cli/generation/ir-migrations/src/migrations/v12-to-v11/migrateFromV12ToV11.ts
```diff
@@ -25,95 +25,59 @@ export const V12_TO_V11_MIGRATION: IrMigration<
...
-                            };
+        const v11Types: Record<string, IrVersions.V11.types.TypeDeclaration> = mapValues(
+            v12.types,
+            (typeDeclaration) => {
+                return {
+                    ...typeDeclaration,
+                    shape: IrVersions.V12.types.Type._visit<IrVersions.V11.types.Type>(typeDeclaration.shape, {
+                        union: IrVersions.V11.types.Type.union,
+                        enum: IrVersions.V11.types.Type.enum,
+                        object: IrVersions.V11.types.Type.object,
```
dsinghvi, line 36: So this is pointing at the constructor?

zachkirsch: yeah exactly

`IrVersions.V11.types.Type.object` is a reference to a function, so it's equivalent to

```ts
object: (...args) => IrVersions.V11.types.Type.object(...args)
```

PR #1467, opened 2023-04-13
add union extensions

dsinghvi:
Now unions can have extensions. Here's an example: 
```yaml

BaseShape: 
  properties: 
    area: double

Shape: 
  extends: BaseShape
  union: 
    triangle: Triangle
    square: Square
```


zachkirsch:
Let’s go!!

PR #1531, opened 2023-05-03
feature: add extensions to objects

dsinghvi:


packages/cli/yaml/yaml-schema/src/ast/visitors/visitTypeDeclarations.ts
```diff
@@ -86,6 +86,7 @@ export async function visitTypeDeclaration({
                         await visitTypeReference(extendedType, nodePathForExtension);
                     }
                 },
+                extensions: noop,
```
mstade, line 89: What's the point of adding `extensions` to the schema, and then not do anything with the data? This could be very useful to [generate extensions](https://spec.openapis.org/oas/v3.0.1#specificationExtensions) in the OAS output. Are there any plans to support this?

PR #1597, opened 2023-05-16
Add fern-go-model generator

amckinney:
Adds the `fernapi/fern-go-model` generator identifier to the `fern` CLI. 

As discussed, this generator might be decoupled into a few different generators (i.e. a separate one for models, clients, and servers), so this generator will only act as the `fernapi/fern-go-model` portion.

### DIFF
--- a/packages/cli/generation/ir-migrations/src/migrations/v20-to-v19/migrateFromV20ToV19.ts
+++ b/packages/cli/generation/ir-migrations/src/migrations/v20-to-v19/migrateFromV20ToV19.ts
@@ -25,7 +25,7 @@ export const V20_TO_V19_MIGRATION: IrMigration<
         [GeneratorName.STOPLIGHT]: AlwaysRunMigration,
         [GeneratorName.POSTMAN]: AlwaysRunMigration,
         [GeneratorName.PYTHON_SDK]: "0.3.6-rc1-2-g6050f7f8",
-        [GeneratorName.GO_MODEL]: AlwaysRunMigration,
+        [GeneratorName.GO_MODEL]: undefined,
     },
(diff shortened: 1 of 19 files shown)


dsinghvi:
@amckinney i think it makes sense to start as `fern-go-model` because its _probably_ just going to be more work to go back and decouple and update versions. But not a strongly held opinion.

dsinghvi (approved):

amckinney:
@dsinghvi Sounds good, updated! Let me know if everything still looks good to you - I verified everything continues to work locally, too.

packages/cli/generation/ir-migrations/src/migrations/v20-to-v19/migrateFromV20ToV19.ts
```diff
@@ -25,6 +25,7 @@ export const V20_TO_V19_MIGRATION: IrMigration<
         [GeneratorName.STOPLIGHT]: AlwaysRunMigration,
         [GeneratorName.POSTMAN]: AlwaysRunMigration,
         [GeneratorName.PYTHON_SDK]: "0.3.6-rc1-2-g6050f7f8",
+        [GeneratorName.GO_MODEL]: AlwaysRunMigration,
```
dsinghvi, line 28: i'm guessing the go model generator will depend on the latest version of the IR, should we put `0.0.0` from here? Basically from the onset the go generator will not support ir version 15 and this will make sure of that

amckinney: Discussed offline - we actually want to use `undefined` in all of the migration files to prevent any migration from occurring - `fern-go-model` will only accept `v20+` to start.

PR #1600, opened 2023-05-16
fix: document example error bodies

dsinghvi:


packages/docs/mintlify/definition/endpoints.mdx
```diff
@@ -325,6 +325,26 @@ examples:
...
+
+If the error has a body, then you must include the body in the example. 
+
+```yaml user.yml
+examples:
+  - path-parameters:
+      userId: missing-user-id
+    response:
+      error: UserNotFoundError # <--
+      body: "User with id `missing-user-id` was not found"
```
zachkirsch, line 342: would move the `# <--` down a line

zachkirsch: and maybe add one to `type: string` too

dsinghvi: got caught with the copy pasta

PR #1605, opened 2023-05-18
Add docs for openapi extensions

zachkirsch:


packages/cli/openapi-parser/src/v3/extensions/extensions.ts
```diff
@@ -1,41 +1,75 @@
...
- *     parameters:
- *       - name: id
- *         in: path
- *         type: string
- *         x-fern-sdk-variable: appName
- */
-export const X_FERN_SDK_VARIABLES = "x-fern-sdk-variables";
-export const X_FERN_SDK_VARIABLE = "x-fern-sdk-variable";
+export const FernOpenAPIExtension = {
+    SDK_METHOD_NAME: "x-fern-sdk-method-name",
```
dsinghvi, line 4: this sets the name of the sdk method (e.g. `getAccessCodes` or `update`)

packages/cli/openapi-parser/src/v3/extensions/extensions.ts
```diff
@@ -1,41 +1,75 @@
...
+    SDK_GROUP_NAME: "x-fern-sdk-group-name",
```
dsinghvi, line 5: this sets the name of the resource / groups the endpoint. e.g `this.accessCodes.update()` it would set accessCodes

dsinghvi: Squidex is a good example that you can possibly link to 

https://github.com/Squidex/sdk-fern/blob/ca5d633d9d08f01ca6ef74c397f028d493b2392c/fern/api/openapi/openapi.json#L89-L90

this would turn into `client.userManagement.getUsers()`

packages/cli/openapi-parser/src/v3/extensions/extensions.ts
```diff
@@ -1,41 +1,75 @@
...
+    STREAMING: "x-fern-streaming",
```
dsinghvi, line 6: this is a boolean 

```
paths: 
  /my-path
    post: 
      x-fern-streaming: true
```
This will set `response-stream` in fern with the success response type (200 status code)

packages/cli/openapi-parser/src/v3/extensions/extensions.ts
```diff
@@ -1,41 +1,75 @@
...
+    REQUEST_NAME_V1: "x-request-name",
+    REQUEST_NAME_V2: "x-fern-request-name",
```
dsinghvi, line 8: sets the request name in fern 

```yaml
endpoints: 
  get
    request: 
      name: GetResourceRequest # <---- sets this name 
      query-parameters: 
        ...
```

dsinghvi: request name is used for wrapped request

packages/cli/openapi-parser/src/v3/extensions/extensions.ts
```diff
@@ -1,41 +1,75 @@
...
+    TYPE_NAME: "x-fern-type-name",
```
dsinghvi, line 9: If you want to add a type to an inlined schema or override an existing type name in OpenAPI, use `x-fern-type-name`

```
components:
  schemas:
     BadName: 
        x-fern-type-name: GoodName # <---- GoodName will be used in the SDK when generating Code
        type: object
        properties: 
           baz: string
```

packages/cli/openapi-parser/src/v3/extensions/extensions.ts
```diff
@@ -1,41 +1,75 @@
...
+    ENUM_NAMES_V1: "x-enum-names",
+    ENUM_NAMES_V2: "x-fern-enum-names",
+    ENUM_VAR_NAMES_V1: "x-enum-varnames",
+    ENUM_VAR_NAMES_V2: "x-fern-enum-var-names",
```
dsinghvi, line 13: lets exclude all enum extensions

packages/cli/openapi-parser/src/v3/extensions/extensions.ts
```diff
@@ -1,41 +1,75 @@
...
+    SERVER_NAME_V1: "x-name",
+    SERVER_NAME_V2: "x-fern-server-name",
```
dsinghvi, line 15: OpenAPI only has fields `url` and `description` for a server. In the generated SDK you need a name you can reference it by which we put into `x-fern-server-name`.

PR #1659, opened 2023-05-28
Set organization flag to not be hidden

ff137:
init can fail because of an invalid organization name, so it's helpful for the user to know that flag exists

### DIFF
--- a/packages/cli/cli/src/cli.ts
+++ b/packages/cli/cli/src/cli.ts
@@ -171,7 +171,7 @@ function addInitCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
                 .option("organization", {
                     alias: "org",
                     type: "string",
-                    description: "Organization name"
+                    description: "Organization name",
                 }),
         async (argv) => {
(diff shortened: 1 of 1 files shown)


zachkirsch:
Thanks @ff137!

packages/cli/cli/src/cli.ts
```diff
@@ -172,7 +172,7 @@ function addInitCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
                     alias: "org",
                     type: "string",
                     description: "Organization name",
```
zachkirsch, line 174:
```suggestion
                    description: "Organization name"
```

packages/cli/cli/src/cli.ts
```diff
@@ -172,7 +172,7 @@ function addInitCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
...
-                    hidden: true,
+                    hidden: false,
```
zachkirsch, line 175:
```suggestion

```

PR #1660, opened 2023-05-28
Update generators.mdx to show latest versions in example

ff137:
Update the version numbers used in the samples to line up with current latest versions.

fastapi sample references version 0.0.33, which is markedly older than the current 0.3.6. This made it unclear to me if the latest release on `fern-python` repo is indeed correct to use in the fastapi generator

Please help validate if the versions I entered are indeed correct


ff137:
Additionally, it would be great if the docs could make it clear that the releases on `fern-python` correspond to the versions that could be used for all python generators.

PR #1705, opened 2023-06-05
Add header blurring

zachkirsch:


zachkirsch:
Notes from aashman: increase blur

zachkirsch:
Ported to https://github.com/fern-api/fern-ui/pull/4

PR #1777, opened 2023-06-13
fix: dependencies support filepaths

dsinghvi:


packages/cli/config-management/dependencies-configuration/src/convertDependenciesConfiguration.ts
```diff
@@ -18,20 +22,35 @@ export async function convertDependenciesConfiguration({
...
-        const organization = splitCoordinate[0];
-        const apiName = splitCoordinate[1];
-        if (organization != null && apiName != null) {
-            dependencies[coordinate] = {
-                organization,
-                apiName,
-                version,
-            };
+    for (const [coordinate, versionOrPath] of Object.entries(rawDependenciesConfiguration.dependencies)) {
+        if (isCoordinate(coordinate)) {
```
zachkirsch, line 26: shouldn't we be checking if `versionOrPath` is a version or path? The coordinate is up to them so they might name it `@org/dep` to match other dependencies

dsinghvi: good point, its what i had before not sure why i changed it

dsinghvi: i went this route because i figured version could also be set by them

Issue #1655, opened 2023-05-28
Improve error message when organization name is invalid


zachkirsch:
We should:
1. Say why it's invalid
2. Suggest the `fern init --organization <org name>` command

ff137:
Thanks for following up!

As a minor mention (not worth creating an issue/PR just for a link fix) - the following link at:https://docs.buildwithfern.com/spec/openapi
> you can run [fern generate](https://docs.buildwithfern.com/compiler/generate).

The link should point to https://docs.buildwithfern.com/compiler/fern-generate

shashank40:
Hello @zachkirsch can you please assign this issue to me? 

Also to get an idea of how specific the message should be when we return an error saying 'Why is it invalid'?

zachkirsch:
I think the message can say something like “organization name must start with a letter contain only lowercase letters, numbers, and dashes”

shashank40:
Hello @zachkirsch. Here is the PR for the same. https://github.com/fern-api/fern/pull/1792

dsinghvi:
Closing since this is no longer a bug

PR #1792, opened 2023-06-21
Issue #1655: Improving error message for incorrect organization name

shashank40:


dsinghvi:
@shashank40 can you try initializing fern with a bad org name? I suspect that it all works fine now?

shashank40:
@dsinghvi I suppose issue no #1655 talks about improving the error message. So It was just an error string addition. Now adding it to business logic can be done separately as a new issue if you'd like me to do that.

Also if the old logic for organization name length 0 works, then this change would work as well, but if this was not working earlier too(as I see it was not working earlier as well, because it made a fern folder for empty organization too), then it needs to be added to business logic as well. I can do that as well if we are done with this issue first

You can let me know which route is better

PR #1803, opened 2023-06-25
Use htmlDescription instead of rendering markdown

zachkirsch:


zachkirsch:
Closing bc no noticeable perf gains

PR #1815, opened 2023-06-26
Implement subpackage endpoints overview box

kafkas:
Implements a new component that shows an overview of all the endpoints in a given subpackage

Screenshot 2023-06-26 at 9 02 59 PM

### DIFF
--- a/packages/ui/app/src/api-page/subpackages/EndpointDescriptor.tsx
+++ b/packages/ui/app/src/api-page/subpackages/EndpointDescriptor.tsx
@@ -37,13 +37,13 @@ export const EndpointDescriptor: React.FC<EndpointDescriptor.Props> = ({ endpoin
         <button className="group flex items-start justify-start space-x-2" onClick={onClick}>
             <div className="flex items-baseline space-x-2">
                 <HttpMethodIcon method={endpointDefinition.method} />
-                <MonospaceText className="text-text-default w-16 break-all text-end transition-colors group-hover:text-white">
+                <MonospaceText className="text-text-default w-16 break-all text-end leading-5 transition-colors group-hover:text-white">
                     {endpointDefinition.method}
                 </MonospaceText>
             </div>
 
             <div className="flex">
-                <MonospaceText className="text-text-default break-words text-start transition-colors group-hover:text-white">
+                <MonospaceText className="text-text-default break-words text-start leading-5 transition-colors group-hover:text-white">
                     {urlWithBreakpoints}
                 </MonospaceText>
             </div>


packages/ui/app/src/api-page/subpackages/SubpackageEndpointsOverview.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const SubpackageEndpointsOverview: React.FC<SubpackageEndpointsOverview.Props> = ({ subpackage }) => {
+    const { navigateToPath, resolvedPathFromUrl } = useDocsContext();
+
+    const handleEndpointClick = useCallback(
+        (endpointDef: FernRegistryApiRead.EndpointDefinition) => {
+            if (resolvedPathFromUrl.type !== "apiSubpackage") {
+                return;
+            }
```
zachkirsch, line 20: I like to avoid these kind of "oh no somehow this is happening" if statements where possible - an alternative is you can pass through the slug from the parent component: 

```jsx
// packages/ui/app/src/api-page/subpackages/ApiSubpackage.tsx
<SubpackageEndpointsOverview subpackage={subpackage} slug={slug} /> 
```

packages/ui/app/src/api-page/subpackages/EndpointDescriptor.tsx
```diff
@@ -0,0 +1,33 @@
...
+                <HttpMethodIcon method={endpointDefinition.method} />
+                <MonospaceText className="text-text-default w-16 break-all text-end transition-colors group-hover:text-white">
+                    {endpointDefinition.method}
+                </MonospaceText>
+            </div>
+
+            <div className="flex">
+                <MonospaceText className="text-text-default break-all text-start transition-colors group-hover:text-white">
+                    {endpointDefinition.path.parts
+                        .map((p) => (p.type === "literal" ? p.value : getPathParameterAsString(p.value)))
```
zachkirsch, line 27: for unions i prefer to use visitors over ternaries in case we add a third variant someday, e.g.
```ts
endpoint.path.parts
            .map((part) =>
                visitDiscriminatedUnion(part, "type")._visit({
                    literal: (literal) => literal.value,
                    pathParameter: (pathParameter) => getPathParameterAsString(pathParameter.value),
                    _other: () => "",
                })
            )
            .join("")
```

then if we add a third variant and you upgrade, you'll get a compile break that you're not handling the new variant

packages/ui/app/src/api-page/subpackages/SubpackageEndpointsOverview.tsx
kafkas: You're right, that's much better. Pushed the change.

PR #1824, opened 2023-06-28
Implement search in docs

kafkas:
Currently looks something like this.

Screenshot 2023-06-28 at 5 18 01 PM

Mobile view currently has some problems due to the sidebar issues that we're already tracking so will probably hide the search bar in mobile view for now.

Screenshot 2023-06-28 at 5 20 22 PM

### DIFF
--- a/packages/ui/app/src/services/useSearchService.ts
+++ b/packages/ui/app/src/services/useSearchService.ts
@@ -11,7 +11,7 @@ export type SearchService =
           isAvailable: false;
       };
 
-if (!process.env.NEXT_PUBLIC_ALGOLIA_APP_ID || !process.env.NEXT_PUBLIC_ALGOLIA_API_KEY) {
+if (process.env.NEXT_PUBLIC_ALGOLIA_APP_ID == null || process.env.NEXT_PUBLIC_ALGOLIA_API_KEY == null) {
     // TODO: Move this validation elsewhere
     throw new Error("Missing Algolia variables.");
 }
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -6968,7 +6968,6 @@ function $$SETUP_STATE(hydrateRuntimeState, basePath) {
             ["@types/node", "npm:18.7.18"],\
             ["@types/react", "npm:18.0.20"],\
             ["@types/react-syntax-highlighter", "npm:15.5.6"],\
-            ["@types/ua-parser-js", "npm:0.7.36"],\
             ["algoliasearch", "npm:4.18.0"],\
             ["classnames", "npm:2.3.1"],\
             ["depcheck", "npm:1.4.3"],\
@@ -6994,7 +6993,6 @@ function $$SETUP_STATE(hydrateRuntimeState, basePath) {
             ["remark-gfm", "npm:3.0.1"],\
             ["stylelint", "npm:14.14.0"],\
             ["typescript", "patch:typescript@npm%3A4.6.4#~builtin<compat/typescript>::version=4.6.4&hash=5d3a66"],\
-            ["ua-parser-js", "npm:1.0.35"],\
             ["vite", "virtual:5859ca6455b1300e5c14a40a0e9dfd22c3c72703e63a89ae86f290b9c13d7e92836c0c40bb85ba0f8a308601ec19634c366ca3f4bec5fed1a79eee5fbf98d0ae#npm:4.3.8"]\
           ],\
           "linkType": "SOFT"\
--- a/packages/commons/core-utils/src/index.ts
+++ b/packages/commons/core-utils/src/index.ts
@@ -15,4 +15,3 @@ export { values, type Values } from "./objects/values";
 export { PLATFORM } from "./platform";
 export type { Digit, Letter, LowercaseLetter, UppercaseLetter } from "./types";
 export { visitDiscriminatedUnion } from "./visitDiscriminatedUnion";
-
--- a/packages/ui/app/package.json
+++ b/packages/ui/app/package.json
@@ -57,8 +57,7 @@
     "react-markdown": "^8.0.5",
     "react-syntax-highlighter": "^15.5.0",
     "rehype-raw": "^6.1.1",
-    "remark-gfm": "^3.0.1",
-    "ua-parser-js": "^1.0.35"
+    "remark-gfm": "^3.0.1"
   },
   "devDependencies": {
     "@babel/core": "^7.19.1",
@@ -70,7 +69,6 @@
     "@types/node": "^18.7.18",
     "@types/react": "^18.0.20",
     "@types/react-syntax-highlighter": "^15.5.6",
-    "@types/ua-parser-js": "^0.7.36",
     "depcheck": "^1.4.3",
     "eslint": "^8.26.0",
     "jest": "^29.4.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3880,7 +3880,6 @@ __metadata:
     "@types/node": ^18.7.18
     "@types/react": ^18.0.20
     "@types/react-syntax-highlighter": ^15.5.6
-    "@types/ua-parser-js": ^0.7.36
     algoliasearch: ^4.18.0
     classnames: ^2.3.1
     depcheck: ^1.4.3
@@ -3906,7 +3905,6 @@ __metadata:
     remark-gfm: ^3.0.1
     stylelint: ^14.14.0
     typescript: 4.6.4
-    ua-parser-js: ^1.0.35
     vite: ^4.3.8
   languageName: unknown
   linkType: soft


packages/ui/app/src/docs-context/DocsContextProvider.tsx
```diff
@@ -138,6 +151,9 @@ export const DocsContextProvider: React.FC<DocsContextProvider.Props> = ({
             resolvedUrlPath,
             scrollToPathListeners.registerListener,
             selectedSlug,
+            isSearchDialogOpen,
```
zachkirsch, line 154: Not a huge deal but I would just make a new context for the search dialog, this context was intended to be for docs content/navigation. For maintainability I generally prefer narrower contexts vs a single-catch all contexts.

kafkas: Gotcha. Will pull it into a separate context.

packages/ui/app/src/App.tsx
```diff
@@ -38,7 +39,9 @@ export const App: React.FC<App.Props> = ({ docs, resolvedUrlPath }) => {
                     <Context>{children}</Context>
                 ),
                 <DocsContextProvider docsDefinition={docs.definition} resolvedUrlPath={resolvedUrlPath}>
-                    <Docs />
+                    <SearchContextProvider>
+                        <Docs />
+                    </SearchContextProvider>
```
zachkirsch, line 44: dope thanks. can also put in [contexts.ts](https://github.com/fern-api/fern/blob/main/packages/ui/app/src/contexts.tsx) since it doesn't need state

```tsx
// contexts.ts
export const CONTEXTS = [
    ({ children }: PropsWithChildren): JSX.Element => <ThemeProvider theme="dark">{children}</ThemeProvider>,
    ({ children }: PropsWithChildren): JSX.Element => <SearchContextProvider>{children}</SearchContextProvider>,
];
```

nbd but a nice pattern as contexts lists grows

packages/ui/app/src/hooks/useKeyboardCommand.ts
```diff
@@ -0,0 +1,27 @@
+import { type Digit, type UppercaseLetter } from "@fern-api/core-utils";
+import { useEffect } from "react";
+import { platform } from "../utils/platform";
+
+/**
+ * Registers a callback that fires on every keyboard command. A keyboard command is a sequence of key
+ * events where the first pressed key is cmd (on Mac) or ctrl (on Windows) and the second one is any
+ * alphanumeric.
+ */
+export function useKeyboardCommand(key: UppercaseLetter | Digit, callback: () => void | Promise<void>): void {
```
zachkirsch, line 10: this can live in react-commons

packages/ui/app/src/search-context/SearchContextProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+    const openSearchDialog = useCallback(() => {
+        setIsSearchDialogOpen(true);
+    }, []);
+
+    const closeSearchDialog = useCallback(() => {
+        setIsSearchDialogOpen(false);
+    }, []);
+
+    const contextValue = useCallback(
```
zachkirsch, line 21: super nbd but just fyi there's a nice hook in the codebase already called `useBooleanState`

```ts
const {
  value: isSearchDialogOpen,
  setTrue: openSearchDialog,
  setFalse: closeSearchDialog,
  // also has toggleValue
} = useBooleanState(false);
```

packages/ui/app/src/services/useSearchService.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!algoliaSearchIndex) {
+        return { isAvailable: false };
+    }
+
+    if (!process.env.NEXT_PUBLIC_ALGOLIA_APP_ID || !process.env.NEXT_PUBLIC_ALGOLIA_API_KEY) {
+        // TODO: Move this validation elsewhere
+        throw new Error("Missing Algolia variables.");
+    }
+
+    const client = algolia(process.env.NEXT_PUBLIC_ALGOLIA_APP_ID, process.env.NEXT_PUBLIC_ALGOLIA_API_KEY);
```
zachkirsch, line 27: this is gonna run on every render, can we bring it outside the hook?

```ts
if (!process.env.NEXT_PUBLIC_ALGOLIA_APP_ID || !process.env.NEXT_PUBLIC_ALGOLIA_API_KEY) {
  // TODO: Move this validation elsewhere
  throw new Error("Missing Algolia variables.");
}

const client = algolia(process.env.NEXT_PUBLIC_ALGOLIA_APP_ID, process.env.NEXT_PUBLIC_ALGOLIA_API_KEY);

export function useSearchService(): SearchService {
  ...
}
```

packages/ui/app/src/search/SearchHits.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const EmptyStateView: React.FC<PropsWithChildren> = ({ children }) => {
+    return <div className="text-text-default justify flex h-24 w-full flex-col items-center py-3">{children}</div>;
+};
+
+export const SearchHits: React.FC = () => {
+    const { hits } = useInfiniteHits<SearchRecord>();
+    const search = useInstantSearch();
+    const [progress, setProgress] = useState<Progress>("nil");
+
+    useEffect(() => {
```
zachkirsch, line 19: this is going to introduce an extra render every time `search` changes:
`search` changes -> render -> this effect runs, which calls `setProgress` -> extra render

I would instead use `useMemo`

```ts
const search = useInstantSearch();
const progress = useMemo((): Progress => {
  switch (search.status) {
    case "idle":
      return "success"
    case "stalled":
    case "loading":
      return "pending";
    case "error":
      return "error";
  }
}, [search])
```

then you can also take out the `"nil"` dummy state from `Progress`.

relevant: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

packages/ui/app/src/sidebar/Sidebar.tsx
```diff
@@ -14,12 +17,17 @@ export declare namespace Sidebar {
...
+    const { openSearchDialog } = useSearchContext();
+    const searchService = useSearchService();
 
     const contextValue = useCallback((): SidebarContextValue => ({ expandAllSections }), [expandAllSections]);
 
     return (
         <SidebarContext.Provider value={contextValue}>
             <div className="border-border bg-background flex min-w-0 flex-1 flex-col justify-between overflow-hidden border-r">
+                <div className="z-10 flex flex-col px-2.5 pt-2.5 shadow-[0_-5px_10px_10px_rgba(18,20,24,1)]">
+                    {searchService.isAvailable && <SidebarSearchBar onClick={openSearchDialog} />}
```
zachkirsch, line 29: is this gonna jump when the search service becomes available?

packages/ui/app/src/services/useSearchService.ts
kafkas: Right, I should have used useMemo here. But sure, can take it out of the function.

packages/ui/app/src/sidebar/SidebarSearchBar.tsx
```diff
@@ -0,0 +1,34 @@
...
+                    {platform === "mac" ? "CMD+K" : "CTRL+K"}
+                </div>
+            )}
+        </button>
+    );
+};
+
+export const SidebarSearchBar = dynamic(() => Promise.resolve(SidebarSearchBarCore), {
+    ssr: false,
+});
```
zachkirsch, line 34: cool! what does this do?

packages/ui/app/src/services/useSearchService.ts
kafkas: Right, I should have used useMemo here. But sure, can take it out of the function.

packages/ui/app/src/types.ts
```diff
@@ -0,0 +1,4 @@
+import type * as FernRegistryDocsRead from "@fern-fern/registry-browser/api/resources/docs/resources/v1/resources/read";
+import { Hit as AlgoliaHit } from "instantsearch.js";
+
+export type SearchRecord = AlgoliaHit<FernRegistryDocsRead.AlgoliaRecord & Record<string, unknown>>;
```
zachkirsch, line 4: nit I would put this file in `src/search/`

packages/ui/app/src/utils/platform.ts
```diff
@@ -0,0 +1,14 @@
+import { UAParser } from "ua-parser-js";
+
+const uaParser = new UAParser();
+
+type Platform = "mac" | "windows" | "other";
+
+export const platform = (function getPlatform(): Platform {
```
zachkirsch, line 7: nit can we do `PLATFORM`

packages/ui/app/src/search/SearchHits.tsx
kafkas: Actually, this was intended. I specifically wanted a `nil` state in order to be able to differentiate it from the `success` state. The goal was to prevent the "No results" flash which we were showing `if (idle && results.length == 0)`. Ideally, we should show it only `if (success && results.length == 0)`.

So I don't think a derivative state solves this issue but there may be a better solution that I'm not yet aware of.

packages/ui/app/src/services/useSearchService.ts
zachkirsch: dope sg. benefit of outside the hook is that it's one shared instance for everyone. With `useMemo` every component that calls `useSearchService` will create its own instance

packages/ui/app/src/sidebar/SidebarSearchBar.tsx
kafkas: It disables ssr for the component. We need to disable it as the content within it relies on platform information (cmd+K on mac vs ctrl+K on windows) which we're retrieving from `User-Agent` with `ua-parser-js` and that information doesn't exist server-side.

PR #1840, opened 2023-07-05
Load algolia env vars from circleci context

kafkas:


dsinghvi (approved):

kafkas:
> @kafkas can you run `yarn test:ete -u` locally, there is a test with some non deterministic behavior and it needs to be run again

There are 3 that fail for me.

PR #1844, opened 2023-07-05
Support 'text' in response-stream

zachkirsch:


dsinghvi:
this should work, i thought we were just going to introduce a new primitive `text` though?

zachkirsch:
> this should work, i thought we were just going to introduce a new primitive text though?

I don't think so, this is more saying "there's something coming over the wire that's not application/json, so don't do JSON.parse". it's not like I can do `myProperty: text` or `myProperty: list<text>` (you can just use `string` for that)

dsinghvi:
> > this should work, i thought we were just going to introduce a new primitive text though?
> 
> I don't think so, this is more saying "there's something coming over the wire that's not application/json, so don't do JSON.parse". it's not like I can do `myProperty: text` or `myProperty: list<text>` (you can just use `string` for that)

Thought it was just going to be like `file`

PR #1847, opened 2023-07-05
[openapi] Move common environment suffix from servers to endpoints

zachkirsch:


packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
```diff
@@ -64,7 +64,7 @@ export function convertEndpoint({
     }
 
     const convertedEndpoint: RawSchemas.HttpEndpointSchema = {
-        path: endpoint.path,
+        path: environments?.type === "single" ? `${environments.endpointPathPrefix}${endpoint.path}` : endpoint.path,
```
dsinghvi, line 67: do the urls need to be joined safely here? dont want a double slash somewhere

zachkirsch: Yeah I trim the trailing slash from the `endpointPathPrefix` https://github.com/fern-api/fern/blob/02169392c4dfd70a3ff50476425c8f6abb75f643/packages/cli/openapi-ir-to-fern/src/getEnvironments.ts#L88-L89

PR #1859, opened 2023-07-06
Add support for logo href

kafkas:
Adds support for header homepage link via logo href.

### DIFF
--- a/packages/ui/app/src/docs/Header.tsx
+++ b/packages/ui/app/src/docs/Header.tsx
@@ -6,8 +6,8 @@ export const Header: React.FC = () => {
     const { resolveFile, docsDefinition } = useDocsContext();
     const { logo, logoHref, navbarLinks } = docsDefinition.config;
 
-    const hasLogo = !!logo;
-    const hasLogoHref = !!logoHref;
+    const hasLogo = logo != null;
+    const hasLogoHref = logoHref != null;
 
     const logoContent = hasLogo && <img className="max-h-5" src={resolveFile(logo)} />;
     const logoSection = hasLogo && (


packages/cli/config-management/docs-configuration/src/convertDocsConfiguration.ts
```diff
@@ -19,20 +19,17 @@ export async function convertDocsConfiguration({
     absolutePathOfConfiguration: AbsoluteFilePath;
     context: TaskContext;
 }): Promise<DocsConfiguration> {
+    const { navigation, colors, favicon, logo, navbarLinks, title } = rawDocsConfiguration;
     return {
-        navigation: convertNavigationConfiguration(rawDocsConfiguration.navigation, absolutePathOfConfiguration),
-        title: rawDocsConfiguration.title,
+        navigation: convertNavigationConfiguration(navigation, absolutePathOfConfiguration),
+        title,
         logo:
```
zachkirsch, line 26: Looks like we're dropping the href? I think you'll need to edit the [`DocsConfiguration` type](https://github.com/fern-api/fern/blob/main/packages/cli/config-management/docs-configuration/src/DocsConfiguration.ts) and then update `publishDocs` to send the href to fdr

packages/cli/config-management/docs-configuration/src/DocsConfiguration.ts
```diff
@@ -1,11 +1,12 @@
 import { Audiences } from "@fern-api/config-management-commons";
 import { AbsoluteFilePath } from "@fern-api/fs-utils";
-import { FernRegistry } from "@fern-fern/registry-node";
+import type { FernRegistry } from "@fern-fern/registry-node";
 
 export interface DocsConfiguration {
     navigation: DocsNavigationConfiguration;
     title: string | undefined;
     logo: ImageReference | undefined;
+    logoHref: FernRegistry.docs.v1.write.Url | undefined;
```
zachkirsch, line 9: The way this is modeled it's possible to have an invalid state:

```ts
logo: undefined,
logoHref: "https://logo.com/image.png"
```

^ this is invalid because you can't have an href without a logo image.

you can do:

```ts
logo: Logo | undefined;

interface Logo {
  reference: ImageReference;
  href: FernRegistry.docs.v1.write.Url | undefined;
}
```

ideally we would do the same thing in fdr but nbd

packages/ui/app/src/docs/Header.tsx
```diff
@@ -4,6 +4,31 @@ import { HeaderButton } from "./HeaderButton";
 
 export const Header: React.FC = () => {
     const { resolveFile, docsDefinition } = useDocsContext();
+    const { logo, logoHref, navbarLinks } = docsDefinition.config;
+
+    const hasLogo = !!logo;
+    const hasLogoHref = !!logoHref;
```
zachkirsch, line 10: nit but let's just always do `!= null` as this can hide some weird bugs. e.g. `""` is coerced to `false` when doing `!!`

zachkirsch: Yeah in general I've been trying to fail fast - e.g. if we're storing urls as `""` that means something is probably wrong upstream and don't wanna swallow it.

zachkirsch:
lgtm!

packages/cli/config-management/docs-configuration/src/DocsConfiguration.ts
dsinghvi: yeah i said not to cause it would make it a much heavier backend change (would need to introduce logoV2 and convert)

zachkirsch: yeah figured

PR #1832, opened 2023-06-29
Update the README

aevitas:
Most notable departure is the positioning of Fern as a toolset that doesn't compete with OpenAPI, but works alongside it and provides strong code generation features.

PR #1874, opened 2023-07-07
README: Command and option tables

aevitas:
Continuation of #1832


aevitas:
@dsinghvi PTAL - adjusted based on your feedback

PR #1887, opened 2023-07-12
feat: customize the documentation font families

codebender828:
- feat: update to stable yarn version and add types for typography configuration
- feat: update to stable yarn version and add types for typography configuration

### DIFF
--- a/packages/ui/app/src/search/SearchHit.tsx
+++ b/packages/ui/app/src/search/SearchHit.tsx
@@ -1,5 +1,6 @@
 import { Icon } from "@blueprintjs/core";
 import { visitDiscriminatedUnion } from "@fern-api/core-utils";
+import Link from "next/link";
 import { Snippet } from "react-instantsearch-hooks-web";
 import { useDocsContext } from "../docs-context/useDocsContext";
 import { useSearchContext } from "../search-context/useSearchContext";
@@ -16,12 +17,13 @@ export const SearchHit: React.FC<SearchHit.Props> = ({ hit }) => {
     const { closeSearchDialog } = useSearchContext();
 
     return (
-        <button
+        <Link
             className="hover:bg-accentHighlight group flex w-full items-center space-x-3 space-y-1 rounded-md p-2"
             onClick={() => {
                 closeSearchDialog();
                 navigateToPath(hit.path);
             }}
+            href={`/${hit.path}`}
         >
             <Icon
                 className="text-text-default group-hover:text-text-stark"
@@ -56,6 +58,6 @@ export const SearchHit: React.FC<SearchHit.Props> = ({ hit }) => {
                     />
                 </div>
             </div>
-        </button>
+        </Link>
     );
 };
--- a/packages/ui/fe-bundle/src/pages/api/revalidate.ts
+++ b/packages/ui/fe-bundle/src/pages/api/revalidate.ts
@@ -1,0 +1,63 @@
+import { NextApiHandler, NextApiResponse } from "next";
+import { REGISTRY_SERVICE } from "../../service";
+import { UrlSlugTree } from "../../url-path-resolver/UrlSlugTree";
+
+export interface Request {
+    url: string;
+}
+
+const handler: NextApiHandler = async (req, res) => {
+    const url = req.body?.url;
+
+    if (url == null) {
+        return res.status(400).send("Property 'url' is missing from request.");
+    }
+    if (typeof url !== "string") {
+        return res.status(400).send("Property 'url' is not a string.");
+    }
+
+    // when we call res.revalidate() nextjs uses
+    // req.headers.host to make the network request
+    if (typeof req.headers["x-fern-host"] === "string") {
+        req.headers.host = req.headers["x-fern-host"];
+    }
+
+    const docs = await REGISTRY_SERVICE.docs.v2.read.getDocsForUrl({
+        url,
+    });
+    if (!docs.ok) {
+        // eslint-disable-next-line no-console
+        console.error("Failed to fetch docs", docs.error);
+        return res.status(500).send("Failed to load docs for: " + url);
+    }
+    const urlSlugTree = new UrlSlugTree(docs.body.definition);
+    const paths = ["/", ...urlSlugTree.getAllSlugs().map((slug) => `/${slug}`)];
+
+    const revalidated: string[] = [];
+    const failures: string[] = [];
+    await Promise.all(
+        paths.map(async (path) => {
+            const didSucceed = await tryRevalidate(res, path);
+            if (didSucceed) {
+                revalidated.push(path);
+            } else {
+                failures.push(path);
+            }
+        })
+    );
+
+    return res.json({ revalidated, failures });
+};
+
+async function tryRevalidate(res: NextApiResponse, path: string): Promise<boolean> {
+    try {
+        await res.revalidate(path);
+        return true;
+    } catch (err) {
+        // eslint-disable-next-line no-console
+        console.error(err);
+        return false;
+    }
+}
+
+export default handler;
--- a/packages/ui/fe-bundle/src/url-path-resolver/UrlSlugTree.ts
+++ b/packages/ui/fe-bundle/src/url-path-resolver/UrlSlugTree.ts
@@ -59,6 +59,37 @@ export class UrlSlugTree {
         return neighbors;
     }
 
+    private allSlugs: string[] | undefined;
+    public getAllSlugs(): string[] {
+        if (this.allSlugs == null) {
+            this.allSlugs = this.getAllSlugsExpensive();
+        }
+
+        return this.allSlugs;
+    }
+
+    private getAllSlugsExpensive(ancestors: string[] = [], root = this.root): string[] {
+        const allSlugs: string[] = [];
+        for (const node of Object.values(root)) {
+            allSlugs.push(...ancestors, node.slug);
+            switch (node.type) {
+                case "api":
+                case "section":
+                case "apiSubpackage":
+                    allSlugs.push(...this.getAllSlugsExpensive([...ancestors, node.slug], node.children));
+                    break;
+                case "clientLibraries":
+                case "endpoint":
+                case "page":
+                case "topLevelEndpoint":
+                    break;
+                default:
+                    assertNever(node);
+            }
+        }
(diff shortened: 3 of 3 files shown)


fern/docs-config/definition/docs.yml
```diff
@@ -28,6 +29,17 @@ types:
...
 
+  DocsTypographyConfig:
+    properties:
+      headingsFont: optional<FontConfig>
+      bodyFont: optional<FontConfig>
+      codeFont: optional<FontConfig>
+
+  FontConfig:
+    properties:
+      name: string
```
zachkirsch, line 40: Do we need the user to supply a name? Or is path ok?

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -218,9 +218,54 @@ async function convertDocsConfiguration({
...
+    typographyConfiguration?: TypographyConfig;
+    docsDefinition: DocsDefinition;
+    uploadUrls: Record<FernRegistry.docs.v1.write.FilePath, FernRegistry.docs.v1.write.FileS3UploadUrl>;
+    context: TaskContext;
+}): FernRegistry.docs.v1.write.DocsTypographyConfig | undefined {
+    if (typographyConfiguration == null) {
+        return;
+    }
+    const result: FernRegistry.docs.v1.write.DocsTypographyConfig = {};
+
```
zachkirsch, line 245: This makes me a little nervous, eg if we add another key to the typography config or there happens to be an extra key at runtime. Can we explicitly convert the three fonts?

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -317,6 +362,16 @@ function getFilepathsToUpload(docsDefinition: DocsDefinition): AbsoluteFilePath[
         filepaths.push(docsDefinition.config.favicon.filepath);
     }
 
+    const typographyConfiguration = docsDefinition.config.typography;
+    if (typographyConfiguration) {
+        for (const key in typographyConfiguration) {
```
zachkirsch, line 367: Same here

packages/cli/config-management/docs-configuration/src/DocsConfiguration.ts
```diff
@@ -9,13 +9,37 @@ export interface DocsConfiguration {
...
 }
 
 export interface Logo {
     reference: ImageReference;
     href: FernRegistry.docs.v1.write.Url | undefined;
 }
 
+export interface FontConfig {
+    name: string;
+    path: string;
```
dsinghvi, line 22: since we don't use this path, I could see us deleting it

zachkirsch: We don’t use it? How does the user supply a font

fern/docs-config/definition/docs.yml
codebender828: The font name is used to declare the name of the font in the eventually generated CSS. We could assign IDs, but it may not be helpful for debugging the front end.

zachkirsch: Maybe we can just default to the filename or something like that? Always nice to minimize config

packages/cli/config-management/docs-configuration/src/DocsConfiguration.ts
codebender828: I believe we do use it to resolve the actually font source file. @dsinghvi

zachkirsch: Can we change it from string to AbsoluteFilePath? This is something we introduced because there were a bunch of bugs with us mixing up relative and absolute filepaths

dsinghvi: there is one on line 23 (right below this line for `AbsoluteFilePath`) which is why i was saying we don't/shouldn't use it

zachkirsch: Ahhh lol

dsinghvi: @codebender828 basically at `loadWorkspace` time we should convert the `path` that the user gave us into an `AbsoluteFilePath` that can later be used by `publishDocs.ts`

fern/docs-config/definition/docs.yml
codebender828: I see. That's also possible. 👍🏽

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
codebender828: 👍🏽

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
codebender828: 👍🏽

packages/cli/config-management/docs-configuration/src/DocsConfiguration.ts
codebender828: After thinking about this, I came up with a solution: I have decided to make the `name` property optional, and the `path` will be required. If no `name` is not supplied, we shall default it to use a randomly generated `name` that will be used to reference the font in the eventually injected CSS. My options for consideration would be the `absoluteFilePath`, or a `randomId` like the `FileId`.

It would be useful in case, the developer wants to inspect what the actual name of the font being used in the published documentation is. I also think it will set us up in the future for better customization in the future as the font name will be used as a token value. So I suggest we keep the `name` option. But it does not need to be explicitly supplied for the docs definition to work. So we keep the definition thin, but provide a bit more visibility if they want to provide a font name. (Will be more useful later on during the local dev setup.)

dsinghvi: that sounds good -- can you add this in the docs for the fern definition `docs-config` ?

codebender828: Done!

zachkirsch: We can delete `path` right? We just need `absolutePath`

packages/cli/config-management/docs-configuration/src/DocsConfiguration.ts
```diff
@@ -9,13 +9,37 @@ export interface DocsConfiguration {
...
 }
 
 export interface Logo {
     reference: ImageReference;
     href: FernRegistry.docs.v1.write.Url | undefined;
 }
 
+export interface FontConfig {
+    path: string;
+    name?: string;
```
zachkirsch, line 22: Nit but for these configs I always do `name: string | undefined` rather than `name?: string` to ensure we don't forget it in the conversion. same for inside the `TypographyConfig`

fern/docs-config/definition/docs.yml
```diff
@@ -28,6 +29,22 @@ types:
...
+
+  FontConfig:
+    properties:
+      path:
+        type: string
+        docs: The relative path of the font file.
+      name:
+        type: optional<string>
+        docs: |
+          If the `name` is not supplied, Fern will default it to a randomly generated name that will be used to reference your custom font in the eventually injected CSS in the docs.
```
zachkirsch, line 46: nit: delete `randomly`

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
zachkirsch: thank you!!

fern/docs-config/definition/docs.yml
codebender828: Gotcha! Done.

packages/cli/config-management/docs-configuration/src/DocsConfiguration.ts
codebender828: Gotcha! Done!

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -218,9 +218,79 @@ async function convertDocsConfiguration({
...
+    docsDefinition: DocsDefinition;
+    uploadUrls: Record<FernRegistry.docs.v1.write.FilePath, FernRegistry.docs.v1.write.FileS3UploadUrl>;
+    context: TaskContext;
+}): FernRegistry.docs.v1.write.DocsTypographyConfig | undefined {
+    if (typographyConfiguration == null) {
+        return;
+    }
+    const result: FernRegistry.docs.v1.write.DocsTypographyConfig = {};
+
+    if (typographyConfiguration.headingsFont != null) {
```
zachkirsch, line 246: Not a huge deal but I could see us simplifying this function to just something like:

```ts
    if (typographyConfiguration == null) {
        return undefined;
    }
    return {
        headingsFont: convertFont({
            font: typographyConfiguration.headingsFont,
            docsDefinition,
            context,
            uploadUrls,
            label: "heading"
        }),
        bodyFont: convertFont({
            font: typographyConfiguration.bodyFont,
            docsDefinition,
            context,
            uploadUrls,
            label: "body"
        }),
        codeFont: convertFont({
            font: typographyConfiguration.codeFont,
            docsDefinition,
            context,
            uploadUrls,
            label: "code"
        }),
    }
```

where convert font is something like

```ts
function convertFont(...) {
    if (font == null) {
        return undefined
    }

    const filepath = convertAbsoluteFilepathToFdrFilepath(
        font.absolutePath,
        docsDefinition
    );
    const file = uploadUrls[filepath];
    if (file == null) {
        return context.failAndThrow(`Failed to locate ${label} file after uploading`);
    }
    result.headingsFont = {
        name: font.name ?? `font:${label}:${file.fileId}`,
        fontFile: file.fileId,
    };    
}
```

jbakebwa: Thanks! This was definitely a lot cleaner. Added changes

PR #1889, opened 2023-07-12
Implement anchor links for atributes

kafkas:
This PR brings support for property-level anchor linking in API documentation. Request and response properties, query, and path parameters are all now linkable.

<p align="center">
Screenshot 2023-07-12 at 1 55 03 PM
</p>

### DIFF
--- a/release-blockers.yml
+++ b/release-blockers.yml
@@ -1,2 +1,3 @@
 # if there are any release blockers in this list, releasing will fail on Circle
-release-blockers: []
+release-blockers:
+  - Anchor computing callbacks need to be made required and added to an anchor context.


packages/ui/app/src/api-page/types/object/ObjectProperty.tsx
```diff
@@ -1,7 +1,8 @@
-import * as FernRegistryApiRead from "@fern-fern/registry-browser/api/resources/api/resources/v1/resources/read";
+import type * as FernRegistryApiRead from "@fern-fern/registry-browser/api/resources/api/resources/v1/resources/read";
```
zachkirsch, line 1: why this change?

packages/ui/app/src/api-page/types/object/ObjectProperty.tsx
```diff
@@ -16,10 +17,11 @@ import { TypeShorthand } from "../type-shorthand/TypeShorthand";
 export declare namespace ObjectProperty {
     export interface Props {
         property: FernRegistryApiRead.ObjectProperty;
+        anchor?: string;
```
zachkirsch, line 20: why optional?

packages/ui/app/src/api-page/types/type-definition/InternalTypeDefinition.tsx
```diff
@@ -22,6 +22,7 @@ export declare namespace InternalTypeDefinition {
     export interface Props {
         typeShape: FernRegistryApiRead.TypeShape;
         isCollapsible: boolean;
+        getPropertyAnchor?: (path: FernRegistryApiRead.ObjectProperty) => string;
```
zachkirsch, line 25: why optional?

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
```diff
@@ -109,16 +139,23 @@ export const EndpointContent = React.memo<EndpointContent.Props>(function Endpoi
                     <div className="mt-8 flex">
                         <div className="flex flex-1 flex-col gap-12">
                             {endpoint.path.pathParameters.length > 0 && (
-                                <PathParametersSection pathParameters={endpoint.path.pathParameters} />
+                                <PathParametersSection
+                                    pathParameters={endpoint.path.pathParameters}
+                                    getParameterAnchor={(param) => computeAnchor("path", param)}
```
zachkirsch, line 144: no need to change but I think these callbacks would've been a good candidate for a context because they get passed around a ton and don't need to be modified (I think?) between here and when they get invoked

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
```diff
@@ -44,6 +54,26 @@ export const EndpointContent = React.memo<EndpointContent.Props>(function Endpoi
...
+            attributeType: "request" | "response" | "path" | "query",
+            attribute:
+                | FernRegistryApiRead.ObjectProperty
+                | FernRegistryApiRead.PathParameter
+                | FernRegistryApiRead.QueryParameter
+        ) => {
+            let anchor = "";
+            if (isSubpackage(package_)) {
+                anchor += snakeCase(package_.urlSlug) + "_";
+            }
```
zachkirsch, line 68: not a big fan of type guards because they're very un-type-safe and can cause bugs down the line when the APIs change.

Also, this will only prepend the urlSlug of the parent subpackage. If the endpoint is nested several subpackages deep, do we need to prepend all the subpackages to the anchor?

Maybe we can just pass in `parentSlug` from the `ApiPackageContents` and use that:

```tsx
<Endpoint
  key={endpoint.id}
  endpoint={endpoint}
  parentSlug={slug}
  slug={joinUrlSlugs(slug, endpoint.urlSlug)}
/>
```

packages/ui/app/src/api-page/types/object/ObjectProperty.tsx
kafkas: When an import is used only for type annotations, I use type-only import since it gets stripped out and is guaranteed to have no effect on runtime.

packages/ui/app/src/api-page/types/object/ObjectProperty.tsx
kafkas: Since the anchor element is absolutely positioned and falls a little outside of the component itself I thought we may want to support a version that doesn't have the anchor. But it looks like the component is currently used in one place only so I probably can make it required. Shall I?

packages/ui/app/src/api-page/types/type-definition/InternalTypeDefinition.tsx
kafkas: Same reason as above.

packages/ui/app/src/api-page/types/object/ObjectProperty.tsx
zachkirsch: Yeah I would also default to required to make sure we're not missing anywhere we're supposed to be passing in. Any usually I'll do `anchor: string | undefined` for the same reason

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
kafkas: Good catch. Ok, I'll see how we can do it with `parentSlug`. For the nested ones, yes we'll probably need to prepend since we need to ensure uniqueness for each anchor.

PR #1891, opened 2023-07-12
Add Fern contributors to readme

aevitas:


zachkirsch:
Pretty sick! Would be awesome if we could show the contributors to the generators as well here

aevitas:
> Pretty sick! Would be awesome if we could show the contributors to the generators as well here

That'd be great, but we'd end up with some duplicate faces. I'll look into it!

dsinghvi:
Filed https://github.com/fern-api/fern/issues/1903 to track the follow up

PR #1906, opened 2023-07-14
Register errors in cli

kafkas:
With this change, we will include error declarations when converting ir endpoints to fdr endpoint definitions.


packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
```diff
@@ -188,6 +191,21 @@ function convertResponse(irResponse: Ir.http.HttpResponse): FernRegistry.api.v1.
     }
 }
 
+function convertResponseError(
+    irResponseError: Ir.http.ResponseError,
+    ir: Ir.ir.IntermediateRepresentation
+): FernRegistry.api.v1.register.ErrorDeclaration | null {
+    const errorDeclaration = ir.errors[irResponseError.error.errorId];
+    // TODO: The errorDeclaration.type == null check here may not be correct. Confirm before merging.
```
kafkas, line 199: We'll want to confirm this before merging the PR.

PR #1918, opened 2023-07-17
Add `fdr` cli command

kafkas:
Adds a new `fdr` command to the cli, which writes the FDR Register API definition associated with a given API to the specified path in the filesystem.

### DIFF
--- a/packages/cli/ete-tests/src/tests/fdr/fdr.test.ts
+++ b/packages/cli/ete-tests/src/tests/fdr/fdr.test.ts
@@ -1,6 +1,6 @@
 import { AbsoluteFilePath, join, RelativeFilePath } from "@fern-api/fs-utils";
 import { GenerationLanguage } from "@fern-api/generators-configuration";
-import { generateFdrAsString } from "./generateFdrAsString";
+import { generateFdrApiDefinitionAsString } from "./generateFdrApiDefinitionAsString";
 
 const FIXTURES_DIR = join(AbsoluteFilePath.of(__dirname), RelativeFilePath.of("fixtures"));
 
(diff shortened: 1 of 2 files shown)


packages/cli/cli/src/cli.ts
```diff
@@ -18,6 +18,7 @@ import { CliContext } from "./cli-context/CliContext";
 import { getLatestVersionOfCli } from "./cli-context/upgrade-utils/getLatestVersionOfCli";
 import { addGeneratorToWorkspaces } from "./commands/add-generator/addGeneratorToWorkspaces";
 import { formatWorkspaces } from "./commands/format/formatWorkspaces";
+import { generateFdrForWorkspaces } from "./commands/generate-fdr/generateFdrForWorkspaces";
```
dsinghvi, line 21: @kafkas i think lets call the method `generateFdrApiDefinitionForWorkspaces`. That's the most readable and makes it easy for someone to understand what is being generated

packages/cli/cli/src/cli.ts
```diff
@@ -328,6 +330,46 @@ function addIrCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
...
+                    demandOption: true,
+                })
+                .option("api", {
+                    string: true,
+                    description: "Only run the command on the provided API",
+                })
+                .option("language", {
+                    choices: Object.values(GenerationLanguage),
+                    description: "Generate FDR API definition for a particular language",
+                })
```
dsinghvi, line 351: The `language` should not matter for the FDR API Definition, so we should be able to remove it. 

This matters for IR because the IR generated for each language is different bc the IR handles escaping reserved keywords for each language, which is configured [here](https://github.com/fern-api/fern/blob/main/packages/cli/generation/ir-generator/src/casings/reserved.ts)

packages/cli/ete-tests/src/tests/fdr/generateFdrAsString.ts
```diff
@@ -0,0 +1,45 @@
+import { AbsoluteFilePath, join, RelativeFilePath } from "@fern-api/fs-utils";
+import { GenerationLanguage } from "@fern-api/generators-configuration";
+import { readFile, rm } from "fs/promises";
+import { runFernCli } from "../../utils/runFernCli";
+
+export async function generateFdrAsString({
```
dsinghvi, line 6: nit: `generateFdrApiDefinitionAsString`

dsinghvi (approved):
LGTM! Will be helpful to debug for the fdr backend!

PR #1920, opened 2023-07-18
feat: inject generated fonts

codebender828:
Mounts fonts to documentation site:

CleanShot 2023-07-18 at 15 10 57@2x

### DIFF
--- a/packages/cli/config-management/generators-configuration/src/convertGeneratorsConfiguration.ts
+++ b/packages/cli/config-management/generators-configuration/src/convertGeneratorsConfiguration.ts
@@ -129,8 +129,8 @@ function convertOutputMode(generator: GeneratorInvocationSchema): FernFiddle.Out
             return FernFiddle.OutputMode.publishV2(
                 FernFiddle.remoteGen.PublishOutputModeV2.pypiOverride({
                     registryUrl: generator.output.url ?? "https://upload.pypi.org/legacy/",
-                    username: generator.output.username ?? "",
-                    password: generator.output.password ?? "",
+                    username: generator.output.token != null ? "__token__" : generator.output.password ?? "",
+                    password: generator.output.token ?? generator.output.password ?? "",
                     coordinate: generator.output["package-name"],
                 })
             );
--- a/packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
+++ b/packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
@@ -84,6 +84,7 @@ export const EndpointContent = React.memo<EndpointContent.Props>(function Endpoi
         },
         [titleHeight]
     );
+    const [selectedErrorIndex, setSelectedErrorIndex] = useState<number | null>(null);
 
     const environmentUrl = useMemo(() => getEndpointEnvironmentUrl(endpoint), [endpoint]);
 
@@ -171,7 +172,13 @@ export const EndpointContent = React.memo<EndpointContent.Props>(function Endpoi
                             )}
                             {endpoint.errors.length > 0 && (
                                 <EndpointSection title="Errors">
-                                    <EndpointErrorsSection errors={endpoint.errors} />
+                                    <EndpointErrorsSection
+                                        errors={endpoint.errors}
+                                        onClickError={(_, idx) => {
+                                            setSelectedErrorIndex(idx === selectedErrorIndex ? null : idx);
+                                        }}
+                                        selectedErrorIndex={selectedErrorIndex}
+                                    />
                                 </EndpointSection>
                             )}
                         </div>
--- a/packages/ui/app/src/api-page/endpoints/EndpointErrorsSection.tsx
+++ b/packages/ui/app/src/api-page/endpoints/EndpointErrorsSection.tsx
@@ -1,22 +1,42 @@
 import { FernRegistry } from "@fern-fern/registry-browser";
+import classNames from "classnames";
 import { TypeReferenceDefinitions } from "../types/type-reference/TypeReferenceDefinitions";
 
 export declare namespace EndpointErrorsSection {
     export interface Props {
         errors: FernRegistry.api.v1.read.ErrorDeclaration[];
+        onClickError: (e: FernRegistry.api.v1.read.ErrorDeclaration, index: number) => void;
+        selectedErrorIndex: number | null;
     }
 }
 
-export const EndpointErrorsSection: React.FC<EndpointErrorsSection.Props> = ({ errors }) => {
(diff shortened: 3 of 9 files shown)


codebender828:
I do have some CORS errors when loading the fonts on localhost. Any ideas around this?

CleanShot 2023-07-18 at 15 12 03@2x

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
```diff
@@ -8,34 +8,58 @@ import Head from "next/head";
...
+export default function Docs({
+    docs,
+    typographyStyleSheet,
+    resolvedUrlPath,
+    nextPath,
+    previousPath,
+}: Docs.Props): JSX.Element {
+    return (
+        <>
+            {typographyStyleSheet != null && <Typography stylesheet={typographyStyleSheet} />}
```
Comment (zachkirsch) [inline packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:43]: should this go in `<Head>`? Feel like that's where I usually see `<style>`.

packages/ui/fe-bundle/src/utils/theme/loadDocsTypography.ts
```diff
@@ -0,0 +1,122 @@
+import type * as FernRegistryDocsRead from "@fern-fern/registry-browser/api/resources/docs/resources/v1/resources/read";
+
+export function getFontExtension(url: URL): string {
```
zachkirsch, line 3: nbd but generally I default to not exporting things unless i need them in other files, makes it easier to change impl detail without breaking consumers

packages/ui/fe-bundle/src/utils/theme/loadDocsTypography.ts
```diff
@@ -0,0 +1,122 @@
...
+    docsDefinition: FernRegistryDocsRead.DocsDefinition,
+    fontType: FontType
+): GenerationFontConfig | undefined {
+    const typographyFontType = docsDefinition.config.typography?.[fontType];
+    if (typographyFontType == null) {
+        return undefined;
+    }
+    const fontName = typographyFontType.name;
+    const fontUrl = docsDefinition.files[typographyFontType.fontFile];
+    const fontExtension = getFontExtension(new URL(fontUrl ?? ""));
```
zachkirsch, line 30: `fontUrl ?? ""` is gonna cause some confusing issues if `fontUrl` is undefined. Can we add an explicit check + error message

packages/ui/fe-bundle/src/utils/theme/loadDocsTypography.ts
```diff
@@ -0,0 +1,122 @@
+import type * as FernRegistryDocsRead from "@fern-fern/registry-browser/api/resources/docs/resources/v1/resources/read";
+
+export function getFontExtension(url: URL): string {
+    const ext = url.pathname.split(".").pop();
+    if (ext == null) {
+        throw new Error("No extension found");
```
zachkirsch, line 6:
```suggestion
        throw new Error("No extension found for font: " + url.pathname);
```

packages/ui/fe-bundle/src/utils/theme/loadDocsTypography.ts
```diff
@@ -0,0 +1,122 @@
...
+    };
+
+    return fontConfig;
+}
+
+/**
+ * Load Documentation Typography configuration
+ * @param docsDefinition
+ * @returns
+ */
```
zachkirsch, line 46: can delete

packages/ui/fe-bundle/src/utils/theme/loadDocsTypography.ts
```diff
@@ -0,0 +1,122 @@
...
+    return ext;
+}
+
+export type FontType = "headingsFont" | "bodyFont" | "codeFont";
+
+export interface GenerationFontConfig {
+    fontType: FontType;
+    fontName: string;
+    fontExtension: string;
+    fontUrl: FernRegistryDocsRead.Url | undefined;
```
zachkirsch, line 17: Let's make this required and throw below if `docsDefinition.files[typographyFontType.fontFile]` is undefined

packages/ui/fe-bundle/src/utils/theme/loadDocsTypography.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    return fontConfig;
+}
+
+/**
+ * Load Documentation Typography configuration
+ * @param docsDefinition
+ * @returns
+ */
+export function loadDocTypography(docsDefinition: FernRegistryDocsRead.DocsDefinition): GenerationFontConfig[] {
```
zachkirsch, line 47: This API of returning `GenerationFontConfig[]` isn't ideal because it can represent invalid states (e.g. multiple headingsFont values). nbd but imo better to have an interface like this:
```ts
interface GenerationFontConfigs {
  headingsFont: GenerationFontConfig | undefined;
  bodyFont: GenerationFontConfig | undefined;
  codeFont: GenerationFontConfig | undefined;
}
```

with that approach it's impossible to represent invalid states

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
```diff
@@ -8,34 +8,58 @@ import Head from "next/head";
...
-            </Head>
-            <App
-                docs={docs}
-                resolvedUrlPath={resolvedUrlPath}
-                nextPath={nextPath ?? undefined}
-                previousPath={previousPath ?? undefined}
-            />
-        </main>
+        // eslint-disable-next-line react/no-unknown-property
+        <style id="fern-font-stylesheet" jsx global>
```
Comment (zachkirsch) [inline packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:28]: what do `jsx` and `global` do?

zachkirsch:
lgtm just some nits. SO stoked about this feature!!

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
Comment (jbakebwa) [reply packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:47]: When using `<styled jsx global>` it injects it in the `head`.  Next.js bundles [`styled-jsx`](https://github.com/vercel/styled-jsx#global-styles) inside it by default.

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
Comment (jbakebwa) [reply packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:32]: When using `<styled jsx global>`, the `global` attribute injects the stylesheet contents inside the document `head`. The `jsx` attribute is a compile-time directive that informs Next.js to process this stylesheet as a [`styled-jsx`](https://nextjs.org/docs/pages/building-your-application/styling/css-in-js) primitive

PR #1924, opened 2023-07-19
Improve endpoint errors section UI

kafkas:
With this PR we'll allow the user to select/unselect an endpoint error, which automatically expands/collapses all the properties associated with it.

Screenshot 2023-07-19 at 4 11 06 PM

### DIFF
--- a/packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
+++ b/packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
@@ -84,6 +84,7 @@ export const EndpointContent = React.memo<EndpointContent.Props>(function Endpoi
         },
         [titleHeight]
     );
+    const [selectedErrorIndex, setSelectedErrorIndex] = useState<number | null>(null);
 
     const environmentUrl = useMemo(() => getEndpointEnvironmentUrl(endpoint), [endpoint]);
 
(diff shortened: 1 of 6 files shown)


dsinghvi:
@kafkas can you file an issue to track the FLUP that clicking outside of the errors region should close the errors ?

kafkas:
> @kafkas can you file an issue to track the FLUP that clicking outside of the errors region should close the errors ?

Done!

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
```diff
@@ -84,6 +84,7 @@ export const EndpointContent = React.memo<EndpointContent.Props>(function Endpoi
         },
         [titleHeight]
     );
+    const [selectedErrorIndex, setSelectedErrorIndex] = useState<number | null>(null);
```
zachkirsch, line 87: nbd but everywhere else in the UI we never use null / always use undefined

zachkirsch: also, jw why is this state hoisted here in EndpointContent and not just in EndpointErrorsSection?

packages/ui/app/src/api-page/types/type-definition/InternalTypeDefinition.tsx
```diff
@@ -147,10 +147,15 @@ export const InternalTypeDefinition: React.FC<InternalTypeDefinition.Props> = ({
...
-                            isHovering ? "text-text-default" : "text-text-default"
+                            "flex gap-1 items-center cursor-pointer px-2 py-1 transition text-text-default border-b border-border",
+                            {
+                                "border-opacity-0": isCollapsed,
+                            }
                         )}
-                        onClick={toggleIsCollapsed}
+                        onClick={(e) => {
+                            toggleIsCollapsed();
+                            e.stopPropagation();
```
zachkirsch, line 157: I always like to add a comment to `e.stopPropagation();` to explain why we need it

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
kafkas: re `null` good to know, will switch to `undefined` for consistency

kafkas: re hoisted state, I default to keeping components stateless. As far as I know, it's generally recommended to have as few stateful components as possible as it's easier to reuse, test and debug them that way. Particularly true if we were using Storybook to build these components

PR #1928, opened 2023-07-20
Dynamic error response examples

kafkas:
This PR brings support for dynamic error response examples in documentation.

Screenshot 2023-07-20 at 9 19 02 PM


packages/ui/app/src/api-page/ApiPackageContents.tsx
```diff
@@ -1,49 +1,50 @@
...
-                            />
-                        );
-                    }),
-                ]}
-            </SeparatedElements>
+            {package_.endpoints.map((endpoint, idx) => (
+                <Endpoint
+                    key={endpoint.id}
+                    endpoint={endpoint}
+                    isLastInApi={isLastInParentPackage && idx === package_.endpoints.length - 1}
```
zachkirsch, line 29: this is def a bit less elegant than what we had before & seems a little more error-prone (eg as we add more api components). what was the issue with `SectionSeparator`?

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
```diff
@@ -85,11 +87,24 @@ export const EndpointContent = React.memo<EndpointContent.Props>(function Endpoi
         [titleHeight]
     );
     const [selectedErrorIndex, setSelectedErrorIndex] = useState<number | null>(null);
+    const selectedError = selectedErrorIndex == null ? null : endpoint.errors[selectedErrorIndex] ?? null;
```
zachkirsch, line 90: nbd but everywhere else in the UI we never use null / always use undefined

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
```diff
@@ -200,11 +217,11 @@ export const EndpointContent = React.memo<EndpointContent.Props>(function Endpoi
                             marginTop: titleHeight - 40,
                         }}
                     >
-                        <EndpointExamples endpoint={endpoint} />
+                        {example && <EndpointExample endpoint={endpoint} example={example} />}
```
zachkirsch, line 220: thought eslint would catch this but i guess not - for clarity I always do `!= null` so it's clear what we're checking for, since loads of other things are falsy

packages/ui/app/src/api-page/endpoints/endpoint-examples/EndpointExample.tsx
```diff
@@ -16,43 +15,44 @@ export const EndpointExample: React.FC<EndpointExample.Props> = ({ endpoint, exa
...
-                            parent={parent}
-                        />
+        <div className="flex min-h-0 min-w-0 flex-1 flex-col">
+            <div className="flex min-h-0 flex-1 flex-col">
+                <div className="grid min-h-0 flex-1 grid-rows-[repeat(auto-fit,_minmax(0,_min-content))] flex-col gap-6">
+                    <TitledExample
+                        title="Request"
+                        type="primary"
+                        onClick={(e) => {
+                            e.stopPropagation();
```
zachkirsch, line 25: i'd add a comment on why we're doing this

test-definitions/fern/api/definition/__package__.yml
```diff
@@ -0,0 +1,37 @@
+types:
```
zachkirsch, line 1: probably a good thing to separate into 2 PRs since we sometimes have to revert stuff when things break (and I think these are unrelated?)

packages/ui/app/src/api-page/ApiPackageContents.tsx
kafkas: Yes, agreed. Definitely less elegant. If you check [this page](https://vellum-staging.docs.dev.buildwithfern.com/api-reference/generate-stream) it should give you an idea. When you select an error you need to be able to unselect it to get back to the success response. Our idea was that the user should be able to unselect it by clicking anywhere outside of the errors section (empty spaces only). `SectionSeparator` reduces that area and it's not straightforward/elegant to attach an `onClick` event handler to it. So ideally we want to replace it with bottom padding and border.

But I just did a quick research and it looks like we can just use Tailwind's [`divide-width`](https://tailwindcss.com/docs/divide-width#add-borders-between-stacked-children) utility for this. May be ideal, wdyt?

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
kafkas: cool will switch to `undefined` for consistency

packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
kafkas: oh right, old habits... will change. i use `&&` only when I know that the value can be either an object or nulllish (e.g. when `""`, `0` etc. are not a possibility)

test-definitions/fern/api/definition/__package__.yml
kafkas: True. Noted. They're slightly related. It's a fern definition playground that helped us test the error response examples.

PR #1930, opened 2023-07-20
Use a smaller font size for api reference pages

kafkas:
Decreases the font size for API reference pages by separating the styles used for custom markdown and API reference pages.

Screenshot 2023-07-20 at 9 53 31 PM

### DIFF
--- a/packages/ui/app/src/api-page/markdown/Markdown.tsx
+++ b/packages/ui/app/src/api-page/markdown/Markdown.tsx
@@ -10,6 +10,11 @@ import styles from "./Markdown.module.scss";
 
 export declare namespace Markdown {
     export interface Props {
+        /**
+         * - `"markdown"`: Markdown pages.
+         * - `"api"`: API reference pages.
+         */
+        type: "markdown" | "api";
         children: string;
         className?: string;
     }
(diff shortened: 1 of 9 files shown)


packages/ui/app/src/api-page/markdown/Markdown.tsx
```diff
@@ -10,6 +10,11 @@ import styles from "./Markdown.module.scss";
 
 export declare namespace Markdown {
     export interface Props {
+        /**
+         * - `"markdown"`: Markdown pages.
+         * - `"api"`: API reference pages.
+         */
+        type: "markdown" | "api";
```
zachkirsch, line 17: I'd probably have made this a bit more declarative (i.e. what is this config actually _doing_? `style="danger"` is better imo than `type="foo"`).

I could see this component staying agnostic to fern (right now it's just a generic markdown component), and utilizing `className` instead. And maybe making a wrapper component to stay DRY.

kafkas: sounds good, yeah def not happy with the `type` solution. was the quickest way to implement the desired functionality (separated font styles for custom markdown and api pages)

PR #1941, opened 2023-07-23
docs and cli release blockers are independent

dsinghvi:
A docs release blocker should not block us from releasing the CLI and vice versa. We now maintain two release blocker files, one for each product.


kafkas (commented):
Looking good to me.

PR #1944, opened 2023-07-23
Display errors based on `NEXT_PUBLIC_DISPLAY_ERRORS`

kafkas:
With this change, we will display the errors section based on whether `NEXT_PUBLIC_DISPLAY_ERRORS` is set to `"true"`.


dsinghvi:
do all the env variables have to be prefixed with `NEXT_PUBLIC` ?

kafkas:
> do all the env variables have to be prefixed with `NEXT_PUBLIC` ?

The ones that need to be used in the browser, yes. Next.js uses it to know which ones to bundle for the browser.

PR #1945, opened 2023-07-23
fix: docs use fdr content-type

dsinghvi:


dsinghvi:
@kafkas going to merge

PR #1946, opened 2023-07-23
feat: add background image to docs ui

codebender828:
- feat: upgrade write and read registry definitions
- feat: add background image to UI

CleanShot 2023-07-23 at 22 04 57@2x

### DIFF
--- a/packages/ui/app/src/docs/Docs.tsx
+++ b/packages/ui/app/src/docs/Docs.tsx
@@ -18,7 +18,7 @@ export const Docs: React.FC = () => {
     useKeyboardCommand({ key: "K", platform: PLATFORM, onCommand: openSearchDialog });
 
     return (
-        <div className="relative flex min-h-0 flex-1 flex-col bg-cover">
+        <div className="relative flex min-h-0 flex-1 flex-col">
             {searchService.isAvailable && <SearchDialog isOpen={isSearchDialogOpen} onClose={closeSearchDialog} />}
             <div className="border-border bg-background sticky inset-x-0 top-0 border-b">
                 <Header />


dsinghvi:
@codebender828 exciting!! 👀

fern/docs-config/definition/docs.yml
```diff
@@ -7,6 +7,9 @@ types:
         docs: used as tab bar title, and in the navbar if no logo is defined
       logo: optional<LogoConfiguration>
       favicon: optional<string>
+      backgroundImage:
```
zachkirsch, line 10: let's make this `background-image` to match `navbar-links`

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
```diff
@@ -20,40 +21,40 @@ export declare namespace Docs {
...
+    backgroundImageStyleSheet,
     resolvedUrlPath,
     nextPath,
     previousPath,
 }: Docs.Props): JSX.Element {
+    const stylesheets = [typographyStyleSheet, backgroundImageStyleSheet].filter((s) => s != null) as string[];
     return (
         <>
-            {typographyStyleSheet != null && <Typography stylesheet={typographyStyleSheet} />}
+            <StyleSheets stylesheets={stylesheets} />
```
Comment (zachkirsch) [inline packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:52]: nitty but maybe just
```tsx
return (
  <>
    {typographyStyleSheet != null && <StyleSheet styleSheet={typographyStyleSheet} />}
    {backgroundImageStyleSheet != null && <StyleSheet styleSheet={backgroundImageStyleSheet} />}
    ...
```

in general I don't love casts (`as string[]`) since it's easy to run into runtime errors if the cast becomes incorrect later when the typings change

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
```diff
@@ -20,40 +21,40 @@ export declare namespace Docs {
...
-        typographyStyleSheet?: string;
+        typographyStyleSheet: string | undefined;
+        backgroundImageStyleSheet: string | undefined;
         nextPath: ResolvedUrlPath | null;
         previousPath: ResolvedUrlPath | null;
     }
 }
 
-function Typography({ stylesheet }: { stylesheet: string }) {
+function StyleSheets({ stylesheets }: { stylesheets: string[] }) {
```
Comment (zachkirsch) [inline packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:31]: can we move this into a separate file?

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
Comment (codebender828) [reply packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:52]: HAha I get you

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
Comment (codebender828) [reply packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:31]: Yeah that's doable. I wanted to keep things simple for now but can move it.

Comment (codebender828) [reply packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:31]: I tried moving this to a sincle instance, but Next JS will ignore the second rendered instance of the global style sheet. The result is to just concatenate all styles into a single string and inject the, globally.

If I were to use a regular style tag, Next, will inject them at the very top of the `<head/>`. This means they will be overriden by other global stylesheets in the head.

Comment (codebender828) [reply packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:31]: @zachkirsch The solution was just to concatenate them to achieve these two outcomes:
- All configuration stylesheets should be serialized into a single stylesheet string.
- The final stylesheet string should be applied at the highest level of the CSS cascade for global styles. (i.e. it should be injected just before the `<head />` closing tag)

Not elegant, but it works. Will probably change this once we look into theming at a more concrete level.

fern/docs-config/definition/docs.yml
codebender828: Done!

dsinghvi:
@codebender828 thanks, @zachkirsch can you review today would love to merge this change in!

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
```diff
@@ -20,40 +21,40 @@ export declare namespace Docs {
...
+                    as styled JSX will only create one instance of global styles
+                    for each component.
+                */}
+                {/* eslint-disable-next-line react/no-unknown-property */}
+                <style jsx global>
+                    {`
+                        ${typographyStyleSheet}
+                        ${backgroundImageStyleSheet}
+                    `}
+                </style>
```
Comment (zachkirsch) [inline packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx:53]: looks great! just curious, would this work?

```tsx
                <style jsx global>
                    {typographyStyleSheet}
                    {backgroundImageStyleSheet}
                </style>
```

and then we don't need the `= ""` in the function signature

packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx
Comment (zachkirsch) [reply packages/ui/fe-bundle/src/pages/[host]/[[...slug]].tsx]: dope sg!

zachkirsch:
right now we are rendering two backgrounds:
1. the one in this pr
2. the nebula image https://github.com/fern-api/fern/blob/a5517cb3d6c8e1d75866966b916eff07aec5fc4d/packages/ui/app/src/docs/Docs.tsx#L24

can we instead have the nebula one be a default - i.e. move that logic to the nextjs app and only render it if there's no custom background image defined?

right now loading the nebula can be kinda slow so I'd love to avoid loading it client-side if we don't need to.

packages/ui/app/src/docs/Docs.tsx
```diff
@@ -8,7 +8,6 @@ import { useSearchService } from "../services/useSearchService";
 import { Sidebar } from "../sidebar/Sidebar";
 import { DocsMainContent } from "./DocsMainContent";
 import { Header } from "./Header";
-import nebulaImage from "./nebula.png";
```
zachkirsch, line 11: Can we delete the image file too?

packages/ui/app/src/docs/Docs.tsx
```diff
@@ -19,10 +18,7 @@ export const Docs: React.FC = () => {
     useKeyboardCommand({ key: "K", platform: PLATFORM, onCommand: openSearchDialog });
 
     return (
-        <div
-            className="relative flex min-h-0 flex-1 flex-col bg-cover"
-            style={{ backgroundImage: `url('${nebulaImage.src}')` }}
-        >
+        <div className="relative flex min-h-0 flex-1 flex-col bg-cover">
```
zachkirsch, line 21: Don’t think we need bg-cover anymore

zachkirsch:
Looks great!!

zachkirsch (approved):

packages/ui/app/src/docs/Docs.tsx
codebender828: Good catch forgot to delete it!

PR #1947, opened 2023-07-23
Improve endpoint URL in api page

kafkas:
This PR improves the URL component that we show in the API endpoint section with better designs and better handling of long URLs. The PR also fixes some layout issues.

Screenshot 2023-07-23 at 9 41 17 PM

Screenshot 2023-07-23 at 9 42 38 PM


dsinghvi:
🔥 @kafkas if a url is blurred out due to length does it have a hover state where you can see the rest of it? if not, thats okay we can track as a follow up

kafkas:
> 🔥 @kafkas if a url is blurred out due to length does it have a hover state where you can see the rest of it? if not, thats okay we can track as a follow up

Not at the moment, unfortunately but we can do that easily. I was actually working on something a little different. A button that allows the user to expand the section. But a tooltip like the one you mentioned would be easier to implement.

PR #1953, opened 2023-07-24
Change navbar link interface in cli

kafkas:


packages/cli/config-management/docs-configuration/src/convertDocsConfiguration.ts
```diff
@@ -189,20 +189,5 @@ function parseHexColorCode(value: string | undefined): number | undefined {
...
-}
-
-function convertNavbarLinkStyle(rawConfig: RawDocs.NavbarLinkStyle): FernRegistry.docs.v1.write.NavbarLinkStyle {
-    switch (rawConfig) {
-        case "primary":
-            return FernRegistry.docs.v1.write.NavbarLinkStyle.Primary;
-        default:
-            assertNever(rawConfig);
-    }
+    return rawConfig as FernRegistry.docs.v1.write.NavbarLink[];
```
dsinghvi, line 192: @kafkas why are we casting here?

PR #1967, opened 2023-07-25
[Breaking] Introducing Docs + API Workspaces

dsinghvi:
The updated fern structure is going to include an `apis` directory if there are multiple apis
```
fern/
  docs.yml
  apis/
    api-one/
    api-two/
    api-three/
```
or if there is only one API then the definition directory will be moved up
```
fern/
  docs.yml
  definition/
    api.yml
  generatiors.yml
```

An automatic migration will be added to a follow up PR. 

Updated CLI Commands: 
```bash
fern init --docs # initializes docs
fern init --api # initializes api
fern generate --api # generates for api
fern generate --docs # generates for docs
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -342,15 +342,15 @@ workflows:
 
       - publish-dev:
           filters:
-            tags:
-              only: /.*/
+            branches:
+              only: main
           requires:
             - check
             - organize-imports
             - compile
             - eslint
             - test
-            # - ete-tests
+            - ete-tests
           context:
             - npm
             - auth0-dev


.circleci/config.yml
```diff
@@ -336,15 +336,15 @@ workflows:
 
       - publish-dev:
           filters:
-            branches:
-              only: main
+            tags:
+              only: /.*/
```
dsinghvi, line 340: remove before merging

packages/cli/cli/src/cli.ts
```diff
@@ -269,18 +274,40 @@ function addGenerateCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext)
...
-                    defaultToAllWorkspaces: false,
-                }),
-                cliContext,
-                version: argv.version,
-                groupName: argv.group,
-                shouldLogS3Url: argv.printZipUrl,
-                keepDocker: argv.keepDocker,
-                useLocalDocker: argv.local,
-            });
+            if (argv.api != null && argv.docs != null) {
```
dsinghvi, line 277: this does not handle `fern generate` if neither `--api` nor `--docs` are specified

packages/cli/cli/src/commands/add-generator/addGeneratorToWorkspaces.ts
```diff
@@ -18,19 +18,14 @@ export async function addGeneratorToWorkspaces({
     groupName: string | undefined;
     cliContext: CliContext;
 }): Promise<void> {
+    if (generatorName === "docs") {
+        await addDocsToProject();
+        return;
+    }
```
dsinghvi, line 24: we should deprecate fern add docs and instead do `fern init --docs` and `fern init --api`

fern/docs-config/definition/docs.yml
```diff
@@ -64,7 +81,10 @@ types:
 
   ApiSectionConfiguration:
     properties:
-      api: string
+      title: string
```
dsinghvi, line 84: We will have to migrate all existing docs configs in a PR

PR #1968, opened 2023-07-25
Add license to GithubConfigurationSchema

amckinney:
> This is very much still a work-in-progress - the code doesn't yet compile because the server-side
> license API definitions have not yet been released. Throwing this up early to show the general flow
> and different `generators.yml` format options.

This introduces a new `license` configuration option for `github` outputs. A couple example configurations are shown below.

#### [SPDX](https://spdx.org/licenses) (MIT + Apache-2.0)

```yaml
default-group: publish
groups:
  publish:
    generators:
      - name: fernapi/fern-go-sdk
        version: 0.0.14
        github:
          repository: org/repo
          license: MIT
```

#### Custom License

```yaml
default-group: publish
groups:
  publish:
    generators:
      - name: fernapi/fern-go-sdk
        version: 0.0.14
        github:
          repository: org/repo
          license: 
            custom: ../../LICENSE
```

### DIFF
--- a/packages/cli/config-management/generators-configuration/src/__test__/convertGeneratorsConfiguration.test.ts
+++ b/packages/cli/config-management/generators-configuration/src/__test__/convertGeneratorsConfiguration.test.ts
@@ -2,8 +2,8 @@ import { AbsoluteFilePath } from "@fern-api/fs-utils";
 import { convertGeneratorsConfiguration } from "../convertGeneratorsConfiguration";
 
 describe("convertGeneratorsConfiguration", () => {
-    it("local-file-system allows absolute download path", () => {
-        const converted = convertGeneratorsConfiguration({
+    it("local-file-system allows absolute download path", async () => {
+        const converted = await convertGeneratorsConfiguration({
             absolutePathToGeneratorsConfiguration: AbsoluteFilePath.of("/path/to/repo/fern/api/generators.yml"),
             rawGeneratorsConfiguration: {
                 groups: {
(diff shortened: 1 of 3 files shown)


dsinghvi:
@amckinney for any user facing configuration, we generallly try to optimize for the easiest authoring experience. other than that, we generally try to optimize for type safety. 

given that, undiscriminated unions sounds like the right path forward. how about something like: 

```
  github:
    repository: org/repo
    license: MIT
```
and 
```
github:
  repository: org/repo
  license: 
    custom: /path/to/file
 ```

**Comment by amckinney** [comment]
I added in a few simple tests to verify that the `github` configuration is being read as we expect. Plus, I did some manual testing using `yarn fern-dev:local` to make sure the error messages look good:

```sh
$ yarn fern-dev:local
...
Failed.
Unsupported license: unknown
```

```sh
$ yarn fern-dev:local
...
Failed.
Failed to read custom license ../../LICENS
```

With that said, there's a little bit of stutter here, but it looks like it's the same structure as the errors used elsewhere (e.g. [semantic version validation](https://github.com/fern-api/fern/blob/a5517cb3d6c8e1d75866966b916eff07aec5fc4d/packages/cli/semver-utils/src/parseVersion.ts#L117)). It might not be that big of a deal in practice, but it's worth calling out.

dsinghvi:
misc fixes: 
- use fs/promises and convert to async
- schema is more type safe (enum as opposed to string)

didnt have time to get tests with exceptions/failed promises working so skipped for now

PR #1969, opened 2023-07-25
Upgrade init behavior

dannysheridan:
* Upgrade generator versions in `fern init` 
* Add Go generator to docs for generators
* Updated tests

### DIFF
--- a/packages/cli/ete-tests/src/tests/add/__snapshots__/add.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/add/__snapshots__/add.test.ts.snap
@@ -91,7 +91,7 @@ groups:
   local:
     generators:
       - name: fernapi/fern-typescript-node-sdk
-        version: 0.6.4
+        version: 0.7.2
         output:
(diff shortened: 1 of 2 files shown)


packages/cli/config-management/manage-generator/src/generatorInvocations.ts
```diff
@@ -2,23 +2,23 @@ import { GeneratorInvocationSchema, GeneratorName } from "@fern-api/generators-c
 
 export const GENERATOR_INVOCATIONS: Record<GeneratorName, Omit<GeneratorInvocationSchema, "name">> = {
     [GeneratorName.JAVA]: {
-        version: "0.0.122",
+        version: "0.3.7",
```
dsinghvi, line 5: this is a legacy generator, will want to keep it at 0.0.122

dannysheridan: roger that. reverted that change in a new commit.

PR #1993, opened 2023-08-01
Add IR ServiceTypeReferenceInfo

amckinney:
This adds a new `serviceTypeReferenceInfo` property to the IR so that generators can recognize what types are only used by a single service, if any. The `serviceTypeReferenceInfo` also defines a `sharedTypes` property that captures all other types (either referenced by zero or multiple services).

With this, generators can consolidate all types that belong to a particular service in a colocated file alongside the client. This significantly reduces the number of files that we'd otherwise need to generate, and simultaneously makes the generated types more discoverable.

### DIFF
--- a/packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
+++ b/packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
@@ -1,4 +1,4 @@
-import { Audiences } from "@fern-api/config-management-commons";
+import { Audiences as ConfigAudiences } from "@fern-api/config-management-commons";
 import { assertNever, noop } from "@fern-api/core-utils";
 import { FernFilepath } from "@fern-fern/ir-model/commons";
 import { ErrorDeclaration } from "@fern-fern/ir-model/errors";
@@ -37,8 +37,8 @@ export class IrGraph {
     private endpointsNeededForAudience: Set<EndpointId> = new Set();
     private subpackagesNeededForAudience: Set<SubpackageId> = new Set();
 
-    public constructor(audiences: Audiences) {
-        this.audiences = audiences;
+    public constructor(audiences: ConfigAudiences) {
+        this.audiences = audiencesFromConfig(audiences);
     }
 
     public addType(declaredTypeName: DeclaredTypeName, descendants: DeclaredTypeName[]): void {
@@ -263,17 +263,17 @@ export class IrGraph {
         return endpointNode;
     }
 
-    public hasAllAudiences(): boolean {
-        return this.audiences.type === "all";
+    public hasNoAudiences(): boolean {
+        return this.audiences.type === "none";
     }
 
     private hasAudience(audiences: AudienceId[]): boolean {
         const configuredAudiences = this.audiences;
         switch (configuredAudiences.type) {
-            case "all":
+            case "none":
                 return true;
-            case "select":
-                return audiences.some((audienceId) => configuredAudiences.audiences.includes(audienceId));
+            case "filtered":
+                return audiences.some((audienceId) => configuredAudiences.audiences.has(audienceId));
             default:
                 assertNever(configuredAudiences);
         }
@@ -287,6 +287,28 @@ export class IrGraph {
             };
             this.subpackagesNeededForAudience.add(IdGenerator.generateSubpackageId(packageFilePath));
         }
+    }
+}
+
+type Audiences = NoAudience | FilteredAudiences;
+
+interface NoAudience {
+    type: "none";
+}
+
+interface FilteredAudiences {
+    type: "filtered";
+    audiences: Set<string>;
+}
+
+function audiencesFromConfig(configAudiences: ConfigAudiences): Audiences {
+    switch (configAudiences.type) {
+        case "all":
+            return { type: "none" };
+        case "select":
+            return { type: "filtered", audiences: new Set(configAudiences.audiences) };
+        default:
+            assertNever(configAudiences);
     }
 }
 
--- a/packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
+++ b/packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
@@ -270,7 +270,7 @@ export async function generateIntermediateRepresentation({
 
     intermediateRepresentation.serviceTypeReferenceInfo = computeServiceTypeReferenceInfo(irGraph);
 
-    const filteredIr = !irGraph.hasAllAudiences() ? irGraph.build() : undefined;
+    const filteredIr = !irGraph.hasNoAudiences() ? irGraph.build() : undefined;
     const intermediateRepresentationForAudiences = filterIntermediateRepresentationForAudiences(
         intermediateRepresentation,
         filteredIr


fern/ir-types-v22/definition/ir.yml
```diff
@@ -38,6 +38,7 @@ types:
       errorDiscriminationStrategy: ErrorDiscriminationStrategy
       sdkConfig: SdkConfig
       variables: list<variables.VariableDeclaration>
+      typeGraph: TypeGraph
```
dsinghvi, line 41: we'll need to make this optional to keep it in v22 (kind of annoying so if you wanted to make it required than we'd have to add it to v23).

fern/ir-types-v22/definition/ir.yml
```diff
@@ -74,3 +75,7 @@ types:
   PackageNavigationConfig:
     properties:
       pointsTo: commons.SubpackageId
+  TypeGraph:
+    properties:
+      typesReferencedOnlyByService: map<commons.ServiceId, list<commons.TypeId>>
+      sharedTypes: list<commons.TypeId>
```
dsinghvi, line 81: lets add some docs about what each of these properties are

packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
```diff
@@ -25,18 +26,21 @@ import {
...
     private endpointsNeededForAudience: Set<EndpointId> = new Set();
-    private audiences: Set<AudienceId> = new Set();
     private subpackagesNeededForAudience: Set<SubpackageId> = new Set();
 
-    public constructor(audiences: AudienceId[]) {
-        this.audiences = new Set(audiences);
+    public constructor(audiences: Audiences) {
+        if (audiences.type !== "all") {
+            this.audiences = new Set(audiences.audiences);
+        }
```
dsinghvi, line 43: what happens when audiences set is empty? looks like the old code was assuming that would never be the case

packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
```diff
@@ -25,18 +26,21 @@ import {
...
 } from "./ids";
 
-export class AudienceIrGraph {
+export class IrGraph {
     private types: Record<TypeId, TypeNode> = {};
-    private typesNeededForAudience: Set<TypeId> = new Set();
     private errors: Record<TypeId, ErrorNode> = {};
     private endpoints: Record<EndpointId, EndpointNode> = {};
+    private audiences: Set<AudienceId> = new Set();
+    private servicesReferencedByType: Record<TypeId, Set<ServiceId>> = {};
```
dsinghvi, line 34: i think naming this `typesUsedByService` might make the code more readable

packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
```diff
@@ -204,6 +219,13 @@ export class AudienceIrGraph {
         });
     }
 
+    private markTypeForService(typeId: TypeId, serviceId: ServiceId): void {
+        if (this.servicesReferencedByType[typeId] == null) {
+            this.servicesReferencedByType[typeId] = new Set();
+        }
+        this.servicesReferencedByType[typeId]?.add(serviceId);
```
dsinghvi, line 226:
```typescript
     const services = this.servicesReferencedByType[typeId];
     if (services == null) {
          this.servicesReferencedByType[typeId] = new Set(serviceId);
      } else {
         services.add(serviceId);
      }
```

packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
```diff
@@ -263,7 +268,9 @@ export async function generateIntermediateRepresentation({
         }
     });
 
-    const filteredIr = audienceIrGraph != null ? audienceIrGraph.build() : undefined;
+    intermediateRepresentation.typeGraph = extractTypeGraph(irGraph);
```
dsinghvi, line 271: `extractTypeGraph` is not the greatest name for the method, i find myself asking wtf is this doing. ideally we should get a high level overview of whats happening just by reading the method name

fern/ir-types-v22/definition/ir.yml
dsinghvi: i could see making this type super specific for now `ServiceTypeUsageInfo` or `ServiceTypeReferenceInfo` and the docs can be about how this object

packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
dsinghvi: https://github.com/fern-api/fern/pull/1993/files#r1281193699 if we do this rename then i could se this being `computeServiceTypeReferenceInfo` and that gives me a little more info

packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
dsinghvi: just caught the `hasAudience()` method. Ideally this would be typed as a union where its `none` or `some`

amckinney: I adapted the behavior of `hasAudience` so that it will always consider the given audiences to be present if none are configured (aka if `all` audiences are configured):

```typescript
    public hasAllAudiences(): boolean {
        return this.audiences.size === 0;
    }

    private hasAudience(audiences: AudienceId[]): boolean {
        if (this.hasAllAudiences()) {
            return true;
        }
        return audiences.some((audienceId) => this.audiences.has(audienceId));
    }
```

I considered adding an additional private field (e.g. `hasAllAudiences`) that would be set to more explicitly signal when all audiences were configured, but it was equivalent to just checking for the size of `this.audiences` so I went with this.

Happy to change to whatever you prefer - I don't mind either way.

fern/ir-types-v22/definition/ir.yml
amckinney: Cool, I'll make it optional here 👍

packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
amckinney: Two options:
1. `newTypeGraphFromIrGraph`
2. Add an `IrGraph.toTypeGraph` method

Both options clear it up quite a bit, but I like the method approach since it encapsulates the logic to the `IrGraph`. I'm happy with either way though.

packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
amckinney: Sweet, given that your suggestion is nearly identical to the `Audiences` already defined in `@fern-api/config-management-common`, we could just use those:

```typescript
export type Audiences = AllAudiences | SelectAudiences;

export interface AllAudiences {
    type: "all";
}

export interface SelectAudiences {
    type: "select";
    audiences: string[];
}
```

```typescript
public constructor(audiences: Audiences) {
    this.audiences = audiences;
}

...

public hasAllAudiences(): boolean {
    return this.audiences.type === "all";
}

private hasAudience(audiences: AudienceId[]): boolean {
    const configuredAudiences = this.audiences;
    switch (configuredAudiences.type) {
        case "all":
            return true;
        case "select":
            return audiences.some((audienceId) => configuredAudiences.audiences.includes(audienceId));
        default:
            assertNever(configuredAudiences);
    }
}
```

It's worth calling out that `SelectAudiences` represents its `audiences` as a list, so this would be slightly less efficient than the `Set`-based approach, but it at least lets us reuse the primary types used elsewhere. I think the tradeoff is worth it, especially because the performance is so negligible for the size of these audience lists.

Thoughts?

amckinney: Pushed up that change (using the `Audiences` that already exists) - let me know what you think.

packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
amckinney: Sweet, I went with `typesReferencedByService` so that its consistent with the other suggestion.

packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
amckinney: I kept it simple and went with your `computeServiceTypeReferenceInfo` suggestion.

packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
dsinghvi: sounds good -- im down to redefine the union as a Set if thats cool with you? we can just redefine the union

dsinghvi (approved):
nit left on audiences union thing, but other than that LGTM

dsinghvi (approved):

PR #2015, opened 2023-08-07
setting up functions to run runLocalGenerationForWorkspace

sahil485:


packages/seed/src/commands/test.ts
```diff
@@ -81,3 +85,67 @@ async function getIntermediateRepresentation({
...
+}: {
+    dockerName: string,
+    absolutePathToOutput: AbsoluteFilePath,
+    language: GenerationLanguage
+}): GeneratorInvocation {
+
+    const dockerNameArray = parseDockerName(dockerName);
+    const generatorInvocation: GeneratorInvocation = {
+        name: "DUMMY",
+        version: "DUMMY",
```
dsinghvi, line 144: version should come from the docker image

packages/seed/src/commands/test.ts
```diff
@@ -81,3 +85,67 @@ async function getIntermediateRepresentation({
...
+    language
+}: {
+    dockerName: string,
+    absolutePathToOutput: AbsoluteFilePath,
+    language: GenerationLanguage
+}): GeneratorInvocation {
+
+    const dockerNameArray = parseDockerName(dockerName);
+    const generatorInvocation: GeneratorInvocation = {
+        name: "DUMMY",
```
dsinghvi, line 143: should be the name of the docker image

packages/seed/src/commands/test.ts
```diff
@@ -81,3 +85,67 @@ async function getIntermediateRepresentation({
...
+        version: "DUMMY",
+        config: "DUMMY",
```
dsinghvi, line 145: we can make this `undefined`

dsinghvi (approved):

PR #2023, opened 2023-08-08
fix: OpenAPI workspaces don't output OpenAPI spec

dannysheridan:
* Remove OpenAPI generator when importing OpenAPI
* Fix Go default file path to be consistent with other generators


amckinney (approved):
Updated Go path looks good - not sure of the context w.r.t the OpenAPI change, but seems reasonable.

PR #2032, opened 2023-08-12
setting up parallel dockers for testing generators with semaphore

sahil485:
"poetry install && poetry run mypy ." command only failing when we include the "single-url-environment-no-default" test definition 

must run rm -rf seed before generating, otherwise it fails as well (idk why)

promise logic -> resolve() seems to be called after the promise is already resolved and the log statements come after "Seed has finished". not sure why, async/awaits all seem fine... :(


packages/seed/src/commands/test.ts
```diff
@@ -18,68 +18,244 @@ export const FIXTURE = {
     BASIC_AUTH: "basic-auth",
     CUSTOM_AUTH: "custom-auth",
     ERROR_PROPERTY: "error-property",
+    MULTI_URL_ENVIRONMENT: "multi-url-environment",
+    NO_ENVIRONMENT: "no-environment",
+    SINGLE_URL_ENVIRONMENT: "single-url-environment-default",
+    SINGLE_URL_ENVIRONMENT_NO_DEFAULT: "single-url-environment-no-default",
```
dsinghvi, line 24: @sahil485 thanks for adding this i was wondering why it wasnt showing up

packages/seed/src/commands/test.ts
```diff
@@ -18,68 +18,244 @@ export const FIXTURE = {
...
 } as const;
 
-export async function runTests({
+export const MAX_NUM_DOCKERS_RUNNING = 4;
+
+function createSemaphore(initialCount: number): {
```
dsinghvi, line 29: 🔥

PR #2038, opened 2023-08-14
debug loggingExeca function and surrounding workflow for 'yarn install' and 'yarn install' commands

sahil485:
if you put in flags for the loggingExeca command as shown here (logging numbers 1-6), you'll see that the loggingExeca command for yarn build finishes

however, when you return control back to seed/src/commands/test/test.ts in line 171 after awaiting the promise, you'll notice that "AFTER" never gets logged to the console

not able to figure out why, but the code after this promise resolves is never executed for either yarn install or yarn build


dsinghvi:
very weird

PR #2047, opened 2023-08-15
add test defs for files upload/download from TS gen to Seed

sahil485:


dsinghvi (commented):
@sahil485 we should be good to delete `file-download-browser` and `file-download-response-headers` and `file-upload-browser`

PR #2067, opened 2023-08-25
Don't leave out `name` when converting webhook group

kafkas:


packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
```diff
@@ -27,6 +27,7 @@ function convertWebhookGroup(webhookGroup: Ir.webhooks.WebhookGroup): FernRegist
             id: FernRegistry.api.v1.register.WebhookId(webhook.name.originalName),
             path: [],
             method: webhook.method,
+            name: webhook.displayName,
```
dsinghvi, line 30: `name: webhook.displayName?? startCase(webhook.name.originalName)`

great catch

dsinghvi (approved):

PR #2129, opened 2023-09-25
Add `EndpointId` to IR

amckinney:
This adds a `commons.EndpointId` alias and includes it in the `HttpEndpoint` type. A `generateEndpointId` method already existed for the audience graph, so this reuses the same form already defined there.

For now, the change remains as simple as possible and includes it in the pre-existing `IRv26`, but we might later want to create another top-level `map<EndpointId, HttpEndpoint` map so that generators can quickly access all of the endpoints without first traversing the services. We might also consolidate the `EndpointName` and `EndpointId` in a single type (similar to the `DeclaredTypeName` object).

### DIFF
--- a/packages/cli/generation/ir-generator/src/IdGenerator.ts
+++ b/packages/cli/generation/ir-generator/src/IdGenerator.ts
@@ -1,5 +1,5 @@
 import { FernIr } from "@fern-fern/ir-sdk";
-import { DeclaredErrorName, DeclaredServiceName, DeclaredTypeName, FernFilepath, Name } from "@fern-fern/ir-sdk/api";
+import { DeclaredErrorName, DeclaredServiceName, DeclaredTypeName, FernFilepath } from "@fern-fern/ir-sdk/api";
 
 export const IdGenerator = {
     generateTypeId: (typeName: Omit<DeclaredTypeName, "typeId">): FernIr.commons.TypeId => {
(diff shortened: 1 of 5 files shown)


packages/cli/generation/ir-generator/src/IdGenerator.ts
```diff
@@ -24,9 +18,9 @@ export const IdGenerator = {
         const joinedFernFilePath = stringifyFernFilepath(fernFilepath);
         return `subpackage_${joinedFernFilePath}`;
     },
-    generateEndpointId: (declaredServiceName: DeclaredServiceName, httpEndpoint: HttpEndpoint): string => {
+    generateEndpointId: (declaredServiceName: DeclaredServiceName, httpEndpointName: Name): string => {
```
dsinghvi, line 21: can we pass the `HttpEndpoint` directly, a lot of things have `Name` so doesnt provide as much type safety

dsinghvi (approved):
LGTM, just one small change

packages/cli/generation/ir-generator/src/IdGenerator.ts
amckinney: I originally wanted to, but we actually use this function while we're constructing the `HttpEndpoint` itself, so it's actually easier to use the name here.

See `convertHttpService.ts` for reference:
```typescript
            const endpointName = file.casingsGenerator.generateName(endpointKey);
            return {
                ...convertDeclaration(endpoint),
                name: file.casingsGenerator.generateName(endpointKey),
                id: IdGenerator.generateEndpointId(serviceName, endpointName),
                name: endpointName,
                displayName: endpoint["display-name"],
                auth: endpoint.auth ?? serviceDefinition.auth,
                baseUrl: endpoint.url ?? serviceDefinition.url,
```

dsinghvi: got it you could make the type `Omit<HttpEndpoint, "id">` ? not the cleanest but i think better than name?

PR #2146, opened 2023-10-02
Title case error name at register

kafkas:
Will `Title Case` the error names when registering the API


dsinghvi:
https://github.com/fern-api/fern-definition-registry/blob/767cb4cdf2e069acc1324305e0cda7b4310e01dc/src/converters/db/convertAPIDefinitionToDb.ts#L145-L160 thought we were going to add to FDR?

PR #2171, opened 2023-10-12
[openapi] Setup example abstractions

amckinney:
This sets up the primary abstraction we'll use to import examples from OpenAPI into Fern


fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,117 @@
...
+
+  PartialExample:
+    union:
+      object: PartialObjectExample
+      oneOf: PartialOneOfExample
+
+  PartialObjectExample:
+    properties:
+      includedProperties: map<string, PartialExample>
+      excludedProperties: list<SchemaInstanceId>
```
dsinghvi, line 104: i think you'll want the key of the properties? what if the properties are an int/string etc? can also punt this till you start writing some code?

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,117 @@
...
+
+  PartialOneOfExample:
+    union:
+      discriminated: PartialDiscriminatedOneOfExample
+      undisciminated: PartialUnDiscriminatedOneOfExample
+
+  PartialDiscriminatedOneOfExample:
+    properties:
+      examples: map<string, PartialExample>
```
dsinghvi, line 113: an example of a discriminated union will only be one of the variants

dsinghvi: so i would think something like: 
```
discriminant: string
value: `PartialExample`
```

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,117 @@
...
+
+  PartialUnDiscriminatedOneOfExample:
+    properties:
+      examples: list<PartialExample>
```
dsinghvi, line 117: I think a PartialUnDiscriminatedOneOfExample = PartialExample? (following from the comments from the discriminated union)

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,117 @@
...
+    properties:
+      examples: map<string, FullExample>
+
+  FullUnDiscriminatedOneOfExample:
+    properties:
+      examples: list<FullExample>
+
+  ReferencedExample:
+    properties:
+      reference: string
```
dsinghvi, line 94: assuming the reference should be to a `SchemaInstanceId` ?

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,117 @@
...
+  IntExample: integer
+  Int64Example: long
+  FloatExample: double
+  DoubleExample: double
+  StringExample: string
+  DatetimeExample: datetime
+  DateExample: date
+  Base64Example: base64
+  BooleanExample: boolean
+  LiteralExample: string
```
dsinghvi, line 17: not sure how useful the aliases are, could also inline if you prefer. i'm not opinionated here so defer to you

fern/openapi-ir/definition/example.yml
amckinney: Definitely, this was intentional but ended up accidentally dropping the key when I was refactoring things with `SchemaInstanceId`. Will update 👍

fern/openapi-ir/definition/example.yml
amckinney: Good catch, I was mistakenly thinking about how this would represent _one or more_ of the possible discriminated union instances for a given definition, but this is already only representing a single example, so your representation is correct.

fern/openapi-ir/definition/example.yml
amckinney: Good catch, that was the intention (re: same reason above), will update 👍

fern/openapi-ir/definition/example.yml
amckinney: Let's in-line em, I agree with the simplicity 👍

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      reference: SchemaInstanceId
+
+  PartialExample:
+    union:
+      object: PartialObjectExample
+      oneOf: PartialOneOfExample
+
+  PartialObjectExample:
+    properties:
+      includedProperties: map<string, FullOrPartialExample>
```
amckinney, line 101: @dsinghvi heads up also made a small adjustment to `includedProperties` so that it can be either a `FullExample` or a `PartialExample` (but not a reference because I _think_ that's impossible in the example syntax).

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      full: FullExample
+      partial: PartialExample
+
+  FullExample:
+    union:
+      primitive: PrimitiveExample
+      object: FullObjectExample
+      array: ArrayExample
+      map: MapExample
+      optional: FullExample
```
dsinghvi, line 32: i think we can actually skip optional (its either there or not)

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      enum: EnumExample
+      reference: FullExample
```
dsinghvi, line 34: i think we can skip reference here, the full example will just have the entire reference inlined

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      literal: LiteralExample
+      oneOf: FullOneOfExample
+      nullable: FullExample
```
dsinghvi, line 37: similar to optional, i think we can delete this

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      unknown: {}
```
dsinghvi, line 38: an `unknown` property should just point to a `FullExample` i believe cause it could be anything

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+  MapExample:
+    properties:
+      pairs: list<MapExamplePair>
+
+  MapExamplePair:
+    properties:
+      key: PrimitiveExample
+      value: FullExample
+
+  EnumExample:
```
dsinghvi, line 69: shouldn't an `EnumExample` just be a string?

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      base64: base64
+      boolean: boolean
+
+  FullObjectExample:
+    properties:
+      properties: map<string, FullExample>
+
+  ArrayExample:
+    properties:
+      elements: list<FullExample>
```
dsinghvi, line 58: down to inline this in `FullExample` as `list<FullExample>`

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+
+  MapExample:
+    properties:
+      pairs: list<MapExamplePair>
```
dsinghvi, line 62: down to inline in FullExample as `list<KeyValuePair>`

fern/openapi-ir/definition/example.yml
dsinghvi: down to inline as `enum: string`

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      value: string
+
+  FullOneOfExample:
+    union:
+      discriminated: FullDiscriminatedOneOfExample
+      undisciminated: FullUnDiscriminatedOneOfExample
+
+  FullDiscriminatedOneOfExample:
+    properties:
+      examples: map<string, FullExample>
```
dsinghvi, line 84: I think this should be 
```yaml
FullDiscriminatedOneOfExample: 
  properties: 
    properties: map<string, FullExample>
```

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      double: double
+      string: string
+      datetime: datetime
+      date: date
+      base64: base64
+      boolean: boolean
+
+  FullObjectExample:
+    properties:
+      properties: map<string, FullExample>
```
dsinghvi, line 54: would be down to inline as `map<string, FullExample>` on `FullExample`

fern/openapi-ir/definition/example.yml
dsinghvi: would be down to inline as well 

```yaml
  PropertyKey: string
 
  FullOneOfExample:
    union:
      discriminated: map<PropertyKey, FullExample>
      undisciminated: FullExample
```

fern/openapi-ir/definition/example.yml
```diff
@@ -0,0 +1,112 @@
...
+      discriminated: FullDiscriminatedOneOfExample
+      undisciminated: FullUnDiscriminatedOneOfExample
+
+  FullDiscriminatedOneOfExample:
+    properties:
+      examples: map<string, FullExample>
+
+  FullUnDiscriminatedOneOfExample:
+    properties:
+      examples: list<FullExample>
```
dsinghvi, line 88: this should be a singleton

fern/openapi-ir/definition/example.yml
dsinghvi: looks great as is!

fern/openapi-ir/definition/example.yml
amckinney: Sweet, I'll reuse the `PropertyKey` alias you defined for the union since it would apply well here, too.

fern/openapi-ir/definition/example.yml
amckinney: Shouldn't this take the same form as the `PartialDiscriminatedOneOfExample`? Otherwise this ends up looking just like a `FullObjectExample`. I'm thinking something more along the lines of the following:

```yaml
  FullDiscriminatedOneOfExample:
    properties:
      discriminant: string
      value: FullExample
```

amckinney: Discussed offline - the full discriminated union will actually _always_ be an object, so we'll go with your suggestion here.

PR #2173, opened 2023-10-14
Initialization wording

jackfischer:
Was looking at fern for an API we may spin up in the near future and clicked from the main page to `Getting started`. I wasn't looking very closely but it appeared superficially like I had accidentally stumbled into some deeper docs and this wouldn't be the getting started guide I was looking for. Almost left the page & then I snapped out of it and identified change as probably the culprit


dsinghvi:
Thanks for the PR Jack!

PR #2178, opened 2023-10-15
move loadOpenApiFromUrl to /init

rishan10:
This function `loadOpenApiFromUrl` can be re-used in `/init`. This change moves it to the shared library under `/init`

### DIFF
--- a/packages/cli/cli/src/cli.ts
+++ b/packages/cli/cli/src/cli.ts
@@ -7,7 +7,7 @@ import {
     GENERATORS_CONFIGURATION_FILENAME,
     getFernDirectory,
     loadProjectConfig,
-    PROJECT_CONFIG_FILENAME
+    PROJECT_CONFIG_FILENAME,
 } from "@fern-api/project-configuration";
 import { loadProject, Project } from "@fern-api/project-loader";
 import { FernCliError } from "@fern-api/task-context";


packages/cli/init/src/utils/loadOpenApiFromUrl.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+class LoadOpenApiResult {
+    public msg: string;
+    public status: LoadOpenApiStatus;
+    public constructor(message = "", status: LoadOpenApiStatus) {
+        this.msg = message;
+        this.status = status;
+    }
+}
```
dsinghvi, line 20: I would use a discriminated union for this 
```
export interface LoadOpenAPIResult = SuccessOpenAPIResult | FailureOpenAPIResult; 

export interface SuccessOpenAPIResult {
  type: "success", 
  path: ""
}

export interface FailedOpenAPIResult {
  type: "failure", 
  errorMessage: ...
}
```

dsinghvi: this will force the user to visit success of error before being able to use those proeprties (can read more here about typescript discriminated unions: https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html)

packages/cli/cli/src/cli.ts
```diff
@@ -204,7 +203,12 @@ function addInitCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
                 let absoluteOpenApiPath: AbsoluteFilePath | undefined = undefined;
                 if (argv.openapi != null) {
                     if (isURL(argv.openapi)) {
-                        const tmpFilepath = await loadOpenAPIFromUrl({ url: argv.openapi, cliContext });
+                        const result = await loadOpenAPIFromUrl({ url: argv.openapi, logger: cliContext.logger });
+                        if (result.status === LoadOpenApiStatus.Failed) {
+                            cliContext.failAndThrow(result.msg);
+                        }
+
+                        const tmpFilepath = result.msg;
```
dsinghvi, line 211: this is pretty undocumented, its not clear to a reader that `result.msg` is going to have the filepath. the response should have a clear `tempFilepath` property (the discriminated unions will help with this)

packages/cli/init/src/utils/loadOpenApiFromUrl.ts
rishan10: Ahh yes, that makes it much cleaner :)

packages/cli/cli/src/cli.ts
rishan10: Thanks for the tip, just pushed a refactor!

dsinghvi (approved):
lgtm

packages/cli/init/src/utils/loadOpenApiFromUrl.ts
```diff
@@ -0,0 +1,46 @@
...
+import { join } from "path";
+import tmp from "tmp-promise";
+import { Logger } from "../../../logger/src/Logger";
+
+export type LoadOpenAPIResult = SuccessLoadOpenAPI | FailedLoadOpenAPI;
+
+export enum LoadOpenAPIStatus {
+    Success = "success",
+    Failure = "failure",
+}
```
dsinghvi, line 13: nit: not too many perks of making this an enum (literal is just fine in ts)

rishan10: i was thinking if we have more statuses down the line, it would be a safer way of defining them here

rishan10: for example, if we wanted to break failures down even further ("failed to download", "failed to locate openapi", etc)

dsinghvi: sounds good

PR #2180, opened 2023-10-15
ReadMe to Fern docs converter: convert openapi specs from readme to fern docs

rishan10:
Given a ReadMe docs url, the cli will init all the openapis (one openapi init failing won't block the rest) it can find along with an init of docs. 

## Changes:
- modified init docs to also init an openapi workspace if we pass in a docs url that contains an open api spec
- added docs utils which has an interface for docs url actions (ex: get all openapis in docs)

## Testing:
- e2e test for successful init of openapi docs

### DIFF
--- a/packages/cli/cli/src/cli.ts
+++ b/packages/cli/cli/src/cli.ts
@@ -199,9 +199,7 @@ function addInitCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
                 if (argv.url != null && !isURL(argv.url)) {
                     return cliContext.failWithoutThrowing("--url must be a valid url");
                 }
-                console.log("Initializing docs 1");
                 await cliContext.runTask(async (context) => {
-                    console.log("Initializing docs");
                     await initializeDocs({
                         organization: argv.organization,
                         versionOfCli: await getLatestVersionOfCli({ cliEnvironment: cliContext.environment }),


packages/cli/cli/src/cli.ts
```diff
@@ -187,16 +188,31 @@ function addInitCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
...
                 return cliContext.failWithoutThrowing("Cannot specify both --api and --docs. Please choose one.");
             } else if (argv.docs != null) {
+                let docsUrl: ReadMeDocsUrl;
+                if (argv.url != null) {
+                    docsUrl = new ReadMeDocsUrl(argv.url);
+                    if (!docsUrl.isUrlValid()) {
+                        return cliContext.failWithoutThrowing("--url must be a valid url");
+                    }
+                    if (argv.organization == null) {
+                        return cliContext.failWithoutThrowing("--organization must be specified when using --url");
```
dsinghvi, line 207: nit: would be sick if it just prompoted user for organization https://stackoverflow.com/questions/33858763/console-input-in-typescript

packages/cli/ete-tests/src/tests/init/init.test.ts
```diff
@@ -60,4 +60,30 @@ describe("fern init", () => {
...
+                cwd: pathOfDirectory,
+            }
+        );
+
+        const command = await runFernCli(["check"], {
+            cwd: pathOfDirectory,
+        });
+
+        expect(command.exitCode).toBe(0);
+    }, 60_000);
```
dsinghvi, line 88: looks great!

packages/cli/init/src/initializeDocs.ts
```diff
@@ -1,41 +1,105 @@
...
+
     return {
         instances: [
             {
                 url: `https://${organization}.${process.env.DOCS_DOMAIN_SUFFIX}`,
             },
         ],
         title: `${organization} | Documentation`,
-        navigation: [{ api: "API Reference" }],
+        navigation: apiSection,
```
dsinghvi, line 102: nit: you can inline `getApiSection()` here

packages/cli/init/src/utils/docsUrlUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+            new URL(this.url);
+            return true;
+        } catch (err) {
+            return false;
+        }
+    }
+}
+
+export class ReadMeDocsUrl extends DocsUrlBase {
+    type: "readme" = "readme";
```
dsinghvi, line 43: nit: would make this public readonly so it can't be modified

dsinghvi: `public readonly type = "readme";`

packages/cli/init/src/utils/docsUrlUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    async getAllOpenApiUrls(): Promise<string[]> {
+        const origin = this.getBaseUrl();
+        const response = await axios.get(origin + "/openapi");
```
dsinghvi, line 47: i would make sure you are safely joining and the leading `/` does not conflict

packages/cli/init/src/utils/docsUrlUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        if (response.status / 100 !== 2) {
+            return [];
+        }
```
dsinghvi, line 51: what is this doing?

packages/cli/init/src/utils/docsUrlUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export type DocsUrl = ReadMeDocsUrl;
+
+export interface DocsUrlInterface {
+    getAllOpenApiUrls(): Promise<string[]>;
+    getBaseUrl(): string;
+    isUrlValid(): boolean;
+}
+
+class DocsUrlBase implements DocsUrlInterface {
```
dsinghvi, line 12: what do you think of changing the naming to: 

```
class DocsInstance implements DocsInterface {

}

class ReadmeInstance extends DocsInstance {

}
```

rishan10: what about 
`class DocsImpl implements DocsInterface {

}

class ReadmeImpl extends DocsInstance {

}`? Noticed [impl](https://sourcegraph.com/github.com/fern-api/fern@add-cli-option/-/blob/packages/cli/cli/src/cli-context/TaskContextImpl.ts?L31) is used in the codebase in other places: https://sourcegraph.com/github.com/fern-api/fern@add-cli-option/-/blob/packages/cli/cli/src/cli-context/TaskContextImpl.ts?L31

packages/cli/init/src/utils/docsUrlUtils.ts
rishan10: ah yes, accomplished this with a URL builder!

packages/cli/init/src/utils/docsUrlUtils.ts
rishan10: Apologies, this was by instinct (handling all non-2XX here). I pushed a commit that should make it more clear: if we don't get a 200, then we log an error and return an empty list, signaling that we were unable to find any openapi urls

packages/cli/cli/src/cli.ts
rishan10: for sure, let me cut a ticket for this (will do in a new pr)

rishan10: https://linear.app/fern-api/issue/FER-547/prompt-user-to-input-an-organization-when-running-init-docs-url

dsinghvi (approved):

PR #2186, opened 2023-10-16
Implement OpenAPI IR -> Fern IR example converter

amckinney:
This adds the last stage of the OpenAPI -> Fern example conversion, such that `FullExample` values are mapped to their Fern-equivalent.

### DIFF
--- a/packages/cli/openapi-ir-to-fern/tsconfig.json
+++ b/packages/cli/openapi-ir-to-fern/tsconfig.json
@@ -4,6 +4,7 @@
   "include": ["./src"],
   "exclude": ["./src/**/__test__/**/generated"],
   "references": [
+    { "path": "../../commons/core-utils" },
     { "path": "../../commons/fs-utils" },
     { "path": "../config-management/project-configuration" },
     { "path": "../logger" },


packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
```diff
@@ -226,13 +229,128 @@ export function convertEndpoint({
...
+    return FullExample._visit<ExampleTypeReferenceSchema>(fullExample, {
+        primitive: (primitive) => convertPrimitiveExampleToExampleTypeReferenceSchema(primitive),
+        object: (object) => convertFullExamplePropertiesToExampleTypeReferenceSchema(object.properties),
+        array: (array) => convertFullArrayExampleToExampleTypeReferenceSchema(array),
+        map: (map) => convertFullMapExampleToExampleTypeReferenceSchema(map),
+        oneOf: (oneOf) => convertFullOneOfExampleToExampleTypeReferenceSchema(oneOf),
+        enum: (value) => value,
+        literal: (value) => value,
+        unknown: (unknown) => convertFullExampleToExampleTypeReferenceSchema(unknown),
+        _unknown: () => "unknown",
```
amckinney, line 300: @dsinghvi would you prefer that I not use the visitor and go with the `if/else` switch and throw an exception in these cases? Or is there another default value we should go with? I don't think we can assume the parsed OpenAPI result is an unknown object, i.e. `{}`, can we? This probably won't even be reachable anyways.

packages/cli/openapi-parser/src/v3/converters/example/ExampleTypeFactory.ts
```diff
@@ -49,6 +49,9 @@ export class ExampleTypeFactory {
     }
 
     private buildFromPartialObjectExample(partialObjectExample: PartialObjectExample): FullExample | undefined {
+        if (Object.keys(partialObjectExample.includedProperties).length === 0) {
+            return undefined;
+        }
```
amckinney, line 54: @dsinghvi Added this in for now so we don't produce examples without any properties. We'll want to remove this when we support generating dummy data for excluded properties so that we can create a `FullExample` from a `PartialObjectExample` with zero included properties.

packages/cli/openapi-parser/src/v3/converters/example/ExampleEndpointFactory.ts
```diff
@@ -31,6 +31,10 @@ export class ExampleEndpointFactory {
             responseSchemaIdResponse != null
                 ? this.exampleTypeFactory.buildExample(responseSchemaIdResponse.schemaInstanceId)
                 : undefined;
+        if (requestExample === undefined && responseExample === undefined) {
+            return undefined;
+        }
```
amckinney, line 36: @dsinghvi This simplifies some of the error handling in the OpenAPI IR -> Fern IR translation. Otherwise we'd need to verify that the example produced should actually be rendered. The Fern IR converter should always assume any example its given is valid / should be written.

dsinghvi: thanks for adding, makes sense

packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
dsinghvi: yeah would be okay for you to throw an exception here

dsinghvi: the `unknown` case shouldnt matter here, nothing needs to be "forward compat"

packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
```diff
@@ -226,13 +228,136 @@ export function convertEndpoint({
...
+    });
+}
+
+function convertPrimitiveExampleToExampleTypeReferenceSchema(
+    primitiveExample: PrimitiveExample
+): RawSchemas.ExampleTypeReferenceSchema {
+    return PrimitiveExample._visit<RawSchemas.ExampleTypeReferenceSchema>(primitiveExample, {
+        int: (value: number) => value.toString(),
```
dsinghvi, line 309: hmm does this actually produce valid yaml, i think you want this to an int

dsinghvi:
```yaml
example: 
  myNumber: 9
  myNumber: "9" 
```

packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
```diff
@@ -226,13 +228,136 @@ export function convertEndpoint({
...
+
+function convertFullArrayExampleToExampleTypeReferenceSchema(
+    fullExamples: FullExample[]
+): RawSchemas.ExampleTypeReferenceSchema {
+    return fullExamples.map((fullExample) => {
+        return convertFullExampleToExampleTypeReferenceSchema(fullExample);
+    });
+}
+
+function convertFullMapExampleToExampleTypeReferenceSchema(
```
dsinghvi, line 341: I think a better method name for all of these is just `convertMapExampleToFern` or something

dsinghvi: `convertFullMapExampleToExampleTypeReferenceSchema` is innacurate because you aren't converting the thing to an "exampleTypeReferenceSchema". You're converting it to an example, and the example has to be typed as that schema

packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
amckinney: Good catch, will update all the number-like cases (and boolean).

packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
amckinney: Totally up for renaming, but I chose `ExampleTypeReferenceSchema` because that's actually the name of the type we're returning from all these cases - you can see it in the signature. But I agree it's still not very clear, I like your suggestion better so let's go with that.

dsinghvi: maybe `convertMapToFernExample`? 🤷 not too opinionated but i think something shorter would be nice

packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
```diff
@@ -289,34 +289,53 @@ function convertNamedFullExamplesToExampleTypeReferenceSchemas(
...
+        case "oneOf":
+            return convertFullOneOfExampleToExampleTypeReferenceSchema(fullExample.oneOf);
+        case "enum":
+            return fullExample.enum;
+        case "literal":
+            return fullExample.literal;
+        case "unknown":
+            return convertFullExampleToExampleTypeReferenceSchema(fullExample.unknown);
+        default:
+            throw new Error("Cannot convert unsupported example type");
```
dsinghvi, line 310: i would do `assertNever(fullExample)` the ts compiler will make sure that the code never gets there

amckinney: Nice, good call 👍

packages/cli/openapi-ir-to-fern/src/converters/convertEndpoint.ts
dsinghvi: yup the type here is a zod schema, and zod helps us do validation of the raw yaml. `ExampleTypeReferenceSchema` makes sense as a name for the zod schema, but its a little confusing for this method cause we aren't returning a schema we are returning the actual example itself.

dsinghvi (approved):
🔥

PR #2188, opened 2023-10-17
Add response property option

amckinney:
This adds a response `property` option that can be used to customize what type is returned in each SDK. For example,

```yaml
types:
  Movie:
    properties:
      id: string

  Response:
    properties:
      data: Movie

service:
  auth: false
  base-path: ""
  endpoints:
    getMovie:
      method: POST
      path: /movie
      request: string
      response:
        type: Response
        property: data
```

With this, SDKs are instructed to return the `data` property defined in the `Response` rather than the response itself. This is particularly useful for APIs that conform to the [JSON API](https://jsonapi.org/) response structure, which requires that response data is returned as `{"data": ...}`.

### DIFF
--- a/packages/cli/yaml/validator/src/rules/no-response-property/no-response-property.ts
+++ b/packages/cli/yaml/validator/src/rules/no-response-property/no-response-property.ts
@@ -61,7 +61,7 @@ function resultToRuleViolations(result: Result, responseProperty: string): RuleV
             return [
                 {
                     severity: "error",
-                    message: `Response does not have a property named ${responseProperty}.`,
+                    message: "Response must be an object in order to return a property as a response.",
                 },
             ];
     }


packages/cli/yaml/yaml-schema/src/schemas/HttpResponseSchema.ts
```diff
@@ -5,6 +5,7 @@ export const HttpResponseSchema = z.union([
     z.string(),
     WithDocsSchema.extend({
         type: z.string(),
+        responseBodyProperty: z.optional(z.string()),
```
amckinney, line 8: Update this should be `returnResponseBodyProperty` (or whatever name we settle on) to match the field defined in the Fern definition.

test-definitions/fern/apis/response-property/definition/service.yml
```diff
@@ -0,0 +1,34 @@
+types:
+  WithDocs:
+    properties:
+      docs: string
+
+  Object:
+    properties:
+      id: string
+
+  Response:
```
dsinghvi, line 10: nit: would be nice to start making these examples more readable something like (e.g. naming the `Resource` something else not `Object`)

fern/ir-types-v26/definition/http.yml
```diff
@@ -145,6 +145,14 @@ types:
...
     properties:
       responseBodyType: types.TypeReference
+      responseProperty:
+        docs: |
+          If set, the SDK will return this property from
+          the response, rather than the response itself.
+
+          This is particularly useful for JSON API structures
+          (e.g. configure 'data' to return 'response.data').
+        type: optional<commons.Name>
```
dsinghvi, line 155: @amckinney i dont want to make you redo a bunch of work -- but i think the right way to represent this in the IR is a union 

```
JsonResponse: 
  union: 
    response: 
    propertyResponse
```
The visitor pattern would force the generators to visit but adding an optional property would certainly be missed by a generator author on upgrade

dsinghvi: This would mean IRV27, a reverse migration, and updating the generateIntermediateRepresentation code

packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
```diff
@@ -42,6 +42,28 @@ export function convertHttpResponse({
...
+function convertJsonResponse(
+    response: RawSchemas.HttpResponseSchema | string,
+    docs: string | undefined,
+    file: FernFileContext,
+    typeResolver: TypeResolver
+): HttpResponse {
+    const responseBodyType = file.parseTypeReference(response);
+    const responseProperty = typeof response !== "string" ? response.property : undefined;
+    if (
+        responseProperty !== undefined &&
```
dsinghvi, line 54: nwe use `!= null` checks everywhere in case the value is ever null at runtime (!= null handles checking null and undefined)

packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
```diff
@@ -53,3 +75,95 @@ function constructStreamingResponseChunkType(
...
+        switch (resolvedType._type) {
+            case "container":
+                return typeReferenceHasProperty(resolvedType.originalTypeReference, property, file, typeResolver);
+            case "named":
+                if (isRawObjectDefinition(resolvedType.declaration)) {
+                    return rawObjectSchemaHasProperty(resolvedType.declaration, property, file, typeResolver);
+                }
+                throw new Error("A custom response property is only supported for objects");
+        }
+    }
```
dsinghvi, line 102: do we need to handle aliases here?

packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
```diff
@@ -53,3 +75,95 @@ function constructStreamingResponseChunkType(
         return StreamingResponseChunkType.json(file.parseTypeReference(typeReference));
     }
 }
+
+function typeReferenceHasProperty(
```
dsinghvi, line 79: historically what we have done is make this a `fern check` rule and then assume its all good when we get here

dsinghvi: this way we can keep validation somewhat separate from `generateIntermediateRepresentation.ts` and we don't have to reimplement -- since you have the code here do you just want to convert this into a fern check rule right now?

packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
amckinney: Aliases are already handled by the `TypeResolver` we're using - it only ever returns the underlying type. You can see that handled [here](https://github.com/fern-api/fern/blob/ff01eefce12d16d0685bc3146cd7fa84038dc82e/packages/cli/generation/ir-generator/src/resolvers/TypeResolver.ts#L256).

packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
dsinghvi: it looks great!

packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
dsinghvi: thanks!

packages/cli/yaml/validator/src/rules/no-response-property/no-response-property.ts
```diff
@@ -0,0 +1,140 @@
...
+            file,
+        });
+        switch (resolvedType._type) {
+            case "container":
+                return typeReferenceHasProperty(resolvedType.originalTypeReference, property, file, typeResolver);
+            case "named":
+                if (isRawObjectDefinition(resolvedType.declaration)) {
+                    return rawObjectSchemaHasProperty(resolvedType.declaration, property, file, typeResolver);
+                }
+                throw new Error("A custom response property is only supported for objects");
```
dsinghvi, line 71: @amckinney if we throw an error here this will cause all of the fern cli to stop and the user will get no output

dsinghvi: i think we should `return false` or the entire function should be updated to return an enum like: 
```
enum Result {
  ContainsProperty,
  DoesNotContainProperty,
  IsNotObject
}


```

dsinghvi: based on the enum the error message can be clearer too

packages/cli/yaml/validator/src/rules/no-response-property/no-response-property.ts
```diff
@@ -29,30 +29,50 @@ export const NoResponsePropertyRule: Rule = {
...
+                {
+                    severity: "error",
+                    message: `Response does not have a property named ${responseProperty}.`,
+                },
+            ];
+        case Result.IsNotObject:
+            return [
+                {
+                    severity: "error",
+                    message: `Response does not have a property named ${responseProperty}.`,
```
dsinghvi, line 64: my guess is you want the error message to be different here, but not too opinionated. 

Something like: `Response must be an object in order to return a property as a response.` ?

dsinghvi (approved):

packages/cli/yaml/validator/src/rules/no-response-property/no-response-property.ts
amckinney: Gah, my mistake (again)! I actually meant to use the old error message we had here:

```
throw new Error("A custom response property is only supported for objects");
```

I like your error message better tho, so I'll update to that 👍

packages/cli/yaml/validator/src/rules/no-response-property/no-response-property.ts
amckinney: Much better, updated 👍

PR #2205, opened 2023-10-25
fix: Support response property for imported types

amckinney:
This fixes an issue where custom response properties were not resolved when the types were imported from another package. This was primarily caused from unnecessarily converting between a `TypeReference` to a `ResolveType` - now we resolve the `ResolvedType` directly from the input using `TypeResolver.resolveType`.

### DIFF
--- a/packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
+++ b/packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
@@ -1,3 +1,4 @@
+import { assertNever } from "@fern-api/core-utils";
 import { isRawTextType, parseRawFileType, parseRawTextType, RawSchemas } from "@fern-api/yaml-schema";
 import { HttpResponse, JsonResponse, ObjectProperty, StreamingResponseChunkType } from "@fern-fern/ir-sdk/api";
 import { FernFileContext } from "../../FernFileContext";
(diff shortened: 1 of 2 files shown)


packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
```diff
@@ -91,38 +85,25 @@ function convertJsonResponse(
...
+            }
+            break;
+        case "named":
+            if (isRawObjectDefinition(resolvedType.declaration)) {
+                return getObjectPropertyFromObjectSchema(resolvedType.declaration, property, file, typeResolver);
+            }
+            break;
     }
-    throw new Error(`Response does not have a property named ${property}.`);
+    throw new Error("Internal error; response must be an object in order to return a property as a response");
```
dsinghvi, line 106: @amckinney i missed this in the initiali review, but thoughts on handling every case and then adding an assertNever -- 

```
case "container": 
 return ...
case "optional":
  return ...
case "other1"
case "other2"
  throw new Error()
default: 
  assertNever();  
```
This will force us to handle the new cases if we ever add to the union

dsinghvi (approved):

packages/cli/generation/ir-generator/src/converters/services/convertHttpResponse.ts
amckinney: Great call, added 👍

PR #2227, opened 2023-11-02
Add support for boolean literals

amckinney:
This adds support for literals in the form of `literal<false>` and `literal<true>` and introduces them in IRv28. 

In terms of migrating between v28 -> v27, all boolean literals are converted into `boolean` properties. This means that any user that specifies a boolean literal in their Fern definition will still have access to a boolean property (regardless of whether or not the SDK has added supported for boolean literals).

Note that this change does not add support for converting OpenAPI sources to boolean literals. The only place that a literal is used today is for single element enums ([link](https://github.com/fern-api/fern/blob/ac7a0cd602688c4364db3fe1ce18f113f5e1020d/packages/cli/openapi-parser/src/v3/converters/convertSchemas.ts#L96)). In other words, there aren't any use cases for them yet (but there will be soon).

There's also a `TODO` left in the translation between the IR types and the FernRegistry types - we'll need to add support for boolean literals in the FernRegsitry API and update the converter.


packages/cli/register/src/ir-to-fdr-converter/convertTypeShape.ts
```diff
@@ -113,6 +113,11 @@ export function convertTypeReference(
                 },
                 literal: (literal) => {
                     return Ir.types.Literal._visit(literal, {
+                        boolean: (booleanLiteral) =>
+                            // TODO: Refactor this when the FernRegsitry API supports boolean literals.
+                            FernRegistry.api.v1.register.TypeReference.literal(
+                                FernRegistry.api.v1.register.LiteralType.stringLiteral(booleanLiteral.toString())
+                            ),
```
amckinney, line 120: Here's the `TODO` for the FernRegistry - for now we just convert the boolean into a string, but we'll need to update this so that we actually capture it as a boolean literal.

packages/cli/generation/ir-migrations/src/migrations/v28-to-v27/__test__/migrateFromV28ToV27.test.ts
```diff
@@ -0,0 +1,14 @@
...
+import { V28_TO_V27_MIGRATION } from "../migrateFromV28ToV27";
+
+const runMigration = createMigrationTester(V28_TO_V27_MIGRATION);
+
+describe("migrateFromV28ToV27", () => {
+    it("maps boolean literals to boolean types", async () => {
+        const migrated = await runMigration({
+            pathToFixture: join(AbsoluteFilePath.of(__dirname), RelativeFilePath.of("./fixtures/literal")),
+        });
+        expect(migrated).not.toBe(null);
```
amckinney, line 12: This test could do a lot more (i.e. actually verify that the properties were mapped into a `PrimitiveType.Boolean` type reference. In its current state it's just verifying that migration runs without failure.

dsinghvi: It would be nice to do a little more verification here just because this logic will run "forever".

dsinghvi:
@amckinney looks great overall, thoughts on creating a `test-definition` that uses boolean literals and we can add it to `seed` as well?

packages/cli/generation/ir-migrations/src/migrations/v28-to-v27/__test__/migrateFromV28ToV27.test.ts
amckinney: Great call, done

packages/cli/register/src/ir-to-fdr-converter/convertTypeShape.ts
amckinney: Discussed offline - made the change in the FernRegistry, but we'll follow-up to migrate to the latest SDK in a follow-up.

dsinghvi (approved):

PR #2229, opened 2023-11-02
Added in property filtering based on audiences

AlexKrupka:


dannysheridan:
closing as we're not moving forward on this right now

PR #2230, opened 2023-11-02
Update snapshot to include new field

tjb9dc:
Fix to https://github.com/fern-api/fern/pull/2228


tjb9dc:
reopening on a not-stale view

PR #2247, opened 2023-11-07
Extend x-fern-streaming functionality

amckinney:
This extends the `x-fern-streaming` OpenAPI extension syntax so that it lets users define an endpoint that has both a unary and streaming version of the same method, which is distinguished with a boolean property set to `true` or `false`, which is beginning to be a common usage pattern seen in several (optionally) streaming APIs.

Example usage is shown below:

```yaml
paths:
  /v1/users:
    post:
      operationId: user
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                id:
                  description: ""
                  type: string
      x-fern-streaming:
        stream-condition: stream
        response: "#/components/schemas/User"
        response-stream: "#/components/schemas/StreamUser"
      responses:
        "200":
          description: ""
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/User"
components:
  schemas:
    User:
      type: object
      properties:
        id:
          type: string
          format: uuid
    StreamUser:
      type: object
      properties:
        id:
          type: string
          format: uuid

```


packages/cli/openapi-parser/src/v3/converters/addStreamConditionProperty.ts
```diff
@@ -0,0 +1,77 @@
...
+            // the objectSchema.properties in-place or return
+            // a new set of properties.
+            property.schema = SchemaWithExample.literal({
+                description: property.schema.description,
+                value: LiteralSchemaValue.boolean(isStreaming),
+            });
+            return {
+                type: "object",
+                ...objectSchema,
+            };
```
amckinney, line 59: Need to fix this mutability issue - the properties aren't updated in the `for ... of ...)` loop. Can verify by adding another test case to the `streaming` fixtures.

packages/cli/openapi-parser/src/v3/converters/convertPathItem.ts
```diff
@@ -409,20 +471,43 @@ function convertToEndpoint({
...
+                content: {
+                    "application/json": {
+                        schema: {
+                            $ref: ref,
+                        },
+                    },
+                },
+            },
+        };
+    }
```
amckinney, line 504: Need to fix this - it's hard-coded as-is. The `code` and `content-type` values are variable strings, so we'll need to map those to the new `operation.responses` value, but it isn't sufficient to just reference the `code` and `cotentType` properties.

packages/cli/openapi-parser/src/v3/extensions/getStreaming.ts
```diff
@@ -0,0 +1,43 @@
...
+    streamConditionProperty: string;
+    responseStream: string;
+    response: string;
+}
+
+export interface StreamConfig {
+    ["stream-condition"]: string;
+    ["response-stream"]: string;
+    response: string;
+}
```
amckinney, line 22: We should probably just rename these `Stream`, `StreamEndpoint`, and `StreamConditionEndpoint` (or some other simpler alternative).

packages/cli/openapi-parser/src/v3/extensions/fernExtensions.ts
```diff
@@ -3,14 +3,39 @@ import { Values } from "@fern-api/core-utils";
...
+     * a boolean propetry that tells the server whether
+     * or not the response should be streamed or not.
+     *
+     * Example usage:
+     *   paths:
+     *     /path/to/my/endpoint:
+     *       x-fern-streaming:
+     *         stream-condition: stream
+     *         response: ./path/to/response/type.yaml
+     *         response-stream: ./path/to/response-stream/type.yaml
```
amckinney, line 35: I haven't actually tested this with file-based references yet, which is what we will see in practice.

amckinney:
Superseded by https://github.com/fern-api/fern/pull/2251

PR #2251, opened 2023-11-07
Generate streaming endpoints that specify stream-condition

amckinney:
This extends the `x-fern-streaming` OpenAPI extension syntax so that it lets users define an endpoint that has both a unary and streaming version of the same method, which is distinguished with a boolean property set to true or false, which is beginning to be a common usage pattern seen in several (optionally) streaming APIs.

Example usage is shown below:

```yaml
paths:
  /v1/users:
    post:
      operationId: user
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                id:
                  description: ""
                  type: string
      x-fern-streaming:
        stream-condition: $request.stream
        response:
          $ref: "#/components/schemas/User"
        response-stream:
          $ref: "#/components/schemas/StreamUser"
      responses:
        "200":
          description: ""
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/User"
components:
  schemas:
    User:
      type: object
      properties:
        id:
          type: string
          format: uuid
    StreamUser:
      type: object
      properties:
        id:
          type: string
          format: uuid
```

### DIFF
--- a/packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/streaming.test.ts.snap
+++ b/packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/streaming.test.ts.snap
@@ -44,7 +44,7 @@ exports[`open api parser streaming simple 1`] = `
             },
             "response": {
               "docs": "",
-              "type": "UserResponse",
+              "type": "User",
             },
           },
           "user_stream": {
@@ -69,7 +69,7 @@ exports[`open api parser streaming simple 1`] = `
             },
             "response-stream": {
               "docs": "",
-              "type": "UserStreamResponse",
+              "type": "StreamUser",
             },
           },
           "user_v2": {
@@ -109,7 +109,7 @@ exports[`open api parser streaming simple 1`] = `
             },
             "response": {
               "docs": "",
-              "type": "UserV2Response",
+              "type": "User2",
             },
           },
           "user_v2_stream": {
@@ -134,7 +134,7 @@ exports[`open api parser streaming simple 1`] = `
             },
             "response-stream": {
               "docs": "",
-              "type": "UserV2StreamResponse",
+              "type": "StreamUser2",
             },
           },
           "user_v3": {
@@ -212,51 +212,7 @@ exports[`open api parser streaming simple 1`] = `
             },
           },
         },
-        "User": {
-          "docs": undefined,
-          "properties": {
-            "id": {
-              "docs": undefined,
-              "type": "optional<string>",
-            },
-          },
-        },
-        "UserResponse": {
-          "docs": undefined,
-          "properties": {
-            "id": {
-              "docs": undefined,
-              "type": "optional<string>",
-            },
-          },
-        },
-        "UserStreamResponse": {
-          "docs": undefined,
-          "properties": {
-            "id": {
-              "docs": undefined,
-              "type": "optional<string>",
-            },
-          },
-        },
-        "UserV2Response": {
-          "docs": undefined,
-          "properties": {
-            "id": {
-              "docs": undefined,
-              "type": "optional<string>",
-            },
-            "name": {
-              "docs": undefined,
-              "type": "optional<string>",
-            },
-            "streamed": {
-              "docs": undefined,
-              "type": "optional<literal<false>>",
-            },
-          },
-        },
-        "UserV2StreamResponse": {
+        "StreamUser2": {
           "docs": undefined,
           "properties": {
             "id": {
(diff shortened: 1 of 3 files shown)


packages/cli/openapi-parser/src/v3/converters/convertSchemas.ts
```diff
@@ -148,6 +148,13 @@ export function convertSchemaObject(
 
     // primitive types
     if (schema === "boolean" || schema.type === "boolean") {
+        const literalValue = getExtension<boolean>(schema, FernOpenAPIExtension.BOOLEAN_LITERAL);
+        if (literalValue !== undefined) {
+            return SchemaWithExample.literal({
+                value: LiteralSchemaValue.boolean(literalValue),
+                description,
+            });
```
dsinghvi, line 156: we're going to want to call something like `wrapPrimitive`

dsinghvi: `wrapPrimitive` handles wrapping in nulllable/optional depending on what has been specified

packages/cli/openapi-parser/src/v3/converters/operation/convertStreamingOperation.ts
```diff
@@ -31,10 +35,191 @@ export function convertStreamingOperation({
...
+                    ...operationContext,
+                    operation: {
+                        ...operationContext.operation,
+                        requestBody: convertRequestBodyToInlinedRequest({
+                            context,
+                            requestBody: operationContext.operation.requestBody,
+                            streamCondition: streamingExtension,
+                            isStreaming: true,
+                        }),
+                    },
```
dsinghvi, line 50: you're probably already tracking but you'll have to also create a similar helper `createOperationResponse(streamingExtension.responseStream)`

dsinghvi: basically this helper will return 
```ts
{
  "200": {
    "application/json": {
        "schema": responseSchema
    }
  }
}
```

amckinney: Yup, tracking that but thanks for calling that out - need to port the implementation over from what we have in https://github.com/fern-api/fern/pull/2247

packages/cli/openapi-parser/src/v3/converters/operation/convertStreamingOperation.ts
```diff
@@ -31,10 +35,191 @@ export function convertStreamingOperation({
...
         }
-        case "streamCondition":
-            break;
+        case "streamCondition": {
+            const streamingOperation = convertHttpOperation({
+                operationContext: {
+                    ...operationContext,
+                    operation: {
+                        ...operationContext.operation,
+                        requestBody: convertRequestBodyToInlinedRequest({
```
dsinghvi, line 44: thoughts on renaming this to `createStreamingRequestBody({ context, requestBody, streamConditionProperty })`; 

`inlinedRequest` is more a fern term so figured we could remove that

amckinney: Totally fine with that, let's do it

packages/cli/openapi-parser/src/v3/converters/operation/convertStreamingOperation.ts
```diff
@@ -31,10 +35,191 @@ export function convertStreamingOperation({
...
+    requestBody,
+    streamCondition,
+    isStreaming,
+}: {
+    context: AbstractOpenAPIV3ParserContext;
+    requestBody?: OpenAPIV3.ReferenceObject | OpenAPIV3.RequestBodyObject;
+    streamCondition: StreamConditionEndpoint;
+    isStreaming: boolean;
+}): OpenAPIV3.RequestBodyObject {
+    if (requestBody === undefined) {
```
dsinghvi, line 93: `== null`

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/streaming.test.ts.snap
```diff
@@ -0,0 +1,338 @@
...
+            "request": {
+              "body": {
+                "properties": {
+                  "id": {
+                    "docs": "",
+                    "type": "optional<string>",
+                  },
+                  "stream": {
+                    "docs": undefined,
+                    "type": "optional<literal<false>>",
```
amckinney, line 40: This shouldn't be optional - the literal we synthesize from `x-fern-streaming` should be sent unconditionally.

amckinney: Done in https://github.com/fern-api/fern/pull/2251/commits/1b763b2f791a071504c9f9a7aa05a668f39400f8

dsinghvi (approved):

PR #2273, opened 2023-11-13
Add ExampleUnionType to IR

amckinney:
This proposes several changes to a few of the example IR types, including:
- Adds a new `ExampleUnionType`.
- Adds the new EscapedString type particularly suited for string literals that contain characters that must be escaped (e.g. `{"foo": "bar"}`).
- Refactors path parameters, query parameters, and headers to include the full `commons.Name` type rather than just their respective wire representations.

We may want to add even more information to the example IR later, but this will at least bridge all the gaps found in the current snippet implementation.

### DIFF
--- a/packages/cli/ete-tests/src/tests/fdr/__snapshots__/fdr.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/fdr/__snapshots__/fdr.test.ts.snap
@@ -512,7 +512,7 @@ exports[`fdr {"name":"simple"} 1`] = `
                         {
                             "path": "/movies/id-123",
                             "pathParameters": {
-                                "undefined": "id-123"
+                                "movieId": "id-123"
                             },
                             "queryParameters": {
                                 "movieName": "hello"
@@ -528,7 +528,7 @@ exports[`fdr {"name":"simple"} 1`] = `
                         {
                             "path": "/movies/id-123",
                             "pathParameters": {
-                                "undefined": "id-123"
+                                "movieId": "id-123"
                             },
                             "queryParameters": {
                                 "movieName": "hello"


fern/ir-types-v29/definition/http.yml
```diff
@@ -207,17 +207,17 @@ types:
...
   ExampleQueryParameter:
     properties:
-      wireKey: string
+      key: commons.Name
       value: types.ExampleTypeReference
 
   ExampleHeader:
     properties:
-      wireKey: string
+      key: commons.Name
```
dsinghvi, line 220: what does `Name` give you that you previously didn't have before?

dsinghvi: `NameAndWireValue`

fern/ir-types-v29/definition/http.yml
```diff
@@ -236,9 +236,13 @@ types:
...
       originalTypeDeclaration:
         docs: |
-          this property may have been brought in via extension. originalTypeDeclaration
+          This property may have been brought in via extension. originalTypeDeclaration
           is the name of the type that contains this property
         type: optional<types.DeclaredTypeName>
+      property:
+        docs: |
+          The property associated with the object type declaration.
+        type: ObjectProperty
```
dsinghvi, line 245: `NameAndWireValue`

fern/ir-types-v29/definition/types.yml
```diff
@@ -184,38 +184,43 @@ types:
...
         docs: |
-          this property may have been brought in via extension. originalTypeDeclaration
-          is the name of the type that contains this property
+          This property may have been brought in via extension. originalTypeDeclaration
+          is the name of the type that contains this property.
         type: DeclaredTypeName
+      property:
+        docs: |
+          The property associated with the object type declaration.
+        type: ObjectProperty
```
dsinghvi, line 213: delete property since you have `NameAndWireValue`

fern/ir-types-v29/definition/types.yml
```diff
@@ -288,3 +293,8 @@ types:
     properties:
       typeName: DeclaredTypeName
       shape: ExampleTypeShape
+
+  ExampleNamedObjectType:
+    properties:
+      typeName: DeclaredTypeName
```
dsinghvi, line 299: `typeId`

amckinney: We can actually just use `ExampleObjectType` directly.

fern/ir-types-v29/definition/types.yml
```diff
@@ -266,7 +271,7 @@ types:
         type: double
         key: double
       string:
-        type: string
+        type: commons.SafeAndUnsafeString
```
dsinghvi, line 274: `EscapedExampleString`

dsinghvi: Decision: turn into object for now

```yaml
   wireString: 
```

fern/ir-types-v29/definition/types.yml
```diff
@@ -184,38 +184,43 @@ types:
...
+      value: commons.NameAndWireValue
 
   ExampleObjectType:
     properties:
       properties: list<ExampleObjectProperty>
 
   ExampleObjectProperty:
     properties:
-      wireKey: string
+      key: commons.NameAndWireValue
```
dsinghvi, line 203: rename all `NameAndWireValue` to `name`

fern/ir-types-v30/definition/http.yml
```diff
@@ -0,0 +1,255 @@
...
+      endpointPathParameters: list<ExamplePathParameter>
+      serviceHeaders: list<ExampleHeader>
+      endpointHeaders: list<ExampleHeader>
+      queryParameters: list<ExampleQueryParameter>
+      request: optional<ExampleRequestBody>
+      response: ExampleResponse
+
+  ExamplePathParameter:
+    properties:
+      name: commons.NameAndWireValue
```
amckinney, line 210: Path parameters don't _really_ have a wire representation - they're just placeholders for the value that's actually sent on the wire (e.g. `/user/{id}` => `/user/123`). However, we could use `commons.NameAndWireValue` here for consistency with the other endpoint parameters (e.g. query parameters and headers), and simply set the `wireValue` to the same thing as the `name`'s original value.

There might even be a use case later to have this representation of the "wire value", such as including the default path structure vs. another name used in the SDKs. For example, we might want to include `/user/{id}` somewhere, and treat `id` as the "wire value", but the example snippet might want to use a `userId` parameter name in the generated SDK signature (which would be the "name").

Alternatively, we just use `commons.Name` here and drop the wire value altogether.

fern/ir-types-v29/definition/types.yml
amckinney: I went with the following structure, using `original` instead of `wireString` because these values aren't necessarily sent on the wire.

```yaml
  EscapedString:
    docs: |
      Defines the original string, and its escaped-equivalent (depending on the target programming language).
      This is paricularly relevant to example string literals.
      For example, in Python we escape strings that contain single or double quotes by using triple quotes,
      in Go we use backticks, etc.
    properties:
      original: string
```

Let me know what you think!

fern/ir-types-v30/definition/http.yml
dsinghvi: > Alternatively, we just use commons.Name here and drop the wire value altogether.

lets do this!

fern/ir-types-v29/definition/types.yml
dsinghvi: sounds great

dsinghvi (approved):

packages/cli/generation/ir-migrations/src/migrations/v30-to-v29/migrateFromV30ToV29.ts
```diff
@@ -0,0 +1,357 @@
+import { assertNever } from "@fern-api/core-utils";
+import { GeneratorName } from "@fern-api/generators-configuration";
+import { IrSerialization } from "../../ir-serialization";
```
dsinghvi, line 3: @amckinney you're going to want to update the exports from `IrSerialization`

fern/ir-types-v30/definition/api.yml
```diff
@@ -0,0 +1,4 @@
+name: ir
+docs: |
+  Changes in this version of IR include:
+  - Several improvements to Example types, including the new ExampleUnionType.
```
dsinghvi, line 4: would you be up to provide a little more detail here? the ts generator will need to be upgraded so would be nice to have some more idea

packages/cli/generation/ir-migrations/src/migrations/v30-to-v29/migrateFromV30ToV29.ts
amckinney: Great catch, updated 👍

PR #2277, opened 2023-11-15
[seed] Add more granular test definitions

amckinney:
This adds more test definitions for all of Fern's model types (e.g. alias, enums, objects, etc), and includes a variety of examples for each type. 

We'll also want to add more granular tests for SDK generators (e.g. services with just query parameters, path parameters, etc), but this starts to move things in the right direction, one piece at a time.


dsinghvi (approved):

dsinghvi:
@amckinney should we delete `exhaustive`?

amckinney:
> @amckinney should we delete `exhaustive`? could also see doing separately

Let's delete `exhaustive` once we get all the granular SDK tests that we're happy with and we're confident everything that was previously covered is covered in the new tests.

PR #2303, opened 2023-11-19
fix: Remove hard failure on `fern check` when endpoints overlap

### LINKED ISSUE #2258: [Bug] OpenAPI importer hard failure when "endpoint name" overlaps
**Describe the bug**
When i ran `fern check` with my OpenAPI spec I got

```
The OpenAPI Spec has conflicting sdk method names. See POST /my/path
```

This error is hard to fix because it doesn't list what endpoints are conflicting + its probably better to emit a warning instead of a hard failure. 

**To Reproduce**

Run `fern check` on https://github.com/Propify-API/propify-docs/blob/74654c181eeff250e49ed43d4a227b0183ba567c/fern/openapi/openapi.yml

**Additional context**
As an initial step, improving the error messaging and suggesting remediation would be better.

charliettaylor:
## Changes
- Pass context into service converter to log error instead of throwing exception
- Update error message to include which endpoint is being overlapped

## Testing
- I had some issues with the Docker tests locally, but passes everything outside of that
- Added ete test to make sure `fern generate` logs the overlapping methods

Resolves #2258


packages/cli/openapi-ir-to-fern/src/converters/convertToServices.ts
```diff
@@ -47,10 +50,11 @@ export function convertToServices({
...
         if (service != null) {
             if (endpointId in service.value.endpoints) {
-                throw new Error(
-                    `The OpenAPI Spec has conflicting sdk method names. See ${endpoint.method.toUpperCase()} ${
+                const firstEnd = endpoints.find((e) => e.sdkName?.methodName === endpointId);
+                context.logger.error(
+                    `The OpenAPI Spec has conflicting SDK method names. See ${endpoint.method.toUpperCase()} ${
                         endpoint.path
-                    }`
+                    }, ${firstEnd?.method.toUpperCase()} ${firstEnd?.path} already has this method name.`
```
dsinghvi, line 57: looks great -- @charliettaylor if you're interested in making a larger change, it would be nice to be able to point to line numbers

dsinghvi (approved):

packages/cli/openapi-ir-to-fern/src/converters/convertToServices.ts
dsinghvi, line 57: Should we log all the conflicting methods + paths, it might be confusing to only log one?

packages/cli/openapi-ir-to-fern/src/converters/convertToServices.ts
```diff
@@ -47,10 +50,11 @@ export function convertToServices({
         const service = services[file];
         if (service != null) {
             if (endpointId in service.value.endpoints) {
-                throw new Error(
-                    `The OpenAPI Spec has conflicting sdk method names. See ${endpoint.method.toUpperCase()} ${
+                const firstEnd = endpoints.find((e) => e.sdkName?.methodName === endpointId);
+                context.logger.error(
+                    `The OpenAPI Spec has conflicting SDK method names. See ${endpoint.method.toUpperCase()} ${
```
dsinghvi, line 55: May be more helpful to log something like: 

```
Multiple endpoints have conflicting names. See below: 
  POST /users has the name `getUser` (line 123)
  PUT /users has the name `getUser` (line 123)
You can change the names using operationId or the extension `x-fern-sdk-method-name`. 
```

charliettaylor: Gotcha, yeah I update this to group all of the conflicts together with more details

dsinghvi: Sounds good -- my guess is that line number is too hard and can follow up with that later

dsinghvi: And then lets add a test to test this case in the `ete-tests` or the `openapi-ir-to-fern` tests

charliettaylor: I updated the conflicts to format like below, and added a test to `ete-tests`
```
Multiple endpoints have conflicting names for 'hello':
- GET /hello
- GET /hellov2
Multiple endpoints have conflicting names for 'goodbye':
- GET /goodbye
- GET /goodbyev2
You can change the names using operationId or the extension `x-fern-sdk-method-name`.
```

PR #2315, opened 2023-11-21
fix: docs preview server returns the proper load docs by url response

dsinghvi:
In this PR, `fern preview` will spin up a server to mimic the docs backend at `localhost:3000`.


dsinghvi:
@abvthecity once this PR is merged, when you run `fern preview` you should get the expected response shape + data.

PR #2348, opened 2023-11-26
documentation: add audiences

dannysheridan:
This change defines how to use `audiences` on:
- endpoints
- types
- properties


dannysheridan:
My VSCode is auto-formatting this codeblock into a single line:

```
<CodeBlock title="api.yml">```yaml name: api audiences: - internal - beta ```</CodeBlock>
```

Can't figure out how to prevent it from doing so. VSCode settings for format on save are `off`.

PR #2351, opened 2023-11-26
feature: seed CLI runs compile commands for verification

dsinghvi:
The Seed CLI now supports running arbitrary commands on the generated source code for verification. The generator author can specify the docker container and a list of commands to run and seed will run those commands as part of the test. 

Here is a sample of the config that the generator author may use: 
```yaml
scripts:
  - docker: sunpeek/poetry:py3.11-slim
    commands:
      - poetry install
      - poetry run mypy .
```


dsinghvi:
@amckinney tagging you so that you can leverage in the generators you work on!

PR #2369, opened 2023-11-28
Fix bug where Fern fails to parse nested maps

### LINKED ISSUE #2316: [Bug] fern fails to parse nested map `map<string, map<string, int>>`
**Describe the bug**
When there is a type with a nested map, `map<string, map<string, Baz>>`, fern fails to parse this and fails with the errors

```bash
[api]: __package__.yml -> types -> SingleUserResourceResponse -> properties -> provider -> type
       Type string, map<string is not defined.
[api]: __package__.yml -> types -> SingleUserResourceResponse -> properties -> provider -> type
       Type Baz> is not defined.
```

**To Reproduce**
Run `fern check` on the following Fern definition

```
# baz.yml
types: 
    MyType:
      properties:
        bar:
          type: map<string, map<string, baz>>
```

**Expected behavior**
Fern should parse the nested map appropriately.

mmolash:
Fixes #2316.

Fixed by updating the regex that matches map types and captures the key/value.

Previously this regex would grab all characters up to the final comma as the first capture group (the key). In the case of `map<string, map<string, int>>`, this would capture `string, map<string` as the first group and `int>` as the second group.

This PR updates the first capture group to look for all characters _except_ for `,`, meaning it will necessarily stop capturing the key once it encounters the first comma. In the case of `map<string, map<string, int>>`, it now captures `string` as the first group and `map<string, int>` as the second group.

### DIFF
--- a/packages/cli/generation/ir-generator/src/__test__/__snapshots__/exhaustive.txt
+++ b/packages/cli/generation/ir-generator/src/__test__/__snapshots__/exhaustive.txt
@@ -1692,6 +1692,7 @@ exports[`generate IR 1`] = `
       "type_types/object:ObjectWithRequiredField",
       "type_types/enum:WeatherReport",
       "type_types/object:ObjectWithOptionalField",
+      "type_types/object:ObjectWithMapOfMap",
       "type_types/object:NestedObjectWithOptionalField",
       "type_types/object:NestedObjectWithRequiredField",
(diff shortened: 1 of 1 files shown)


dsinghvi:
@mmolash you'll need to update some of the tests in the `ir-generator` package. You can do that by running 
`yarn workspace @fern-api/ir-generator test -u`

PR #2390, opened 2023-11-29
improvement: don't run CI unnecessarily on docs

dannysheridan:
It'll save a contributor to Fern's documentation ~4 minutes of waiting to remove these CI checks that are unnecessary when only docs are being changed.


dannysheridan:
for example, this PR is blocked on failed tests: https://github.com/fern-api/fern/pull/2393

PR #2401, opened 2023-11-30
[seed] Add optional output version configuration

amckinney:
Re: https://github.com/fern-api/fern-go/issues/97#issuecomment-1834008812

This adds a new `outputVersion` configuration so that individual seed fixture tests can test against different release versions (e.g. `v1.0.0` vs. `v2.0.0`).

### DIFF
--- a/packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
+++ b/packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
@@ -15,8 +15,6 @@ import tmp, { DirectoryResult } from "tmp-promise";
 import { LocalTaskHandler } from "./LocalTaskHandler";
 import { runGenerator } from "./run-generator/runGenerator";
 
-const DEFAULT_OUTPUT_VERSION = "0.0.1";
-
 export async function runLocalGenerationForWorkspace({
     organization,
     workspace,
@@ -191,7 +189,7 @@ async function writeFilesToDiskAndRunGenerator({
         absolutePathToWriteConfigJson,
         workspaceName: workspace.name,
         organization,
-        outputVersion: outputVersionOverride ?? DEFAULT_OUTPUT_VERSION,
+        outputVersion: outputVersionOverride,
         keepDocker,
         generatorInvocation
     });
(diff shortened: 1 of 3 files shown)


packages/seed/src/commands/test/testCustomFixture.ts
```diff
@@ -24,6 +25,7 @@ export async function testCustomFixture({
     workspace: SeedWorkspace;
     generatorType: GeneratorType;
     irVersion: string | undefined;
+    outputVersion: string | undefined;
```
dsinghvi, line 28: looks like we always pass `undefined` here -- b/c we do that thoughts on removing the arg and just passing it as undefined to `acquireLocksAndRunTest` ?

packages/seed/src/commands/test/testWorkspaceFixtures.ts
```diff
@@ -106,6 +106,10 @@ export async function testWorkspaceFixtures({
                         lock,
                         generatorType,
                         irVersion,
+                        outputVersion:
+                            fixtureConfigInstance.outputVersion != null
+                                ? fixtureConfigInstance.outputVersion
+                                : undefined,
```
dsinghvi, line 112: should just be able to do `outputVersion: fixtureConfigInstance.outputVersion`

packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
```diff
@@ -183,6 +191,7 @@ async function writeFilesToDiskAndRunGenerator({
         absolutePathToWriteConfigJson,
         workspaceName: workspace.name,
         organization,
+        outputVersion: outputVersionOverride != null ? outputVersionOverride : DEFAULT_OUTPUT_VERSION,
```
dsinghvi, line 194: `outputVersionOverride ?? DEFAULT_OUTPUT_VERSION`

packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
```diff
@@ -143,6 +150,7 @@ async function writeFilesToDiskAndRunGenerator({
     keepDocker: boolean;
     context: TaskContext;
     irVersionOverride: string | undefined;
+    outputVersionOverride: string | undefined;
```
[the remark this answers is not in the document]
dsinghvi: not sure if thats possible just asking -- having `override` fields feels like its going to cause hard to root cause bugs at some point

PR #2407, opened 2023-11-30
fix: example properties for imported types are properly serialized

dsinghvi:
https://github.com/fern-api/fern/pull/2404 introduced some issues with parsing example types that have been imported. This PR fixes those introduced bugs.

### DIFF
--- a/packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
+++ b/packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
@@ -550,10 +550,6 @@ function convertSingleUnionType({
             if (!isPlainObject(example)) {
                 throw new Error("Example is not an object");
             }
-            const resolvedType = typeResolver.resolveTypeOrThrow({
-                type: rawValueType,
-                file: fileContainingType
-            });
             return {
                 wireDiscriminantValue,
                 shape: ExampleSingleUnionTypeProperties.singleProperty(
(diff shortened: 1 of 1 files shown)


packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
```diff
@@ -407,32 +407,28 @@ function convertObject({
...
-                                          propertyKey: wireKey,
-                                          property: originalTypeDeclaration.rawPropertyType
-                                      }).name,
-                                      wireValue: wireKey
-                                  }),
-                                  value: valueExample,
-                                  originalTypeDeclaration: originalTypeDeclaration.typeName
-                              });
-                          } catch (e) {
-                              // skip properties that fail to convert (undiscriminated unions)
```
dsinghvi, line 434: we remove this catch all -- its possible stuff will throw on upgrade, but if it does folks can either downgrade or file an issue.

packages/cli/generation/ir-generator/src/resolvers/TypeResolver.ts
```diff
@@ -47,7 +47,9 @@ export class TypeResolverImpl implements TypeResolver {
     }): RawTypeDeclarationInfo {
         const declaration = this.getDeclarationOfNamedType({ referenceToNamedType, file });
         if (declaration == null) {
-            throw new Error("Cannot find declaration of type: " + referenceToNamedType);
+            throw new Error(
+                "Cannot find declaration of type: " + referenceToNamedType + " in file " + file.relativeFilepath
```
dsinghvi, line 51: log file in addition to type not being found

Issue #2417, opened 2023-12-01
[Feature] `x-fern-ignore` OpenAPI extension works on schemas


dsinghvi:
### Problem description

`x-fern-ignore` should work on schemas and the converter should ignore bringing those schemas in

### Why would it be useful?

This would be useful if folks want to exclude certain redundant or internal schemas from Fern. 

### Describe the solution (optional)

```
components: 
  schemas: 
     MyType: 
        x-fern-ignore: true # <----- ignore
        proeprties: 
          ...
```

dannysheridan:
this is now supported

PR #2428, opened 2023-12-03
feature: support x-fern-ignore at the schema level

dsinghvi:
This PR adds an extension where you can configure fern to ignore schemas from the OpenAPI spec by placing `x-fern-ignore: true` at the schema level.


dsinghvi:
Fixes https://github.com/fern-api/fern/issues/2417

PR #2446, opened 2023-12-06
feature: OpenAPI importer supports `audio/mpeg` content type

dsinghvi:
The `audio/mpeg` content type returns a byte stream. In the fern definition this is just a `file` type so the importer converts as a file.


dsinghvi:
@armandobelardo fyi small PR to support the `audio/mpeg` type!

PR #2492, opened 2023-12-14
[feat] For seed only run one container per script for all fixtures

armandobelardo:


packages/seed/src/commands/test/testWorkspaceFixtures.ts
```diff
@@ -228,21 +246,55 @@ async function testWithWriteToDisk({
...
+                ["cp", `${outputDir}/.`, `${script.containerId}:/${workDir}/generated/`],
+                {
+                    doNotPipeOutput: true
+                }
+            );
+            if (copyScriptCommand.failed || mkdirCommand.failed || copyCommand.failed) {
+                taskContext.logger.error("Failed to run copy of scripts. See ouptut below");
+                taskContext.logger.error(copyCommand.stdout);
+                taskContext.logger.error(copyCommand.stderr);
+                return { type: "failure", reason: "Failed to run script...", fixture };
```
dsinghvi, line 283: @armandobelardo small bug here that we are printing just the copyCommand.stdout but checking failure for all of the 3 commands; i think it might be better to have 3 separate if statements for each command in case it fails

armandobelardo: That's fair, will do! I was originally trying to keep it a bit more concise, but that doesn't make much sense haha

Issue #1268, opened 2023-02-02
Output `All checks passed` when fern check succeeds


dsinghvi:

rscottlundgren:
Per Discord conversation w/ @dsinghvi:
> When you run `fern check` today there is no output signifying that everything succeeded. It'd be nice if it would output `All checks passed`

Relevant File (where change should be made): https://github.com/fern-api/fern/blob/ec16010ab28b86ea2fb2d3cfb7733cf3f21ab0ce/packages/cli/cli/src/commands/validate/validateFernWorkspaceAndLogIssues.ts#L38

I'll pick this up @dsinghvi and will reach out via Discord if I have any questions once I've played a little to get used to the program itself.

dsinghvi:
@rscottlundgren are you still planning on implementing this?

23solo:
hey @dsinghvi if @rscottlundgren is not working on this can i take this ? I've just gone through the code and workings of it,  also i've tested the above case locally (Attaching the images for ref.)
Also, I couldn't find any tests associated with it.
![fernProject](https://github.com/fern-api/fern/assets/71120053/4dfc7271-259e-4542-9868-224ac41cdbe4)
![userApi](https://github.com/fern-api/fern/assets/71120053/91984bc3-d7f2-4525-a2e2-b30b332c739e)

23solo:
hey @dsinghvi if @rscottlundgren is not working on this can i take this ? I've just gone through the code and workings of it, i've tested the above case locally (Attaching the images for ref.)
Also, I couldn't find any tests associated with it.
![fernUserDir](https://github.com/fern-api/fern/assets/71120053/3fc9286a-7847-4c5a-ad64-98d0ec289b84)

![fernDir](https://github.com/fern-api/fern/assets/71120053/d2a3d0dc-fa73-42b6-860a-c5681c061938)

dannysheridan:
This has been released! https://github.com/fern-api/fern/pull/2501

PR #2499, opened 2023-12-17
fix: `fern write-definition` doesn't throw on non-OpenAPI workspaces

dsinghvi:
Previously, we would throw if you ran `fern write-definition` on a non-OpenAPI workspace which would stop the cli from running on other workspaces. Now, we just skip non OpenAPI workspaces and output when the definition directory is written.


dsinghvi:
@armandobelardo fyi

dsinghvi:
Also fixes https://github.com/fern-api/fern/issues/1268

Issue #2434, opened 2023-12-04
[Feature] Better CLI output for Fern Check


damonbanks:
### Problem description

I get several irrelivant warnings when running `fern check`. If there is a real error, id like to only see that. When the terminal window is small/short the error gets lost.

### Why would it be useful?

The warnings are cluttering the output when there is an actual error.

### Describe the solution (optional)

If there are error(s) show those only. If not, then show some warnings. 

### Additional context

Add any other context or screenshots.
<img width="1186" alt="Screenshot 2023-12-04 at 10 42 29 AM" src="https://github.com/fern-api/fern/assets/149414575/ff57fbfb-7631-4d01-ab01-ddf663e73a63">
![Screenshot 2023-12-04 at 10 36 37 AM](https://github.com/fern-api/fern/assets/149414575/bb2ba17e-32a7-4c48-8d0d-51941f0abe11)

dsinghvi:
Suggested fix: 
- `fern check` does not print out warnings by default
- `fern check --warnings` will print out both errors + warnings

lesjonesiv:
> Suggested fix:
> 
> * `fern check` does not print out warnings by default
> * `fern check --warnings` will print out both errors + warnings

+1 would love to see something like this

damonbanks:
> Suggested fix:
> 
> * `fern check` does not print out warnings by default
> * `fern check --warnings` will print out both errors + warnings

Yes. This would be good. Maybe a `fern check --log-level warning` if it needs to be more robust for future iterations for logs such as `info, warning, error, debug, etc` But that might be overkill for the use case

dsinghvi:
@damonbanks @lesjonesiv this is releases in `0.16.9`. To grab the latest, upgrade fern by running `fern upgrade`

PR #2501, opened 2023-12-17
fix: improve `fern check` logging

dsinghvi:
This PR improvs `fern check` logging to make it more digestible: 
- We now log a summary first with the number of errors + warnings. 
- We don't log warnings by default (if there are warnings, we tell the user to run `fern check --warnings` if they are curious)
- We log all errors first, then we log all warnings if `logWarnings` is specified
- When there are no errors and no warnings, we log a `All checks passed`

See an example of the output below: 
Screenshot 2023-12-16 at 7 50 39 PM


dsinghvi:
fixes https://github.com/fern-api/fern/issues/2434

dannysheridan:
HUGE for user visibility

PR #2551, opened 2024-01-02
fix: Allow generating SDKs locally without Fern token

NathanFlurry:
Using Fern in an open source repo can be tricky since outside contributors currently need to have their own Fern token.

The `fern generate --local` does not require a Fern token to function. This PR disables prompting the user to sign in when using the local Docker generation.


zdunecki:
Hi,

Any update on this? I see it's one of the factors that may help to run fern locally without any token/login required. 

Best

packages/cli/cli/src/commands/generate/generateAPIWorkspace.ts
```diff
@@ -59,6 +59,9 @@ export async function generateWorkspace({
             context
         });
     } else {
+        if (!token) {
+            return context.failAndThrow("Must provide token if 'useLocalDocker' is false");
```
dsinghvi, line 63: `useLocalDocker` is not a flag the user knows about -- they only know about `--local`

packages/cli/cli/src/commands/generate/generateAPIWorkspace.ts
```diff
@@ -25,7 +25,7 @@ export async function generateWorkspace({
     version: string | undefined;
     groupName: string | undefined;
     shouldLogS3Url: boolean;
-    token: FernToken;
+    token?: FernToken;
```
dsinghvi, line 28: can we type this as FernToken | undefined, so that consumers are forced to provide a value

packages/cli/cli/src/commands/generate/generateAPIWorkspace.ts
dsinghvi: how about `Please run fern login.`

AngelOnFira, line 63: @dsinghvi what are your thoughts on this?
```suggestion
            return context.failAndThrow("Please run fern login. No token was found, and `--local` is not enabled.");
```

AngelOnFira: Not sure if "no token was found" or "no token was provided" makes more sense here.

packages/cli/cli/src/commands/generate/generateAPIWorkspace.ts
AngelOnFira, line 28:
```suggestion
    token: FernToken | undefined;
```

Amenocy:
is there any update regarding this pr?! i would be useful to merge it :)

Amenocy (approved):

PR #2553, opened 2024-01-03
Add alert about `x-fern-*` headers and CORS policies

NathanFlurry:
Strict CORS policies will return throw an error if the `X-Fern-*` headers are not explicitly permitted.


NathanFlurry:
@dsinghvi  Any action needed from me on this? Can't see the output from `pull-request / preview-docs` that's giving an error.

dannysheridan:
Changed made in https://github.com/fern-api/fern/pull/3173 ! Thanks for contributing @NathanFlurry <3 

We are deprecating the browser SDK due to improvements in out typescript sdk that now runs in multiple runtimes (e.g., node, browser, deno, bun, cloudflare, lambdas)

PR #2572, opened 2024-01-08
[seed] Add support for testing snippets

amckinney:
This updates `seed` so that it verifies that the generated snippets are valid (according to the same configured `scripts` used to ensure the generated SDK is valid).

The new `SnippetTestBuilder` is used to format the snippets written in the `snippet.json` into the appropriate format required to validate the result (i.e. with `dmypy`). For Python, the generated snippets are placed adjacent to the generated SDK files, and a separate `pyproject.toml` is created so that the generated SDK is consumed as a dependency (so that the module successfully compiles).

**Note:** This branch was originally based on an old version of `main` - it still needs to be rebased with the latest changes introduced in https://github.com/fern-api/fern/pull/2492. The refactor isn't as straightforward as I'd hope so I'm putting this up early to discuss the general approach before spending the time required to consolidate everything together.


packages/seed/src/SnippetTestBuilder.ts
```diff
@@ -0,0 +1,90 @@
...
+        await fs.writeFile(
+            path.join(dir.path, "src", "sync_client.py"),
+            snippet.syncClient,
+        );
+        await fs.writeFile(
+            path.join(dir.path, "src", "async_client.py"),
+            formatAsyncPythonSnippet(snippet.asyncClient),
+        );
+    }
+}
```
dsinghvi, line 76: can we have the generator author specify this in `seed.yml` ? They can also just point to a file so that it doesn't clutter up `seed.yml` itself

amckinney: Good call - if we continue to go down this route then that's a good call. @armandobelardo suggested moving even more of this into the Python generator so I'll continue to explore that in the meantime.

dannysheridan:
@amckinney is this stale?

PR #2612, opened 2024-01-12
cleanup: remove deprecated generators

dannysheridan:
@dsinghvi know you will have thoughts about this one and back compat!


dsinghvi:
closing b/c not backcompat, there's no win in getting rid of old ones

PR #2618, opened 2024-01-14
Add fern mock command

### LINKED ISSUE #1116: [Feature Request] Generate fake implementation
It would be really cool if Fern could auto-generate a fake HTTP service that I can easily access from a client (whether it actually runs on the server or is mocked on the client) that returns notional data based on the response shape types.

To me, this would be valuable for front-end (client) development where when working with a BE counter-part, we start with Fern definitions and from the get-go, I have endpoints I can hit that 200 and give me data (albeit they may not be perfect/semantically correct). Alternatively, if I can "just use Fern" and specify "fake" mode and Fern handles not actually sending any HTTP requests but resolving local promises with notional data.

This saves me time as I don't spend tens of minutes/hours to create a fake data service in my client / probably also useful for writing unit or integration tests.

EDIT: it could even be based on the examples that a developer provides in the documentation

amckinney:
Fixes #1116

This adds the `fern mock` command which can be used to spin up a local mock server for any of the user's Fern APIs.


packages/cli/cli/src/cli-context/TtyAwareLogger.ts
```diff
@@ -11,7 +11,7 @@ export class TtyAwareLogger {
     private tasks: TaskContextImpl[] = [];
     private lastPaint = "";
     private spinner = ora({ spinner: "dots11" });
-    private interval: NodeJS.Timer | undefined;
+    private interval: NodeJS.Timeout | undefined;
```
dsinghvi, line 14: weird -- why did we have to change this?

packages/cli/mock/src/runMockServer.ts
```diff
@@ -0,0 +1,208 @@
...
+        audiences: { type: "all" },
+        generationLanguage: undefined
+    });
+
+    const app = express();
+    app.use(express.json({ limit: "50mb", strict: false }));
+
+    for (const service of Object.values(ir.services)) {
+        for (const endpoint of service.endpoints) {
+            const endpointPath = getFullPathForEndpoint(endpoint);
```
dsinghvi, line 42: i think we'll need to take account any top level api base path, service path and endpoint path for this to be accurate

packages/cli/mock/src/runMockServer.ts
```diff
@@ -0,0 +1,208 @@
...
+            context.logger.debug(`Registering ${endpoint.method} ${endpointPath} ...`);
```
dsinghvi, line 43: i could see this being info actually

packages/cli/mock/src/runMockServer.ts
```diff
@@ -0,0 +1,208 @@
...
+import { ParamsDictionary, Query } from "express-serve-static-core";
+import getPort from "get-port";
+import { IncomingHttpHeaders } from "http";
+import { isEqual } from "lodash-es";
+import qs from "qs";
+import urlJoin from "url-join";
+
+type RequestHandler = (req: Request, res: Response) => void;
+
+export async function runMockServer({
```
dsinghvi, line 22: today examples don't include auth / global headers -- don't have to solve in this PR but we're going to need to figure that out in order for the sdk testing to work

dsinghvi:
what do you think about adding an ete-test for `mock` that spins up the server and makes a single request?

packages/cli/mock/src/runMockServer.ts
```diff
@@ -0,0 +1,208 @@
...
+    context,
+    workspace,
+    port
+}: {
+    context: TaskContext;
+    workspace: FernWorkspace;
```
dsinghvi, line 28: my guess is the entry point here should be IR, not a fern workspace -- that way anything can call into this as long as they have IR

packages/cli/mock/src/runMockServer.ts
amckinney: I'm OK with starting with `info`, but I figured this would end up looking like a massive wall of text for larger APIs. We can always change it to `debug` later though, so I'm fine with changing this here.

packages/cli/mock/src/runMockServer.ts
dsinghvi: resolving due to use of `fullPath`

amckinney: We're actually using the full path here, so it will already include those details.

packages/cli/mock/src/runMockServer.ts
amckinney: Good call, I'll add a TODO in-line for now (along with a mention to some of the other features we're eventually need to support like streaming, file upload, etc).

packages/cli/mock/src/runMockServer.ts
dsinghvi: yeah i think we could always limit to 25 paths or something but kind of nice to see all the paths that the server is serving

packages/cli/cli/src/cli-context/TtyAwareLogger.ts
amckinney: Looks like it's something with my local setup - reverting this continues to pass compilation in CI.

packages/cli/mock/src/runMockServer.ts
amckinney: Created a new issue [here](https://github.com/fern-api/fern/issues/2620).

dsinghvi (approved):

PR #2666, opened 2024-01-19
💎 Ruby: Containerize and run seed tests

armandobelardo:


dsinghvi (approved):
Can we make the publish-generator.yml very specific to ruby: 

- lives in a ruby folder ? 
- only triggered when there is a change in the ruby dir ?

armandobelardo:
@dsinghvi it's manually triggered, you want it changed to release every time there's a change to the ruby directory? That sounds off

I was about to make it take in an input for which generator to publish and put them all in this one file

PR #2675, opened 2024-01-20
Improve docs.yml CLI navigation contents

minaelee:
- rename CLI section contents page **Overview** to **CLI Overview** to avoid confusion with general **Overview** tab 
  - this especially looks misleading at the bottom of the [Generators](https://docs.buildwithfern.com/overview/welcome/generators) page, where it indicates that the next page is simply **Overview**.
- fix page name **fern write definition** to **fern write-definition**

### DIFF
--- a/fern/docs.yml
+++ b/fern/docs.yml
@@ -43,7 +43,7 @@ navigation:
             slug: generators
       - section: CLI
         contents:
-          - page: Overview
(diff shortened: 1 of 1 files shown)


dannysheridan:
will the change of `fern write-definition` page name impact any links to that page elsewhere in our docs?

PR #2677, opened 2024-01-20
feat(generate-support-irversion): impl

zdunecki:
`irVersion` flag passed for `generate` command allows the use of explicit IR version, e.g `v32`.


packages/cli/cli/src/commands/generate/generateAPIWorkspace.ts
```diff
@@ -61,6 +63,7 @@ export async function generateWorkspace({
             workspace,
             generatorGroup: group,
             keepDocker,
+            irVersion,
```
dsinghvi, line 66: @zdunecki can we pass the `irVersion` in both local or remote ?

dsinghvi:
@zdunecki looks great! if you wire it up with the remote code generation as well, that would be great!

fern/docs/pages/cli/fern-generate.mdx
```diff
@@ -44,6 +44,17 @@ $ fern generate --version
...
 
+### Intermediate Representation(IR) Version
+
+**Type**: `string`
+**Description:** Specify a version of Intermediate Representation(IR).
+
+**Example**
+
+```bash
+$ fern generate --irVersion v32
```
dsinghvi, line 55: @zdunecki looks great -- thoughts on calling this `--irVersionOverride` to signify that you are overriding typical CI behavior ? 

For example if you run the python generator with `--irVersion 33` even though it actually depends on 32, the CLI would now allow it

zdunecki:
@dsinghvi 

I've introduced your suggestions.

dsinghvi:
Solved by https://github.com/fern-api/fern/pull/3212

Issue #2649, opened 2024-01-17
[Fern's Documentation] See example of Tabs


SK-Sam:
### What can be improved
On [this page](https://docs.buildwithfern.com/generate-docs/overview/writing-content#tabs) the Tabs feature is being detailed, but as a new user I'm having to guess what the end result looks like. Can we show a visible example of what the Tabs example looks like?

### Relevant Screenshots

<img width="787" alt="Screenshot 2024-01-17 at 2 24 13 PM" src="https://github.com/fern-api/fern/assets/10618376/0efe1895-058a-4c55-9899-26f89dbf8362">

PR #2691, opened 2024-01-22
chore: Include visual example of Tabs in action.

SK-Sam:
[Targeting this issue](https://github.com/fern-api/fern/issues/2649)

This PR pitches an improvement to Fern docs. Fern shows how to include Tabs but doesn't give an example of what the end result of their suggested code looks like.

This solution uses an image hosted on Imgur, but ultimately should be an image hosted on Fern's AWS S3. Can we move this img into Fern's AWS? I can remove the image from local dir.

image

### DIFF
--- a/fern/docs/pages/docs/writing-content.mdx
+++ b/fern/docs/pages/docs/writing-content.mdx
@@ -223,7 +223,7 @@ tabs:
 ```
 </CodeBlock>
 
-![Let's replace this with the same img hosted on Fern's AWS S3 instead of Imgur](https://imgur.com/a/zQGqQWv)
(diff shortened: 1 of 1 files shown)


fern/docs/pages/docs/writing-content.mdx
```diff
@@ -223,6 +223,8 @@ tabs:
 ```
 </CodeBlock>
 
+![Let's replace this with the same img hosted on Fern's AWS S3 instead of Imgur](https://imgur.com/a/zQGqQWv)
```
SK-Sam, line 226: @dannysheridan I'd like to move this img into Fern's AWS S3 instance. Can you send me back a URL of this image hosted on Fern's AWS?

Issue #2669, opened 2024-01-19
[Fern's Documentation] Codeblock example leads to error


SK-Sam:
### What can be improved

[On this section](https://docs.buildwithfern.com/generate-docs/component-library/code-block) when copying down the example, the validations pass yet leads to an unintended code block. This is most likely due to a backslash before the backticks in the codeblock: https://github.com/fern-api/fern/blob/main/fern/docs/pages/docs/components/codeblock.mdx#L14

### Relevant Screenshots

When copied down: 
![Screenshot 2024-01-19 at 3 54 24 PM](https://github.com/fern-api/fern/assets/10618376/07e1920f-f559-4e16-84de-1b40d5fc1624)

When removing backslash from Fern docs example:
![Screenshot 2024-01-19 at 3 55 42 PM](https://github.com/fern-api/fern/assets/10618376/12cde77e-caa1-4551-8933-95873cd8def2)

dannysheridan:
resolved by https://github.com/fern-api/fern/pull/2748

PR #2692, opened 2024-01-22
chore: Remove unwanted backslashes in docs.

SK-Sam:
This PR removes backslashes that result in a `Badly formatted code block` text.

[Link to issue](https://github.com/fern-api/fern/issues/2669)


dsinghvi (approved):

minaelee:
Removing backslashes only would result in incorrect rendering since they are an attempt to escape the inner set of backticks
Must also add an extra backtick to the outer set of backticks as in https://github.com/fern-api/fern/pull/2748

dannysheridan:
closing in favor of https://github.com/fern-api/fern/pull/2748

PR #2720, opened 2024-01-22
Fix: Ruby: Await linting as well

armandobelardo:


dsinghvi (approved):

packages/generators/ruby/model/src/GeneratorCli.ts
```diff
@@ -110,6 +114,10 @@ export class RubyModelGeneratorCli extends AbstractGeneratorCli<RubyModelCustomC
             })
         );
         // Run lint and generate lockfile
-        exec(`rubocop --auto-correct-all ${config.output.path}`);
+        try {
+            execSync(`rubocop --autocorrect-all ${config.output.path}`);
```
dsinghvi, line 118: checkout how we do it in other parts of the repo: https://github.com/fern-api/fern/blob/aa9d8f96750d5d82ab6e5ef6e71fb921688052e9/packages/seed/src/commands/test/testWorkspaceFixtures.ts#L67

dsinghvi: generally want to not use `<>Sync` and use promise based for everything

packages/generators/ruby/model/src/GeneratorCli.ts
```diff
@@ -110,6 +114,10 @@ export class RubyModelGeneratorCli extends AbstractGeneratorCli<RubyModelCustomC
...
+        } catch {
+            // NOOP, ignore warns
+        }
```
dsinghvi, line 121: can we add a debug log or info log ?

PR #2721, opened 2024-01-22
Update generators.mdx with openapi export generator

minaelee:
Seems like `fernapi/fern-openapi` should be added to the list of generators.


fern/docs/pages/welcome/generators.mdx
```diff
@@ -35,6 +35,16 @@ Whether you call them server interfaces, server stubs, or API implementation stu
 
 _Looking for a framework that's not listed? [Let us know.](https://github.com/fern-api/fern/issues/new?title=%5BFramework%20Request%5D%20)_
 
+## OpenAPI export
+
+Export your [Fern Definition](https://docs.buildwithfern.com/api-definition/fern-definition/overview) to an OpenAPI Description file.
```
dannysheridan, line 40: OpenAPI is referred to many different things in different places.

Description, Definition, File, Specification, Spec... which do you think we should use?

PR #2728, opened 2024-01-23
Update broken link to .net server boilerplate in generators.mdx

minaelee:


dannysheridan:
this feels like a mistake in `docs.yml` I think the url should be "dotnet" and not just "net". want to try adjusting the slug?

minaelee:
> this feels like a mistake in `docs.yml` I think the url should be "dotnet" and not just "net". want to try adjusting the slug?

Sure, https://github.com/fern-api/fern/pull/2746

PR #2748, opened 2024-01-23
 Fix incorrect backticks rendering in a fenced MD codeblock in codeblock.mdx

minaelee:
Also remove unnecessary spacing around props in component tag.

Using a backslash doesn't work to escape the backticks. It shows the backslash in the code block, as seen here:

image

Should be resolved by using a 4-backtick fenced code block instead of 3, as demonstrated below:

````mdx
<CodeBlock title = "Node.js SDK">
```bash
$ npm install --save @fern-api/node-sdk
```
</CodeBlock>
````

### DIFF
--- a/fern/docs/pages/docs/components/codeblock.mdx
+++ b/fern/docs/pages/docs/components/codeblock.mdx
@@ -9,15 +9,15 @@ CodeBlock Props:
 
 This would look like:
 
-```mdx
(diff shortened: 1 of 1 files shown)


dannysheridan:
@minaelee think we should document this trick in case others run into it? or too much of a detail?

PR #2764, opened 2024-01-24
Fix missing closing tag for CodeBlock in extensions.mdx

minaelee:
It looks like this right now without this fix:
image


dannysheridan:
appreciate the clear sreenshot!

PR #2769, opened 2024-01-24
chore: Move Ruby generator to top level (match new repo structure)

armandobelardo:
Moved Ruby generator + it's seed workspace
Also moved the generator to an MRLint CLI to clean up the implementation some


.github/CODEOWNERS
```diff
@@ -1 +1,5 @@
-* dsinghvi
\ No newline at end of file
+* @dsinghvi
+/generators/cli/* @armandobelardo
+/generators/commons/* @armandobelardo
+/generators/docker/* @armandobelardo
```
dsinghvi, line 4: i wonder if we should leave these in packages -- i like the idea of clicking on generators and just seeing recoganizable folders: postman, python, openapi, ruby, etc.

.github/workflows/seed.yml
```diff
@@ -36,19 +36,20 @@ jobs:
...
+          yarn seed:local test --workspace ruby-model --parallel 8
 
       - name: Ensure no changes to git-tracked files
         run: git --no-pager diff --exit-code
 
   pydantic-model:
     runs-on: ubuntu-latest
     needs: changes
-    if: ${{ contains(needs.changes.outputs.generators_changed, 'ruby') }} || ${{ contains(needs.changes.outputs.generators_changed, 'seed') }}
+    if: ${{ contains(needs.changes.outputs.generators_changed, 'python') }} || ${{ contains(needs.changes.outputs.generators_changed, 'seed') }}
```
dsinghvi, line 52: thankyou 🙏

.github/CODEOWNERS
armandobelardo: yeah I can do that! pls hold

PR #2796, opened 2024-01-25
chore: consolidate ruby publish to new format

armandobelardo:


.github/workflows/publish-generator.yml
```diff
@@ -235,3 +241,65 @@ jobs:
...
+        run: |
+          projectVersion=$(cat generators/ruby/sdk/VERSION)
+          echo $projectVersion
+          echo "VERSION=$projectVersion" >> $GITHUB_ENV
+
+      - name: Build and push Ruby SDK Docker image
+        run: |
+          projectVersion=${{ env.VERSION }}
+          yarn workspace @fern-api/fern-ruby-sdk dockerTagVersion "$projectVersion"
+          docker push fernapi/fern-ruby-sdk:"$projectVersion"
```
dsinghvi, line 305: can we publish multiplatform dockers?

dsinghvi (approved):

.github/workflows/publish-generator.yml
armandobelardo: good call! I didn't see we were doing that!

PR #2811, opened 2024-01-26
(release): support scanning env variable for auth in python sdk generator 0.8.1 

dsinghvi:
In this PR we setup a new release for the python `0.8.1` generator which supports scanning environment variables for auth. Instead of passing in `api_key` for example, the sdk consumer can just set the configured environment variable (e.g. `FERN_API_KEY`).


dannysheridan:
should these two destinations be updated with this PR?

* https://github.com/fern-api/fern?tab=readme-ov-file#generators
* https://github.com/fern-api/fern/blob/main/generators/python/pydantic/CHANGELOG.md

PR #2825, opened 2024-01-28
fix: increase python generator recursion depth to allow for deeply nested examples

dsinghvi:
Temporary fix to allow for deeply nested examples, this is a shortterm workaround while we investigate a more condensed example data model


dsinghvi:
@armandobelardo up to update the `changelog` for python sdk and bump version -- other than that this is g2g

armandobelardo:
> @armandobelardo up to update the `changelog` for python sdk and bump version -- other than that this is g2g

Totally! Just added!

PR #2841, opened 2024-01-30
(improvement): Add common initialisms to naming converter

amckinney:
This improves Fern's IR naming convention generator to handle common initialisms like `API`, `URL`, and `ID`. For example, users will now have their SDK constructs with better names like:

- `WithAPIKey` instead of `WithApiKey`
- `ErrorInstanceID` instead of `ErrorInstanceId`
- `GetUserIDs` instead of `GetUserIds`.

Any occurrence of adjacent initialisms (e.g. `api_url`) will use the original naming convention because the format would otherwise be worse (i.e. `ApiUrl` continues to be used instead of `APIURL`).

Note that this would be a breaking change for any SDK that defines one of these common initialisms (because the names of these types and/or methods will change), so the feature is gated behind the `smart-casing` configuration in the `generators.yml`. It can be specified like so:

```yaml
      - name: fernapi/fern-go-sdk
        version: 0.12.0
        smart-casing: true
```

### DIFF
--- a/packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
+++ b/packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
@@ -8,10 +8,13 @@ export interface CasingsGenerator {
     generateNameAndWireValue(args: { name: string; wireValue: string }): NameAndWireValue;
 }
 
-export function constructCasingsGenerator(
-    generationLanguage: GenerationLanguage | undefined,
-    smartCasing: boolean
-): CasingsGenerator {
+export function constructCasingsGenerator({
+    generationLanguage,
+    smartCasing
+}: {
+    generationLanguage: GenerationLanguage | undefined;
+    smartCasing: boolean;
+}): CasingsGenerator {
     const casingsGenerator: CasingsGenerator = {
         generateName: (name) => {
             const generateSafeAndUnsafeString = (unsafeString: string): SafeAndUnsafeString => ({
--- a/packages/cli/generation/ir-generator/src/__test__/casingsGenerator.test.ts
+++ b/packages/cli/generation/ir-generator/src/__test__/casingsGenerator.test.ts
@@ -2,7 +2,7 @@ import { Name } from "@fern-fern/ir-sdk/api";
 import { constructCasingsGenerator } from "../casings/CasingsGenerator";
 
 describe("casingsGenerator", () => {
-    const casingsGenerator = constructCasingsGenerator(undefined, true);
+    const casingsGenerator = constructCasingsGenerator({ generationLanguage: undefined, smartCasing: true });
     it("simple", () => {
         const expected: Name = {
             originalName: "hello",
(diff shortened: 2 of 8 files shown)


amckinney:
Discussed offline - given the significant breaking change potential for this one, we'll want to guard it behind a feature flag. We imagine it will be some sort of `smart-casing` variant in the `generators.yml`.

amckinney:
Added the `special-casing` generator invocation configuration so that users can specify the following:

```yaml
      - name: fernapi/fern-go-sdk
        version: 0.11.0
        special-casing: true
```

In its current state, this required a fair number of function signature changes. I opted to make the parameter explicit in a variety of places (i.e. requiring an explicit `false`), but it's effectively set to `false` in every scenario except the `fern generate` flow. The IR tests explicitly set this value to `true` so that the snapshots use the special casing form.

Also, we'll want to improve name generation a bit to handle the plural form of each of these initialisms (e.g. `GetUserIDs`). We want the singular form to contain the initialism form, but the trailing `s` should not. This will be an easy change that we should handle here.

packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
```diff
@@ -1,30 +1,66 @@
...
 export interface CasingsGenerator {
     generateName(name: string): Name;
     generateNameAndWireValue(args: { name: string; wireValue: string }): NameAndWireValue;
 }
 
-export function constructCasingsGenerator(generationLanguage: GenerationLanguage | undefined): CasingsGenerator {
+export function constructCasingsGenerator(
+    generationLanguage: GenerationLanguage | undefined,
+    specialCasing: boolean
+): CasingsGenerator {
```
dsinghvi, line 14: nit: i think using named params here will be better 

```
{ generationLanguage, specialCasing }: { generationLanguage: GenerationLanguage | undefined | specialCasing: boolean } 
```

packages/cli/config-management/generators-configuration/src/convertGeneratorsConfiguration.ts
```diff
@@ -91,6 +91,7 @@ async function convertGenerator({
         version: generator.version,
         config: generator.config,
         outputMode: await convertOutputMode({ absolutePathToGeneratorsConfiguration, generator }),
+        specialCasing: generator["special-casing"] ?? false,
```
dsinghvi, line 94: i think `smart-casing` here is probably better than `special-casing` but ik thats a really annoying change

dsinghvi (approved):

packages/cli/config-management/generators-configuration/src/convertGeneratorsConfiguration.ts
amckinney: Let's do it, won't be too bad with some `sed`!

packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
amckinney: Good call, always a better move to do this when we can.

PR #2842, opened 2024-01-30
Add support for custom code samples

abvthecity:


packages/cli/openapi-parser/src/v3/extensions/fernExtensions.ts
```diff
@@ -197,7 +197,32 @@ export const FernOpenAPIExtension = {
...
+     *        description: This is a code sample that logs the response
+     */
+    CUSTOM_CODE_SAMPLES: "x-fern-code-samples",
+
+    /**
+     * Migrate from readme extensions to fern IR.
+     * 
+     * https://docs.readme.com/main/docs/openapi-extensions
+     */
+    README_EXT: "x-readme",
```
dsinghvi, line 225: i would put this in a `readmeExtensions.ts`

dsinghvi: just so we can manage accordingly

dsinghvi (approved):

PR #2865, opened 2024-02-02
improvement: use commit number for version

armandobelardo:
Ensures version number is monotonically increasing


dsinghvi (approved):
can we do this for more than the ruby gen -- happy to merge in and follow up with that

armandobelardo:
> can we do this for more than the ruby gen -- happy to merge in and follow up with that

Yeah I can just add it to all of them

PR #2886, opened 2024-02-04
(java): idempotency header sdks compile properly

dsinghvi:
This PR fixes Java SDKs with idempotency headers which previously didn't compile. Specifically, the `ClientOptions.java` needs to provide a `headers` method that takes `IdempotentRequestOptions` as input.


dsinghvi:
@armandobelardo if you revert this commit you should be able to reproduce seed test not working

PR #2904, opened 2024-02-06
fix: improve handling color config for dark vs light themes

abvthecity:
### Before this PR
* unless dark/light mode is clearly defined, the theme is always "dark"
* if the background is undefined and accent primary is dark, then the theme defaults to light (which is wrong)

### After this PR
* defaults to darkAndLight
* automatically lightens and darkens colors to improve contrast
* returns a readability warning if accessibility standards aren't met.


dsinghvi (approved):
thank you for writing the tests 🙏

PR #2912, opened 2024-02-07
fix: use brightness not luminance to flip the color theme

abvthecity:
HSL (luminance) is the wrong scale for flipping the color from light-to-dark and dark-to-light. Instead, we'll use HSV (brightness-value).

bonus: @dannysheridan's updated warning message:
> Alert: The chosen shade, 'accentPrimary' in dark mode, fails to meet minimum contrast requirements. To enhance accessibility and ensure content readability, Fern will adjust from #XXX to a more contrast-rich #XXX.

todo: document this behavior and add a permalink


dsinghvi (commented):
Would love to move this into a `fern-check` rule for docs at some point (https://github.com/fern-api/fern/tree/main/packages/cli/yaml/docs-validator)

dsinghvi (approved):

PR #2916, opened 2024-02-08
(fix): update OpenAPI example generation

dsinghvi:
In this PR, we fix multiple things about example generation: 
1) If an example for a property or a full object is present, we do our best to include it. 
2) If no example is present, we only show required fields for a request. 
3) If no example is present, we show 3 levels of optional properties for a response. 
4) We default to the property name instead of using "string" for string examples. 
5) We autogenerate 1 entry to a list, if no list examples are provided.


dannysheridan:
This PR is going to pay dividends over the coming years! 💰

PR #2926, opened 2024-02-09
(python, feat): add in request options to python

armandobelardo:
I wanted to go with a TypedDict here to start that pattern, however in python 3.7 (which we support), there's no easy way to not include keys within the dict (e.g. here you'd always need to include every key: `{timeout_in_seconds: None, additional_headers: None, additional_query_parameters: None, additional_body_parameters: None, }`) which felt like poor UX.
This can be done better with `NotRequired` over `Optional`, but this requires python >= 3.8 to leverage `typing_extensions` (or `typing` in >= 3.10)

### DIFF
--- a/seed/python-sdk/audiences/src/seed/core/request_options.py
+++ b/seed/python-sdk/audiences/src/seed/core/request_options.py
@@ -7,22 +7,22 @@ try:
 except ImportError:
     from typing_extensions import NotRequired  # type: ignore
 
-"""
-Additional options for request-specific configuration when calling APIs via the SDK.
-This is used primarily as an optional final parameter for service functions.
-
-Attributes:
-    - timeout_in_seconds: int. The number of seconds to await an API call before timing out.
-
-    - additional_headers: typing.Dict[str, typing.Any]. A dictionary containing additional parameters to spread into the request's header dict
-
-    - additional_query_parameters: typing.Dict[str, typing.Any]. A dictionary containing additional parameters to spread into the request's query parameters dict
-
-    - additional_body_parameters: typing.Dict[str, typing.Any]. A dictionary containing additional parameters to spread into the request's body parameters dict
-"""
-
 
 class RequestOptions(typing.TypedDict):
+    """
+    Additional options for request-specific configuration when calling APIs via the SDK.
+    This is used primarily as an optional final parameter for service functions.
+
+    Attributes:
+        - timeout_in_seconds: int. The number of seconds to await an API call before timing out.
+
+        - additional_headers: typing.Dict[str, typing.Any]. A dictionary containing additional parameters to spread into the request's header dict
+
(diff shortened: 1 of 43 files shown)


dsinghvi:
I think reasonable to bump the dep to `python >= 3.8`, dont think we should use base model

armandobelardo:
@dsinghvi would rather not introduce a breaking change in this PR

dsinghvi:
https://github.com/merge-api/merge-python-client/pull/69 example of a user trying to upgrade us to python 3.8 so that we aren't locking the pydantic version either

armandobelardo:
> [merge-api/merge-python-client#69](https://github.com/merge-api/merge-python-client/pull/69) example of a user trying to upgrade us to python 3.8 so that we aren't locking the pydantic version either

Addressed by https://github.com/fern-api/fern/pull/2967

armandobelardo:
> > @dsinghvi would rather not introduce a breaking change in this PR
> 
> We made some questionable decisions around forcing the use of `pydantic.BaseModel` in a ton of places and this continues that trajectory, rather than improving the client. I'd rather have an untyped `requestOptions` if we want to not make a breaking change and can bump the dep in the next release

Updated the PR to use TypedDicts

seed/python-sdk/audiences/src/seed/core/request_options.py
```diff
@@ -0,0 +1,15 @@
...
+
+import typing
+
+try:
+    from typing import NotRequired  # type: ignore
+except ImportError:
+    from typing_extensions import NotRequired  # type: ignore
+
+
+class RequestOptions(typing.TypedDict):
```
dsinghvi, line 11: can we document this class and every parameter?

seed/python-sdk/audiences/src/seed/resources/folder_a/resources/service/client.py
```diff
@@ -33,9 +56,28 @@ class AsyncServiceClient:
     def __init__(self, *, client_wrapper: AsyncClientWrapper):
         self._client_wrapper = client_wrapper
 
-    async def get_direct_thread(self) -> Response:
+    async def get_direct_thread(self, *, request_options: typing.Optional[RequestOptions] = None) -> Response:
+        """
+        Parameters:
+            - request_options: typing.Optional[RequestOptions]. Additional options for request-specific configuration when calling APIs via the SDK.
```
dsinghvi, line 62: do you think just making the docs `Request-specific configuration` is enough?

dsinghvi (approved):
LFG

seed/python-sdk/audiences/src/seed/resources/folder_a/resources/service/client.py
armandobelardo: Done!

PR #2932, opened 2024-02-11
improvement: loosen generated Ruby SDK dependencies on Faraday

armandobelardo:
Previously, with the `~>` specifier, we were depending on faraday >= 2.7 and < 3.0 which is a tight window, this PR keeps the same upperbound but loosens the lowerbound to support a full major version below


dsinghvi (approved):

dsinghvi:
@armandobelardo before merging in i think it may be worth adding this to an `# unreleased` section of the ruby sdk generator changelog. Then once you release you can pull that note below a version

PR #2953, opened 2024-02-12
(feature): go generator supports whitelabelling

dsinghvi:


dsinghvi:
@amckinney do you think the go generator would benefit from passing structs around instead of long params? feels easy to make mistakes

generators/go/internal/generator/generator.go
```diff
@@ -783,6 +795,7 @@ func newClientTestFile(
 		"client/client_test.go",
 		"client",
 		baseImportPath,
+		false,
```
dsinghvi, line 798: `false` for the test file for now but we'll want to plumb it in there as well

dsinghvi:
Merging due to time sensitivity but can address comments in a FLUP @amckinney

amckinney:
> @amckinney do you think the go generator would benefit from passing structs around instead of long params? feels easy to make mistakes

In many places we prefer structs, but creating an excessive number of params structs anytime we have a large list of parameters can be cumbersome in Go - it requires a separate structure for every function that requires it. I usually tend to reserve parameter structs for the public-most, external-facing APIs, and use in-line parameters for internal functions.

With that said, there are probably some good candidates where a struct makes sense for internal functions (given the sheer number of parameters involved), but it's equally as easy to forget to include a parameter and the default value will be used (unlike other languages like Typescript that enforce every parameter is set).

I'll take a look at this again later and see what we can do to improve the most important use cases.

PR #2959, opened 2024-02-13
Fix recursive loop in OpenAPI example generation

amckinney:
This fixes an issue where the `ExampleTypeFactory` would not stop based on the configured depth. This change moves the depth check inside the `hasExample` helper so that we make sure to check the depth at the beginning of every call and immediately return `false` if the depth has been reached.


amckinney:
> GitGuardian has uncovered 1 secret following the scan of your pull request.

This is just example data included in the API spec.

packages/cli/openapi-parser/src/v3/converters/example/ExampleTypeFactory.ts
```diff
@@ -372,23 +380,51 @@ export class ExampleTypeFactory {
...
+        options,
+        depth
+    }: {
+        schema: SchemaWithExample;
+        options: ExampleTypeFactory.Options;
+        depth: number;
+    }): boolean {
+        if (this.exceedsMaxDepth(depth, options)) {
+            return false;
+        }
```
dsinghvi, line 394: the idea here was that if an example is present in the OAI spec, we should ignore max depth

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/apiture.test.ts.snap
```diff
@@ -1266,18 +1274,15 @@ Note: This operation requires an identity challenge if the financial institution
...
                     "scheduledOn": "2022-06-28",
                   },
                   "sourceAccount": {
                     "id": "bd9b7af2-6f9b",
-                    "label": "Checking *1008",
                     "location": "internal",
                   },
                   "targetAccount": {
                     "id": "88b1ca3e-d0f3",
-                    "label": "Checking *1008",
```
dsinghvi, line 1280: do you know why these examples are being removed?

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/apiture.test.ts.snap
```diff
@@ -1488,7 +1488,6 @@ Note: This operation requires an identity challenge if the financial institution
                         "memo": "school expenses",
                         "processedAt": "2024-01-15T09:30:00Z",
                         "schedule": {
-                          "amountLimit": "3456.78",
```
amckinney, line 1491: It's interesting to see that this previously existed here before this change. The example is not present in the original declaration:

```yaml
- id: d62c0701-0d74
  amount: "100.00"
  state: scheduled
  memo: school expenses
  ...
  schedule:
    scheduledOn: "2023-08-01"
    frequency: monthly
    recurreneType: fixed
    count: 9
    creditsOn: "2023-08-01"
    debitsOn: "2026-08-01"
```

The `monetaryValue` type (which is what `amountLimit` is) does have the `3456.78` example associated with it, so the OpenAPI -> Fern converter should still fill in the gap (like it used to), regardless of the `ExampleTypeFactory` change.

I'll need to take a closer look at why the change affected this, but including this comment in the meantime in case anything is obvious (cc @dsinghvi)

amckinney:
Superseded by https://github.com/fern-api/fern/pull/3086

Issue #2094, opened 2023-09-06
[Bug] Generator fails when a discriminant value starts with a number


kikones34:
I have the following enum type in my API definition:

```json
"DetectorType": {"type": "string", "enum": ["person", "face", "4w_vehicle"]
```

When attempting to generate a Java SDK, it fails with the following error:

```
[api]: fernapi/fern-java-sdk not a valid name: 4WVehicle
```

Note that the same specification works without issue when generating a Python SDK.
I'm using the latest stable versions Fern 0.14.3 and Java generator 0.3.7.

----

As a side note, it was impossible for me to see where the issue was when executing the generator with the `--local` option, as it just says:

```
[api]: fernapi/fern-java-sdk Docker exited with a non-zero exit code.
```
And using `--log-level debug` yielded the following undecipherable traceback:

```
DEBUG 2023-09-06T14:01:02.032Z [api]: fernapi/fern-java-sdk Error: Docker exited with a non-zero exit code.
                                                                at tryRunDocker (/home/k34/.nvm/versions/node/v20.5.0/lib/node_modules/fern-api/bundle.cjs:549185:11)
                                                                at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
                                                                at async runDocker (/home/k34/.nvm/versions/node/v20.5.0/lib/node_modules/fern-api/bundle.cjs:549159:5)
                                                                at async runGenerator (/home/k34/.nvm/versions/node/v20.5.0/lib/node_modules/fern-api/bundle.cjs:549253:3)
                                                                at async writeFilesToDiskAndRunGenerator (/home/k34/.nvm/versions/node/v20.5.0/lib/node_modules/fern-api/bundle.cjs:549311:3)
                                                                at async /home/k34/.nvm/versions/node/v20.5.0/lib/node_modules/fern-api/bundle.cjs:549272:9
                                                                at async TaskContextImpl.runInteractiveTask (/home/k34/.nvm/versions/node/v20.5.0/lib/node_modules/fern-api/bundle.cjs:534309:7)
                                                                at async Promise.all (index 0)
                                                                at async runLocalGenerationForWorkspace (/home/k34/.nvm/versions/node/v20.5.0/lib/node_modules/fern-api/bundle.cjs:549267:19)
                                                                at async generateWorkspace (/home/k34/.nvm/versions/node/v20.5.0/lib/node_modules/fern-api/bundle.cjs:550066:5)
```

kikones34:
[...]

PR #2961, opened 2024-02-13
Use safe name to generate discriminator wrapper class

kikones34:
Fixes https://github.com/fern-api/fern/issues/2094

We cannot use the unsafe name because it may start with a digit.


dsinghvi (approved):

dsinghvi:
@kikones34 do you mind adding to the changelog and bumping the patch VERSION of the java generator

armandobelardo:
@kikones34 do you mind merging in main and bumping one last time? We just released version `0.8.2` of the Java generator

kikones34:
> @kikones34 do you mind merging in main and bumping one last time? We just released version `0.8.2` of the Java generator

Done

PR #2964, opened 2024-02-13
fix: pipe through optionality in OpenAPI parser

armandobelardo:
note this also fixes an issue in ts-sdk where we were dropping query parameters

### DIFF
--- a/seed/ts-sdk/api-wide-base-path/package.json
+++ b/seed/ts-sdk/api-wide-base-path/package.json
@@ -1,5 +1,7 @@
 {
+    "name": "",
     "version": "0.0.1",
+    "private": false,
     "repository": "https://github.com/api-wide-base-path/fern",
     "main": "./index.js",
     "types": "./index.d.ts",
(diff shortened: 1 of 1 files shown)


generators/typescript/sdk/client-class-generator/src/endpoint-request/GeneratedFileUploadEndpointRequest.ts
```diff
@@ -64,7 +64,10 @@ export class GeneratedFileUploadEndpointRequest implements GeneratedEndpointRequ
         this.generatedSdkClientClass = generatedSdkClientClass;
         this.targetRuntime = targetRuntime;
 
-        if (requestBody.properties.some((property) => property.type === "bodyProperty")) {
+        if (
+            requestBody.properties.some((property) => property.type === "bodyProperty") ||
+            endpoint.queryParameters.length > 0
+        ) {
```
dsinghvi, line 70: what does this do?

.github/workflows/seed.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       python: ${{ steps.filter.outputs.python }}
       postman: ${{ steps.filter.outputs.postman }}
       java: ${{ steps.filter.outputs.java }}
-      typescript: ${{ steps.filter.outputs.ts }}
+      typescript: ${{ steps.filter.outputs.typescript }}
```
dsinghvi, line 22: thanks

dsinghvi (approved):

generators/typescript/sdk/client-class-generator/src/endpoint-request/GeneratedFileUploadEndpointRequest.ts
armandobelardo: So previously, if there was no bodyParameter, we would never initialize the request parameter or query parameter objects here. This would make it if you had only files as properties and then query parameters, those would get dropped from the request

This coincidentally also impacted this seed test: https://github.com/fern-api/fern/blob/ab/oafileparse/test-definitions/fern/apis/file-upload/definition/service.yml#L34-L51

PR #2985, opened 2024-02-14
(feature): Add pagination (IRv35)

amckinney:
This adds the `pagination` configuration to the IR as well as the user-facing Fern definition and OpenAPI extensions.

## Fern Defintion

The `pagination` can be declared at the API-level, and simply toggled on or off at the endpoint level like so:

```yaml
# api.yml
pagination:
  page: $request.cursor
  next: $response.next
  results: $response.results
```

```yaml
# service.yml
service:
  endpoints:
    listOrganizations:
      path: /organizations
      request:
        name: ListOrganizationsRequest
        method: GET
        pagination: true
        query-parameters:
          cursor: optional<string>
      response: LisOrganizationsResponse
```

Alternatively, this can be specified at the endpoint-level like so:

```yaml
# service.yml
service:
  endpoints:
    listOrganizations:
      path: /organizations
      request:
        name: ListOrganizationsRequest
        method: GET
        pagination:
          page: $query.cursor
          next: $response.next
          results: $response.data
        query-parameters:
          cursor: optional<string>
      response: LisOrganizationsResponse
```

## OpenAPI Extension

The `x-fern-pagination` extension is used just like the Fern `pagination` configuration. It can be specified like so:

```yaml
paths:
  "/issues":
    get:
      x-fern-pagination:
        page: $query.cursor
        next: $responses.next
        results: $responses.results
      parameters:
      ...
      - in: query
        name: cursor
        schema:
          type: string
```

This also works if specified on the document-level like so:

```yaml
x-fern-pagination:
  page: $query.cursor
  next: $responses.next
  results: $responses.results
paths:
  "/issues":
    get:
      x-fern-pagination: true
      parameters:
      ...
      - in: query
        name: cursor
        schema:
          type: string
```


packages/cli/yaml/yaml-schema/src/ast/visitors/services/visitHttpService.ts
```diff
@@ -269,7 +269,8 @@ async function visitEndpoint({
                     example
                 });
             }
-        }
+        },
+        pagination: noop
```
dsinghvi, line 273: @amckinney do we want to add a `fern check` rule to verify that all the parameters exist?

packages/cli/yaml/yaml-schema/src/schemas/PaginationSchema.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const PaginationSchema = z.object({
+    page: z.string(),
+    next: z.string(),
+    results: z.string()
+});
```
dsinghvi, line 7: Thoughts on start to using the [.describe()](https://zod.dev/?id=describe) method on these parameters -- would love to eventually start adding auto-generated reference documentation and I think this will help more folks in the codebase

packages/cli/openapi-parser/src/v3/extensions/fernExtensions.ts
```diff
@@ -62,6 +62,20 @@ export const FernOpenAPIExtension = {
...
+     *
+     * Example usage:
+     *   paths:
+     *     /path/to/my/endpoint:
+     *       x-fern-pagination:
+     *         page: $request.page
+     *         next: $response.next
+     *         results: $response.results
+     */
+    PAGINATION: "x-fern-pagination",
```
dsinghvi, line 77: should we also support at the root level? 

Then the OpenAPI importer can even detect which endpoints should be paginated automatically. (can do this in a follow-up).

dsinghvi (approved):

packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
```diff
@@ -198,6 +199,18 @@ types:
...
+    docs: |
+      If set, the endpoint will be generated with auto-pagination features.
+
+      The page must be defined as a query parameter included in the request,
+      whereas the next page and results are resolved from properties defined
+      on the response.
+    properties:
+      page: QueryParameter
+      next: types.ObjectProperty
+      results: types.ObjectProperty
```
amckinney, line 212: Thinking this through some more -

This might not actually be enough information for the generators to deduce where the property is defined. This works fine if the property is a top-level property within the response object like so:

```yaml
types:
  Response:
    properties:
      next: string
      results: list<User>
```

However, if the property is nested within another property like so:

```yaml
types:
  PageInfo:
    properties:
      next: string

  Response:
    properties:
      pageInfo: PageInfo
      results: list<User>
```

In this case, we still want to support a pagination configuration that permits nested properties, i.e.

```yaml
pagination:
  page: $request.cursor
  next: $response.pageInfo.next
  results: $response.results
```

--

We'll either need to leave in some breadcrumbs for the generator to identify where the `next` and `results` are located, or something else similar. Something like,

```yaml
PageProperty:
  properties:
    breadcrumbs: list<commons.Name> # Empty if the response defines the property at the top-level.
    property: types.ObjectProperty
```

amckinney: I went ahead went this change for now.

dsinghvi (approved):

packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
amckinney: Discussed offline (as shown by the image above). In the future we'll reduce the size of the IR by using IDs wherever possible. For now, this is fine as-is.

packages/cli/yaml/yaml-schema/src/ast/visitors/services/visitHttpService.ts
amckinney: Discussed offline, we'll follow-up with the `fern check` and fern compiler changes in a follow-up PR. This at least gets the interface into IRv35 so the generators can start to act upon the shape.

PR #2988, opened 2024-02-15
python, fix: revert regressions in writing circular references

armandobelardo:
We're currently scoping a rewrite for circular references now, so reverting https://github.com/fern-api/fern/pull/2719 which had solved some issues but introduced more

### DIFF
--- a/seed/fastapi/api-wide-base-path/resources/service/service/service.py
+++ b/seed/fastapi/api-wide-base-path/resources/service/service/service.py
@@ -68,7 +68,7 @@ class AbstractServiceService(AbstractFernService):
         wrapper.__globals__.update(cls.post.__globals__)
 
         router.post(
-            path="//test/{path_param}/{service_param}/{endpoint_param}",
+            path="/test/{path_param}/{service_param}/{endpoint_param}",
             response_model=None,
             status_code=starlette.status.HTTP_204_NO_CONTENT,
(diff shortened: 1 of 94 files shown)


seed/fastapi/api-wide-base-path/resources/service/service/service.py
```diff
@@ -68,7 +68,7 @@ def wrapper(*args: typing.Any, **kwargs: typing.Any) -> None:
         wrapper.__globals__.update(cls.post.__globals__)
 
         router.post(
-            path="/test/{path_param}/{service_param}/{endpoint_param}",
+            path="//test/{path_param}/{service_param}/{endpoint_param}",
```
dsinghvi, line 71: seems like we introduced a bug with the slashes

seed/fastapi/basic-auth/resources/basic_auth/service/service.py
```diff
@@ -121,7 +121,7 @@ def wrapper(*args: typing.Any, **kwargs: typing.Any) -> bool:
         wrapper.__globals__.update(cls.post_with_basic_auth.__globals__)
 
         router.post(
-            path="//basic-auth",
+            path="/basic-auth",
```
dsinghvi, line 124: and we removed the bug here.. would be good to figure out whats changing

dsinghvi (approved):

PR #2993, opened 2024-02-15
improvement: allow files to be arrays within the IR

armandobelardo:
will put up a separate PR to consume this in Python

### DIFF
--- a/packages/cli/yaml/validator/src/rules/no-undefined-type-reference/__test__/no-undefined-type-reference.test.ts
+++ b/packages/cli/yaml/validator/src/rules/no-undefined-type-reference/__test__/no-undefined-type-reference.test.ts
@@ -33,12 +33,6 @@ describe("no-undefined-type-reference", () => {
                 severity: "error"
             },
             {
-                message: "The file type can only be used as properties in inlined requests.",
-                nodePath: ["service", "endpoints", "listOfFiles", "request", "body", "properties", "file"],
-                relativeFilepath: RelativeFilePath.of("file-upload.yml"),
-                severity: "error"
-            },
-            {
                 message: "Type MissingType is not defined.",
                 nodePath: ["types", "MyType"],
                 relativeFilepath: RelativeFilePath.of("simple.yml"),


packages/ir-sdk/fern/apis/ir-types-latest/generators.yml
```diff
@@ -18,7 +18,7 @@ groups:
         output:
           location: npm
           url: npm.buildwithfern.com
-          package-name: '@fern-fern/ir-sdk'
+          package-name: '@fern-fern/ir-v34-sdk'
```
dsinghvi, line 21: lets keep this `ir-sdk` otherwise will break a shit ton of imports in typescript bc you import from the package directly

armandobelardo: Updated!

packages/cli/generation/ir-generator/src/converters/services/convertHttpRequestBody.ts
```diff
@@ -43,13 +44,27 @@ export function convertHttpRequestBody({
...
-                        isOptional: property.isOptional
-                    });
+                    if (property.isArray) {
+                        return FileUploadRequestProperty.file(
+                            FileProperty.fileArray({
+                                key: file.casingsGenerator.generateNameAndWireValue({
+                                    wireValue: property.key,
+                                    name: property.key
+                                }),
+                                isOptional: property.isOptional
```
dsinghvi, line 54: I don't think we'll support `list<optional<file>>`, thoughts on disallowing that ?

dsinghvi: `optional<list<file>>` would make sense though

armandobelardo: This is for optional<list<file>>, not list<optional<file>>

dsinghvi:
It looks like all the relevant parts are updated, but i think we should add some more test cases to the file-upload test [definition](https://github.com/fern-api/fern/tree/main/test-definitions/fern)

Specifically would want to test: 
- optional<file>
- list<file>
- optional<list<file>>

Once you add these as a test definition you can also run `yarn workspace @fern-api/ir-generator test -u` and this will update the IR snapshots

packages/cli/generation/ir-generator/src/converters/services/convertHttpRequestBody.ts
armandobelardo: It tries to follow the existing paradigm which is isOptional flags theproperty as optional:
image

armandobelardo: You can also see where this is consumed, it makes `optional<list<file>>`, not `list<optional<file>>`

packages/cli/generation/ir-migrations/src/migrations/v34-to-v33/migrateFromV34ToV33.ts
```diff
@@ -0,0 +1,78 @@
...
+        [GeneratorName.RUBY_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR,
+        [GeneratorName.CSHARP_MODEL]: GeneratorWasNeverUpdatedToConsumeNewIR,
+        [GeneratorName.CSHARP_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR
+    },
+    jsonifyEarlierVersion: (ir) =>
+        IrSerialization.V33.IntermediateRepresentation.jsonOrThrow(ir, {
+            unrecognizedObjectKeys: "strip",
+            skipValidation: true
+        }),
+    migrateBackwards: (V34): IrVersions.V33.ir.IntermediateRepresentation => {
```
dsinghvi, line 43: looks great -- if you look at the other migrations they have a simple test so may make sense to add a test here as well

packages/cli/generation/ir-generator/src/converters/services/convertHttpRequestBody.ts
dsinghvi: ah thank you!

packages/cli/generation/ir-migrations/src/migrations/v34-to-v33/migrateFromV34ToV33.ts
dsinghvi: thoughts on adding a log here for which endpoint is affected 

```
GET /foo/bar accepts a list of files however the {generator.name}@{generatior.version} only supports accepting a single file. File an [issue](https://github.com/fern-api/fern/issues) to add request for file arrays in {generator.name}@{generatior.version}! 
```

armandobelardo: Added this log!

dsinghvi:
@armandobelardo i think you'll have to update some seed snapshots

armandobelardo:
> @armandobelardo i think you'll have to update some seed snapshots

Updating! Forgot to when I added the tests here

dsinghvi (approved):
One of the tests is failing but other than that this is g2g. FYI @amckinney that this may cause some merge conflicts with your pagination PR.

PR #3010, opened 2024-02-18
improvement, python: update python file type to be more reflective or HTTPX types and allow lists of files

armandobelardo:
Note from: https://www.python-httpx.org/advanced/#multipart-file-encoding

That to allow for lists of files to be uploaded we need to move from dicts of files to a list of tuples, where each element in the list is added as a member of the tuple with the first element being it's key


seed/python-sdk/alias/src/seed/core/file_types.py
```diff
@@ -0,0 +1,32 @@
...
+    typing.Tuple[typing.Optional[str], FileContent],
+    # (filename, file (or bytes), content_type)
+    typing.Tuple[typing.Optional[str], FileContent, typing.Optional[str]],
+    # (filename, file (or bytes), content_type, headers)
+    typing.Tuple[typing.Optional[str], FileContent, typing.Optional[str], typing.Mapping[str, str]],
+]
+
+# The format we use is a list of tuples, where the first element is the name of the file and the second is the file object
+# Typically HTTPX wants a dict, but to be able to send lists of files, you have to use the list approach (which also works for non-lists)
+# https://github.com/encode/httpx/pull/1032
```
dsinghvi, line 21: nit: let's reduce the length of the docs here / use a multiline string (whatever makes it look prettier)
Screenshot 2024-02-19 at 11 24 35 AM

seed/python-sdk/alias/src/seed/core/file_types.py
```diff
@@ -0,0 +1,32 @@
+# This file was auto-generated by Fern from our API Definition.
+
```
dsinghvi, line 2: nit: can we call this file.py `core/file.py`

seed/python-sdk/file-upload/src/seed/resources/service/client.py
```diff
@@ -46,9 +49,13 @@ def post(
 
             - integer: int.
 
-            - file: typing.IO.
+            - file: FileTypes.
```
dsinghvi, line 52: I think it would be nice to append this string `See core.FileTypes for more documentation` to any file upload property. Ideally the `core.FileTypes` should link to the class or type directly.

generators/python/src/fern_python/snippet/type_declaration_snippet_generator.py
```diff
@@ -22,6 +23,8 @@ def __init__(
         self._generate_enum = enum
         self._generate_object = object
         self._generate_discriminated_union = discriminated_union
+        # TODO: implement undiscriminated_union snippets
+        # self._generate_undiscriminated_union = undiscriminated_union
```
dsinghvi, line 27: Thanks for adding a TODO here -- do you mind adding this TODO in linear as well so we can track

dsinghvi (changes requested):
Added a couple small comments on improving small parts of the generated code. 

2 larger things: 
- Were you able to test this some place ETE ? 
- CHANGELOG entry + VERSION bump

seed/python-sdk/file-upload/src/seed/core/file_types.py
```diff
@@ -0,0 +1,32 @@
+# This file was auto-generated by Fern from our API Definition.
+
+import typing
+
+# File typing inspired by the flexibility of types within the httpx library
+# https://github.com/encode/httpx/blob/master/httpx/_types.py
+FileContent = typing.Union[typing.IO[bytes], bytes, str]
+FileTypes = typing.Union[
```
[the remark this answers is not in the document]
armandobelardo: I don't think Python has a File type, but this would confusing me into thinking it has one and we're using it

seed/python-sdk/file-upload/src/seed/resources/service/client.py
armandobelardo: Updated, though it doesn't match the rest of the usage of classes from `core`

generators/python/sdk/CHANGELOG.md
```diff
@@ -5,47 +5,70 @@ All notable changes to this project will be documented in this file.
...
+      FileContent,
+      # (filename, file (or bytes))
+      typing.Tuple[typing.Optional[str], FileContent],
+      # (filename, file (or bytes), content_type)
+      typing.Tuple[typing.Optional[str], FileContent, typing.Optional[str]],
+      # (filename, file (or bytes), content_type, headers)
+      typing.Tuple[typing.Optional[str], FileContent, typing.Optional[str], typing.Mapping[str, str]],
+  ]
+  ```
+- Fix: Python now supports API specifications that leverage lists for file upload. Previously, Fern incorrectly made all `list<file>` type requests simply `file`.
```
dsinghvi, line 24: Also mind adding a function signature of an endpoint to show

generators/python/sdk/CHANGELOG.md
```diff
@@ -5,47 +5,70 @@ All notable changes to this project will be documented in this file.
...
+      # file (or bytes)
+      FileContent,
+      # (filename, file (or bytes))
+      typing.Tuple[typing.Optional[str], FileContent],
+      # (filename, file (or bytes), content_type)
+      typing.Tuple[typing.Optional[str], FileContent, typing.Optional[str]],
+      # (filename, file (or bytes), content_type, headers)
+      typing.Tuple[typing.Optional[str], FileContent, typing.Optional[str], typing.Mapping[str, str]],
+  ]
+  ```
```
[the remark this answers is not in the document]
dsinghvi: and i think we can include examples of calling it using a couple of the different options for `httpx.File`

generators/python/sdk/CHANGELOG.md
```diff
@@ -5,47 +5,70 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.4] - 2024-02-19
```
dsinghvi, line 8: let's make this a minor version release `0.11.0`

dsinghvi (approved):
Approving since it mostly LGTM -- can you pull from main and regenerate the seed snapshots? I think your test definitions are out of date

PR #3018, opened 2024-02-20
(internal) rename Websocket to WebSocket and bump fdr

abvthecity:
* rename Websocket to WebSocket everywhere
* bump fdr (see [packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts](https://github.com/fern-api/fern/pull/3018/files#diff-4bfb17a7005fccf04ed68d284bf6761c6b4f2b2bd97da4d91f13f00abb3c5040R133))

### DIFF
--- a/packages/cli/generation/ir-generator/src/__test__/__snapshots__/websocket.txt
+++ b/packages/cli/generation/ir-generator/src/__test__/__snapshots__/websocket.txt
@@ -176,6 +176,7 @@ exports[`generate IR 1`] = `
     "channel_realtime": {
       "auth": true,
       "availability": null,
+      "displayName": null,
       "docs": null,
       "examples": [
         {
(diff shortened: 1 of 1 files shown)


packages/cli/generation/ir-generator/src/converters/services/convertExampleEndpointCall.ts
```diff
@@ -514,6 +514,7 @@ function buildUrl({
             ...pathParams.servicePathParameters,
             ...pathParams.rootPathParameters
         ]) {
+            // TODO: should we URL encode the value?
```
dsinghvi, line 517: I think it makes sense to only do this on the frontend? That way if you just want to render the value somewhere where it doesn't make sense to url encode, you don't need to

dsinghvi (approved):
Looking 🔥 , once you add the test definition and run `yarn workspace @fern-api/ir-generator test -u` you'll see the snapshots come through. 

You will also need to run seed `yarn local:seed test --fixture websocket --skip-scripts` (i can also help with this one once you're at this stage)

PR #3025, opened 2024-02-20
feature, python: allow for users to define custom exports from __init__.py

armandobelardo:
This also adds docstrings to root clients for usability
And adds in flags for `fern write-overrides` to be initialize model names, as well as point to an existing overrides file to start from

### DIFF
--- a/packages/cli/cli/src/commands/generate-overrides/writeOverridesForWorkspaces.ts
+++ b/packages/cli/cli/src/commands/generate-overrides/writeOverridesForWorkspaces.ts
@@ -66,7 +66,7 @@ async function writeDefinitionForOpenAPIWorkspace({
     const openApiIr = await getOpenAPIIRFromOpenAPIWorkspace(workspace, context);
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    let existingOverrides: Record<string, any> = {};
+    let existingOverrides: any = {};
     if (overridesFilepath !== undefined) {
         existingOverrides = await readExistingOverrides(overridesFilepath, context);
     }
@@ -92,7 +92,7 @@ async function writeDefinitionForOpenAPIWorkspace({
             }
             sdkMethodNameExtensions["x-fern-sdk-method-name"] = endpointLocation.endpointId;
             pathItem[endpoint.method.toLowerCase()] = sdkMethodNameExtensions;
-        } else if (existingOverrides === undefined) {
+        } else if (existingOverrides == null) {
             context.logger.warn(`Endpoint ${endpoint.path} ${endpoint.method} is defined multiple times`);
         }
     }


packages/cli/cli/src/cli.ts
```diff
@@ -720,20 +720,32 @@ function addWriteOverridesCommand(cli: Argv<GlobalCliOptions>, cliContext: CliCo
...
             yargs.option("api", {
                 string: true,
                 description: "Only run the command on the provided API"
             }),
+            yargs.option("include-models", {
+                boolean: true,
+                description:
+                    "When generating the initial overrides, also stub the models (in addition to the endpoints)"
+            }),
+            yargs.option("existing-overrides", {
```
dsinghvi, line 733: can we detect this by looking at their `generators.yml` ? You should be able to get access to that from the TaskContext

packages/cli/cli/src/cli.ts
```diff
@@ -720,20 +720,32 @@ function addWriteOverridesCommand(cli: Argv<GlobalCliOptions>, cliContext: CliCo
     cli.command(
         "write-overrides",
         "Generate a basic openapi overrides file.",
-        (yargs) =>
+        (yargs) => [
             yargs.option("api", {
                 string: true,
                 description: "Only run the command on the provided API"
             }),
+            yargs.option("include-models", {
```
dsinghvi, line 728: Let's just always default this to true for now

dsinghvi: Can always delete what we don't like

generators/python/sdk/CHANGELOG.md
```diff
@@ -5,6 +5,45 @@ All notable changes to this project will be documented in this file.
...
+    """
+    Use this class to access the different functions within the SDK. You can instantiate any number of clients with different configuration that will propogate to these functions.
+    ---
+    from seed.client import SeedTest
+
+    client = SeedTest(
+        token="YOUR_TOKEN",
+        base_url="https://yourhost.com/path/to/api",
+    )
+    """
```
dsinghvi, line 44: do you think it's easy to add the parameter level documentation as well just like we do for endpoints? 

```python
token: Optional[str]. The token. 

base_url: Optional[str]. The Sever URL used by the SDK. 
```

dsinghvi (changes requested):
Couple small comments. Let's all add a snapshot test in `cli/ete-tests` that calls `write-overrides` !

generators/python/sdk/CHANGELOG.md
armandobelardo: Let me check!

packages/cli/cli/src/cli.ts
armandobelardo: totally

generators/python/sdk/CHANGELOG.md
armandobelardo: Updated!

dsinghvi (approved):

PR #3035, opened 2024-02-22
(feature): Add Go snippet generation

amckinney:
This introduces Go snippet generation for any API definition that defines examples, such as the following:

```go
import (
  context "context"
  time "time"
  acme "github.com/acme/acme-go"
  acmeclient "github.com/acme/acme-go/client"
  option "github.com/acme/acme-go/option"
)

client := acmeclient.NewClient(
  option.WithToken(
    "<YOUR_AUTH_TOKEN>",
  ),
)
response, err := client.User.Create(
  context.TODO(),
  &acme.CreateUserRequest{
    Username: "john",
    Language: acme.LanguageEnglish,
  },
)
```

Given that this was such a substantial feature, this PR also moved the Go generator's seed tests into the top-level `seed` directory so that all of the generated Go fixtures exist alongside the rest of the generators.

This also resolved a small bug that caused package documentation to not be generated into the correct `doc.go` file, which sometimes resulted in non-deterministic generation results (i.e. the last `doc.go` won).

### DIFF
--- a/seed/python-sdk/alias/poetry.lock
+++ b/seed/python-sdk/alias/poetry.lock
@@ -85,13 +85,13 @@ files = [
 
 [[package]]
 name = "httpcore"
-version = "1.0.3"
+version = "1.0.4"
 description = "A minimal low-level HTTP client."
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "httpcore-1.0.3-py3-none-any.whl", hash = "sha256:9a6a501c3099307d9fd76ac244e08503427679b1e81ceb1d922485e2f2462ad2"},
-    {file = "httpcore-1.0.3.tar.gz", hash = "sha256:5c0f9546ad17dac4d0772b0808856eb616eb8b48ce94f49ed819fd6982a8a544"},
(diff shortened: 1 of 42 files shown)


generators/go/sdk/CHANGELOG.md
```diff
@@ -7,11 +7,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ## Unreleased -->
 
+## Unreleased
+
+- Fix: Package documentation is now generated into the correct package's `doc.go`.
+- Feature: Add support for generated endpoint snippets.
```
dsinghvi, line 13: nit: I think it will be useful to specify where the code snippets will show up and in this case it will be for the documentation and over the snippets API but not embedded in the SDK itself

dsinghvi (approved):
Thanks for moving the seed snapshots to the root

generators/go/sdk/CHANGELOG.md
amckinney: Great call, will update 👍

PR #3040, opened 2024-02-22
fix, python: move from lists to sequences when using lists in function signatures

armandobelardo:
Transition from lists to sequences within function calls, this is a fix as a result of how mypy handles type variance.
This fix is only for function calls as testing shows that we do not hit the same issue within mypy with list[union[*]] fields on pydantic objects.
This issue outlines it well: https://stackoverflow.com/questions/76138438/incompatible-types-in-assignment-expression-has-type-liststr-variable-has

Also worth noting that in my testing you can pass in lists just fine (even when hard typed:)
```
l: list[str] = ["string"]
func(l)
```


dsinghvi (approved):

armandobelardo:
The TS issues are unrelated to this PR, so going to merge and look into the TS issue separately

PR #3043, opened 2024-02-23
fix: ensure we apply audience-based filtering to examples as well

armandobelardo:
Also allows for specifiying audiences in your seed tests, and also pass configuration to any custom fixture you're running
Since we can only run one custom fixture right now, this is just specified as a single config block in your seed workspace
Ex:
```
irVersion: v31
docker: fernapi/fern-typescript-node-sdk:local
dockerCommand: yarn workspace @fern-typescript/sdk-generator-cli dockerTagLocal:node
language: typescript
generatorType: SDK
customFixtureConfig:
  customConfig: 
    useBrandedStringAliases: true
    neverThrowErrors: true
    namespaceExport: "Fiddle"
    bundle: true
    includeUtilsOnUnionMembers: true
    includeOtherInUnionTypes: true
    extraDependencies: 
      lodash-es: "^4.17.21"
    extraDevDependencies: 
      jest: "^29.7.0"
  outputFolder: dev-dependencies
```

### DIFF
--- a/packages/cli/generation/ir-generator/src/__test__/__snapshots__/enum.txt
+++ b/packages/cli/generation/ir-generator/src/__test__/__snapshots__/enum.txt
@@ -131,6 +131,7 @@ exports[`generate IR 1`] = `
               "request": {
                 "jsonExample": {
                   "operand": ">",
+                  "operandOrColor": "red",
                 },
                 "properties": [
                   {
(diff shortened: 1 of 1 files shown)


dsinghvi (approved):

packages/seed/src/commands/test/runDockerForWorkspace.ts
```diff
@@ -40,7 +42,7 @@ export async function runDockerForWorkspace({
 }): Promise<void> {
     const generatorGroup: GeneratorGroup = {
         groupName: "test",
-        audiences: ALL_AUDIENCES,
+        audiences: selectAudiences !== undefined ? { type: "select", audiences: selectAudiences } : ALL_AUDIENCES,
```
dsinghvi, line 45: `!= null` for safety

packages/cli/generation/ir-generator/src/filterExamples.ts
```diff
@@ -0,0 +1,336 @@
...
+                ? {
+                      ...s,
+                      object: {
+                          properties: s.object.properties
+                              .filter((p) => filteredIr.hasProperty(p.originalTypeDeclaration.typeId, p.name.wireValue))
+                              .map((p) => ({
+                                  ...p,
+                                  value: filterExampleTypeReference(filteredIr, p.value)
+                              }))
+                              .filter((p) => p.value !== undefined) as ExampleObjectProperty[]
```
dsinghvi, line 36: Why do we have to do a cast here: `as ExampleObjectProperty[]` ?

armandobelardo: It still thinks there are undefined values without the cast, this is a pretty common thing

packages/cli/generation/ir-generator/src/filterExamples.ts
```diff
@@ -0,0 +1,336 @@
...
+                  })
+                : undefined,
+        unknown: () => exampleTypeReference,
+        _other: () => exampleTypeReference
+    });
+}
+
+function filterExamplePathParameters(
+    filteredIr: FilteredIr,
+    pathParameters: ExamplePathParameter[]
```
dsinghvi, line 198: you'll notice that a lot of the typescript functions in the CLI don't take parameters but take a single object with name properties instead: 

```ts
filterExamplePathParameters({ filteredIr, pathParameters }: { filteredIr: FilteredIr, pathParameters: ExamplePathParameter[]);
```

dsinghvi: this makes it easier to add more parameters down the road (if necessary) and also makes it more readable when you have complex expressions for the value of the parameter

packages/cli/generation/ir-generator/src/filterExamples.ts
dsinghvi: whenever i've encountered this pattern (having to filter the nulls at the end) I've typically found that its better to use a normal `for` loop

armandobelardo: Yeah! I think it's a matter of preference 
I like the map and filter but I can change it if you'd prefer

dsinghvi (approved):
3 things we have followed in the codebase so far: 

1. Use `!= null` instead of `!== undefined` b/c types don't exist at runtime so it's safer
2. Minimize usage of casts (such as the `as Type[]`) 
3. Pass objects into functions instead of parameters so that its easier to add parameters over time and the code is more readable. 

Those practices make sense to me so would be nice to continue with them

packages/cli/generation/ir-generator/src/filterExamples.ts
dsinghvi: Yeah we've just stayed away from any usage of `as` to prevent any weird bugs and have the compiler help as much as possible. (the situation I'm concerned about is what if someone modifies the map but the compiler doesn't give them an error because of the cast)

armandobelardo: Cool, removed them

PR #3050, opened 2024-02-24
(fix): Fix Go snippets for optional primitive aliases

amckinney:
Simple bug fix for supporting aliases to optional primitive values. With this, the generated snippet will include the top-level pointer helpers (e.g. `acme.String(...)`).

Note that this includes a new test case within the `examples` definition, so all of the `example` snapshots are updated for every generator.


amckinney:
> [enum]:                              Found 7 errors and 0 warnings. Run fern check --warnings to print out the warnings.

Looks like `fern check` is unhappy with the `enum` test definition. We can fix this in a follow-up PR targeted at that fix (especially given that it will update fixtures across every generator).

dsinghvi (approved):

PR #3054, opened 2024-02-24
fix: build seed docker multiplatform

armandobelardo:
Ensure seed containers are built to run on diff platforms


.github/workflows/build-seed.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - ab/fixSeed
```
dsinghvi, line 7: lets take this out before merging!

armandobelardo: good catch! removed!

dsinghvi:
@armandobelardo it took 49m to build the seed container for python, should we just not make it multiplatform?

armandobelardo:
> @armandobelardo it took 49m to build the seed container for python, should we just not make it multiplatform?

Yeah it's horrible, nothing I tried made it faster though

When it's not multiplatform we just can't run it locally, the container that's used for seed in Actions is different from the one we use on macs, so we need to build for both archs

We could change the base runners for the actions though so architectures match, I don't have much of a preference since this image shouldn't change

dsinghvi:
> We could change the base runners for the actions though so architectures match, I don't have much of a preference since this image shouldn't change

Let's do this, 49m on the Seed Runner is quite expensive for Github actions

dsinghvi (approved):

PR #3055, opened 2024-02-26
(fix): `fern check` no longer throws when an undiscriminated union is a list of primitives

dsinghvi:
Previously when an undiscriminated union was used as a path parameter, query parameter, or header `fern check` would throw. Now, if the undiscriminated union is just a list of primitives then the CLI will accept that as a schema for http parameters.


dsinghvi:
@armandobelardo I believe this [PR](https://github.com/fern-api/fern/pull/3044) introduced `fern check` failures to the test definition as well (generally want to make sure that `fern check` passes on the test definitions whenever we edit them).

dsinghvi:
> > @armandobelardo I believe this [PR](https://github.com/fern-api/fern/pull/3044) introduced `fern check` failures to the test definition as well (generally want to make sure that `fern check` passes on the test definitions whenever we edit them).
> 
> Can we make fern check run in ci for changes to test-definitions? I fix this in my autogenerated examples PR as well

It does run -- 
Screenshot 2024-02-25 at 7 23 41 PM


> > @armandobelardo I believe this [PR](https://github.com/fern-api/fern/pull/3044) introduced `fern check` failures to the test definition as well (generally want to make sure that `fern check` passes on the test definitions whenever we edit them).
> 
> Can we make fern check run in ci for changes to test-definitions? I fix this in my autogenerated examples PR as well

Great idea, fixed

PR #3059, opened 2024-02-26
[wip] feat: add flag to disable Pydantic validation and keep extra fields on the Pydantic model

armandobelardo:
This is meant to add flexibility to our models, which have very strict validation right now (strict type checking on fields as well as dropping additional fields)

This approach appeared to work when testing, but I still need to test the generated version.
Known issues to fix:
- containers
- primitives


armandobelardo:
Closing in favor of https://github.com/fern-api/fern/pull/3311

PR #3072, opened 2024-02-27
(chore): Refactor Pagination IR to support offset

amckinney:
WIP - will still need to carve out the OpenAPI extension to support both pagination forms.

### DIFF
--- a/fern.schema.dev.json
+++ b/fern.schema.dev.json
@@ -643,23 +643,52 @@
                   "pagination": {
                     "anyOf": [
                       {
-                        "type": "object",
-                        "properties": {
-                          "page": {
-                            "type": "string",
-                            "description": "The request property name that represents the page cursor or offset."
+                        "anyOf": [
+                          {
+                            "type": "object",
+                            "properties": {
+                              "type": {
+                                "type": "string",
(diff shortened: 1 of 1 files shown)


packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
```diff
@@ -201,6 +201,13 @@ types:
       error: errors.DeclaredErrorName
 
   Pagination:
+    docs: |
+      If set, the endpoint will be generated with auto-pagination features.
+    union:
+      cursor: CursorPagination
+      offset: OffsetPagination
+
+  CursorPagination:
```
dsinghvi, line 210: this will be a breaking change so we'll have to do an IR migration (assume you're already tracking)

amckinney: I figured we would be able to get away with it because there are zero consumers of it (both in the `fern` compiler as well as all the other generators). As long as we update the respective IR SDKs this should be ok, no? I might be missing something though.

dsinghvi (approved):

PR #3082, opened 2024-02-29
improvement, python: swap to literals instead of enums

armandobelardo:
This allows for us to have forward and backward compatibility with "enums" without needing for unions that would make it annoying to manage as a user


dsinghvi (approved):

armandobelardo:
Just tested and things look good, going to merge this in!

PR #3096, opened 2024-03-03
feat, python: add in max_retries with exponential backoff

armandobelardo:


seed/python-sdk/audiences/src/seed/core/client_wrapper.py
```diff
@@ -24,10 +26,10 @@ def get_base_url(self) -> str:
 class SyncClientWrapper(BaseClientWrapper):
     def __init__(self, *, base_url: str, httpx_client: httpx.Client):
         super().__init__(base_url=base_url)
-        self.httpx_client = httpx_client
+        self.httpx_client = HttpClient(httpx_client=httpx_client)
```
dsinghvi, line 29: nit, should we call this `self.client` now ?

dsinghvi (approved):
excited to get this in 🚀

seed/python-sdk/audiences/src/seed/core/client_wrapper.py
armandobelardo: I kept it httpx_ since it's a thin wrapper and we use client so much that I wanted it to have some prefix haha

PR #3104, opened 2024-03-04
chore: cache docker builds in github actions

mscolnick:
This uses https://docs.docker.com/build/cache/backends/gha/ to cache the docker builds via in GitHub actions

### DIFF
--- a/.github/workflows/publish-generator.yml
+++ b/.github/workflows/publish-generator.yml
@@ -186,6 +186,8 @@ jobs:
           context: ./generators/python
           file: ./generators/python/fastapi/Dockerfile
           platforms: linux/amd64,linux/arm64
+          cache-from: type=gha
+          cache-to: type=gha,mode=min
           push: true
           tags: fernapi/fern-fastapi-server:${{env.VERSION}}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/seed-dockers.yml
```diff
@@ -22,12 +25,12 @@ jobs:
...
-            ts: 
+            ts:
               - 'docker/seed/Dockerfile.ts'
-              - '.github/workflows/build-seed.yml'
-            java: 
+              - '.github/workflows/seed-dockers.yml'
+            java:
               - 'docker/seed/Dockerfile.java'
-              - '.github/workflows/build-seed.yml'
+              - '.github/workflows/seed-dockers.yml'
```
mscolnick, line 33: i think there was a file namechange that was missed

.github/workflows/publish-generator.yml
```diff
@@ -130,6 +133,8 @@ jobs:
           context: ./generators/python
           file: ./generators/python/pydantic/Dockerfile
           platforms: linux/amd64,linux/arm64
+          cache-from: type=gha
```
armandobelardo, line 136: Meganit: do we want to specify a cache-from here?

Also should we do this for every docker push step?

armandobelardo (approved):
Approving, just the one comment on doing this for every generator container!

.github/workflows/publish-generator.yml
mscolnick: oh woops didnt realize there were so many more of them. ill add them below too.

> Meganit: do we want to specify a cache-from here?
are you suggesting putting it somewhere else? i think this is the best spot for it.

armandobelardo (approved):

PR #3110, opened 2024-03-05
chore: Use turborepo

mscolnick:
This migrates some scripts and GitHub flows to use [turborepo](https://turbo.build/repo). 
The reason for this is to cache tasks across builds using turborepo's remote caching. 

Some other benefits to gain down the road:
* better task management (and cleaner scripts)
* local caching (improved devX for contributors)

Couple blockers/improvements:
- [ ] turbo caches builds per package. but a `yarn compile` may compile other packages (based off project references). this wont get included in the cached outputs, so when we have a "cache hit", we don't get the compiled deps.
- [ ] turbo and yarn pnp doesn't work well together. need to do more testing to see what the limitations are


mscolnick:
> Do we need to change any of these: https://github.com/fern-api/fern/blob/main/.circleci/config.yml ?

@dsinghvi only if we want caching. but the fallback still works as is. maybe something we can wait for until migrated to GHA (it seems like its not part of the PR flow, just main/tags)

dannysheridan:
@dsinghvi are you planning to close this PR?

PR #3113, opened 2024-03-05
docs: add titles and descs

minaelee:
add titles and descs to all frontmatter and update existing to use title appended from docs.yml
also:
- fix some links for generators due to slugs
- add slug 'commands' to cli-commands section and update links
- update ferndef-overview with petstore example and information about converting from openapi
- delete definition.mdx as replaced by ferndef-overview

### DIFF
--- a/fern/docs.yml
+++ b/fern/docs.yml
@@ -76,29 +76,29 @@ navigation:
             slug: commands
             contents:
               - page: fern generate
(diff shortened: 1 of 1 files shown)


dannysheridan:
@minaelee can you check the `check-definitions` CI failure?

minaelee:
> @minaelee can you check the `check-definitions` CI failure?

done! it's passing now.

PR #3115, opened 2024-03-06
(feature): introduce more layout options for docs configuration

abvthecity:


packages/docs-config-sdk/fern/definition/docs.yml
```diff
@@ -341,8 +399,60 @@ types:
 
   ColorsConfiguration:
     properties:
-      accentPrimary: optional<ColorConfig>
```
dsinghvi, line 344: @abvthecity this is a break which is why the ete tests are failing

PR #3122, opened 2024-03-07
chore: add bones for csharp snippets

armandobelardo:


dsinghvi (approved):
Should the AST be SDK aware? My initial thought was no but also see the value

PR #3137, opened 2024-03-09
fix: don't require a schema to exist under `application/octet-stream`

armandobelardo:
This reflects OAS specs: https://spec.openapis.org/oas/v3.1.0#considerations-for-file-uploads


seed/python-sdk/exhaustive/additional_init_exports
armandobelardo, line 1: I'm not sure what's causing this

dsinghvi (approved):

PR #3139, opened 2024-03-10
improvement: add in query param representation to the IR

armandobelardo:
Updates query parameters to specify the way they should be represented within the resultant HTTP request.
 Specifically, for parameters that are lists, we allow for two types of representations:
 - `exploded` = `a="one"&a="two"&a="three"`
 - `comma-separated` = `a=one,two,three,four`


packages/cli/openapi-parser/src/openapi/v3/extensions/fernExtensions.ts
```diff
@@ -317,7 +317,28 @@ export const FernOpenAPIExtension = {
...
+     *      parameters:
+     *          - name: id
+     *            in: query
+     *            description: Ids to retrieve
+     *            x-fern-query-parameter-representation: EXPLODED
+     *            schema:
+     *              type: array
+     *              items: string
+     */
+    FERN_QUERY_PARAMETER_REPRESENTATION: "x-fern-query-parameter-representation"
```
dsinghvi, line 341: OpenAPI should already have fields for this actually: https://swagger.io/docs/specification/serialization/

packages/cli/yaml/yaml-schema/src/schemas/HttpQueryParameterSchema.ts
```diff
@@ -4,7 +4,8 @@ import { extendTypeReferenceSchema } from "./TypeReferenceSchema";
 
 export const HttpQueryParameterSchema = extendTypeReferenceSchema(
     DeclarationWithNameSchema.extend({
-        "allow-multiple": z.optional(z.boolean())
+        "allow-multiple": z.optional(z.boolean()),
+        "query-parameter-representation": z.optional(z.enum(["EXPLODED", "COMMA_SEPARATED"]))
```
dsinghvi, line 8: - `encoding` or `format` are probably better keys than `query-parameter-representation` 
- lets not make the value screaming snake case (`exploded` and `comma_delimited`) 
- you can make the typing a union here where if `allow-multiple` is true then `encoding` needs to be specified (by leveraging z.union())

packages/cli/openapi-parser/src/openapi/v3/extensions/fernExtensions.ts
armandobelardo: Oh perfect, will update and have the options match for the fern def + IR

packages/cli/generation/ir-migrations/src/migrations/v38-to-v37/migrateFromV38ToV37.ts
```diff
@@ -0,0 +1,47 @@
...
+        [GeneratorName.CSHARP_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR
+    },
+    jsonifyEarlierVersion: (ir) =>
+        IrSerialization.V37.IntermediateRepresentation.jsonOrThrow(ir, {
+            unrecognizedObjectKeys: "strip",
+            skipValidation: true
+        }),
+    migrateBackwards: (V38, _context): IrVersions.V37.ir.IntermediateRepresentation => {
+        return {
+            ...V38
```
dsinghvi, line 44: if someone specifies the query parameter format as an unexploded list, should we reverse migrate to a string and add docs that they can specify list by doing `a,b,c` or something?

dsinghvi (dismissed):
Couple small changes requested! Another question -- do you know why the extra parameters end up getting added to older versions of the IR?

packages/cli/generation/ir-migrations/src/migrations/v38-to-v37/migrateFromV38ToV37.ts
armandobelardo: I was kinda thinking leaving it to revert to our current model of however the SDKs treat it today was best, but I'm not strongly opinionated here

packages/cli/yaml/yaml-schema/src/schemas/HttpQueryParameterSchema.ts
armandobelardo: Updated!

packages/cli/openapi-parser/src/openapi/v3/extensions/fernExtensions.ts
armandobelardo: Updated!

dannysheridan:
@armandobelardo can you merge or close this PR?

PR #3144, opened 2024-03-11
import float as unknown from openapi spec

buie:


packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -285,6 +285,16 @@ export function convertSchemaObject(
...
+    if (schema === "float") {
+        const inferredValue = schema.example ?? schema.default;
+        return SchemaWithExample.unknown({
+            nameOverride,
+            generatedName,
+            description,
+            groupName,
+            example: inferredValue
+        });
+    }
```
dsinghvi, line 297: we can likely convert to `number` 
```
return convertNumber({
            nameOverride,
            generatedName,
            format: schema.format,
            description,
            wrapAsNullable,
            example: getExampleAsNumber(schema),
            groupName
        });
```

dsinghvi: and in `format` you can put `float`

dsinghvi (approved):

PR #3168, opened 2024-03-13
fixing unit tests

armandobelardo:
a number of fixes to the mock server and python: 
- Stringifies headers (required as HTTPX does not do this, but rather hard fails when a non-encodable resource comes through)
- Unstringifies query parameters
- gets .text when receiveing a text response, as opposed to returning a response object
- adds a multibaseurl object to tests when needed
- fixes example generation for unions 
- a lot more


generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
```diff
@@ -709,9 +709,9 @@ def _get_headers_for_endpoint(
             if literal_header_value is not None and type(literal_header_value) is str:
                 headers.append((header.name.wire_value, AST.Expression(f'"{literal_header_value}"')))
             elif literal_header_value is not None and type(literal_header_value) is bool:
-                headers.append((header.name.wire_value, AST.Expression(f"{literal_header_value}")))
+                headers.append((header.name.wire_value, AST.Expression(f'"{str(literal_header_value).lower()}"')))
```
armandobelardo, line 712: boolean headers are pretty non-standard, but we need to stringify somehow

dsinghvi (approved):

generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
dsinghvi: str() makes sense to me

PR #3189, opened 2024-03-16
(docs): update fern write-definition docs

minaelee:
- update that write-definition outputs a `.definition/` folder (not `definition/`) and how to rename it for use
- add warning about hidden folders & tip about converting back to openapi

### DIFF
--- a/fern/docs/pages/overview/cli/cli-commands/fern-write-definition.mdx
+++ b/fern/docs/pages/overview/cli/cli-commands/fern-write-definition.mdx
@@ -3,7 +3,7 @@ title: Convert OpenAPI to Fern Definition
 description: The command "fern write-definition" converts your OpenAPI specification to a Fern Definition.
 ---
 
-The `fern write-definition` command converts your OpenAPI specification into a Fern Definition. To use it, you must have a `fern/openapi/` folder that contains an OpenAPI specification file in `json` format.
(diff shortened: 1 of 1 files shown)


fern/docs/pages/overview/cli/cli-commands/fern-write-definition.mdx
```diff
@@ -3,12 +3,18 @@ title: Convert OpenAPI to Fern Definition
 description: The command "fern write-definition" converts your OpenAPI specification to a Fern Definition.
 ---
 
-`fern write-definition` will convert your OpenAPI specification into a Fern Definition. A pre-requisite is a folder within your `fern/` folder named `openapi` that contains a `json` or `yml` file.
+The `fern write-definition` command converts your OpenAPI specification into a Fern Definition. To use it, you must have a `fern/openapi/` folder that contains an OpenAPI specification file in `json` format.
```
dannysheridan, line 6: I'm surprised by this. It feels like a bug, not a feature, of YAML isn't allowed. Can you double check?

minaelee: Surprised me too! It doesn't work, with either .yaml or .yml files, unless there's some undocumented way to use it that's different from simply running `fern write-definition` while in the fern folder. That's the way it works if the file is .json. 

I submitted a bug in Linear - https://linear.app/buildwithfern/issue/FER-1175/bug-fern-write-definition-only-works-for-json-files

minaelee: odd - wasn't working earlier for me, but working now. will update PR!

PR #3205, opened 2024-03-18
feat: API navigation overrides

abvthecity:
Steps

- [x] Add `navigation` field to generators schema
- [x] Plumb through navigation schema to docs (included via workspace?)
- [x] Convert items in the navigation schema from sdk group/method names to subpackageId and endpointId
- [ ] Add a fern check rule
- [ ] Add to write-overrides (optional)
- [ ] Write a test

### DIFF
--- a/packages/cli/cli/src/commands/register/registerWorkspacesV2.ts
+++ b/packages/cli/cli/src/commands/register/registerWorkspacesV2.ts
@@ -25,7 +25,7 @@ export async function registerWorkspacesV2({
                         context,
                         token,
                         audiences: { type: "all" },
-                        snippetsConfig: {}, 
+                        snippetsConfig: {},
                         navigation: undefined
                     });
                     context.logger.info(chalk.green("Registered API"));


abvthecity:
@dsinghvi some context and learnings for handoff:
- In the FDR.json, the `id` for endpoints/websockets/webhooks are the originalName, whereas the `id` for subpackages are generated. So convertIrToNavigation handles an exact match for endpoints/websockets/webhooks but looks up the ID for subpackages
- convertIrToNavigation should gracefully ignores incorrect config, so we'll offload the work of validating the config to fern check

where i ended (incomplete):
- adding generatorYaml to the fern check
- the IR is also required for validation (does it exist at this step?)

what is left to do:
- adding a utility to write-overrides (or some other command, like write-navigation). this should be relatively simple to do (since this is almost fully implemented in convertIrToNavigation). Otherwise it'd be incredibly painful to write the navigation structure manually.

PR #3220, opened 2024-03-20
(internal): document syntax highlighting

abvthecity:
Also

* adds "edit this page"
* moves component library up one level
* rename `front matter` to `frontmatter`
* add descriptions to frontmatter

----
Screenshot 2024-03-19 at 8 34 31 PM


dsinghvi:
Not sure if you've seen this https://mintlify.com/docs/content/components/code but i think the page layout here is easier to follow because of the navigation, then schema, then example. (ik fern docs doesn't support this layout yet but wanted to call out)

dsinghvi:
Thoughts on just copying Nextra docs here and not using accordions https://nextra.site/docs/guide/syntax-highlighting?

PR #3226, opened 2024-03-20
(chore): rename `syntax highlighting` to `code snippets`

dsinghvi:
This PR also starts to create an `API References` section that will explain to users how to configure API References in Fern's Docs.

### DIFF
--- a/fern/docs.yml
+++ b/fern/docs.yml
@@ -14,11 +14,11 @@ tabs:
   overview:
     display-name: Overview
     icon: "globe"
-  fern-sdks:
+  sdks:
     display-name: SDKs
     slug: sdks
     icon: code
-  fern-docs:
+  docs:
     display-name: Docs
     slug: docs
     icon: book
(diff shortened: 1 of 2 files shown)


fern/docs.yml
```diff
@@ -201,20 +201,20 @@ navigation:
...
         slug: quickstarts
         contents:
-          - page: Fern Docs Quickstart with OpenAPI
+          - page: Use OpenAPI
             slug: openapi
             path: ./docs/pages/fern-docs/docs-quickstarts/docs-quickstart-openapi.mdx
-          - page: Fern Docs Quickstart with Fern Def
+          - page: Use a Fern Definition
             slug: ferndef
             path: ./docs/pages/fern-docs/docs-quickstarts/docs-quickstart-ferndef.mdx
```
dsinghvi, line 216: needs to eventually be a single quickstart that leverages two cards and a stepper (prisma does this) https://www.prisma.io/docs/getting-started/quickstart

PR #3233, opened 2024-03-20
(docs): update links due to recent docs changes

minaelee:
I left README.md alone for now assuming that should be changed after new docs site is live.

### DIFF
--- a/fern/docs.yml
+++ b/fern/docs.yml
@@ -20,6 +20,7 @@ tabs:
     icon: code
   docs:
     display-name: Docs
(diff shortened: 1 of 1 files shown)


fern/docs.yml
```diff
@@ -20,7 +20,6 @@ tabs:
     icon: code
   docs:
     display-name: Docs
-    slug: docs
```
dsinghvi, line 23: lets just keep the explicit slug so that it doesn't get changed

PR #3235, opened 2024-03-21
FER-1215: Consolidate GeneratorNotificationService implementations

omarrida:


packages/generators/commons/src/GeneratorNotificationService.ts
```diff
@@ -6,33 +6,50 @@ export interface GeneratorNotificationService {
 }
 
 export class GeneratorNotificationServiceImpl implements GeneratorNotificationService {
```
dsinghvi, line 8: Let's actually get rid of the interface/impl pattern here -- one rule we want to start imposing is that an `interface` is only relevant if there are more than 1 impls (in this case there aren't)

packages/generators/commons/src/GeneratorNotificationService.ts
```diff
@@ -6,33 +6,50 @@ export interface GeneratorNotificationService {
...
     public async sendUpdate(update: FernGeneratorExec.GeneratorUpdate): Promise<void> {
+        if (!this.client) {
+            return;
+        }
+
         this.buffer.push(update);
         await this.flush();
     }
 
     private async flush(): Promise<void> {
```
dsinghvi, line 48: should we make flush a public method too?

yarn.lock
```diff
@@ -3064,6 +3064,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@fern-api/openapi-generator@workspace:generators/openapi"
   dependencies:
+    "@fern-api/generator-commons": "workspace:*"
     "@fern-fern/generator-exec-sdk": 0.0.675
```
dsinghvi, line 3068: i think a successful PR here means that `"@fern-fern/generator-exec-sdk"` is not used in any of the generators.

dsinghvi: what would it take to get to that?

generators/csharp/cli/src/AbstractGeneratorCli.ts
```diff
@@ -50,12 +49,17 @@ export abstract class AbstractGeneratorCli<CustomConfig> {
...
 
                 // kick off log, but don't wait for it
-                void generatorNotificationService?.sendUpdateAndSwallowError(
-                    FernGeneratorExec.GeneratorUpdate.log({
-                        message: message.join(" "),
-                        level: LOG_LEVEL_CONVERSIONS[level]
-                    })
-                );
+                try {
+                    void generatorNotificationService?.sendUpdate(
```
dsinghvi, line 53: nit: do we still need the `?` here ? (guessing no)

dsinghvi (commented):
Overall looks great -- can we update the commons library to also export something that reads and parses the generator config? Would love to share that logic. 

Ideally no generator should have to depend on `@fern-fern/generator-exec-sdk` and they just bring in the utils package

dsinghvi (approved):

PR #3247, opened 2024-03-22
fix depth check to prevent max call stack exceeded issue

omarrida:


test-definitions/fern/apis/circular-references/definition/ast.yml
```diff
@@ -14,3 +9,14 @@ types:
...
+  FieldName: string
+  FieldValue:
+    union:
+      primitive_value: PrimitiveValue
+      object_value: ObjectValue
+      container_value: ContainerValue
+  ObjectFieldValue:
+    properties:
+      name: FieldName
+      value: FieldValue
```
dsinghvi, line 21: is it worth capturing the edge case this was testing?

armandobelardo (approved):
Approving given seed passes locally

dsinghvi (approved):

PR #3248, opened 2024-03-22
fix: update python defaults to be the user provided number and not th…

armandobelardo:
…e global default


dsinghvi (approved):
The only thing to fix is not applying the timeout when a custom client is provided

PR #3259, opened 2024-03-24
(fix): support schema references in OpenAPI that aren't just Schema Ids

omarrida:


packages/cli/openapi-parser/src/openapi/v3/AbstractOpenAPIV3ParserContext.ts
```diff
@@ -46,38 +46,86 @@ export abstract class AbstractOpenAPIV3ParserContext implements SchemaParserCont
     }
 
     public resolveSchemaReference(schema: OpenAPIV3.ReferenceObject): OpenAPIV3.SchemaObject {
+        const startsWithSchemaReferencePrefix = schema.$ref.startsWith(SCHEMA_REFERENCE_PREFIX);
+        const endsWithSchemaSuffix = schema.$ref.endsWith("/schema");
```
dsinghvi, line 50: If we think about these cases, they aren't actually that different -- 

In one case you point to `#/components/schemas/MySchema` and in another case you might point to `#/components/schemas/MySchema/application~json/200/schema`

dsinghvi: Feels like we should just update the code to take the reference, convert it into an array of keys `string[]` and then index into the OpenAPI spec accordingly. At any point if we hit undefined, we should throw an error.

packages/cli/openapi-parser/src/openapi/v3/AbstractOpenAPIV3ParserContext.ts
```diff
@@ -46,38 +46,86 @@ export abstract class AbstractOpenAPIV3ParserContext implements SchemaParserCont
...
+                throw new Error(`${splitSchemaKey[0]} is undefined`);
+            }
+            if (isReferenceObject(resolvedSchema)) {
+                resolvedSchema = this.resolveSchemaReference(resolvedSchema);
+            }
+
+            if (splitSchemaKey[1] === "properties" && splitSchemaKey[2] != null) {
+                const resolvedProperty = resolvedSchema.properties?.[splitSchemaKey[2]];
+                if (resolvedProperty == null) {
+                    throw new Error(`${schema.$ref} is undefiened. Property does not exist on object.`);
```
dsinghvi, line 80: nit: typo from before your commit

packages/cli/openapi-parser/src/openapi/v3/AbstractOpenAPIV3ParserContext.ts
```diff
@@ -46,38 +46,86 @@ export abstract class AbstractOpenAPIV3ParserContext implements SchemaParserCont
...
+                } else if (isReferenceObject(resolvedProperty)) {
+                    resolvedSchema = this.resolveSchemaReference(resolvedProperty);
+                } else {
+                    resolvedSchema = resolvedProperty;
+                }
+            }
+
+            const schemaId = schema.$ref;
```
dsinghvi, line 123: The names assigned to schemas directly have an impact on the quality of the fern-generated SDK. We should be smarter about how we generate names: 

So if the ref points to a schema (`#/components/schemas/MyUser`) we should use the `MyUser`. If the ref points to a response (`#/components/responses/MyResponse/application~json/200/scheam` we should use `MyResponse`. If the ref points to a nested property (`#/components/schemas/MyUser/properties/ssn` we should make it `MyUserSsn`. 

I think it may be worthwhile to create a static helper function called `getSchemaIdFromRef` and unit test that as well.

dsinghvi (changes requested):
Overall looking great! Couple comments to adjust the code

packages/cli/openapi-parser/src/openapi/v3/AbstractOpenAPIV3ParserContext.ts
omarrida: Updated in second attempt

packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -617,6 +618,10 @@ export function convertSchemaObject(
 }
 
 export function getSchemaIdFromReference(ref: OpenAPIV3.ReferenceObject): string {
+    if (ref.$ref.startsWith(SCHEMA_INLINE_REFERENCE_PREFIX)) {
+        return ref.$ref.replace(SCHEMA_INLINE_REFERENCE_PREFIX, "").split("/")[0] ?? "unknown";
```
omarrida, line 622: Not sure about falling back on `unknown` or if all the inline referenced schemas will actually conform to starting with `#/components/responses/`.

test-definitions-openapi/fern/apis/code-samples-open-api/.definition/api.yml
```diff
@@ -0,0 +1,8 @@
+name: api
+error-discrimination:
```
dsinghvi, line 2: @omarrida can you make sure to not commit the `.definition` folders? maybe you can just add to `.gitignore`

packages/cli/openapi-parser/src/schema/convertSchemas.ts
dsinghvi: yeah lets scope the `unknown` change to another PR, for now we should just throw

packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -616,23 +630,59 @@ export function convertSchemaObject(
...
-        throw new Error(`Cannot get schema id from reference: ${ref.$ref}`);
+        return undefined;
     }
     return ref.$ref.replace(SCHEMA_REFERENCE_PREFIX, "");
 }
 
+export function getBreadcrumbsFromReference(ref: OpenAPIV3.ReferenceObject): string[] {
+    let breadcrumbs: string[] = ref.$ref
+        .replace(SCHEMA_REFERENCE_PREFIX, "")
+        .replace(SCHEMA_INLINE_REFERENCE_PREFIX, "")
```
dsinghvi, line 643: what about `#/components/schemas/requests` ?

packages/cli/openapi-parser/src/schema/utils/getBreadcrumbsFromReference.ts
```diff
@@ -0,0 +1,205 @@
...
+    isApplicable: (refUnderProcessing: string[]) => refUnderProcessing[0] === "oneOf",
+    applyStrategy: (refUnderProcessing: string[]) => {
+        return {
+            refUnderProcessing: refUnderProcessing.slice(1),
+            newBreadcrumbs: []
+        };
+    }
+};
+
+// export const ENDS_WITH_SCHEMA: Strategy = {
```
omarrida, line 115: TODO: Clean up commented code

packages/cli/openapi-parser/src/schema/utils/getBreadcrumbsFromReference.ts
```diff
@@ -0,0 +1,205 @@
...
+//         }
+//         return breadcrumbs;
+//     }
+// };
+
+export function getBreadcrumbsFromReference(reference: string): string[] {
+    let refUnderProcessing = reference.split("/").map((breadcrumb) => breadcrumb.replace(/~1/g, "/"));
+    let breadcrumbs: string[] = [];
+
+    // create map of key value pairs where the key is a function that takes in the breadcrumbs array,
```
omarrida, line 169: Clean up comments

dsinghvi (approved):
🔥 the only thing i would do is add some clear descriptions in each of your rules in `getBreadcrumbsFromReference`

dsinghvi:
fyi @abvthecity

PR #3262, opened 2024-03-25
improvement, python: add __version__ variable

armandobelardo:


seed/python-sdk/alias/src/seed/version.py
```diff
@@ -0,0 +1,4 @@
+
+from importlib import metadata
+
+__version__ = metadata.version("fern_alias")
```
dsinghvi, line 4: will this normally keep the version string?

dsinghvi (approved):

seed/python-sdk/alias/src/seed/version.py
armandobelardo: Yeah, this is meant to evaluate to the version within the toml file

PR #3272, opened 2024-03-26
(break): redo IA for docs

dsinghvi:
This PR starts to propose a new layout for the docs. Note that there are some feature requests that we'll need to prioritize in order to make the docs look visually appealing, but this takes a stab at putting the right content in the right hierarchy. 

A preview URL is [here](https://fern-preview-2cf52780-0075-4624-9b3f-f6d4b679accb.docs.buildwithfern.com/home/introduction)

### Goals
1. **Professsional**: Our docs look professional and reflect our engineering quality
2. **Self-serve docs**: We are increasingly seeing users that want to get started using the docs product on their own 
3. **Credit for SDK Features**:  Our SDKs do a lot but most folks only understand the features once they schedule a call. 

### Features Required 
- Horizontal Tab Layout: https://docs.stripe.com/. The horizontal tabs will be `Home`,  `SDKs`, `Docs`, `Configuration`, and `Changelog`. 
- Section toggles are towards the end of the nav: https://nextra.site/docs/docs-theme. This will allow sections that contain single pages and nested sections will be aligned which will feel more natural. 
- Icons next to pages: https://linear.app/docs. This will make our docs jump out more  + allow us to signal paid from free in the SDK features section

PR #3278, opened 2024-03-26
chore(docs): alphabetize docs components in the navigation sidebar

abvthecity:


dsinghvi:
@abvthecity would you be down to make another PR so that these changes get captured here https://github.com/fern-api/fern/pull/3272

PR #3281, opened 2024-03-26
(fix): add option to preserve property casing while keeping serde layer on

omarrida:


dsinghvi:
hey @omarrida did you test this works? also we need to make sure it works across unions, primitives, headers, path params, basically everything

PR #3282, opened 2024-03-26
fix: remove form-data dep

mscolnick:
- [ ] need to run across seeds


dsinghvi:
@mscolnick we'll need to modify the endpoint generator to not use `FormData` right?

PR #3283, opened 2024-03-26
improvement: default literal values for unions

armandobelardo:


dsinghvi (approved):
@armandobelardo do you mind updating the snippet generator as well -- should be a simple change [here](https://github.com/fern-api/fern/blob/e7ccaa4c0161851c16c896869b000cf03550ba73/generators/python/src/fern_python/generators/pydantic_model/type_declaration_handler/discriminated_union/simple_discriminated_union_generator.py#L231-L236)

PR #3288, opened 2024-03-28
commit ir json to seed folders

mscolnick:
Helps with unit testing without needing to run Docker


dsinghvi:
@mscolnick should we `.gitignore` the IRs? Can we also commit the `config.json` as well?

PR #3303, opened 2024-04-01
improvement: allow header auth extension to specify auth prefix

armandobelardo:
Also fixes: FER-1321


dsinghvi:
up to add a snapshot test, maybe to `x-fern-headers` ? And then can we document this here https://buildwithfern.com/learn/overview/define-your-api/openapi/extensions#authentication-overrides

dsinghvi (approved):
approving so you can merge in once those two things are added

PR #3304, opened 2024-04-01
(feature): Implement pagination

amckinney:
Re: https://github.com/fern-api/fern/pull/2985

This implements the pagination validation rule (run w/ `fern check`), and updates the `fern` CLI to include `pagination` in the generated IR. Note that this change does _not_ include the implementation required to map the `x-fern-pagination` OpenAPI extension to the Fern IR.

### DIFF
--- a/seed/csharp-model/mixed-case/.inputs/ir.json
+++ b/seed/csharp-model/mixed-case/.inputs/ir.json
@@ -2817,10 +2817,9 @@
                         }
                     },
                     "displayName": null,
-                    "auth": false,
-                    "idempotent": false,
-                    "baseUrl": null,
                     "method": "GET",
+                    "headers": [],
+                    "baseUrl": null,
                     "path": {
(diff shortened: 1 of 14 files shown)


dsinghvi:
@amckinney think we'll need to run seed here

dsinghvi (approved):
Discussed Offline -- 

1. Refactor into validation for different pagination schemes (e.g. `cursor-pagination.ts`, `link-pagination.ts`, `offset-pagination.ts`)
2. Potentially a good use for a `PropertyResolver` or `PropertyValidator` class that handles doing cross file property resolution and then running a validate function. The validate function has some context where it gets the resolved property type and the consumer can validate based on the property type.

amckinney:
Split everything out into separate files so that it's a lot easier to follow. I will explore the `PropertyValidator` refactor when we introduce another pagination type (e.g. link pagination).

PR #3310, opened 2024-04-01
(feature): Add retainOriginalCasing option to TypeScript generators

amckinney:
This adds a new `retainOriginalCasing` generator option so that users can preserve the casing convention expressed in their Fern definition rather than using the camelCase-equivalent (e.g. `page_limit` vs `pageLimit`).


dsinghvi (commented):
@amckinney have you thought about cases where we retain original casing but it is an unsafe word? 

I think the only way to handle that would eventually be an IR change where the ir passes in a list of unsafe words that you check against?

PR #3311, opened 2024-04-02
feat: add flag to disable Pydantic validation and keep extra fields on the Pydantic model

armandobelardo:
Will actually add in a wrapper to the convert method to deal with containers and primitives, etc


dsinghvi:
Unit tests that are required: 
- list
- map
- union (discriminated)
- undiscriminated
- object (contains required and optional properties)
- object with extends
- primitive values (uuid/int/float/string/literals)

dsinghvi (approved):

PR #3314, opened 2024-04-02
(fix): send auth prefix to docs

dsinghvi:
Auth prefix should be sent to the docs.


packages/cli/configuration/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@fern-api/core-utils": "workspace:*",
-    "@fern-api/fdr-sdk": "0.71.0",
+    "@fern-api/fdr-sdk": "^0.71.0-37390eacb",
```
dsinghvi, line 32: `0.72.0-547602b7d`

armandobelardo (approved):

PR #3318, opened 2024-04-03
fix: force pydantic.v1 only if pydantic v2, this is needed due to a p…

armandobelardo:
…atch version break from pydantic


dsinghvi (approved):

dsinghvi:
- Changelog
- Version

dsinghvi:
Screenshot 2024-04-03 at 3 01 18 PM
I don't think we're handling the case where someone sets the pydantic version to `v2`

armandobelardo:
> Screenshot 2024-04-03 at 3 01 18 PM I don't think we're handling the case where someone sets the pydantic version to `v2`

Ah this is pydantic seed which I hadn't run yet! Let me push that up

PR #3323, opened 2024-04-04
(feature): Generate snippets locally

amckinney:
This adds the ability to generate snippets locally with the following `snippets` configuration:

```yaml
groups:
  go-sdk:
    generators:
      - name: fernapi/fern-go-sdk
        version: 0.18.2
        snippets:
          path: ../generated/snippet.json
        output:
          location: local-file-system
          path: ../generated/acme-go
```

This is compatible with local `local-file-system` generation for customers that are authorized to generate snippets. Any customer that has access to `github` generation will also have access to do so with the following:

```yaml
groups:
  go-sdk:
    generators:
      - name: fernapi/fern-go-sdk
        version: 0.18.2
        snippets:
          path: ../generated/snippet.json
        github:
          repository: acme/acme-go
```

Note that snippets are supported by Fern docs by default and enabled with the `docs.yml` configuration shown below:

```yaml
navigation:
  - api: API Reference
    snippets: 
      typescript: acme
      python: acme
      go: https://github.com/acme/acme-go
```


dsinghvi:
>Note that snippets are generated for Fern's documentation by default - this simply allows users to retrieve the snippets locally so that they can use them for a variety of other use cases. 

what does this mean?

amckinney:
> what does this mean?

Terribly worded 🤦‍♂️ - I actually meant to say that snippets are supported by Fern docs by default with the `docs.yml` configuration (this new `snippets` configuration is not required). I'll clean this up and include an example of the `docs.yml` config so it's more clear.

PR #3332, opened 2024-04-05
fix: allow unchecked models to just pass through any

armandobelardo:
This also adds a check to make sure we're not passing in a non-class to our issubclass checks and barfing

Confirmed working at Cohere after this change: https://github.com/cohere-ai/cohere-python/pull/462


dsinghvi (approved):

generators/python/tests/utils/test_construct_type.py
```diff
@@ -73,7 +76,8 @@ def test_construct_unset() -> None:
         "map": {"1": "string"},
         "union": {"type": "square", "id": "string", "length": 1.1},
         "undiscriminated_union": {"id": "string2", "length": 6.7},
-        "enum": "red"
+        "enum": "red",
+        "any": "something here"
```
dsinghvi, line 80: is this to check additional props?

dsinghvi: nvm i see you added to the fern def thanks

armandobelardo: Yeah! Just adding an any type

PR #3368, opened 2024-04-12
(feature): Add allowExtraFields configuration to TypeScript generators

amckinney:
This adds the `allowExtraFields` configuration option to the TypeScript generators. This permits the user to include extra fields in the request and/or response wire representation rather than being stripped.

When enabled, this makes _every_ request/response allow extra fields. We will later follow-up with a more granular option that can be specified on individual types in the OpenAPI / Fern definitions directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 ![License](https://img.shields.io/badge/License-MIT-blue)
 
 [![Discord](https://img.shields.io/badge/Join%20Our%20Community-black?logo=discord)](https://discord.com/invite/JkkXumPzcG)
-[![Documentation](https://img.shields.io/badge/Read%20our%20Documentation-black?logo=book)](https://docs.buildwithfern.com?utm_source=fern-api/fern/readme-read-our-documentation)
+[![Documentation](https://img.shields.io/badge/Read%20our%20Documentation-black?logo=book)](https://buildwithfern.com/learn/home?utm_source=fern-api/fern/readme-read-our-documentation)
 
 </div>
 
--- a/fern/docs.yml
+++ b/fern/docs.yml
@@ -58,6 +58,9 @@ navigation:
             path: ./pages/sdks/features/multipart-form-data.mdx
           - page: Forward compatibility
             path: ./pages/sdks/features/forward-compatibility.mdx
+          - page: Registry publishing
+            path: ./pages/sdks/features/publish-to-registries.mdx
+            icon: pro
           - page: Auto-Pagination
             path: ./pages/sdks/features/auto-pagination.mdx
             icon: pro
--- a/fern/pages/sdks/features/publish-to-registries.mdx
+++ b/fern/pages/sdks/features/publish-to-registries.mdx
@@ -1,0 +1,133 @@
+---
+title: Automated publishing to registries
+excerpt: Fern will automatically publish your SDKs to registries like NPM, PyPI, and Maven
+---
+
+<Warning title='Pro Feature'>
+  This feature is only available on paid plans. Please schedule a [demo](https://buildwithfern.com/contact) 
+  or [email us](mailto:support@buildwithfern.com) to get started. 
+</Warning>
+
+Fern will automatically publish your SDKs to registries like NPM, PyPI, and Maven. This means 
+that you can easily distribute your SDKs to your users without having to worry 
+about the publishing process.
+
+<Steps>
+  ### Step 1: Configure your output location
+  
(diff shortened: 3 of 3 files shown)


generators/typescript/express/express-endpoint-type-schemas-generator/src/GeneratedExpressEndpointTypeSchemasImpl.ts
```diff
@@ -173,6 +195,17 @@ export class GeneratedExpressEndpointTypeSchemasImpl implements GeneratedExpress
...
                     throw new Error("No response schema was generated");
                 }
+                if (this.allowExtraFields) {
+                    return this.generatedResponseSchema
+                        .getReferenceToZurgSchema(context)
+                        .jsonOrThrow(referenceToParsedResponse, {
+                            unrecognizedObjectKeys: "passthrough",
+                            allowUnrecognizedEnumValues: true,
+                            allowUnrecognizedUnionMembers: true,
+                            skipValidation: true,
```
dsinghvi, line 205: should we be skipping validation here?

generators/typescript/express/express-error-schema-generator/src/GeneratedExpressErrorSchemaImpl.ts
```diff
@@ -59,23 +68,43 @@ export class GeneratedExpressErrorSchemaImpl
...
         switch (this.type.type) {
             case "named":
+                if (this.allowExtraFields) {
+                    return context.typeSchema
+                        .getSchemaOfNamedType(this.type, { isGeneratingSchema: false })
+                        .jsonOrThrow(referenceToBody, {
+                            unrecognizedObjectKeys: "passthrough",
+                            allowUnrecognizedEnumValues: true,
+                            allowUnrecognizedUnionMembers: true,
+                            skipValidation: true,
```
dsinghvi, line 78: same

generators/typescript/express/express-endpoint-type-schemas-generator/src/GeneratedExpressEndpointTypeSchemasImpl.ts
amckinney: No, good catch - updated with a `getSchemaOptions` function (as discussed).

generators/typescript/utils/commons/src/codegen-utils/getSchemaOptions.ts
```diff
@@ -0,0 +1,11 @@
+import { SchemaOptions } from "@fern-typescript/zurg";
+
+export function getSchemaOptions({ allowExtraFields }: { allowExtraFields: boolean }): Required<SchemaOptions> {
```
amckinney, line 3: @dsinghvi I noticed that we actually set `skipValidation` to `false` and `breadcrumbsPrefix` to `[]` in _every_ instance of this function, so I included those properties here too (simply by specifying `Required<SchemaOptions>`. No `Pick` was necessary after all, but we can easily adapt this later to restrict specifics keys later if need be.

For what it's worth, users can still easier override the spread with any overrides they need like we normally do in migrations, i.e.

```typescript
return this.generatedRequestSchema
    .getReferenceToZurgSchema(context)
    .jsonOrThrow(referenceToParsedRequest, {
        ...getSchemaOptions({
            allowExtraFields: this.allowExtraFields
        }),
        skipValidation: true,
    });
```

generators/typescript/express/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.0-rc0] - 2024-04-12
+
+- Feature: Add `allowExtraFields` option to permit extra fields in the returned response.
```
dsinghvi, line 10: can we add an example of consuming this in `generators.yml`?

generators/typescript/utils/commons/src/codegen-utils/getSchemaOptions.ts
dsinghvi: interesting do you know where we set `skipValidation` to true?

dsinghvi (approved):

generators/typescript/utils/commons/src/codegen-utils/getSchemaOptions.ts
dsinghvi: i guess we only do it on response fields which this doesn't change

PR #3370, opened 2024-04-13
feat: introduce snippets for Ruby SDKs

armandobelardo:
Fixes FER-994


dsinghvi (approved):
can we print the json multiline for `snippet.json` ?

PR #3373, opened 2024-04-14
feat: introduce oauth configs into IR, Fern def and OAS extensions

armandobelardo:
The general idea here is to allow the SDKs to generate helper methods for oauth. We're taking different paths for the different types of oauth:
- Client credentials
- Access code
    - Since this flow typically requires a user to get a short lived access code prior to requesting an access token and refresh token we'd limit this to static helpers:
        -  A static helper to generate the authorization URL, the URL a user visits to get the access code, this URL is just _constructed_ by us leveraging the details held within `authorization-endpoint`
        - A static helper to get the access token from the access code
        - A static helper to refresh a token and get a new access + refresh token
        - If you want to go the extra mile, you could make an object-oriented AccessToken object that takes in the access code with a .get and .refresh token that handle this for the user


packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
```diff
@@ -22,13 +23,83 @@ types:
...
+  OauthAuthorizationEndpointParameters:
+    properties:
+      name: commons.NameAndWireValue
+      type: types.TypeReference
+  OauthAuthorizationEndpoint:
+    properties:
+      path:
+        type: string
+        docs: The path to append to the base URL to get the authorization endpoint a user is meant to visit within the browser.
+      queryParameters: optional<list<OauthAuthorizationEndpointParameters>>
```
amckinney, line 82: Given that the property is named `queryParameters`, should we just use `http.QueryParameter` here instead of defining `OauthAuthorizationEndpointParameters`?

```
  QueryParameter:
    extends: commons.Declaration
    properties:
      name: commons.NameAndWireValue
      valueType: types.TypeReference
      allowMultiple: boolean
```

packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
```diff
@@ -22,13 +23,83 @@ types:
...
+  OAuthScheme:
+    extends: commons.WithDocs
+    docs: Note that for the oauth scheme we are currently assuming the resultant token is leveraged as a bearer token, e.g. "Authorization Bearer
+    properties:
+      configuration: OAuthConfiguration
+  OAuthConfiguration:
+    union:
+      authorizationCode: AuthorizationCodeOAuthScheme
+      clientCredentials: ClientCredentialsOAuthScheme
+  OauthAccessTokenFields:
```
amckinney, line 44: nit: Maybe best to standardize on either `OAuth*` or `Oauth`. I prefer `OAuth*`, but fine with either way.

PR #3374, opened 2024-04-15
improvement: pass snippets version to fdr to register docs with snippets at a specific version

armandobelardo:
Fixes FER-1377


dsinghvi (approved):

packages/cli/configuration/fern/definition/docs.yml
```diff
@@ -439,12 +439,23 @@ types:
...
 
+  VersionedSnippetLanguageConfiguration:
+    properties:
+      version: string
+      name: string
+
+  SnippetLanguageConfiguration:
+    discriminated: false
+    union:
+      - string
```
dsinghvi, line 450: nit: can we add some docs here that this is supposed to be the name?

packages/cli/configuration/fern/definition/docs.yml
```diff
@@ -439,12 +439,23 @@ types:
       link: string
       href: string
 
+  VersionedSnippetLanguageConfiguration:
+    properties:
+      version: string
+      name: string
```
dsinghvi, line 445: nit: should this be package?

armandobelardo: Yeah let me rename

PR #3386, opened 2024-04-16
(fix): discriminated union schema examples don't contain discriminants

dsinghvi:
When Fern's OpenAPI importer detects a discriminated union it removes the discriminant from the underlying type. For example, the following OpenAPI

```yaml
schemas: 
  Shape: 
    oneOf: 
      - $ref: #/components/schemas/Circle
      - $ref: #/components/schemas/Square
   Square: 
     properties: 
        type: 
          const: "square"
        side: 
          type: integer
```

is translated to  the following Fern Definition. 

```yaml
Shape: 
  discriminated: true
  union: 
    circle: Circle
    square: Square

Square: 
  properties: 
    a: integer
```

This means that any examples of `Square` should contain no discriminant but examples of Shape should.

### DIFF
--- a/packages/cli/openapi-parser/src/openapi/v3/generateIr.ts
+++ b/packages/cli/openapi-parser/src/openapi/v3/generateIr.ts
@@ -124,7 +124,19 @@ export function generateIr({
             })
             .filter((entry) => entry.length > 0)
     );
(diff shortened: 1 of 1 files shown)


dsinghvi:
fyi @abvthecity this was the issue with `type` example validation you were seeing

PR #3389, opened 2024-04-17
(feature): Go supports environment variable scanning

amckinney:
The Go generator now supports environment variable scanning. For example, consider the following `api.yml`:

  ```yaml
  name: api
  auth: Bearer
  auth-schemes:
    Bearer:
      scheme: bearer
      token:
        name: apiKey
        env: ACME_API_KEY
  ```

The client reads this environment variable and sets the value in the `Authorization` header if the `APIKey` option is not explicitly specified like so:

  ```go
  func NewClient(opts ...option.RequestOption) *Client {
    options := core.NewRequestOptions(opts...)
    if options.APIKey == "" {
      options.APIKey = os.Getenv("ACME_API_KEY")
    }
    return &Client{
      baseURL: options.BaseURL,
      caller: core.NewCaller(
        &core.CallerParams{
          Client:      options.HTTPClient,
          MaxAttempts: options.MaxAttempts,
        },
      ),
      // The header associated with the client will contain
      // the ACME_API_KEY value.
      //
      // It can still be overridden by endpoint-level request
      // options.
      header:  options.ToHeader(),
    }
  }
  ```

Note that this also adds a `basic-auth-environment-variables` test definition because this code path was previously unexercised.

### DIFF
--- a/generators/go/sdk/CHANGELOG.md
+++ b/generators/go/sdk/CHANGELOG.md
@@ -30,8 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ```go
   func NewClient(opts ...option.RequestOption) *Client {
     options := core.NewRequestOptions(opts...)
-    if options.ApiKey == "" {
-      options.ApiKey = os.Getenv("FERN_API_KEY")
+    if options.APIKey == "" {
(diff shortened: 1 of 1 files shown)


dsinghvi (approved):
Would be great to follow this up with some documentation on the auth features in https://buildwithfern.com/learn/sdks/introduction/overview

PR #3393, opened 2024-04-17
(fix): updating sdk invocation versions

bsinghvi:
Used the generator versions here - https://github.com/fern-api/fern?tab=readme-ov-file + the changelog to decide on the default version.

### DIFF
--- a/packages/cli/configuration/src/generators-yml/generatorInvocations.ts
+++ b/packages/cli/configuration/src/generators-yml/generatorInvocations.ts
@@ -16,10 +16,10 @@ export const GENERATOR_INVOCATIONS: Record<GeneratorName, Omit<GeneratorInvocati
         version: "0.5.7"
     },
     [GeneratorName.JAVA_SDK]: {
-        version: "0.8.7"
+        version: "0.9.3"
     },
     [GeneratorName.JAVA_SPRING]: {
-        version: "0.5.7",
+        version: "0.8.0-rc0",
         output: {
             location: "local-file-system",
             path: "../src/main/java/com/fern"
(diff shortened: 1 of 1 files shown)


dsinghvi:
@bsinghvi you'll need to fix ete tests here `yarn workspace @fern-api/ete-tests test -u` and then we are g2g

dannysheridan:
is there a way to make this dynamic (inspired by what Armando did here: https://github.com/fern-api/fern/blob/8633c289e09088ad865303c8b74005dea260dda9/README.md?plain=1#L79)

bsinghvi:
@dannysheridan there would still have to be some level of manual intervention to determine which version we should mark as safe to be the default invocation version. right now, Armando's just grabs the latest, which could potentially not be release safe

dsinghvi:
> is there a way to make this dynamic? I'm inspired by what Armando did here:
> 
> https://github.com/fern-api/fern/blob/8633c289e09088ad865303c8b74005dea260dda9/README.md?plain=1#L79
> 
> )



> @dannysheridan there would still have to be some level of manual intervention to determine which version we should mark as safe to be the default invocation version. right now, Armando's just grabs the latest, which could potentially not be release safe

That's right -- in fact the latest TS release has some issues so have to wait to merge this

bsinghvi:
@dsinghvi this PR doesn't update it to the latest for the TypeScript SDK - are there concerns with 12.9?

dannysheridan:
image
getting an error when I try to run: `yarn workspace @fern-api/ete-tests test -u`

@dsinghvi can you land this PR?

PR #3399, opened 2024-04-18
(fix, typescript): do file upload snippet generation

dsinghvi:
In this PR, we update the TypeScript generator to support snippet generation for file upload.


generators/typescript/sdk/generator/src/__test__/__snapshots__/multi-url-environment.txt
```diff
@@ -435,8 +435,8 @@ export * from "./requests";
  * This file was auto-generated by Fern from our API Definition.
  */
 
-import * as serializers from "../../../..";
-import * as TestSdk from "../../../../../api";
+import * as serializers from "../../../../index";
```
dsinghvi, line 438: for deno

PR #3405, opened 2024-04-19
(improvement): appending type for type exports

bsinghvi:
Adding type explicitly into the exports when generating type declarations.

### DIFF
--- a/generators/typescript/sdk/CHANGELOG.md
+++ b/generators/typescript/sdk/CHANGELOG.md
@@ -7,8 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.15.1-rc1] - 2024-04-24
 
-- Improvement: In an effort to make the generated code JSR compatible, the TS generator
-  will append the `type` explicitly for request exports.
+- Improvement: Earlier for inlined request exports, we were doing the following:
+
+```ts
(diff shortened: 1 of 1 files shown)


generators/typescript/sdk/generator/src/SdkGenerator.ts
```diff
@@ -409,7 +409,8 @@ export class SdkGenerator {
                 run: ({ sourceFile, importsManager }) => {
                     const context = this.generateSdkContext({ sourceFile, importsManager });
                     context.type.getGeneratedType(typeDeclaration.name).writeToFile(context);
-                }
+                },
+                addExportTypeModifier: true
```
dsinghvi, line 413: can we add this to request wrappers instead of types declarations?

dsinghvi: https://github.com/hookdeck/hookdeck-typescript-sdk/blob/ecda54dde80df5286f460d5f15833ae17f299d99/src/api/resources/bookmark/client/requests/index.ts#L1-L4 is an example of the first problem to fix

dsinghvi: all request wrappers are interfaces so we know that this is safe

dsinghvi (approved):
LFG

PR #3410, opened 2024-04-19
(feature): Add OAuth to IR

amckinney:
This builds upon the changes introduced in https://github.com/fern-api/fern/pull/3373, but refactors the IR to match the properties specified in the YAML schema. For now, we only support the `client_credentials` OAuth flow.

Other changes include the following:
* Refactors the pagination IR using the `PropertyResolver` abstraction that we introduce here. 
  * Both OAuth and pagination now take the same approach with respect to resolving request/response properties.
* A small issue was also fixed with the pagination IR's `propertyPath`, such that it omits the final property element (for ease of use).

There are still some improvements we can make (most notably in the `convertProperty.ts` utility file), which will be handled in a follow-up.


dsinghvi (approved):
+1ing

packages/ir-sdk/fern/apis/ir-types-v38/changelog/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [v38] - 2024-04-17
+
+- Feature: Add support for Server-Sent-Events to Streaming HTTP Responses
+  Read more about SSE here: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events.
```
dsinghvi, line 11: @amckinney would you be up to bring this back up and add ir v39 here?

dsinghvi: eventual plan is to add this to the docs

amckinney: I added a new `CHANGELOG.md` under `ir-types-latest`, I just mistakenly copied this over to `ir-types-v38` instead of merging them together 🤦 I'll update here!

PR #3411, opened 2024-04-19
(feat, ts): support server-sent events

dsinghvi:
In this PR, we update the `Stream` interface in Fern to support Server-Sent-Events and JSON streams. The code to process both is the same and relies on an internal data model of `prefix` and `messageTerminator`.

### DIFF
--- a/seed/ts-sdk/streaming/allow-custom-fetcher/.inputs/ir.json
+++ b/seed/ts-sdk/streaming/allow-custom-fetcher/.inputs/ir.json
@@ -386,9 +386,9 @@
                     },
                     "response": {
                         "type": "streaming",
-                        "dataEventType": {
+                        "value": {
                             "type": "json",
-                            "json": {
+                            "payload": {
                                 "_type": "named",
                                 "name": {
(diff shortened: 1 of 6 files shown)


generators/typescript/utils/core-utilities/fetcher/src/stream/Stream.ts
```diff
@@ -0,0 +1,134 @@
...
+        messageTerminator: string;
+    }
+
+    interface SseEvent {
+        type: "sse";
+        streamTerminator?: string;
+    }
+}
+
+const DATA_PREFIX = "data:";
```
amckinney, line 27: nit: Thoughts on `SSE_DATA_PREFIX`? At first I was concerned this was accidentally being used for the original JSON streaming approach.

amckinney (approved):

generators/typescript/utils/core-utilities/fetcher/src/stream/Stream.ts
dsinghvi: sounds great

PR #3413, opened 2024-04-20
(feat, docs): create a api definition tab before sdks and docs

dsinghvi:
In this PR we: 
- Introduce a tab called `API Definition`. The biggest change here is that now there is a page that explains what an api definition is, which ones Fern integrates with and why creating an API definition is important. 
(The accordiion component was key)
- Delete the Fern CLI documentation (Redo coming soon)

Screenshot 2024-04-20 at 6 45 20 PM


dsinghvi:
The next step here will be to add documentation about how to get to an API definition (code first generation, manually writing it) as well as a quality guide

PR #3430, opened 2024-04-22
fix, python: in best-case formatting add missing tab

armandobelardo:


dsinghvi (approved):

armandobelardo:
ended up folding this into another fix, closing out for now

PR #3440, opened 2024-04-23
(fix): OAuth is migrated back to bearer

amckinney:
This fixes the IRv39 -> IRv38 migration so that we treat the `oauth` auth scheme as bearer if the generator doesn't support IRv39 yet.


dsinghvi (approved):
@amckinney can we also update [here](https://github.com/fern-api/fern/blob/main/packages/cli/register/src/ir-to-fdr-converter/covertAuth.ts)

PR #3450, opened 2024-04-24
feat, ts: introduce snippet template creation

armandobelardo:


dsinghvi (approved):
@armandobelardo should we bump the VERSION ?

dsinghvi:
ts-sdk will be fixed with bhaskar's latest snapshot updates so good to merge whenever

PR #3454, opened 2024-04-24
(fix, python): Use kwargs for all httpx params

amckinney:
This fixes a few issues:
1. Sets `mypy` version to `1.9.0` in generated `pyproject.toml`.
2. Updates the `httpx.Client` request call to use named parameters for everything (i.e. `kwargs`).
3. Temporarily disables unit test generation.


dsinghvi (approved):
the seed tests look good to me

PR #3466, opened 2024-04-25
feat: support multiple custom domains

abvthecity:


packages/cli/configuration/fern/definition/docs.yml
```diff
@@ -39,6 +39,7 @@ types:
     properties:
       url: string
       "custom-domain": optional<string>
+      "custom-domains": optional<list<string>>
```
armandobelardo, line 42: should we just make custom-domain a list or a string (union) instead of adding a new key to the config?

PR #3472, opened 2024-04-26
(fix): backfill SSE events as streaming json

dsinghvi:


amckinney:
Should we update the [migration](https://github.com/fern-api/fern/blob/fc378f2884ae1ad5c3df4988db444f2f809c8de4/packages/cli/generation/ir-migrations/src/migrations/v38-to-v37/migrateFromV38ToV37.ts#L67), too?

dsinghvi:
> Should we update the [migration](https://github.com/fern-api/fern/blob/fc378f2884ae1ad5c3df4988db444f2f809c8de4/packages/cli/generation/ir-migrations/src/migrations/v38-to-v37/migrateFromV38ToV37.ts#L67), too?

I didn't do that b/c the actual generated SDKs would fail to parse SSE (but for docs its okay)

PR #3488, opened 2024-04-30
(fix, go): Add URL encoding to path parameters

amckinney:
This fixes an issue where particular path parameters would create invalid endpoint URLs (e.g. `\\example`). This also refactors the endpoint URL mapping logic with a new `core.EncodeURL` function so that the generated code reads a little better. See the `CHANGELOG.md` for more details.


dsinghvi (approved):

generators/go/internal/testdata/sdk/environments-core/fixtures/core/core.go
```diff
@@ -22,6 +23,16 @@ type HTTPClient interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
+// EncodeURL encodes the given arguments into the URL, escaping
+// values as needed.
+func EncodeURL(urlFormat string, args ...interface{}) string {
```
bsinghvi, line 28: Could this potentially be a helper function given it's re-used many places?

amckinney: It is! This is actually included in every generated SDK's `core` package, which is why we see it in all these different `seed` and `internal/testdata` snapshots.

bsinghvi: 🤦‍♂️totally missed that, my bad!

PR #3497, opened 2024-04-30
feat: tabs with href

abvthecity:
Example in docs.yml

```yaml
tabs:
  docs:
     display-name: Docs
  api:
     display-name: API Reference
  github:
    display-name: Github
    href: https://github.com/fern-api/fern

navigation:
  - tab: docs
     layout: []
  - tab: api
     layout: []
  - tab: github
```

^^ docs, api are normal tabs, whereas github links out.

Also in this PR:
- fern check rule that `tab.href != null XOR tab.layout != null`

### DIFF
--- a/fern/pages/fern-docs/config/navigation.mdx
+++ b/fern/pages/fern-docs/config/navigation.mdx
@@ -137,7 +137,7 @@ You can use a trackable link in your URL. For example, if using [UTM parameters]
 
 ## Tabs
 
-Within the navigation, you can add `tabs`. Tabs are used to group sections together. The example below shows tabs for `Help Center` and `API Reference`. Each tab has a `title` and `icon`. [Browse the icons available](/learn/docs/components/icons) from FontAwesome.
+Within the navigation, you can add `tabs`. Tabs are used to group sections together. The example below shows tabs for `Help Center`, `API Reference`, and an external link to `Github`. Each tab has a `title` and `icon`. [Browse the icons available](/learn/docs/components/icons) from FontAwesome.
 
 <CodeBlock title="docs.yml">
 ```yaml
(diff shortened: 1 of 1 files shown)


packages/cli/configuration/fern/definition/docs.yml
```diff
@@ -142,7 +148,9 @@ types:
   TabbedNavigationItem:
     properties:
       tab: TabId
-      layout: list<NavigationItem>
+      layout:
```
armandobelardo, line 151: [not blocking] for my SA, why is `layout` not on the tab object? It seems tied to a tab and then you could make it so that you can't have both by making a union

packages/cli/configuration/fern/definition/docs.yml
```diff
@@ -34,6 +34,12 @@ types:
       display-name: string
       icon: optional<string>
       slug: optional<string>
+      href:
+        type: optional<string>
+        docs: |
+          If `href` is set, clicking on the tab will redirect to the given URL.
+
+          Tabs with `href` must not have children in the navigation config.
```
dsinghvi, line 42: Are you up to add some docs here: https://buildwithfern.com/learn/docs/config/navigation ?

packages/cli/configuration/fern/definition/docs.yml
abvthecity: this is because tabs are specified at the top level. i'm not a huge fan of this personally but we can't change it now.

packages/cli/configuration/fern/definition/docs.yml
abvthecity: done

dsinghvi (approved):

PR #3514, opened 2024-05-02
feat: support multipart form data encoding

abvthecity:
* parse `encoding` field from openapi
* introduce `content-type` field in openapi-ir, fern yaml-schema, and fern-ir
* plumb through encoding.contentType -> fern-ir
---
remaining steps:
- plumb through in FDR and UI

### DIFF
--- a/packages/cli/yaml/yaml-schema/src/schemas/ObjectPropertySchema.ts
+++ b/packages/cli/yaml/yaml-schema/src/schemas/ObjectPropertySchema.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
-import { TypeReferenceDeclarationWithNameSchema } from "./TypeReferenceSchema";
+import { TypeReferenceDeclarationWithContentType } from "./TypeReferenceSchema";
 
-export const ObjectPropertySchema = TypeReferenceDeclarationWithNameSchema;
+export const ObjectPropertySchema = TypeReferenceDeclarationWithContentType;
 
 export type ObjectPropertySchema = z.infer<typeof ObjectPropertySchema>;
--- a/packages/cli/yaml/yaml-schema/src/schemas/TypeReferenceSchema.ts
+++ b/packages/cli/yaml/yaml-schema/src/schemas/TypeReferenceSchema.ts
@@ -34,3 +34,12 @@ export const TypeReferenceDeclarationWithEnvOverride = extendTypeReferenceSchema
     }).shape
 );
 export type TypeReferenceDeclarationWithEnvOverride = z.infer<typeof TypeReferenceDeclarationWithEnvOverride>;
+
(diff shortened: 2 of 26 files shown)


packages/cli/yaml/yaml-schema/src/schemas/ObjectPropertySchema.ts
```diff
@@ -1,6 +1,6 @@
 import { z } from "zod";
-import { TypeReferenceDeclarationWithNameSchema } from "./TypeReferenceSchema";
+import { TypeReferenceDeclarationWithContentType } from "./TypeReferenceSchema";
 
-export const ObjectPropertySchema = TypeReferenceDeclarationWithNameSchema;
```
dsinghvi, line 4: @abvthecity because we don't want to affect all object properties I would create a separate file called `InlinedObjectPropertySchema` and leverage the `TypeReferenceDeclarationWithContentType` type. 

Then `HttpInlineRequestBodySchema` can depend on the `InlinedObjectPropertySchema` instead.

packages/cli/yaml/yaml-schema/src/schemas/TypeReferenceSchema.ts
```diff
@@ -34,3 +34,12 @@ export const TypeReferenceDeclarationWithEnvOverride = extendTypeReferenceSchema
     }).shape
 );
 export type TypeReferenceDeclarationWithEnvOverride = z.infer<typeof TypeReferenceDeclarationWithEnvOverride>;
+
+export const TypeReferenceDeclarationWithContentType = extendTypeReferenceSchema(
+    DeclarationWithNameSchema.extend({
+        ["content-type"]: z.optional(z.string(), {
```
dsinghvi, line 40: @abvthecity Is this as simple as a string value that the generator needs to append?

abvthecity: not exactly.
the `content-type` here can be a list of acceptable content-types, i.e. `image/png, image/jpeg`.

the sdk should be able to infer the content-type from whatever file is being provided.

for string values (application/json), however, yes.

dsinghvi (changes requested):
Beyond the comment about introducing a `InlinedRequestBodyPropertySchema`




1) @abvthecity can you update the docs for the Fern Definition ([link](https://buildwithfern.com/learn/api-definition/fern/endpoints#multipart-form-uploads))
2) Can we add a test for the OpenAPI parsing to make sure it works. You can use the existing [fixture](https://github.com/fern-api/fern/blob/main/packages/cli/openapi-parser/src/__test__/file-upload.test.ts)
3) Ideally a validation rule that make sure `encoding` isn't used unless it is a MultipartFileUploadRequest (something like [this](https://github.com/fern-api/fern/tree/main/packages/cli/yaml/validator/src/rules/no-extensions-with-file-upload))

PR #3518, opened 2024-05-02
(feature): expose `x-fern-property-name` extension

dsinghvi:
This PR exposes an extension for `x-fern-property-name` so that OpenAPI users can control the name of the generated properties.

### DIFF
--- a/packages/cli/openapi-ir-to-fern/src/buildTypeDeclaration.ts
+++ b/packages/cli/openapi-ir-to-fern/src/buildTypeDeclaration.ts
@@ -107,21 +107,21 @@ export function buildObjectTypeDeclaration({
             context,
             fileContainingReference: declarationFile
         });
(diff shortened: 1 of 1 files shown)


packages/cli/openapi-ir-to-fern/src/buildTypeDeclaration.ts
```diff
@@ -107,7 +107,17 @@ export function buildObjectTypeDeclaration({
             context,
             fileContainingReference: declarationFile
         });
-        if (property.audiences.length > 0) {
+        if (property.audiences.length > 0 && property.nameOverride != null) {
```
amckinney, line 110: nit: We might be able to reduce nesting here if we resolve the property name and audiences outside of the condition, then set the value in the `properties` map from there.

Not a big deal either way though, so not blocking.

amckinney (approved):

PR #3519, opened 2024-05-03
(fix): remove `api.yml` not found error when the openapi folder is present

dsinghvi:
If the CLI can't find an OpenAPI spec and the `openapi` folder is present it used to return an error: 
```
Missing file: api.yml not found
```
Now the CLI sends a message
```
Missing file: `openapi.yml` not found
```


packages/cli/openapi-parser/src/__test__/__snapshots__/apiture.test.ts.snap
```diff
@@ -8446,6 +8446,7 @@ This error response may have one of the following \`type\` values:
           "conflict": {},
           "generatedName": "abstractPagedBodyStart",
           "key": "start",
+          "nameOverride": null,
```
dsinghvi, line 8449: my fault for not running theses tests earlier

PR #3521, opened 2024-05-03
(improvement, fern): Add better error for invalid generators.yml

amckinney:
This improves the `fern` compiler's error reporting when the user's `generators.yml` is invalid. For example, suppose you have the following configuration with duplicate map keys (i.e. `publish`):

```yaml
groups:
  publish:
    generators:
      - name: fernapi/fern-go-sdk
        version: 0.21.1
        github:
          repository: acme/acme-go
  publish:
    generators:
        version: 1.4.0
        github:
          repository: acme/acme-python
```

The CLI would previously fail with an unhelpful error message like the following:

```sh
$ fern generate
Missing file: api.yml
```

This was because we would ignore any errors thrown by the `generators.yml`, even if the file existed. This refactors the approach so that we actually act upon the errors thrown by the `generators.yml` parser and surface it to the user. Now, the user will see the following:

```sh
$ fern generate
Failed to parse generators.yml: duplicated mapping key
```


dsinghvi (approved):
Thank you! 🚀 If there are multiple generators.yml files should we be logging the filepath of the erroneous one?

amckinney:
Great call, updated!

PR #3535, opened 2024-05-06
feat, cli: add `fern generator upgrade` command

armandobelardo:


dsinghvi:
Would be great to add some end-to-end tests: https://github.com/fern-api/fern/blob/f965520030b6a1f99dda0b1caeeabc6fdef5462a/packages/cli/ete-tests/src/tests/add/add.test.ts#L10-L12 

I think it should be pretty easy with the `runFernCli` utility

packages/cli/configuration/src/generators-yml/addGenerator.ts
```diff
@@ -7,8 +8,70 @@ import {
...
+
+    if (image.length === 0) {
+        throw new Error(`No image found for ${generatorName}`);
+    } else if (image.length > 1) {
+        throw new Error(`Multiple images found at latest tag for ${generatorName}`);
+    }
+
+    // This assumes we have a label of the form version=x.y.z
+    // specifically adding a label to do this to be able to more easily get the version without regex
+    const generatorVersion = image[0]?.Labels["version"];
```
dsinghvi, line 36: assuming this means folks will get upgraded to `rc` versions?

packages/cli/cli/src/commands/upgrade/upgradeGenerator.ts
```diff
@@ -0,0 +1,93 @@
...
+        groupName: group,
+        context
+    });
+    await writeFile(
+        workspace.generatorsConfiguration?.absolutePathToConfiguration ??
+            generatorsYml.getPathToGeneratorsConfiguration({
+                absolutePathToWorkspace: workspace.absoluteFilepath
+            }),
+        yaml.dump(newConfiguration)
+    );
```
dsinghvi, line 36: not a big deal but worth calling out that i think this will delete comments https://github.com/vellum-ai/vellum-client-generator/blob/a63b5f8ee60c0d8c2e10ce063adeaecc048e7c93/fern/generators.yml#L10-L15

packages/cli/cli/src/cli.ts
```diff
@@ -644,41 +645,48 @@ function addUpgradeCommand({
...
-//                 cliContext,
-//                 generator: argv.generator,
-//                 targetVersion: argv.version
-//             });
-//         }
-//     );
-// }
+function addUpgradeGeneratorCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
+    cli.command(
+        "generator upgrade",
```
amckinney, line 650: nit: Not actionable here, but is this the best way to handle sub-commands with `yargs`? I'm curious what the user will see if they type `fern generator --help` in order to discover what sub-commands they have available to them. It's unclear from a quick search. Seems like it might be tucked away in [here](https://github.com/yargs/yargs/blob/main/docs/advanced.md).

Not at all important for the first iteration, so feel free to ignore.

packages/cli/cli/src/commands/upgrade/upgradeGenerator.ts
armandobelardo: ah nice ok will do a different way

packages/cli/configuration/src/generators-yml/addGenerator.ts
armandobelardo: chatted live, the CI file makes it so we only tag "latest" if it's not an RC
then this file pulls "latest", grabs the version off of it, and uses that to upgrade the generator

packages/cli/cli/src/cli.ts
```diff
@@ -644,41 +645,48 @@ function addUpgradeCommand({
...
+        `Upgrades the specified generator in ${GENERATORS_CONFIGURATION_FILENAME} to the latest version, or latest if one is not specified.`,
```
bsinghvi, line 651: This isn't clear imo - upgrading to latest version vs latest.

packages/cli/cli/src/commands/upgrade/upgradeGenerator.ts
```diff
@@ -0,0 +1,92 @@
...
+            await cliContext.runTaskForWorkspace(workspace, async (context) => {
+                const generatorsConfiguration =
+                    (await generatorsYml.loadRawGeneratorsConfiguration({
+                        absolutePathToWorkspace: workspace.absoluteFilepath,
+                        context
+                    })) ?? {};
+                if (generatorsConfiguration == null || generatorsConfiguration.groups == null) {
+                    return;
+                } else {
+                    if (generator != null && group != null) {
```
amckinney, line 61: nit: If we `return` here we can reduce nesting slightly.
```suggestion
                if (generatorsConfiguration == null || generatorsConfiguration.groups == null) {
                    return;
                }
                if (generator != null && group != null) {
                  ...
                }
```

packages/cli/configuration/src/generators-yml/addGenerator.ts
```diff
@@ -1,14 +1,68 @@
...
     GeneratorName,
     GeneratorsConfigurationSchema,
     updateGeneratorGroup
 } from ".";
 import { GENERATOR_INVOCATIONS } from "./generatorInvocations";
+import { upgradeGeneratorVersion } from "./upgradeGeneratorVersion";
 
-export function upgradeGenerator(): void {
-    // TODO
+function getOrThrowGeneratorName(generatorName: string, context: TaskContext): GeneratorName {
```
amckinney, line 12: nit: I think the convention I've seen elsewhere is `getGeneratorNameOrThrow`.

amckinney (approved):
This is awesome 👏

packages/cli/configuration/src/generators-yml/addGenerator.ts
armandobelardo: Whooooops, yeah this was not a great name choice on my part

packages/cli/cli/src/commands/upgrade/upgradeGenerator.ts
armandobelardo: updated!

dsinghvi:
A basic `fern add fern-openapi --group my-group` fails now and incorrectly updates the `generators.yml`. Also pretty tough to revert since there have been changes on top of this

PR #3536, opened 2024-05-06
(fix, openapi): Recursively visit nested anyOf schemas

amckinney:
This resolves an issue where the OpenAPI parser would fail to create a discriminated union for an `anyOf` that contained a nested `anyOf`. For example, consider the following OpenAPI schema:

```yaml
"Animal": {
  "anyOf": [
    {
      "$ref": "#/components/schemas/Dog"
    },
    {
      "$ref": "#/components/schemas/Reptile"
    },
  ]
},
...
"Reptile": {
  "anyOf": [
    {
      "$ref": "#/components/schemas/Snake"
    },
    {
      "$ref": "#/components/schemas/Iguana"
    },
  ]
},
...
```

if all of `Dog`, `Snake`, and `Iguana` all define a `type` field, the generated `Animal` type can be a discriminated union between all of them.

### DIFF
--- a/packages/cli/openapi-parser/src/schema/convertSchemas.ts
+++ b/packages/cli/openapi-parser/src/schema/convertSchemas.ts
@@ -897,12 +897,11 @@ function getPossibleDiscriminantsForSchemaObject({
     const possibleDiscrimimants: Record<string, string> = {};
     if (schema.anyOf != null) {
         for (const elem of schema.anyOf) {
-            for (const [key, value] of Object.entries(
-                getPossibleDiscriminants({
-                    schema: elem,
-                    context
-                }) ?? {}
-            )) {
+            const possibleDiscriminantsForVariant = getPossibleDiscriminants({
+                schema: elem,
+                context
+            });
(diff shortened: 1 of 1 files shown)


packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -877,11 +879,35 @@ function getDiscriminant({
...
     context: SchemaParserContext;
 }): Record<string, string> {
     const possibleDiscrimimants: Record<string, string> = {};
+    if (schema.anyOf != null) {
+        for (const elem of schema.anyOf) {
+            for (const [key, value] of Object.entries(
+                getPossibleDiscriminants({
+                    schema: elem,
+                    context
+                }) ?? {}
```
dsinghvi, line 904:
```ts
const possibleDiscriminantsForVariant = getPossibleDiscriminants({
                    schema: elem,
                    context
}) ?? {};
for (const [key, value] of possibleDiscriminantsForVariant) {
 ...
 }
```

would be easier to read

PR #3545, opened 2024-05-07
(fix) [wip] java empty response body instead of null

dcb6:


seed/java-sdk/audiences/src/main/java/com/seed/audiences/resources/foo/FooClient.java
```diff
@@ -70,7 +70,8 @@ public ImportingType find(FindRequest request, RequestOptions requestOptions) {
             }
             throw new ApiError(
                     response.code(),
-                    ObjectMappers.JSON_MAPPER.readValue(response.body().string(), Object.class));
+                    ObjectMappers.JSON_MAPPER.readValue(
+                            response.body() != null ? response.body().toString() : "{}", Object.class));
```
dsinghvi, line 74:
```java
Response response = client.newCall(okhttpRequest).execute();
String text = response.body().string();
if (response.isSuccessful()) {
  return ObjectMappers.JSON_MAPPER.readValue(text, ImportingType.class);
}
throw new ApiError(
   response.code(),
   text == null ? "" : text)
```

dsinghvi: @dcb6  what do you think about getting the text into a variable before hand -- this will set us up in a better place to add custom thrown errors to java imo and handle json decode errors as well. 

Stealing from [TypScript](https://github.com/fern-api/fern/blob/91deacfd43be76321340482b52774e2cfbd71301/generators/typescript/utils/core-utilities/fetcher/src/fetcher/Fetcher.ts#L137-L149)

dcb6: i think the NPE is actually on the `response.body().string()` call itself -- `response.body()` itself can bell null, intellij is even warning me about it:

CleanShot 2024-05-07 at 09 55 58

so i could make a variable out of response.body() but that's about it

(also noted that i erroneously changed it to `toString()`, will fix)

dan-thera: i see what you're pointing out though:
CleanShot 2024-05-07 at 12 41 04

how about this:
```
            Response response = client.newCall(okhttpRequest).execute();
            ResponseBody responseBody = response.body();
            String responseBodyString = responseBody != null ? responseBody.toString() : "{}";
            if (response.isSuccessful()) {
                return ObjectMappers.JSON_MAPPER.readValue(responseBodyString, ImportingType.class);
            }
            throw new ApiError(
                    response.code(),
                    ObjectMappers.JSON_MAPPER.readValue(responseBodyString, Object.class));
```

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/AbstractEndpointWriter.java
```diff
@@ -346,7 +349,7 @@ protected final MethodSpec getEnvironmentToUrlMethod() {
     }
 
     public final String getVariableName(String variable) {
-        if (this.endpointParameterNames.contains("body")) {
+        if (this.endpointParameterNames.contains(variable)) {
```
dan-thera, line 352: this was a small separate fix -- method was only being used with a `variable` of `body` so there were no bugs, but could led have led to future oens

dsinghvi (approved):
@dcb6 are you up to add a `CHANGELOG` and `VERSION` ?

dan-thera:
@dsinghvi oh yeah :) on it

generators/java/sdk/VERSION
```diff
@@ -1 +1 @@
-0.8.8-rc0
```
dan-thera, line 1: why would we have already released an rc?

PR #3553, opened 2024-05-07
fix: treat multipart form as form

abvthecity:
* adds a more robust way to parse and check media types (case-insensitive, and aware of parameters, i.e. charset=utf-8)
* updates `parseFileUploadRequest` to check if the media type is a multipart formdata rather than isFile=true

### DIFF
--- a/packages/cli/yaml/yaml-schema/src/utils/parseFileUploadRequest.ts
+++ b/packages/cli/yaml/yaml-schema/src/utils/parseFileUploadRequest.ts
@@ -1,5 +1,5 @@
 import { MediaType } from "@fern-api/core-utils";
-import { HttpRequestSchema, ObjectPropertySchema } from "../schemas";
+import { HttpInlineRequestBodySchema, HttpRequestSchema, ObjectPropertySchema } from "../schemas";
 import { isInlineRequestBody } from "./isInlineRequestBody";
 import { parseRawFileType } from "./parseRawFileType";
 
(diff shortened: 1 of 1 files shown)


packages/cli/yaml/yaml-schema/src/utils/parseFileUploadRequest.ts
```diff
@@ -35,7 +35,8 @@ export function parseFileUploadRequest(request: HttpRequestSchema | string): Raw
         request.body == null ||
         typeof request.body === "string" ||
         !isInlineRequestBody(request.body) ||
-        request.body.properties == null
+        request.body.properties == null ||
+        (request["content-type"] != null && request["content-type"] !== "multipart/form-data")
```
dsinghvi, line 39: where does this get used/how does this help?

dsinghvi (approved):

packages/cli/openapi-parser/src/openapi/v3/converters/endpoint/convertResponse.ts
```diff
@@ -154,30 +145,39 @@ function convertResolvedResponse({
...
+        if (mimeType == null) {
+            continue;
+        }
 
-    if (resolvedResponse.content?.[TEXT_PLAIN_CONTENT] != null) {
-        const textPlainSchema = resolvedResponse.content[TEXT_PLAIN_CONTENT]?.schema;
-        if (textPlainSchema == null) {
-            return ResponseWithExample.text({ description: resolvedResponse.description });
+        if (mimeType.isOctetStream()) {
+            return ResponseWithExample.file({ description: resolvedResponse.description });
```
dsinghvi, line 155: is this duplicative

dsinghvi:
@abvthecity looks like an ete test is failing, otherwise this looks great

PR #3561, opened 2024-05-07
(fix): handle `allow-mulitple` query parameter validation

dsinghvi:
### Goal
`allow-multiple` query parameters should support list examples. 

### Approach
Validating examples in Fern is pretty complex. So the way I approached this was to validate the query param as a list and as a non-list. If there are no errors in one of the two cases, then `fern check` passes. 

If there are errors, then we list them for the `non-list` case.


packages/cli/yaml/validator/src/rules/valid-example-endpoint-call/valid-example-endpoint-call.ts
```diff
@@ -73,24 +73,53 @@ export const ValidExampleEndpointCallRule: Rule = {
...
+                                let rawType = typeof queryParameter === "string" ? queryParameter : queryParameter.type;
+                                if (
+                                    schema === "array" &&
+                                    typeof queryParameter !== "string" &&
+                                    queryParameter["allow-multiple"] === true
+                                ) {
+                                    rawType = rawType.startsWith("optional")
+                                        ? `optional<list<${rawType}>>`
+                                        : `list<${rawType}>`;
+                                }
```
dsinghvi, line 100: tracking that this is weird

packages/cli/yaml/validator/src/rules/valid-example-endpoint-call/__test__/valid-example-endpoint-call.test.ts
```diff
@@ -481,4 +496,18 @@ describe("valid-example-endpoint-call", () => {
...
+            rule: ValidExampleEndpointCallRule,
+            absolutePathToWorkspace: join(
+                AbsoluteFilePath.of(__dirname),
+                RelativeFilePath.of("fixtures"),
+                RelativeFilePath.of("allow-multiple-query-param")
+            )
+        });
+
+        const expectedViolations: ValidationViolation[] = [];
+        expect(violations).toEqual(expectedViolations);
```
dsinghvi, line 511: the goal was to get this test case to pass which works

packages/cli/yaml/validator/src/rules/valid-example-endpoint-call/__test__/valid-example-endpoint-call.test.ts
```diff
@@ -177,7 +177,22 @@ describe("valid-example-endpoint-call", () => {
...
+                severity: "error"
+            },
+            {
+                message: 'Example is missing required query parameter "queryParam"',
+                nodePath: [
+                    "service",
+                    "endpoints",
+                    "get",
+                    {
+                        arrayIndex: 3,
```
dsinghvi, line 195: lines 180-195 are incorrect, ideally the diff should be 0 here

PR #3569, opened 2024-05-08
(internal, fix): rewrite inputs and run seed on ir changes

dsinghvi:
Whenever the IR changes, we need to make sure that ir inputs are regenerated in seed otherwise all of the seed tests will start failing.


amckinney (approved):

dsinghvi:
Will look at the `customConfig: null` error in a separate PR

PR #3572, opened 2024-05-08
fix, ir: fall back to the generated name when creating schemas if the…

armandobelardo:
… name override has conflicts

---

Right now we prefer title when creating objects, but the title can have conflicts since it's just a string as opposed to a map key like generated name

So this PR keeps the preference to nameOverride, but falls back to generatedName in the event nameOverride is already taken

### DIFF
--- a/packages/cli/openapi-parser/tsconfig.json
+++ b/packages/cli/openapi-parser/tsconfig.json
@@ -6,7 +6,6 @@
   "references": [
     { "path": "../../commons/core-utils" },
     { "path": "../../commons/fs-utils" },
-    { "path": "../configuration" },
     { "path": "../logger" },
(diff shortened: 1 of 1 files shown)


dsinghvi:
@armandobelardo this may create some non-determinism b/c its unclear if we have 2 schemas that claim the same `title` then its unclear who wins. Thoughts on going a different route where the user decides whether or not not we should use titles? 

```yaml
# generators.yml
api: 
  - path: ./openapi.yml
     overrides: ./openapi-overrides.yml
     settings: 
       useTitles: false
```

dsinghvi (approved):
looks great can we add a snapshot test for this in `openapi-ir-to-fern` ?

PR #3573, opened 2024-05-08
(improvement, yaml): Update default OAuth configuration

amckinney:
This builds upon the changes introduced in https://github.com/fern-api/fern/pull/3563, and adds the following:
* Update YAML configuration to expose `clientId`, `clientSecret` and `scopes` request properties.
* Add default configuration to simplify the expected use case.
* Refactor `fern check` to include more validation around the added configuration.

With this, the simplest OAuth configuration available to the user is the following:

```yaml
name: api
imports:
  auth: auth.yml

auth: OAuthScheme
auth-schemes:
  OAuthScheme:
    scheme: oauth
    type: client-credentials
    get-token:
      endpoint: auth.getToken
```

### DIFF
--- a/seed/csharp-model/oauth-client-credentials-default/src/SeedOauthClientCredentialsDefault/SeedOauthClientCredentialsDefault.csproj
+++ b/seed/csharp-model/oauth-client-credentials-default/src/SeedOauthClientCredentialsDefault/SeedOauthClientCredentialsDefault.csproj
@@ -7,6 +7,7 @@
 
     <ItemGroup>
       <PackageReference Include="OneOf" Version="3.0.263" />
+      <PackageReference Include="System.Text.Json" Version="8.0.3" />
     </ItemGroup>
 
 </Project>
--- a/seed/csharp-model/oauth-client-credentials-environment-variables/.mock/definition/api.yml
+++ b/seed/csharp-model/oauth-client-credentials-environment-variables/.mock/definition/api.yml
@@ -1,5 +1,5 @@
 name: oauth-client-credentials-environment-variables
-imports: 
+imports:
   auth: auth.yml
 
 auth: OAuthScheme
--- a/seed/csharp-model/oauth-client-credentials-environment-variables/.mock/definition/auth.yml
+++ b/seed/csharp-model/oauth-client-credentials-environment-variables/.mock/definition/auth.yml
@@ -14,7 +14,7 @@ service:
     getTokenWithClientCredentials:
       path: /token
       method: POST
-      request: 
+      request:
         name: GetTokenRequest
         body:
           properties:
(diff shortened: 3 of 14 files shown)


packages/cli/yaml/validator/src/rules/valid-oauth/__test__/valid-oauth.test.ts
```diff
@@ -71,4 +153,46 @@ describe("valid-oauth", () => {
...
+        const expectedViolations: ValidationViolation[] = [
+            {
+                message: `For details on OAuth configuration, see ${OAUTH_DOCS_LINK}`,
+                nodePath: ["service", "endpoints", "getToken"],
+                relativeFilepath: RelativeFilePath.of("auth.yml"),
+                severity: "warning"
+            },
+            {
+                message:
+                    "OAuth configuration for endpoint getToken specifies 'client-id' $request.client_id, which is not a valid 'client-id' type.",
```
dsinghvi, line 176: Better UX for the user who hasn't specified a value is probably something along the lines of `client_id property is missing in oauth config.`

packages/cli/yaml/validator/src/rules/valid-oauth/valid-oauth.ts
```diff
@@ -11,6 +11,8 @@ import { Rule } from "../../Rule";
 import { CASINGS_GENERATOR } from "../../utils/casingsGenerator";
 import { validateClientCredentials } from "./validateClientCredentials";
 
+export const OAUTH_DOCS_LINK = "https://buildwithfern.com/learn/api-definition/fern/api-yml-reference#authentication";
```
dsinghvi, line 14: we've talked a lot about linking folks out to `docs` -- one thing that could make this more manageable to ensure we don't break is to create a top level `docsLinks.ts` 

```ts
export interface DocsLinks {
  oauth: string;
  ....
}

expornt const DocsLinks: DocsLinks {
   oauth: "https://", 
}
```

More parts of the code base can then import this utility and rely on it

packages/cli/yaml/validator/src/rules/valid-oauth/__test__/valid-oauth.test.ts
```diff
@@ -71,4 +153,46 @@ describe("valid-oauth", () => {
...
+            absolutePathToWorkspace: join(
+                AbsoluteFilePath.of(__dirname),
+                RelativeFilePath.of("fixtures"),
+                RelativeFilePath.of("invalid"),
+                RelativeFilePath.of("types")
+            )
+        });
+        const expectedViolations: ValidationViolation[] = [
+            {
+                message: `For details on OAuth configuration, see ${OAUTH_DOCS_LINK}`,
```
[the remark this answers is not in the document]
amckinney: Good call, added. Error messages now look like the following (in compatible terminals):

image

dsinghvi (approved):

PR #3577, opened 2024-05-08
feat: support Stream and SSE in ExampleResponseSchema

abvthecity:
The response schema now accepts an undiscriminated union of body or stream, where stream itself can either be a list of data or a list of server side events, depending on the streaming format in the endpoint definition.

https://github.com/fern-api/fern/blob/1390e867dea2e89f728a0fa7c0ff8b7ebe49a9c8/packages/cli/yaml/yaml-schema/src/schemas/ExampleResponseSchema.ts#L4-L63

The fern check rule is updated to check that endpoint response-stream format (json or sse) matches the example response format, before also validating that the schema of the example aligns.

### DIFF
--- a/packages/cli/yaml/validator/src/rules/valid-example-endpoint-call/validateResponse.ts
+++ b/packages/cli/yaml/validator/src/rules/valid-example-endpoint-call/validateResponse.ts
@@ -28,7 +28,15 @@ export function validateResponse({
     errorResolver: ErrorResolver;
 }): RuleViolation[] {
     if (example == null) {
-        return [{ severity: "warning", message: "Response example is missing." }];
+        return validateBodyResponse({
+            example: {},
+            endpoint,
+            typeResolver,
+            exampleResolver,
+            file,
+            workspace,
+            errorResolver
+        });
     }
     return visitExampleResponseSchema(endpoint, example, {
         body: (example) =>


packages/cli/yaml/yaml-schema/src/schemas/ExampleResponseSchema.ts
```diff
@@ -1,9 +1,73 @@
...
+    stream: z.array(ExamplSseEventSchema)
+});
+
+export type ExampleSseResponseSchema = z.infer<typeof ExampleSseResponseSchema>;
+
+export const ExampleResponseSchema = z.union([
+    ExampleBatchResponseSchema,
+    ExampleStreamResponseSchema,
+    ExampleSseResponseSchema
+]);
```
dsinghvi, line 71: can we add an `sse` test defnition [here](https://github.com/fern-api/fern/tree/main/test-definitions/fern/apis) to make sure this works.

packages/cli/yaml/yaml-schema/src/schemas/ExampleResponseSchema.ts
```diff
@@ -1,9 +1,73 @@
...
+ */
+export const ExampleStreamResponseSchema = z.strictObject({
+    stream: z.array(ExampleTypeReferenceSchema)
+});
+
+export type ExampleStreamResponseSchema = z.infer<typeof ExampleStreamResponseSchema>;
+
+export const ExamplSseEventSchema = z.strictObject({
+    event: z.string(),
+    data: z.optional(ExampleTypeReferenceSchema)
```
dsinghvi, line 41: Can we update this `fern check` rule to check if the stream or data examples are actually valid? (https://github.com/fern-api/fern/tree/main/packages/cli/yaml/validator/src/rules/valid-example-endpoint-call)

packages/cli/yaml/validator/src/rules/valid-example-endpoint-call/validateResponse.ts
```diff
@@ -26,15 +26,44 @@ export function validateResponse({
     file: FernFileContext;
     workspace: FernWorkspace;
     errorResolver: ErrorResolver;
+}): RuleViolation[] {
+    if (example == null) {
+        return [{ severity: "warning", message: "Response example is missing." }];
+    }
```
dsinghvi, line 32: @abvthecity why did you add this?

abvthecity: ah this is probably not needed now that i'm reading the original code.
`return [];`

PR #3578, opened 2024-05-08
(feature, ts): Support oauth client credentials flow

amckinney:
This adds support for the OAuth client credentials flow in the TypeScript generator. The new `OAuthTokenProvider` automatically resolves the access token and refreshes it as needed. The resolved access token is then used as the bearer token in all client requests.


dsinghvi:
This looks great to me -- the only thing I could see us adding is some doc strings in the `OAuthTokenProvider` which explains some of the functionality

dsinghvi (approved):
the codegen looks great to me

PR #3592, opened 2024-05-10
feat: streaming and sse examples

abvthecity:
* create V44 IR
* ResponseExample is now a union between Body, Stream, SSE
* default generated stream examples [will have 2 events](https://github.com/fern-api/fern/pull/3592/files#diff-df1c75b317625fc418d0d811ac28d0db00d73c87c8ee88520c9c507d02155427R410-R424)
* IR for stream + sse are converted to FDR


packages/cli/generation/ir-migrations/src/migrations/v38-to-v37/__test__/__snapshots__/migrateFromV38ToV37.test.ts.snap
```diff
@@ -175,293 +175,7 @@ exports[`migrateFromV38ToV37 snapshot 1`] = `
               },
               "response": {
                 "_visit": [Function],
-                "body": {
```
abvthecity, line 178: I'm accepting this snapshot update because this is how the V44->V43 migrator is meant to behave:
- keep only response body examples
- wipe away stream and sse response examples

The v38toV37 migrator previously generated a dummy example from the streaming endpoint, which is being treated as a body JSON example, which is incorrect.

packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
```diff
@@ -600,9 +608,36 @@ function convertHttpEndpointExample(
                 throw new Error("Unknown ExampleResponse: " + irExample.response.type);
             }
         }),
-        responseBody: irExample.response.body?.jsonExample,
-        responseBodyV3:
-            irExample.response.body != null ? { type: "json", value: irExample.response.body.jsonExample } : undefined,
+        responseBody: irExample.response._visit({
```
abvthecity, line 611: responseBody is deprecated anyway in favor of responseBodyV3, so this part of the code is technically useless (and should be deletd)

packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
```diff
@@ -390,8 +390,15 @@ types:
       error: ExampleEndpointErrorResponse
 
   ExampleEndpointSuccessResponse:
-    properties:
+    union:
       body: optional<types.ExampleTypeReference>
+      stream: list<types.ExampleTypeReference>
```
abvthecity, line 395: this is where the magic is happening

dsinghvi (approved):

dsinghvi:
Looks like this change may have entirely busted examples in the TypeScript generator

PR #3595, opened 2024-05-10
fix: ensure formdata utils work cross-runtime

armandobelardo:
this is the lib impl, now to add it to the generated sdks

fixes FER-1512


generators/typescript/utils/core-utilities/form-data-utils/src/FormDataWrapper.ts
```diff
@@ -0,0 +1,67 @@
+import { RUNTIME } from "@fern-typescript/runtime";
```
dsinghvi, line 1: @armandobelardo for anything that uses `RUNTIME` we end up having to put inside the `fetcher` utils -- see `stream` as an example

armandobelardo: updated!

PR #3601, opened 2024-05-12
fix: ensure formdata utils work cross-runtime

armandobelardo:
This PR leverages `formdata-node` and `form-data-encoder` when in a node environment, falling back to `form-data` otherwise (the current state of the world). `form-data` should work when there is a global `FormData` instance or one present within the window (e.g. in-browser)

fixes FER-1512


remorses:
Can we revert this PR? FormData is already available in Node 18, which is the oldest Node version currently maintained

PR #3604, opened 2024-05-13
feat: introduce pagination to python

armandobelardo:
Needs testing but this should be the bulk of it
Fixes FER-1522

This consumes the pagination IR
The generated `client.Users.List` can then be used as a `User` generator (effectively the "default"):

  ```python
  for user in client.users.list(...):
    print user
  ```

  a page-by-page generator:

  ```python
  for page in client.users.list(...).iter_pages():
    print(page.items)
  ```

  or statically calling `next_page` to perform the pagination manually:

  ```python
  pager = client.users.list(...)
  # First page
  print(pager.items)
  # Second page
  pager = pager.next_page()
  print(pager.items)
  ```


dsinghvi:
The client api looks great -- can you document it [here](https://buildwithfern.com/learn/sdks/features/auto-pagination) once we finalize it?

generators/python/core_utilities/sdk/pagination.py
```diff
@@ -0,0 +1,71 @@
...
+
+class SyncPage(BasePage, typing.Generic[T]):
+    get_next: typing.Optional[typing.Callable[[], typing.Optional[Self]]]
+
+class AsyncPage(BasePage, typing.Generic[T]):
+    get_next: typing.Optional[typing.Callable[[], typing.Awaitable[typing.Optional[Self]]]]
+
+# ----------------------------
+
+class SyncPaginator(SyncPage[T], typing.Generic[T]):
```
dsinghvi, line 32: Should we align the naming here with Go like `SyncPager`, `AsyncPager` ?

generators/python/sdk/CHANGELOG.md
```diff
@@ -5,6 +5,66 @@ All notable changes to this project will be documented in this file.
...
+
+  or statically calling `next_page` to perform the pagination manually:
+
+  ```python
+  pager = client.users.list(...)
+  # First page
+  print(pager.items)
+  # Second page
+  pager = pager.next_page()
+  print(pager.items)
```
dsinghvi, line 65: @armandobelardo we'll want to gate this feature as well

generators/python/sdk/VERSION
```diff
@@ -1 +1 @@
-1.7.0-rc0
+2.0.1
```
dsinghvi, line 1: should we do a minor since its a feature release

seed/python-sdk/alias/src/seed/core/pagination.py
```diff
@@ -0,0 +1,79 @@
+# This file was auto-generated by Fern from our API Definition.
+
+import asyncio
```
dsinghvi, line 3: can we only include `pagination.py` if there are paginated endpoints?

seed/python-sdk/pagination/src/seed/users/client.py
```diff
@@ -101,7 +103,21 @@ def list_with_cursor_pagination(
...
+                _has_next = _parsed_next is not None
+                _get_next = lambda: self.list_with_cursor_pagination(
+                    page=page,
+                    per_page=per_page,
+                    order=order,
+                    starting_after=_parsed_next,
+                    request_options=request_options,
+                )
+            _items = _parsed_response.data
+            return SyncPaginator(has_next=_has_next, items=_items, get_next=_get_next)
```
dsinghvi, line 120: how does this work if `_parsed_next` and `_has_next` aren't lambdas and aren't recomputed on every page load?

seed/python-sdk/pagination/src/seed/users/client.py
```diff
@@ -116,7 +132,7 @@ def list_with_offset_pagination(
         order: typing.Optional[Order] = None,
         starting_after: typing.Optional[str] = None,
         request_options: typing.Optional[RequestOptions] = None,
-    ) -> ListUsersPaginationResponse:
+    ) -> SyncPaginator[User]:
```
dsinghvi, line 135: i don't see `snippets.json` updated for this endpoint, but would be nice to include the `for ... if`

generators/python/core_utilities/sdk/pagination.py
armandobelardo: Updating!

seed/python-sdk/pagination/src/seed/users/client.py
armandobelardo: It might be easier to include in a different PR

seed/python-sdk/pagination/src/seed/users/client.py
armandobelardo: discussed offline, but the `get_next` is the API call wrapped in a lambda, so it's passing in the function call it was just in, with updated parameters
so the full SyncPager is recomputed and returned on the `get_next` call

seed/python-sdk/alias/src/seed/core/pagination.py
armandobelardo: Will add!

generators/python/sdk/CHANGELOG.md
armandobelardo: Will add a flag, I think it might be easier to just omit from the IR, but will think about it more

armandobelardo: PR here: https://github.com/fern-api/venus/pull/150

test-definitions/fern/apis/pagination/ir.json
```diff
@@ -0,0 +1,6992 @@
+{
```
dsinghvi, line 1: is this file intentional?

dsinghvi (approved):
looks great -- can we run `yarn seed:build and yarn seed write-inputs` for all of the generators. 

And can we make this a `2.1.0` bump

test-definitions/fern/apis/pagination/ir.json
armandobelardo: ah no it's not! removing

PR #3607, opened 2024-05-13
(fix, openapi): Consolidate enums into discriminants

amckinney:
Similar to This fixes an issue in the OpenAPI parser, where we mistakenly categorize a discriminated union as undiscriminated. With this, we consolidate any `enum` variants (e.g. `foo` vs. `FOO`) into a single discriminant, which makes it possible to identify a discriminated union.


packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -937,6 +937,27 @@ function getPossibleDiscriminantsForSchemaObject({
...
+    const set = new Set<any>();
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    enums.forEach((item: any) => {
+        if (typeof item === "string") {
+            set.add(item.toLowerCase());
+        } else {
+            set.add(item);
+        }
+    });
```
dsinghvi, line 956: this makes sense -- can you add a comment about why we are doing this

PR #3608, opened 2024-05-13
[WIP] Upgrade Java Generator to IR 42

dcb6:
todo:
- [ ] update `migrateFromV31ToV30.ts` type files -- double check that versions i've chosen for generators are correct first + check that I should put that new version in all those files

refererence pr: https://github.com/fern-api/fern/pull/2884/files


generators/java/sdk/VERSION
```diff
@@ -1 +1 @@
-0.8.11
\ No newline at end of file
+0.9.0-rc0
```
dcb6, line 1: wasn't sure on these versions -- pr i based this off of bumped minor version + made it an rc so i imitated

PR #3613, opened 2024-05-14
(fix, openapi): OpenAPI importer now parses list examples that are specific to a field

dsinghvi:
Previously we would skip parsing list examples that were specified at the field level. Now those are parsed!

### DIFF
--- a/packages/cli/openapi-parser/src/schema/convertArray.ts
+++ b/packages/cli/openapi-parser/src/schema/convertArray.ts
@@ -41,7 +41,7 @@ export function convertArray({
         itemSchema,
         wrapAsNullable,
         description,
-        example: example
+        example
     });
 }
 


packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -376,7 +376,8 @@ export function convertSchemaObject(
             description,
             wrapAsNullable,
             context,
-            groupName
+            groupName,
+            example: getExampleAsArray(schema)
```
dsinghvi, line 380: we were never passing through examples on the array field itself

packages/cli/openapi-parser/src/schema/examples/ExampleTypeFactory.ts
```diff
@@ -287,8 +287,23 @@ export class ExampleTypeFactory {
                             itemExamples.push(itemExample);
                         }
                     }
-                } else {
+                    // Otherwise, use a schema level example
+                } else if (schema.example != null && schema.example.length > 0) {
```
dsinghvi, line 291: if the full example doesn't contain the array example, then we default to a field specific example

PR #3615, opened 2024-05-14
(docs): Add OAuth SDK docs

amckinney:
This adds a new `OAuth` section to the SDK docs. A simple preview is shown below:

image


dsinghvi (approved):
Thanks for adding; capturing our discussion: 
1) may be more intuitive to only show the client consumption code
2) may be worth linking to real life examples (sayari + webflow)

PR #3617, opened 2024-05-14
(fix, ts): Client credentials are optional with env vars

amckinney:
This updates the OAuth client constructor to specify `clientId` and `clientSecret` as optional parameters when environment variables are specified.


generators/typescript/sdk/client-class-generator/src/GeneratedSdkClientClassImpl.ts
```diff
@@ -880,6 +880,9 @@ export class GeneratedSdkClientClassImpl implements GeneratedSdkClientClass {
         }
 
         if (this.isRoot && this.oauthAuthScheme != null && context.generateOAuthClients) {
+            const hasQuestionToken =
```
armandobelardo, line 883: chatted offline, it might be nicer if these null checks were divorced so you can require one and not the other

amckinney: Good call, ended up also refactoring the other relevant call-site that tied them together.

armandobelardo (approved):

PR #3618, opened 2024-05-14
fix: fern-aware pydantic models now effectively 'exclude_optional' in…

armandobelardo:
…stead of 'exclude_unset'


amckinney (approved):
Discussed offline, lgtm

PR #3621, opened 2024-05-14
chore: add nuget config for csharp sdks

armandobelardo:


seed/csharp-model/alias/.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v3
+
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 7.0.404
+
+      - name: Publish
+        env:
+          NUGET_API_KEY: ${{ secrets. }}
```
dsinghvi, line 44: @armandobelardo small typo here

seed/csharp-model/api-wide-base-path/.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v3
+
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 7.0.404
+
+      - name: Publish
+        env:
+          NUGET_API_KEY: ${{ secrets. }}
```
dsinghvi, line 44: we make the key here `NUGET_API_KEY` and then reference `secrets.NUGET_API_TOKEN` in the actual command

generators/csharp/codegen/src/project/CsharpProject.ts
```diff
@@ -71,6 +71,16 @@ export class CsharpProject {
...
 
+        const githubWorkflowTemplate = (await readFile(getAsIsFilepath(AsIsFiles.CiYaml))).toString();
+        const githubWorkflow = template(githubWorkflowTemplate)({
+            projectName: this.name,
+            shouldWritePublishBlock: this.context.publishConfig != null,
+            nugetTokenEnvvar: this.context.publishConfig?.apiKeyEnvironmentVariable ?? "NUGET_API_TOKEN"
+        }).replaceAll("\\{", "{");
+        const ghDir = join(this.absolutePathToOutputDirectory, RelativeFilePath.of(".github/workflows"));
+        await mkdir(ghDir, { recursive: true });
+        await writeFile(join(ghDir, RelativeFilePath.of("ci.yml")), githubWorkflow);
```
dsinghvi, line 82: how do we make sure this only happens when the generator is in `github` mode?

armandobelardo: The template has a conditional that's if the publish block is present, which is a GitHub output block

I tested it and it works well, the whole block is added or removed

Check the template evaluation a couple lines above

armandobelardo: https://github.com/fern-api/fern/pull/3621/files#diff-2e2d2dd0dc8bd56571b1e52014eab736a44e2dc8f6d0f4306c98f7ababb197daR44-R53

seed/csharp-model/api-wide-base-path/.github/workflows/ci.yml
armandobelardo: Updated!

seed/csharp-model/alias/.github/workflows/ci.yml
armandobelardo: Updated!

armandobelardo: This was because I set the envvar name to empty string in seed

dsinghvi (approved):
🙏

PR #3622, opened 2024-05-14
(fix, openapi): Fix nameOverride resolution

amckinney:
This fixes an issue where the `nameOverride` was not respecting the `x-fern-type-name` extension. 

Previously, we would evaluate the extension _and_ the `shouldUseTitleName` configuration together before determining whether or not we should use the title fallback. This groups the conditions separately so that we get the desired behavior.


dsinghvi (approved):
thanks, this fixes scorecard too

PR #3625, opened 2024-05-15
(feat): Add default values, validation rules, and big integer to primitives

dsinghvi:


dsinghvi:
@amckinney this looks good to me -- would be great to add a section [here](https://buildwithfern.com/learn/api-definition/fern/overview) about 
- bigint
- validation rules

amckinney:
Good call - I'll add those docs when I introduce the YAML schema changes for the new default and validation rule schema in my next PR.

amckinney (approved):

PR #3629, opened 2024-05-15
improvement: python now respects deep object query parameters

armandobelardo:


generators/python/core_utilities/sdk/query_encoder.py
```diff
@@ -0,0 +1,35 @@
...
+  for k, v in dict_flat.items():
+    key = f'{key_prefix}[{k}]' if key_prefix is not None else k
+    if isinstance(v, dict):
+      result.update(traverse_query_dict(v, key))
+    else:
+      result[key] = v
+  return result
+
+
+def single_query_encoder(query_key: str, query_value: Any) -> dict:
```
dsinghvi, line 21: can we add a couple pytest unit tests for this function -- but othwerise it looks good

armandobelardo: some tests added

dsinghvi (approved):

PR #3635, opened 2024-05-16
feat: add seo and metadata configuration in docs.yml

abvthecity:
added:
- footer-links
- metadata (for SEO)
- redirects
- integrations


dsinghvi (approved):
can we document these new properties?

PR #3639, opened 2024-05-16
refactor: share common logic between publishDocs and previewDocs

abvthecity:
### Before this PR:

publishDocs and previewDocs were independent implementations doing nearly the same thing, and growing out-of-sync

### After this PR:

* publishDocs is refactored into a stateful `DocsDefinitionResolver` class (introducing `@fern-api/docs-resolver`)
* certain features of publishDocs were extracted, so that DocsDefinitionResolver can be shared with previewDocs
* updated previewDocs to reuse DocsDefinitionResolver + `fdr-sdk`
* update the preview server to watch for file changes, and communicate via websocket

### DIFF
--- a/packages/cli/docs-preview/tsconfig.json
+++ b/packages/cli/docs-preview/tsconfig.json
@@ -6,7 +6,6 @@
     { "path": "../../commons/fs-utils" },
     { "path": "../../ir-sdk" },
     { "path": "../docs-resolver" },
-    { "path": "../generation/ir-generator" },
     { "path": "../logger" },
     { "path": "../project-loader" },
     { "path": "../register" },


packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -41,184 +44,146 @@ export async function publishDocs({
...
+        }
+    );
+
+    const docsDefinition = await resolver.resolve();
 
-    // postprocess markdown files after uploading all images to replace the image paths in the markdown files with the fileIDs
-    for (const [relativePath, markdown] of Object.entries(parsedDocsConfig.pages)) {
-        parsedDocsConfig.pages[RelativeFilePath.of(relativePath)] = replaceImagePaths(markdown, collectedFileIds);
+    if (docsRegistrationId == null) {
+        return context.failAndThrow("Failed to publish docs.", "Docs registration ID is missing.");
```
dsinghvi, line 179: Don't think we ever want to show the error `"Docs registration ID is missing."` is a user

packages/cli/docs-resolver/src/__test__/__snapshots__/parseImagePaths.test.ts.snap
```diff
@@ -2,25 +2,25 @@
...
+  "/Volume/git/fern/tutorials/end_node.png",
+  "/Volume/git/fern/tutorials/kb_node.png",
+  "/Volume/git/fern/tutorials/webhook_node.png",
+  "/Volume/git/fern/tutorials/add_global_prompt.png",
+  "/Volume/git/fern/tutorials/live_call_logs.png",
+  "/Volume/git/fern/tutorials/pathway_chat.png",
+  "/Volume/git/fern/tutorials/extract_variables.png",
+  "/Volume/git/fern/tutorials/pathwaylog_finetune.png",
+  "/Volume/git/fern/tutorials/finetuned.png",
+  "/Volume/git/fern/tutorials/finetune.png",
```
dsinghvi, line 23: is this intentional?

dsinghvi: looks like it is, resolving

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
dsinghvi: will fix in a follow up

dsinghvi:
@abvthecity i've been testing this PR out and the live-reload almost never works for me, but we can address that in a follow up -- the rest of the refactor looks great

PR #3640, opened 2024-05-16
(feature): Add support for default values and validation rules

amckinney:
This adds support for default values and validation rules for integer, double, and string types.

The Fern definition looks like the following, where each of the relevant primitive types has a full example:

```yaml
types:
  Type:
    docs: |
      Defines properties with default values and validation rules.
    properties:
      decimal:
        type: double
        default: 1.1
        validation:
          exclusiveMin: true
          exclusiveMax: false
          min: 1.1
          max: 2.2
          multipleOf: 1.1
      even:
        type: integer
        default: 42
        validation:
          exclusiveMin: false
          exclusiveMax: true
          min: 0
          max: 100
          multipleOf: 2
      name:
        type: string
        default: fern
        validation:
          minLength: 3
          maxLength: 10
          pattern: ^[a-z]+$
```

### DIFF
--- a/seed/go-sdk/validation/snippet.json
+++ b/seed/go-sdk/validation/snippet.json
@@ -3,7 +3,8 @@
         {
             "id": {
                 "path": "/",
-                "method": "GET"
+                "method": "GET",
+                "identifier_override": "endpoint_.get"
             },
             "snippet": {
                 "type": "go",
@@ -13,7 +14,8 @@
         {
             "id": {
                 "path": "/create",
-                "method": "POST"
+                "method": "POST",
+                "identifier_override": "endpoint_.create"
             },
             "snippet": {
                 "type": "go",
--- a/seed/java-model/seed.yml
+++ b/seed/java-model/seed.yml
@@ -17,6 +17,7 @@ allowedFailures:
   - enum
   - exhaustive
   - extends
+  - multi-url-environment
   - reserved-keywords
   - trace
   - undiscriminated-unions
--- a/seed/python-sdk/validation/snippet-templates.json
+++ b/seed/python-sdk/validation/snippet-templates.json
@@ -7,7 +7,8 @@
         },
         "endpointId": {
             "path": "/create",
-            "method": "POST"
+            "method": "POST",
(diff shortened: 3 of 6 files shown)


packages/cli/yaml/yaml-schema/src/utils/visitRawTypeReference.ts
```diff
@@ -20,28 +23,79 @@ export interface RawTypeReferenceVisitor<R> {
...
+                            : undefined
+                })
+            });
+        }
+        case RawPrimitiveType.string: {
+            const maybeStringValidation = validation != null ? (validation as StringValidationSchema) : undefined;
+            return visitor.primitive({
+                v1: PrimitiveTypeV1.String,
+                v2: PrimitiveTypeV2.string({
+                    default: _default != null ? (_default as string) : undefined,
```
dsinghvi, line 81: do we have a `fern check` rule that checks the default values against the `type` ? otherwise this cast can thread through incorrect types

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/devrev.test.ts.snap
```diff
@@ -93,8 +93,10 @@ exports[`open api parser devrev docs 1`] = `
               "body": {
                 "properties": {
                   "file_name": {
+                    "default": undefined,
                     "docs": "The name of the file that's being uploaded.",
                     "type": "string",
+                    "validation": undefined,
```
dsinghvi, line 99: can we update the openapi-ir-to-fern to only add `default` and `validation` if they are non empty? `fern write-definition` shouldn't be writing out `undefined` values

packages/cli/yaml/yaml-schema/src/utils/visitRawTypeReference.ts
amckinney: I was originally planning to add it in a follow-up, but I went ahead and included it here

dsinghvi (approved):

PR #3645, opened 2024-05-17
Tidy up python generator docs

fabubaker:
A small fix to the example in the `pydantic_config.include_union_utils` section.

### DIFF
--- a/generators/python/README.md
+++ b/generators/python/README.md
@@ -19,7 +19,7 @@
 This repository contains the source for the various generators that produce Python artifacts for [Fern](https://github.com/fern-api/fern):
 
 - `fernapi/fern-python-sdk`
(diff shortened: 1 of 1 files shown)


fabubaker:
Hey folks,

Just wanted to bring to your attention that even though the docs say the default for `pydantic_config.include_union_utils` is `False`, this is not true when using the fastAPI generators (that override happens [here](https://github.com/fern-api/fern/blob/cd0da26ca09f067a7ac75a8ecbcaa61ddf96339c/generators/python/src/fern_python/generators/fastapi/fastapi_generator.py#L60)). 

Not sure if this is intentional or not. I noticed this because there was a difference in the types produced by our fastapi generator vs the python sdk generator.

armandobelardo (approved):

PR #3651, opened 2024-05-20
replaced deprecated 'excerpt' frontmatter property with 'description'

atwooddc:


dannysheridan:
@abvthecity can you confirm we want to make this change? Was my ask.

PR #3664, opened 2024-05-21
(fix, python): change author format and fix query encoder

armandobelardo:


amckinney (approved):
As long as the `pyproject.toml` snapshots look good, then lgtm.

PR #3675, opened 2024-05-21
(feature, openapi): Map additionalProperties to extra-properties

amckinney:
This updates the OpenAPI parser so that we map the `additionalProperties` setting (when set to `{}` or `true`) to Fern's `extra-properties` setting, which is available on both object declarations and in-lined requests. For example, consider the following type in OpenAPI:

```yaml
components:
  schemas:
    User:
      type: object
      properties:
        name:
          type: string
          description: The name of the user.
      additionalProperties: true
      description: Defines a user that contains arbitrary metadata.
```

The `User` gets mapped to the following Fern definition:

```yaml
types:
  User:
    docs: >-
      Defines a user that contains arbitrary metadata.
    properties:
      name:
        type: optional<string>
        docs: The name of the user.
    extra-properties: true
```


packages/cli/openapi-ir-sdk/fern/definition/finalIr.yml
```diff
@@ -454,6 +455,7 @@ types:
       allOf: list<ReferencedSchema>
       properties: list<ObjectProperty>
       allOfPropertyConflicts: list<AllOfPropertyConflict>
+      extraProperties: boolean
```
dsinghvi, line 458: nit: i try to keep "openapi concepts" in the OpenAPI parser so calling these `additionalProperties` would make more sense and we map to `extra-properties` in openapi-ir-to-fern

dsinghvi (approved):

packages/cli/openapi-ir-sdk/fern/definition/finalIr.yml
amckinney: Totally makes sense, will update!

PR #3676, opened 2024-05-21
(fix, python): correct snippets for optional referenced requests when…

armandobelardo:
… inlined_request_params is True

### DIFF
--- a/generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
+++ b/generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
@@ -1026,14 +1026,14 @@ class EndpointFunctionGenerator:
             )
         return query_parameter_type_hint
 
+
 def _is_type_reference_optional(type_reference: ir_types.TypeReference) -> bool:
-    return (type_reference.get_as_union().type == "reference"
-                and type_reference.get_as_union().request_body_type.get_as_union().type == "container"
-                and type_reference.get_as_union()
-                .request_body_type.get_as_union()
-                .container.get_as_union()
-                .type
-                == "optional")
+    return (
+        type_reference.get_as_union().type == "reference"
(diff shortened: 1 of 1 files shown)


generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
```diff
@@ -1107,13 +1109,19 @@ def generate_snippet(self) -> AST.Expression:
                 )
 
         if self.example.request is not None:
+            # For some reason the example type refernce is not marking it's type as optional, so we need to specify it so the
```
amckinney, line 1112: nit: 
```suggestion
            # For some reason the example type reference is not marking it's type as optional, so we need to specify it so the
```

generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
```diff
@@ -1107,13 +1109,19 @@ def generate_snippet(self) -> AST.Expression:
...
+            # snippets (and thus unit tests) write correctly
+            is_optional = False
+            if self.endpoint.request_body is not None and self.endpoint.request_body.get_as_union().type == "reference" and self.endpoint.request_body.get_as_union().request_body_type.get_as_union().type == "container" and self.endpoint.request_body.get_as_union().request_body_type.get_as_union().container.get_as_union().type == "optional":
+                is_optional = True
```
amckinney, line 1116: nit: Might be worthwhile to put this in a helper, unless we plan to remove this check for something better later (re: your comment).

amckinney (approved):
lgtm, only nits

generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
armandobelardo: updated!

PR #3678, opened 2024-05-22
(fix, csharp): package in LICENSE in `.csproj`

dsinghvi:


generators/csharp/codegen/src/project/CsharpProject.ts
```diff
@@ -304,4 +306,16 @@ class CsProj {
...
+    private getAdditionalItemGroups(): string[] {
+        const result: string[] = [];
+
+        if (this.license != null) {
+            result.push(`<ItemGroup>
+    <None Include="..\\..\\README.md" Pack="true" PackagePath=""/>
+</ItemGroup>`);
+        }
+
+        return result;
```
dsinghvi, line 319: this is the key part

armandobelardo (approved):

PR #3682, opened 2024-05-22
java: oauth improvements including token refresh

dcb6:
improvements:
* implement token refresh (via client credentials only, not by refresh token)
* handle token prefix
* throw better exceptions

### DIFF
--- a/generators/java/sdk/CHANGELOG.md
+++ b/generators/java/sdk/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.9.3] - 2024-05-23
 
-- Feature: Support OAuth token refresh.
+- Feature: Generated SDK clients with an OAuth security scheme will now automatically refresh access tokens before they
+  expire.
 
 ## [0.9.2] - 2024-05-21
 


generators/java/sdk/src/main/java/com/fern/java/client/Cli.java
```diff
@@ -233,6 +223,17 @@ public GeneratedRootClient generateClient(ClientGeneratorContext context, Interm
...
         });
 
+        Optional<GeneratedJavaFile> generatedOAuthTokenSupplier =
+                maybeOAuthScheme.map(it -> new OAuthTokenSupplierGenerator(
+                                context,
+                                it.getConfiguration()
+                                        .getClientCredentials()
+                                        .orElseThrow(() ->
+                                                new RuntimeException("Only client credentials oAuth scheme supported")),
+                                generatedFiles)
```
dcb6, line 233: change is just adding generated files here

seed/java-sdk/oauth-client-credentials-default/src/main/java/com/seed/oauthClientCredentialsDefault/core/OAuthTokenSupplier.java
```diff
@@ -37,6 +37,6 @@ public String get() {
             TokenResponse authResponse = fetchToken();
             this.accessToken = authResponse.getAccessToken();
         }
-        return "Bearer " + (accessToken != null ? accessToken : fetchToken());
```
dcb6, line 40: small separate improvement -- i already check above if access token is null

dsinghvi (approved):

PR #3684, opened 2024-05-23
internal: specify the example ID within the generated snippet

armandobelardo:


generators/typescript/sdk/generator/src/SdkGenerator.ts
```diff
@@ -806,7 +806,8 @@ export class SdkGenerator {
                             },
                             snippet: FernGeneratorExec.EndpointSnippet.typescript({
                                 client: snippet
-                            })
+                            }),
+                            exampleIdentifier: example.name?.originalName
```
dsinghvi, line 810: On line 784 we do `const example = endpoint.examples[0]`, can we turn that into a for loop now?

armandobelardo: missed that, updated

generators/ruby/codegen/src/ast/ExampleGenerator.ts
```diff
@@ -106,7 +106,8 @@ export class ExampleGenerator {
                 },
                 snippet: FernGeneratorExec.EndpointSnippet.ruby({
                     client: exampleNode.write({ startingTabSpaces: 0 })
-                })
+                }),
+                exampleIdentifier: endpoint.name.originalName
```
dsinghvi, line 110: we are setting the exampleIdentifier to the original name of the endpoint -- this should be the name of the example. 

I'd also expect to see a for loop somewhere where we iterate over each example, does that already exist?

armandobelardo: Yeah fixing ruby right now

generators/python/src/fern_python/generators/sdk/client_generator/root_client_generator.py
```diff
@@ -247,14 +247,14 @@ def _create_class_declaration(
...
+                if generated_endpoint_function.snippet is not None and len(generated_endpoint_function.snippet) > 0:
                     if is_async:
                         self._snippet_registry.register_async_client_endpoint_snippet(
-                            endpoint=endpoint, expr=generated_endpoint_function.snippet
+                            endpoint=endpoint, expr=generated_endpoint_function.snippet[0].snippet, example_id=None
                         )
                     else:
                         self._snippet_registry.register_sync_client_endpoint_snippet(
-                            endpoint=endpoint, expr=generated_endpoint_function.snippet
+                            endpoint=endpoint, expr=generated_endpoint_function.snippet[0].snippet, example_id=None
```
dsinghvi, line 257: @armandobelardo any idea why we don't loop over the snippets here?

armandobelardo: I think just an oversight, it should be a similar look to the service-level snippets

seed/go-sdk/examples/snippet.json
```diff
@@ -88,6 +187,17 @@
...
+        {
+            "id": {
+                "path": "/response",
+                "method": "POST",
+                "identifier_override": "endpoint_service.getResponse"
+            },
+            "snippet": {
+                "type": "go",
+                "client": "import (\n\tcontext \"context\"\n\tfern \"github.com/examples/fern\"\n\tfernclient \"github.com/examples/fern/client\"\n\toption \"github.com/examples/fern/option\"\n)\n\nclient := fernclient.NewClient(\n\toption.WithToken(\n\t\t\"\u003cYOUR_AUTH_TOKEN\u003e\",\n\t),\n\toption.WithBaseURL(\n\t\tfern.Environments.Production,\n\t),\n)\nresponse, err := client.Service.GetResponse(\n\tcontext.TODO(),\n)\n"
+            }
```
dsinghvi, line 199: I don't think the example ids are still coming though in Go -- is there a place that i should be checking @armandobelardo ?

dsinghvi:
the merge conflicts are brutal and i can take a stab at them for slowing this down

PR #3693, opened 2024-05-24
feat: markdown-in-markdown - load markdown from another markdown file.

abvthecity:
`<Markdown src="./path/to/markdown.mdx" />` will be replaced by the contents of that file.


dsinghvi (approved):
should this be documented?

Issue #2710, opened 2023-08-25
[Bug] FastAPI generator outputs paths with an extra backslash /


ngoldack:
When using fernapi/fern-fastapi-server (version 0.4.4) it registers the paths with a wrong path in FastAPI.

`generators.yaml`:
```yaml
groups:
    server:
        generators:
            - name: fernapi/fern-fastapi-server
              version: 0.4.4
              output:
                  location: local-file-system
                  path: ../../apps/api_new/src/api/generated
              config:
                  include_validators: false
```

I am using following `openapi.yaml`:

```yaml openapi.yaml
...
paths:
    /health:
        get:
            summary: Health
            description: |
                "This endpoint is used to check the health of the API."
            operationId: getHealth
            tags:
                - health
            responses:
                "200":
                    description: OK
                    content:
                        application/json:
                            schema:
                                $ref: "#/components/schemas/Health"
                "400":
                    $ref: "#/components/responses/BadRequest"
                "500":
                    $ref: "#/components/responses/InternalServerError"
...
```

Generated AbstractService (error is near the bottom):
```py
...
@classmethod
    def __init_get_health(cls, router: fastapi.APIRouter) -> None:
        endpoint_function = inspect.signature(cls.get_health)
        new_parameters: typing.List[inspect.Parameter] = []
        for index, (parameter_name, parameter) in enumerate(endpoint_function.parameters.items()):
            if index == 0:
                new_parameters.append(parameter.replace(default=fastapi.Depends(cls)))
            else:
                new_parameters.append(parameter)
        setattr(cls.get_health, "__signature__", endpoint_function.replace(parameters=new_parameters))

        @functools.wraps(cls.get_health)
        def wrapper(*args: typing.Any, **kwargs: typing.Any) -> Health:
            try:
                return cls.get_health(*args, **kwargs)
            except (BadRequestError, InternalServerError) as e:
                raise e
            except FernHTTPException as e:
                logging.getLogger(f"{cls.__module__}.{cls.__name__}").warn(
                    f"Endpoint 'get_health' unexpectedly threw {e.__class__.__name__}. "
                    + f"If this was intentional, please add {e.__class__.__name__} to "
                    + "the endpoint's errors list in your Fern Definition."
                )
                raise e

        # this is necessary for FastAPI to find forward-ref'ed type hints.
        # https://github.com/tiangolo/fastapi/pull/5077
        wrapper.__globals__.update(cls.get_health.__globals__)

        router.get(
            path="//health", # <---
[...]

PR #3699, opened 2024-05-27
fix, fastapi: fixes path prefixes and construction

armandobelardo:
Should resolve https://github.com/fern-api/fern/issues/2710


amckinney (approved):
Nice, great to see that we had snapshots that already demonstrated the problem too.

PR #3710, opened 2024-05-27
improvement: filter out nulls after merging API specs

armandobelardo:
This would break if there's ever a place we want to handle an explicit null value, but I wasn't tracking any


dsinghvi (approved):
@armandobelardo this looks good to me but looks like CI is broken

PR #3719, opened 2024-05-28
java, fix: de-conflict undiscriminated unions

dcb6:
Undiscriminated unions are now generated with de-conflicted method signatures. Previously, certain undiscriminated unions would have failed to compile due to Java's type erasure causing conflicts.


amckinney (approved):
Very clear and easy to follow, nicely done

dsinghvi (approved):
This could potentially be a breaking change -- any thoughts on how to deal with that?

generators/java/model/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.8.0] - 2024-05-28
+
+- Fix: Undiscriminated unions are now generated with de-conflicted method signatures. Previously, certain
+  undiscriminated unions would have failed to compile due to Java's type erasure causing conflicts.
```
dsinghvi, line 11: if you have entirely deconflicted undiscriminated unions (this is going to cause the SDK to break). Here is an [example](https://github.com/merge-api/merge-java-client/blob/6cc427ce66b2d8964677a110443d311e11f2027f/src/main/java/com/merge/api/resources/ats/types/OfferCreator.java#L60-L68)

dsinghvi:
I think the best approach is probably: 
- introduce a feature flag
- default feature flag to off (as it works now)
- major version release with feature flag to true 
- existing users can upgrade how they see fit

PR #3740, opened 2024-05-30
improvement: add a flag to allow python to generate discriminated unions as undiscriminated unions

armandobelardo:
This PR adds a config for API settings (`use-undiscriminated-unions-with-literals`) that creates discriminated unions as undiscriminated unions, forcing the discriminant as a literal property on the members

This PR also allows the OpenAPI parsing/conversion step to now take into account the generation language

### DIFF
--- a/packages/cli/configuration/src/generators-yml/convertGeneratorsConfiguration.ts
+++ b/packages/cli/configuration/src/generators-yml/convertGeneratorsConfiguration.ts
@@ -98,7 +98,7 @@ async function parseAPIConfiguration(
                         audiences: definition.audiences,
                         settings: {
                             shouldUseTitleAsName: definition.settings?.["use-title"],
-                            shouldUseUndiscriminatedUnionsWithLiterals: definition.settings?.["unions"] === "v1"
+                            shouldUseUndiscriminatedUnionsWithLiterals: definition.settings?.unions === "v1"
                         }
                     });
                 }
@@ -111,7 +111,7 @@ async function parseAPIConfiguration(
                 audiences: apiConfiguration.audiences,
                 settings: {
                     shouldUseTitleAsName: apiConfiguration.settings?.["use-title"],
-                    shouldUseUndiscriminatedUnionsWithLiterals: apiConfiguration.settings?.["unions"] === "v1"
+                    shouldUseUndiscriminatedUnionsWithLiterals: apiConfiguration.settings?.unions === "v1"
                 }
             });
         }
@@ -130,7 +130,7 @@ async function parseAPIConfiguration(
                 audiences: [],
                 settings: {
                     shouldUseTitleAsName: settings?.["use-title"],
-                    shouldUseUndiscriminatedUnionsWithLiterals: settings?.["unions"] === "v1"
+                    shouldUseUndiscriminatedUnionsWithLiterals: settings?.unions === "v1"
                 }
             });
         } else if (openapi != null) {
@@ -141,7 +141,7 @@ async function parseAPIConfiguration(
                 audiences: [],
                 settings: {
                     shouldUseTitleAsName: openapi.settings?.["use-title"],
-                    shouldUseUndiscriminatedUnionsWithLiterals: openapi.settings?.["unions"] === "v1"
+                    shouldUseUndiscriminatedUnionsWithLiterals: openapi.settings?.unions === "v1"
                 }
             });
         }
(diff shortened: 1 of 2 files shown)


packages/cli/workspace-loader/src/types/Workspace.ts
```diff
@@ -36,7 +36,9 @@ export interface Spec {
 
 export interface SpecImportSettings {
     audiences: string[];
+    sdkLanguage: generatorsYml.GenerationLanguage | undefined;
```
dsinghvi, line 39: why do we add the sdk language here ? we should just pass it from the generator invocation

packages/cli/workspace-loader/src/types/Workspace.ts
```diff
@@ -36,7 +36,9 @@ export interface Spec {
...
     shouldUseTitleAsName: boolean;
+    shouldUseUndiscriminatedUnionsForDiscriminated: boolean;
```
dsinghvi, line 41: thoughts on a better name here? just brutal for a user to add and we actually need to roll this out broadly at vocode, vellum, all new demos etc

packages/cli/configuration/src/generators-yml/schemas/APIConfigurationSchema.ts
```diff
@@ -14,6 +14,11 @@ export const APIDefinitionSettingsSchema = z.object({
         .optional(z.boolean())
         .describe(
             "Whether to use the titles of the schemas within an OpenAPI definition as the names of the types within Fern. Defaults to true."
+        ),
+    "use-undiscriminated-unions-with-literals": z
```
dsinghvi, line 18: nevermind this looks like a reasonable name

dsinghvi: might be good to just use this across the whole codebase

packages/cli/openapi-parser/src/schema/SchemaParserContext.ts
```diff
@@ -6,11 +7,20 @@ export interface SchemaParserContext {
...
     shouldUseTitleAsName: boolean;
 
     referenceExists(ref: string): boolean;
     resolveSchemaReference(schema: OpenAPIV3.ReferenceObject): OpenAPIV3.SchemaObject;
     markSchemaAsReferencedByNonRequest(schemaId: SchemaId): void;
     markSchemaAsReferencedByRequest(schemaId: SchemaId): void;
     markReferencedByDiscriminatedUnion(schema: OpenAPIV3.ReferenceObject, discrminant: string, times: number): void;
+
+    shouldUseUndiscriminatedUnionsForDiscriminated: boolean;
+    getShouldUseUndiscriminatedUnionsForDiscriminated(): boolean;
```
dsinghvi, line 20: nit: why do we need both of these functions and not just one?

packages/cli/openapi-parser/src/schema/SchemaParserContext.ts
```diff
@@ -6,11 +7,20 @@ export interface SchemaParserContext {
...
+    storeDiscriminatedUnionMetadata(
+        schema: OpenAPIV3.ReferenceObject,
+        discrminant: string,
+        discriminantValue: string
+    ): void;
```
dsinghvi, line 25: thoughts on making this very specific? 
`markSchemaWithDiscriminant()`  and add docs that we will append literal values here

packages/cli/openapi-parser/src/schema/SchemaParserContext.ts
armandobelardo: ones the flag and the other is a function which evals if it should beused (really just checks the flag as well as the language)

packages/cli/cli/src/cli.ts
```diff
@@ -872,20 +878,27 @@ function addWriteDefinitionCommand(cli: Argv<GlobalCliOptions>, cliContext: CliC
...
         async (argv) => {
             cliContext.instrumentPostHogEvent({
                 command: "fern write-definition"
             });
             await writeDefinitionForWorkspaces({
                 project: await loadProjectAndRegisterWorkspacesWithContext(cliContext, {
                     commandLineApiWorkspace: argv.api,
-                    defaultToAllApiWorkspaces: true
+                    defaultToAllApiWorkspaces: true,
+                    sdkLanguage: argv.language
```
dsinghvi, line 898: should we delete `sdkLanguage` from here now that its no longer reequired

dsinghvi:
@armandobelardo cloning this now to play around, will merge after

PR #3741, opened 2024-05-30
java, improvement: run seed faster using local mode

dcb6:
This PR improves seed performance by allowing you to run java seed in local mode rather than in Docker. As part of this change, we are now gitignoring certain generated gradle files that are difficult to match exactly when running locally.


seed/java-sdk/seed.yml
```diff
@@ -7,6 +7,15 @@ dockerCommand:
 language: java
 generatorType: SDK
 defaultOutputMode: github
```
dcb6, line 9: key changes in this file

dsinghvi (approved):

PR #3749, opened 2024-05-31
fix: address a number of papercuts in the mock server and python unit tests

armandobelardo:
This PR:
- Plumbs through the optional type on examples
- Fixes path ordering in the express server to avoid path registration conflicts
- Generates text responses correctly on the example generation side
- Fixes oauth tests


packages/cli/mock/src/runMockServer.ts
```diff
@@ -37,33 +36,95 @@ export class MockServer {
...
-                        assertNever(endpoint.method);
                 }
             }
         }
     }
 
+    // Sort the paths with a priority for literal path parts over parameters
+    // e.g. /movie/:movieId should come after /movie/123, such that express
+    // can match the exact endpoint before the variable one (/movie/:movieId)
+    sortExpressPaths(
```
amckinney, line 83: nit: Should we mark this `private`?
```suggestion
    private sortExpressPaths(
```

packages/cli/mock/src/runMockServer.ts
```diff
@@ -37,33 +36,95 @@ export class MockServer {
...
+
+                const existingGroup = endpointGroups.get(endpointPath) ?? new Map();
+                const existingEndpoints = existingGroup.get(endpoint.method) ?? [];
+
+                endpointGroups.set(endpointPath, existingGroup.set(endpoint.method, [...existingEndpoints, endpoint]));
+            }
+        }
+
+        const listGroups = Array.from(endpointGroups);
+        const sortedEndpoints = listGroups.sort(this.sortExpressPaths);
```
amckinney, line 55: nit: It might be worth making a private function devoted to creating the `sortedEndpoints` variable (i.e. the `for` loop above).

amckinney (approved):
lgtm - only nits, take em or leave em

PR #3751, opened 2024-05-31
(chore, go): Update docker builds to consume private modules

amckinney:
This updates the Go docker images so that we can consume private GitHub repositories.


amckinney:
This approach won't work well for our local development environments.

PR #3754, opened 2024-05-31
feat: skip urls in tabs, and scaffold tab changelog

abvthecity:


abvthecity:
Replaced by https://github.com/fern-api/fern/pull/3780

PR #3756, opened 2024-05-31
improvement: add literal example type and add id to example

armandobelardo:


packages/cli/ete-tests/src/tests/ir/__snapshots__/ir.test.ts.snap
```diff
@@ -30011,6 +30035,7 @@ exports[`ir {"name":"simple"} 1`] = `
                         },
                         {
                             "exampleType": "userProvided",
+                            "id": null,
```
dsinghvi, line 30038: @armandobelardo im guessing we want ids here as well

packages/cli/generation/ir-generator/src/examples/ExampleGenerator.ts
```diff
@@ -186,6 +186,7 @@ export class ExampleGenerator {
     ): Omit<HttpEndpointExample, "response" | "type" | "_visit" | "exampleType"> {
         const examples = endpoint.examples;
         return {
+            id: "5de04949-6f1b-4560-b4d4-c454c473cf33",
```
dsinghvi, line 189: are we intentionally hardcoding this?

packages/cli/ete-tests/src/tests/ir/__snapshots__/ir.test.ts.snap
armandobelardo: I don't think we want to generate for user provided right? Cause then it changes with every generation

So I made it exposed in the yaml schema and a user can specify the id

PR #3757, opened 2024-05-31
java: make sure oauth gated properly

dcb6:
This change checks generator config to ensure that oauth is only generated if the user has permission. If not, we replace this auth scheme with a default bearer token auth scheme.

As part of this change, we now store a `resolvedAuthSchemes` in the context and read from this rather than IR in our code paths.

### DIFF
--- a/generators/java/generator-utils/src/main/java/com/fern/java/FeatureResolver.java
+++ b/generators/java/generator-utils/src/main/java/com/fern/java/FeatureResolver.java
@@ -60,7 +60,8 @@ public class FeatureResolver {
         if (maybeOAuthScheme.isEmpty() || generatorConfig.getGenerateOauthClients()) return schemes;
         generatorExecClient.sendUpdate(GeneratorUpdate.log(LogUpdate.builder()
                 .level(LogLevel.ERROR)
-                .message("OAuth is not supported in your current Java SDK plan; falling back to bearer auth. Please reach out to the Fern team!")
+                .message(
+                        "OAuth is not supported in your current Java SDK plan; falling back to bearer auth. Please reach out to the Fern team!")
                 .build()));
         List<AuthScheme> resolvedSchemes = schemes.stream()
                 .filter(authScheme -> !authScheme.isOauth() && !authScheme.isBearer())


generators/java/sdk/src/main/java/com/fern/java/client/Cli.java
```diff
@@ -198,19 +198,21 @@ public GeneratedRootClient generateClient(ClientGeneratorContext context, Interm
         generatedTypes.getTypes().values().forEach(this::addGeneratedFile);
         generatedTypes.getInterfaces().values().forEach(this::addGeneratedFile);
 
-        Optional<OAuthScheme> maybeOAuthScheme = context.getIr().getAuth().getSchemes().stream()
-                .map(AuthScheme::getOauth)
-                .flatMap(Optional::stream)
-                .findFirst();
+        Optional<GeneratedJavaFile> generatedOAuthTokenSupplier = Optional.empty();
+        if (context.getGeneratorConfig().getGenerateOauthClients()) {
+            Optional<OAuthScheme> maybeOAuthScheme = context.getIr().getAuth().getSchemes().stream()
```
dcb6, line 203: should i throw if someone has oauth configured but `context.getGeneratorConfig().getGenerateOauthClients()` is false?

as written, just doesn't generate the oauth stuff

dsinghvi: good question i think not generating is okay or falling back to bearer, @amckinney what do you do in the go / ts generators?

dcb6: looks to me like we don't throw in TS:
https://github.com/fern-api/fern/blob/main/generators/typescript/sdk/generator/src/SdkGenerator.ts#L382

couldn't find code falling back to bearer but not 100%, would need @amckinney to confirm

amckinney: Discussed offline - 

We'll go with falling back to bearer in all generators. We'll follow-up with test cases / additional handling in all of the generators that support OAuth.

generators/java/generator-utils/src/main/java/com/fern/java/FeatureResolver.java
```diff
@@ -0,0 +1,76 @@
...
+    public List<AuthScheme> getResolvedAuthSchemes() {
+        List<AuthScheme> schemes = ir.getAuth().getSchemes();
+        Optional<OAuthScheme> maybeOAuthScheme = schemes.stream()
+                .map(AuthScheme::getOauth)
+                .flatMap(Optional::stream)
+                .findFirst();
+        if (maybeOAuthScheme.isEmpty() || generatorConfig.getGenerateOauthClients()) return schemes;
+        generatorExecClient.sendUpdate(GeneratorUpdate.log(LogUpdate.builder()
+                .level(LogLevel.ERROR)
+                .message("OAuth is not supported in your current Java SDK plan. Please reach out to the Fern team!")
```
dcb6, line 68: Could use pointers on copy here

amckinney: We might also want to include a message that describes we're using bearer instead.
```suggestion
                .message("OAuth is not supported in your current Java SDK plan; falling back to bearer auth. Please reach out to the Fern team!")
```

amckinney (approved):
Looks great!

PR #3764, opened 2024-06-03
wip: introduce oauth clients into Ruby generator

armandobelardo:


armandobelardo:
Implemented in https://github.com/fern-api/fern/pull/3842

PR #3769, opened 2024-06-03
(feature): Add keywords configuration

amckinney:
This adds a new `keywords` configuration option to the generator invocation. With this, users can now specify their own custom list of keywords that the language should treat as reserved. If any name in their API matches this type, it will be handled in a safe, language-specific manner.

An example `generators.yml` is shown below:

```yaml
  version: 2.8.0
  keywords:
    - True
    - False
    ...
```

Note that the `keywords` configuration is not intended to be used in general because each generator takes care of this automatically. This feature is primarily used to preserve backwards compatibility whenever the list of keywords changes.

### DIFF
--- a/packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
+++ b/packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
@@ -24,15 +24,13 @@ export function constructCasingsGenerator({
     keywords: string[] | undefined;
     smartCasing: boolean;
 }): CasingsGenerator {
-    const keywordSet: Set<string> | undefined = keywords != null ? new Set(keywords) : undefined;
     const casingsGenerator: CasingsGenerator = {
         generateName: (name, opts) => {
             const generateSafeAndUnsafeString = (unsafeString: string): SafeAndUnsafeString => ({
                 unsafeName: unsafeString,
-                safeName: sanitizeNameForLanguage({
+                safeName: sanitizeName({
                     name: unsafeString,
-                    generationLanguage,
-                    keywords: keywordSet
+                    keywords: getKeywords({ generationLanguage, keywords })
                 })
             });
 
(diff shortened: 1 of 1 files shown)


packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
```diff
@@ -17,16 +17,23 @@ const CAPITALIZE_INITIALISM: generatorsYml.GenerationLanguage[] = ["go", "ruby"]
...
 export function constructCasingsGenerator({
     generationLanguage,
+    keywords,
     smartCasing
 }: {
     generationLanguage: generatorsYml.GenerationLanguage | undefined;
+    keywords: string[] | undefined;
     smartCasing: boolean;
 }): CasingsGenerator {
+    const keywordSet: Set<string> | undefined = keywords != null ? new Set(keywords) : undefined;
```
dsinghvi, line 27: nit: its probably easier to read the code if it was: 
```ts
const keywordSet: Set<string> = keywords != null ? new Set(keywords) : RESERVED_KEYWORDS[generationLanguage];
```

otherwise its less apparent that we _do_ actually use predefined reserved keywords

dsinghvi (approved):

packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
amckinney: Good call - it's possible `generationLanguage` is undefined at this layer, so I'll refactor this with a helper instead:

```ts
function getKeywords({
    generationLanguage,
    keywords
}: {
    generationLanguage: generatorsYml.GenerationLanguage | undefined;
    keywords: string[] | undefined;
}): Set<string> | undefined {
    if (keywords != null) {
        return new Set(keywords);
    }
    if (generationLanguage != null) {
        return RESERVED_KEYWORDS[generationLanguage];
    }
    return undefined;
}
```

PR #3777, opened 2024-06-04
improvement: add casing version and allow for "preserved" casing words

armandobelardo:


packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
```diff
@@ -37,8 +39,52 @@ export function constructCasingsGenerator({
...
             let snakeCaseName = snakeCase(name);
-            const camelCaseWords = words(camelCaseName);
-            if (smartCasing) {
+
+            // generatorsYml.CasingVersion.V0 should effectively be treated as smartCasing
+            if (
+                smartCasing ||
+                casingVersion === generatorsYml.CasingVersion.V0 ||
+                casingVersion === generatorsYml.CasingVersion.V1
+            ) {
```
dsinghvi, line 48: i would have thought that `v0` would not default to smart casing? 

Some context on the PR about what is `v0` and what is `v1` would be helpful

packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts
```diff
@@ -18,11 +18,13 @@ const CAPITALIZE_INITIALISM: generatorsYml.GenerationLanguage[] = ["go", "ruby"]
...
     generationLanguage,
     keywords,
-    smartCasing
+    smartCasing,
+    casingVersion
 }: {
     generationLanguage: generatorsYml.GenerationLanguage | undefined;
     keywords: string[] | undefined;
     smartCasing: boolean;
+    casingVersion: generatorsYml.CasingVersion | undefined;
```
dsinghvi, line 27: do you think we can delete smartCasing altogether and just rely on `casingVersion` now  here? 

i would have expected the code to read something like: 
```ts
switch (version) {
  case "v0": 
    // do a vanilla case
  case "v1": 
    // handle initialisms and numbers
  default: 
    assertNever(version)
}
```

armandobelardo: So you want to have a yaml migration to move smartCasing to v0?

dsinghvi: I think internally in the codebase we can have `casingVersion` (in `ir-generator` for example), but externally we can listen to version first and then fall back to smart-casing -> v1

dsinghvi: so no yaml migration needed

PR #3778, opened 2024-06-04
(docker): Install node in every generator image

amckinney:
In preparation for calling the `generator-cli` from within every generator, this updates all the generator docker images to install `node` and `npm`. Note that images that use `alpine` base images (e.g. Go, C#) use different installation instructions than those that do not (e.g. Python, Java).

This also includes the beginning of a `generator-cli` installer in Go to verify that the command is actually installed. The `generator-cli` behavior is a no-op for now.


amckinney:
We'll add this on a case-by-case basis for each generator as needed.

PR #3779, opened 2024-06-04
java, improvement: error types

dcb6:
We now generate Exception types for all errors that are defined in the IR. Generated clients with an error discrimination strategy of "status code" will throw one of these typed Exceptions based on the status code of error responses. 

Example error type:
  ```java
  public final class MyCompanyAuthBadRequest extends MyCompanyAuthApiError {
    public MyCompanyAuthBadRequest(Object body) {
        super("BadRequest", 400, body);
    }
  }
  ```
Suggest looking at `CustomAuthClient.java` to see the changes and their variations.

### DIFF
--- a/seed/java-sdk/alias/src/main/java/com/seed/alias/core/SeedAliasApiError.java
+++ b/seed/java-sdk/alias/src/main/java/com/seed/alias/core/SeedAliasApiError.java
@@ -3,30 +3,43 @@
  */
 package com.seed.alias.core;
 
+/**
+ * This exception type will be thrown for any non-2XX API responses.
+ */
 public class SeedAliasApiError extends SeedAliasError {
-    private final String message;
-
+    /**
+     * The error code of the response that triggered the exception.
+     */
     private final int statusCode;
 
+    /**
+     * The body of the response that triggered the exception.
+     */
     private final Object body;
 
     public SeedAliasApiError(String message, int statusCode, Object body) {
         super(message);
-        this.message = message;
         this.statusCode = statusCode;
         this.body = body;
     }
 
+    /**
+     * @return the statusCode
+     */
     public int statusCode() {
         return this.statusCode;
     }
 
+    /**
+     * @return the body
+     */
     public Object body() {
         return this.body;
     }
 
     @java.lang.Override
     public String toString() {
-        return "SeedAliasApiError{" + "message: " + message + ", statusCode: " + statusCode + ", body: " + body + "}";
+        return "SeedAliasApiError{" + "message: " + getMessage() + ", statusCode: " + statusCode + ", body: " + body
+                + "}";
     }
 }
--- a/seed/java-sdk/alias/src/main/java/com/seed/alias/core/SeedAliasError.java
+++ b/seed/java-sdk/alias/src/main/java/com/seed/alias/core/SeedAliasError.java
@@ -3,6 +3,9 @@
  */
 package com.seed.alias.core;
 
+/**
+ * This class serves as the base exception for all errors in the SDK.
+ */
 public class SeedAliasError extends RuntimeException {
     public SeedAliasError(String message) {
         super(message);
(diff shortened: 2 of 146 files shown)


seed/java-sdk/custom-auth/src/main/java/com/seed/customAuth/resources/errors/errors/SeedCustomAuthUnauthorizedRequest.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public SeedCustomAuthUnauthorizedRequest(UnauthorizedRequestErrorBody body) {
+        super("UnauthorizedRequest", 401, body);
+        this.body = body;
+    }
+
+    @Override
+    public UnauthorizedRequestErrorBody body() {
+        return this.body;
+    }
```
dsinghvi, line 20: @dcb6 what happens when you `toString` one of these errors?

dsinghvi: or `System.out.printlin(e)`

dcb6: Same thing right? 
```java
    @Test
    public void test() {
        // Add tests here and mark this file in .fernignore
        UnauthorizedRequestErrorBody body = UnauthorizedRequestErrorBody.builder().message("Oh no!").build();
        SeedCustomAuthUnauthorizedRequest error = new SeedCustomAuthUnauthorizedRequest(body);
        System.out.println(error);
        assert true;
    }
```
prints:
```
SeedCustomAuthApiError{statusCode: 401, body: {
  "message" : "Oh no!"
}}
```

dcb6: was wondering if we should just do a `writeValueAsString` with an object mapper instead?

dsinghvi (approved):

seed/java-sdk/custom-auth/src/main/java/com/seed/customAuth/resources/errors/errors/SeedCustomAuthUnauthorizedRequest.java
dcb6: I think the best way to improve the `toString` is just to add a message field like so:
```java
public class SeedCustomAuthApiError extends SeedCustomAuthError {

    private final String message;
    private final int statusCode;

    private final Object body;

    public SeedCustomAuthApiError(String message, int statusCode, Object body) {
        super(message);
        this.message = message;
        this.statusCode = statusCode;
        this.body = body;
    }

    public int statusCode() {
        return this.statusCode;
    }

    public Object body() {
        return this.body;
    }

    @java.lang.Override
    public String toString() {
        return "SeedCustomAuthApiError{" + "statusCode: " + statusCode + ", body: " + body +", message: " + message + "}";
    }
}
```

Not exactly sure how i'd cleanly refactor the `createToStringMethod` here to instead call getMessage() if we wanted to go that route: https://github.com/fern-api/fern/blob/dc03bc01865cfdeb9640b4b88d5dae106f0c36da/generators/java/generator-utils/src/main/java/com/fern/java/ObjectMethodFactory.java#L87

dcb6: `SeedCustomAuthError` as is prints`com.seed.customAuth.core.SeedCustomAuthError: Oh no!` for example (roughtly, `<className> <message>`)

so i don't think any improvements are needed there (i believe you mentioned adding a `message` field)

dsinghvi (approved):

PR #3804, opened 2024-06-06
Add Extra Field Support for FastAPI

jmedway614:
Added extra field support for fastapi. Set default to "forbid" which preserves current functionality. To enable custom fields we will need to request customers to set "extra_fields" to "allow"

### DIFF
--- a/generators/python/src/fern_python/generators/pydantic_model/custom_config.py
+++ b/generators/python/src/fern_python/generators/pydantic_model/custom_config.py
@@ -24,7 +24,3 @@ class PydanticModelCustomConfig(BasePydanticModelCustomConfig):
     use_str_enums: bool = True
     # Skip validation of fields (automatically includes additional fields)
     skip_validation: bool = False
-
-
-class FastApiPydanticModelCustomConfig(BasePydanticModelCustomConfig):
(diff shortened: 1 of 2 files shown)


generators/python/src/fern_python/generators/pydantic_model/custom_config.py
```diff
@@ -24,3 +24,7 @@ class PydanticModelCustomConfig(BasePydanticModelCustomConfig):
     use_str_enums: bool = True
     # Skip validation of fields (automatically includes additional fields)
     skip_validation: bool = False
+
+
+class FastApiPydanticModelCustomConfig(BasePydanticModelCustomConfig):
```
armandobelardo, line 29: Mind putting this in the fastapi config file? Just so all the fastapi config is in one place (it's also how we do it with SDKs today)

armandobelardo (approved):
LGTM! Thank you!

generators/python/src/fern_python/generators/pydantic_model/custom_config.py
jmedway614: Makes sense! Done

PR #3805, opened 2024-06-06
improvement: support pydantic v2 outright

armandobelardo:
Because the seed diff is large, it is in another PR: https://github.com/fern-api/fern/pull/4069

--

Instead of leveraging pydantic.v1 when in pydantic v2, we now support pydantic v2 outright and use all the configuration and functionality in v2 as opposed to v2
This most importantly has our models leverage v2 models when in pydantic v2 which unblocks compatibility issues with FastAPI and pydantic.v1

A follow up to this PR will be to appropriately respect the pydantic version config instead of using `IS_PYDANTIC_V2` everywhere. This should be essentially making copies of `pydantic_utilities.py` as well as `unchecked_base_model.py` and then just copying over the right set of files based on the specified compatibility. 
(there are a couple references outside of these two files, which we should investigate folding into here if possible, though that might be hard)

Some breaking changes for simplicity:
- remove wrapped aliases
- remove custom root validators for models with a `__root__` type (e.g. only unions with utils still leverages root models)
- public fields previously prefixed with `_` are now prefixed with `f_` (pydantic does not allow for _ prefixes and python does not allow for a numeric prefix)


dsinghvi (approved):

generators/python/src/fern_python/generators/pydantic_model/fern_aware_pydantic_model.py
```diff
@@ -197,65 +190,42 @@ def add_method(
...
-        self._override_json()
-        self._override_dict()
-        if self._model_contains_forward_refs:
+        if self._model_contains_forward_refs or self._force_update_forward_refs:
             self._pydantic_model.update_forward_refs()
         self._pydantic_model.finish()
 
     def _get_validators_generator(self) -> ValidatorsGenerator:
-        root_type = self._pydantic_model.get_root_type()
-        if root_type is not None:
```
armandobelardo, line 244: FER-2480: This will need to be added back for allowing root overrides

generators/python/src/fern_python/generators/pydantic_model/type_declaration_handler/discriminated_union/discriminated_union_with_utils_generator.py
```diff
@@ -1,9 +1,13 @@
+from re import A
```
armandobelardo, line 1: FER-2479: This can be greatly simplified if we respect the Pydantic version config

PR #3813, opened 2024-06-07
fix: unchecked base model respects dicts as well as objects

armandobelardo:


dsinghvi (dismissed):
I know we had some tests for `unchecked_base_model.py`, can we add a unit test for this edge case?

amckinney (approved):

dsinghvi (approved):

PR #3822, opened 2024-06-10
(fix, python): Unions with single element and/or no properties

amckinney:
This fixes a couple small issues with the Python union generator:
1. Unions that included types with no properties are now correctly generated.
2. Unions that define a single type now use a valid type alias (rather than an invalid `typing.Union` with a single element).


armandobelardo (approved):

test-definitions/fern/apis/unions/definition/types.yml
```diff
@@ -64,6 +64,10 @@ types:
       date: optional<date>
       dateimte: optional<datetime>
 
+  UnionWithSingleElement:
+    union:
+      foo: Foo
```
dsinghvi, line 69: 🔥 thanks for adding the test

PR #3830, opened 2024-06-10
(fix, openapi): Handle more `allow-multiple` oneOf cases

amckinney:
This updates the OpenAPI parser to better handle query parameters that can either be a single element or a list of elements. For example,

```json
{
  "in": "query",
  "name": "user_id",
  "schema": {
    "anyOf": [
      {
        "type": "string"
      },
      {
        "type": "array",
        "items": {
          "type": "string"
        }
      }
    ]
  }
}
```

For now, we special case this so that we set the `allow-multiple` field so that the caller will always be able to set multiple elements (in the `explode` format).


packages/cli/openapi-ir-to-fern/src/buildQueryParameter.ts
```diff
@@ -202,6 +246,50 @@ function getQueryParameterTypeReference({
...
+                                nameOverride: schema.nameOverride,
+                                generatedName: schema.generatedName,
+                                value: firstSchema,
+                                description: schema.description,
+                                groupName: undefined
+                            }),
+                            context,
+                            fileContainingReference
+                        }),
+                        allowMultiple: true
```
dsinghvi, line 288: should we be checking the [explode](https://github.com/fern-api/fern/blob/512b16a989ccd08a2cb0b1b0318a815bfa5aa13c/packages/cli/openapi-parser/src/__test__/fixtures/hume/openapi.yaml#L99) key in OAI ?

dsinghvi (approved):
1 comment but approving so you can merge/release whenever it makes sense

packages/cli/openapi-ir-to-fern/src/buildQueryParameter.ts
amckinney: Discussed offline - 

It looks like the current handling was primarily meant to handle comma-separated query parameters, so we should definitely gate this feature on the `explode` key (even if it requires editing OpenAPI schemas to specify this explicitly). Otherwise we would most likely break a variety of other use cases.

I'll follow-up with that change here before merging.

amckinney: On second thought, it looks like this is the default behavior for _any_ query parameter array (e.g. [reference](https://github.com/fern-api/fern/blob/512b16a989ccd08a2cb0b1b0318a815bfa5aa13c/packages/cli/openapi-ir-to-fern/src/buildQueryParameter.ts#L98)). Either way, we'll want to revisit this once we actually support different query parameter schemes (e.g. comma-separated, explode, etc).

PR #3833, opened 2024-06-11
[FER-1986] Fix two DiscriminatedUnion bugs in dynamic Typescript snippets

ppod1991:
Address two DiscriminatedUnion bugs in the dynamic Typescript snippets for Cohere:

1) When you try to generate snippets using discriminated unions in TypeScript, it prefixes them with ' which causes the code to be invalid. This removes the extra `
2) Prevent extra nesting of subunion types which caused invalid JSON in the snippet

Will be followed up with a PR updating https://github.com/fern-api/fern-platform/blob/main/packages/template-resolver/src/__test__/unit-tests/__snapshots__/cohere.test.ts.snap


generators/typescript/sdk/generator/src/TemplateGenerator.ts
```diff
@@ -291,7 +295,9 @@ export class TemplateGenerator {
         }
         return FdrSnippetTemplate.Template.generic({
             imports: [],
-            templateString: this.getAsNamedParameterTemplate(name, `{\n${childTabs}${TEMPLATE_SENTINEL}\n${selfTabs}}`),
+            templateString: isUnionType
+                ? this.getAsNamedParameterTemplate(name, `${TEMPLATE_SENTINEL}\n`)
```
armandobelardo, line 299: Do you not need to persist the tabs here?
I think this might need to be something like  `this.getAsNamedParameterTemplate(name, ${childTabs}${TEMPLATE_SENTINEL}\n)`

generators/typescript/sdk/generator/src/TemplateGenerator.ts
```diff
@@ -291,7 +295,9 @@ export class TemplateGenerator {
...
+                : this.getAsNamedParameterTemplate(name, `{\n${childTabs}${TEMPLATE_SENTINEL}\n${selfTabs}}`),
```
dsinghvi, line 300: is it worth adding a comment here to understand why the template string changes if it is a union? Definitely a little hard to grok what's happening here

generators/typescript/sdk/generator/src/TemplateGenerator.ts
```diff
@@ -488,14 +498,16 @@ export class TemplateGenerator {
...
+        indentationLevel,
+        isUnionType
     }: {
         typeReference: TypeReference;
         name: string | undefined;
         location: FdrSnippetTemplate.PayloadLocation;
         wireOrOriginalName: string | undefined;
         nameBreadcrumbs: string[] | undefined;
         indentationLevel: number;
+        isUnionType?: boolean;
```
armandobelardo, line 510: Nit: mind making this `| undefined` as opposed to `?`? Just makes it harder to miss passing it in!

generators/typescript/sdk/generator/src/TemplateGenerator.ts
```diff
@@ -258,14 +260,16 @@ export class TemplateGenerator {
...
+        indentationLevel,
+        isUnionType
     }: {
         otd: ObjectTypeDeclaration;
         name: string | undefined;
         location: FdrSnippetTemplate.PayloadLocation;
         wireOrOriginalName: string | undefined;
         nameBreadcrumbs: string[] | undefined;
         indentationLevel: number;
+        isUnionType: boolean;
```
armandobelardo, line 272: Another nit: you could call this `isObjectInlined` instead of `isUnionType` and that could add some clarity as to what's happening (in addition to a comment)

generators/typescript/sdk/generator/src/TemplateGenerator.ts
ppod1991: Made this required instead!

generators/typescript/sdk/generator/src/TemplateGenerator.ts
ppod1991: I don't think we do since the value will be on the same line rather than the next

armandobelardo: You should run some tests! I think you might have to since this template corresponds to multiple properties which will be on separate lines

dsinghvi: hey @armandobelardo -- it doesn't look like this breaks existing unions. since we need to move a little fast on this, planning to fix in a follow up and release an rc of the generator

PR #3834, opened 2024-06-11
(fix, openapi): fall back to default status code if none provided

dsinghvi:
The OpenAPI parser now handles the `default` status code

In OpenAPI, you can use `default` to cover all unlisted status codes. The example below shows how `default` can be used as an error schema. 
Screenshot 2024-06-11 at 3 43 49 PM

If you specify default with no other status codes, its technically both a success + error schema. This PR handles this case.

### DIFF
--- a/packages/cli/ete-tests/src/tests/dependencies/__snapshots__/dependencies.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/dependencies/__snapshots__/dependencies.test.ts.snap
@@ -841,4 +841,4 @@ exports[`dependencies correctly incorporates dependencies 1`] = `
 }"
 `;
 
-exports[`dependencies file dependencies 1`] = `1400746`;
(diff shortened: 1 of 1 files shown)


packages/cli/openapi-parser/src/openapi/v3/converters/endpoint/convertResponse.ts
```diff
@@ -13,7 +13,7 @@ import { getApplicationJsonSchemaMediaObject, getSchemaMediaObject } from "./get
 
 // The converter will attempt to get response in priority order
 // (i.e. try for 200, then 201, then 204)
-const SUCCESSFUL_STATUS_CODES = ["200", "201", "204"];
+const SUCCESSFUL_STATUS_CODES = ["200", "201", "204", "default"];
```
dsinghvi, line 16: this is the singular change

amckinney (approved):

PR #3836, opened 2024-06-11
(fix, go): Handle deepObject query parameter arrays

amckinney:
This fixes an issue in the Go SDK and adds support for serializing an array of struct types in the `deepObject` query parameter format. The format is now consistent with the TypeScript SDK (and all others).

For posterity, an example of a `deepObject` array is shown below:

```go
  user, err := client.Users.Get(
    ...,
    acme.GetUserRequest{
      Filters: []*acme.Filter{
        {
          Key: "age",
          Value: "42",
        },
        {
          Key: "firstName",
          Value: "john",
        },
      },
    },
  )
  // Serialized as: ?filters[key]=age&filters[key]=firstName&filters[value]=42&filters[value]=john
```

### DIFF
--- a/generators/go/sdk/CHANGELOG.md
+++ b/generators/go/sdk/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
       },
     },
   )
-  // The query string is serialized as: ?filters[key]=hello&filters[key]=firstName&filters[value]=42&filters[value]=john
+  // The query string is serialized as: ?filters[key]=age&filters[key]=firstName&filters[value]=42&filters[value]=john
   ```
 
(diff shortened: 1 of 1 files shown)


generators/go/sdk/CHANGELOG.md
```diff
@@ -7,6 +7,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+          Value: "42",
+        },
+        {
+          Key: "firstName",
+          Value: "john",
+        },
+      },
+    },
+  )
+  // The query string is serialized as: ?filters[key]=hello&filters[key]=firstName&filters[value]=42&filters[value]=john
```
dsinghvi, line 31: chat from offline: lets put this into the PR description so we can track this in the future

dsinghvi (approved):

PR #3843, opened 2024-06-13
fix: image path parsing from markdown considers MDX children

abvthecity:
**Before this PR:** image path parser ignores images that are embedded within markdown that's a child of JSX elements.

This PR adds additional test cases that make the parser more robust.


abvthecity:
Approved 👍

armandobelardo (approved):
Approved 👍

PR #3845, opened 2024-06-13
java, feature: pagination

dcb6:
Add support for cursor and offset pagination. 

  For example, consider the following endpoint `/users` endpoint:

  ```yaml
  types:
    User:
      properties:
        name: string

    ListUserResponse:
      properties:
        next: optional<string>
        data: list<User>

  service:
    auth: false
    base-path: /users
    endpoints:
      list:
        path: ""
        method: GET
        pagination:
          cursor: $request.starting_after
          next_cursor: $response.next
          results: $response.data
        request:
          name: ListUsersRequest
          query-parameters:
            starting_after: optional<string>
        response: ListUsersResponse
  ```
  
  The generated `SyncPagingIterable<User>` can then be used to traverse through the `User` objects:
  ```java
  for (User user : client.users.list(...)) {
      System.out.println(user);
  }
  ```
  
  Or stream them:
  ```java
  client.users.list(...).streamItems().map(user -> ...);
  ```
  
  Or statically calling `nextPage()` to perform the pagination manually:
  ```java
  SyncPagingIterable<User> pager = client.users.list(...);
  // First page
  System.out.println(pager.getItems());
  // Second page
  pager = pager.nextPage();
  System.out.println(pager.getItems());
  ```

### DIFF
--- a/generators/java/sdk/CHANGELOG.md
+++ b/generators/java/sdk/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
         response: ListUsersResponse
   ```
   
-  The generated `SyncPagingIterable<User>` and then be used to traverse through the `User` objects:
+  The generated `SyncPagingIterable<User>` can then be used to traverse through the `User` objects:
   ```java
   for (User user : client.users.list(...)) {
       System.out.println(user);


generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/AbstractEndpointWriter.java
```diff
@@ -714,6 +881,229 @@ private boolean isAliasContainer(com.fern.irV42.model.types.TypeReference respon
...
+        private SnippetAndResultType(
+                com.fern.irV42.model.types.TypeReference typeReference, TypeName typeName, CodeBlock codeBlock) {
+            this.typeReference = typeReference;
+            this.typeName = typeName;
+            this.codeBlock = codeBlock;
+        }
+    }
+
+    private class NestedPropertySnippetGenerator
+            implements com.fern.irV42.model.types.TypeReference.Visitor<GetSnippetOutput> {
```
dcb6, line 914: This turned out very not elegant -- there may be a better way, but I couldn't figure it out

dsinghvi (approved):

PR #3849, opened 2024-06-14
(fix): handle code samples without accompanying examples

dsinghvi:
Currently the following fern definition fails fern check, because every example is required to have a structured yaml config describing what query, body, path params are required for the endpoint. In this PR, the goal is to relax that requirement so that users can only specify code samples if they wanted to.  
```yaml
service: 
  endpoints: 
    get: 
      path: /users
      ....
      examples: 
        - code-samples: 
             - language: typescript
               code: | 
                  var client = new Client();
```


packages/cli/yaml/yaml-schema/src/ast/visitors/services/visitHttpService.ts
```diff
@@ -299,6 +299,17 @@ async function visitExampleEndpointCall({
...
+    // if an example is entirely empty, just skip it
+    if (
+        example.headers == null &&
+        example["path-parameters"] == null &&
+        example["query-parameters"] == null &&
+        example.request == null &&
+        example.response == null
+    ) {
+        return;
+    }
```
dsinghvi, line 311: don't validate examples that are entirely empty

packages/cli/generation/ir-generator/src/converters/services/convertCodeSamples.ts
```diff
@@ -0,0 +1,52 @@
...
+            ExampleCodeSample.sdk({
+                name: sdkScheme.name != null ? file.casingsGenerator.generateName(sdkScheme.name) : undefined,
+                docs: sdkScheme.docs,
+                sdk: removeSdkAlias(sdkScheme.sdk),
+                code: sdkScheme.code
+            })
+    });
+}
+
+function removeSdkAlias(sdk: RawSchemas.SupportedSdkLanguageSchema): SupportedSdkLanguage {
```
amckinney, line 31: Are we referencing these cases from somewhere? It might be helpful to include a link to whatever markdown and/or code samples document we're (presumably) referencing. Otherwise they feel arbitrary and it's difficult to know how to update in the future.

I know this is just copy/pasted from another file in this change, so can handle this later if we don't know the original source.

amckinney (approved):
Discussed offline - general direction makes sense to me

PR #3864, opened 2024-06-17
docs: add fern definition display-name property

chdeskur:


fern/pages/api-definition/fern-definition/endpoints.mdx
```diff
@@ -58,7 +59,24 @@ The full path for the endpoint is the concatenation of:
 - The service `base-path`
 - The endpoint `path`
 
-### Path parameters
+## Display Name
+
+By default, the display name of an endpoint will be automatically generated. If you would like to customize the endpoint name, you can **set the display name**.
```
amckinney, line 64: It's unclear to the reader where the display name has any effect. I know this is the API definition documentation, but I wonder if it's worth either:

1. Linking to a docs page that shows how the display name is rendered
2. Including an image of the display name here (e.g. the image [here](https://buildwithfern.com/learn/docs/api-references/sdk-snippets)).

Also, can we speak to how the display name is generated by default? It might be simpler to say:

```
By default, the display name is equal to the 'Title Case' of the endpoint name. If you would like to ...
```

I'd also like to hear @dsinghvi's thoughts though.

chdeskur: Definitely agree - I've added your default description. 

Not sure what the best approach would be between the two options for showing the example? Image would be more descriptive, but we don't use them often in the docs. 

Maybe something like this: 
The display name will appear as the title of an endpoint [like this one](https://buildwithfern.com/learn/cli-api/api-reference/docs/generate).

dsinghvi (approved):

fern/pages/api-definition/fern-definition/endpoints.mdx
amckinney: That looks great to me!

PR #3869, opened 2024-06-18
Update overrides.mdx

dannysheridan:
* Improves the slug of overrides
* Removes slugs from the OpenAPI section in `docs.yml` and puts them in each page's frontmatter (cleaner IMO)

### DIFF
--- a/fern/pages/api-definition/openapi/extensions.mdx
+++ b/fern/pages/api-definition/openapi/extensions.mdx
@@ -1,7 +1,7 @@
 ---
 title: OpenAPI extensions for SDKs
 description: Learn about Fern's OpenAPI extensions for SDK method names, authentication overrides, global headers, enum descriptions and names, audiences, and more.
-slug: api-definition/openapi/extensions-sdks
+slug: api-definition/openapi/sdk-extensions
 ---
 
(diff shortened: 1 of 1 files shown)


dannysheridan:
@chdeskur do you think we should 

A/ split out OpenAPI extensions into two pages: Docs and SDKs?
B/ change the page title to include both?

This is for Docs, but the title says SDKs:
> api-definition/openapi/extensions-sdks#audiences-for-docs

chdeskur:
@dannysheridan 

Two pages: 
Pros: Someone developing SDKs doesn't have to navigate off the SDK tab to learn, can highlight specific information (display-name: relevant for Docs but not so much for SDKs)
Cons: Might have a lot of duplicated information, people developing both might feel like they're jumping around a lot

One page, different title: 
Pros: Avoid duplicating information, have one source of truth for all-things-overrides. If the page is titled well, it shouldn't be that hard to find/search for. 
Cons: The page is already quite long. I notice a lot of questions we get are often buried somewhere on this page. Might be hard for someone developing SDKs to find overrides information?

Based on this, I might lean towards two pages. Thoughts?

dannysheridan:
Let's do 2 pages. Want to make it happen?

chdeskur:
I can tackle this after the Syndicate migration

PR #3871, opened 2024-06-18
fix: python and ts generators only add publish block if they have cre…

armandobelardo:
…dentials


dsinghvi (approved):

seed/ts-sdk/variables/.github/workflows/ci.yml
```diff
@@ -28,24 +28,23 @@ jobs:
...
-
-      - name: Set up node
-        uses: actions/setup-node@v3
-
-      - name: Install dependencies
-        run: yarn install
-
-      - name: Build
-        run: yarn build
+        needs: [ compile, test ]
```
dsinghvi, line 32: i think the indenting may be off here

armandobelardo: fixed!

PR #3872, opened 2024-06-18
fix: validate files to be uploaded

trevorblades:
This PR is part one of two, ensuring that only files of specific types are uploaded to our system. This does file upload validation on the client side, but we must also implement some server-side validation.

Fixes FER-2070

### DIFF
--- a/packages/cli/yaml/docs-validator/src/rules/valid-file-types/valid-file-types.ts
+++ b/packages/cli/yaml/docs-validator/src/rules/valid-file-types/valid-file-types.ts
@@ -1,3 +1,4 @@
+import { doesPathExist } from "@fern-api/fs-utils";
 import { fileTypeFromBuffer } from "file-type";
 import { readFile } from "fs/promises";
 import isSvg from "is-svg";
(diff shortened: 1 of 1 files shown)


dsinghvi (commented):
@trevorblades thoughts on adding this to the `docs-validation` package ([link](https://github.com/fern-api/fern/tree/main/packages/cli/yaml/docs-validator/src/rules))? 

This way when a user runs `fern check` they will get a validation error that will automatically notify them that they can't upload those specific files.

trevorblades:
> @trevorblades thoughts on adding this to the `docs-validation` package ([link](https://github.com/fern-api/fern/tree/main/packages/cli/yaml/docs-validator/src/rules))?
> 
> This way when a user runs `fern check` they will get a validation error that will automatically notify them that they can't upload those specific files.

That's a good idea! I can do that. Does the docs validator run before docs publishing, and would its failure stop a publish command? What I'm getting at is if you see this as something that lives in the docs validator _and_ here in the publish command, or just in the validator.

dsinghvi:
> > @trevorblades thoughts on adding this to the `docs-validation` package ([link](https://github.com/fern-api/fern/tree/main/packages/cli/yaml/docs-validator/src/rules))?
> > This way when a user runs `fern check` they will get a validation error that will automatically notify them that they can't upload those specific files.
> 
> That's a good idea! I can do that. Does the docs validator run before docs publishing, and would its failure stop a publish command? What I'm getting at is if you see this as something that lives in the docs validator _and_ here in the publish command, or just in the validator.

Just in the validator works! the validator runs and has to be passed in order to hit publish

trevorblades:
@dsinghvi I moved the file type check to the validator CLI and added some tests to confirm it's working as expected. Lmk if that looks like what you had in mind!

dsinghvi (approved):
🚀 thank you!

packages/cli/yaml/docs-validator/src/rules/valid-file-types/valid-file-types.ts
```diff
@@ -27,16 +28,20 @@ export const ValidFileTypes: Rule = {
...
+
+                    if (!isValid) {
+                        return [
+                            {
+                                severity: "error",
+                                message: `File type of ${value} is invalid`
+                            }
+                        ];
+                    }
                 }
```
abvthecity, line 45: should we add a warning or error if the file doesn't exist?

dsinghvi: there's another rule that checks that!

PR #3886, opened 2024-06-19
(fix, csharp): ToString() Datetimes must be explicitly iso encoded

dsinghvi:
When calling `ToString()` on a `Datetime` in C#, there must be a `"o"` added to serialize it as an ISO 8601.


amckinney (approved):
Ah of course, `.ToString("o")` is obviously ISO format.

PR #3901, opened 2024-06-21
(feat, csharp): support sending bytes requests

dsinghvi:
The design for this in the SDK is: 
1) `bytes` parameters are typed as `Stream` so that users can pass in a file stream, byte stream, in memory byte stream, etc. 
2) The RawClient now accepts either a `JsonApiRequest` or a `StreamApiRequest` as input so each endpoint can pass in what they want

### DIFF
--- a/generators/csharp/sdk/src/endpoint/EndpointGenerator.ts
+++ b/generators/csharp/sdk/src/endpoint/EndpointGenerator.ts
@@ -115,7 +115,11 @@ export class EndpointGenerator {
         const parameters: csharp.Parameter[] = [];
         const service = this.context.getHttpServiceOrThrow(serviceId);
         const pathParameterReferences: Record<string, string> = {};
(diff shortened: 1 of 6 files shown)


generators/csharp/sdk/src/endpoint/EndpointGenerator.ts
```diff
@@ -38,6 +38,7 @@ export class EndpointGenerator {
         endpoint: HttpEndpoint;
         rawClientReference: string;
     }): csharp.Method {
+        console.log(`Generating endpoint id ${endpoint.id}`);
```
armandobelardo, line 41: nit: should this use the generator exec context logger? or should it be removed?

armandobelardo (approved):

generators/csharp/sdk/src/endpoint/EndpointGenerator.ts
dsinghvi: should be removed, my b

PR #3903, opened 2024-06-21
feat: changelog on tabs and sections

abvthecity:


packages/cli/configuration/src/docs-yml/parseDocsConfiguration.ts
```diff
@@ -115,7 +117,7 @@ export async function parseDocsConfiguration({
         pages,
 
         /* navigation */
-        tabs,
+        // tabs,
```
dsinghvi, line 120: is this intentional ?

dsinghvi (approved):

PR #3908, opened 2024-06-22
(fix, python): SDK doesn't leak `JSONDecodeError` to users

dsinghvi:


generators/python/fastapi/Dockerfile
```diff
@@ -6,10 +6,13 @@ ENV _TYPER_STANDARD_TRACEBACK=1
...
 COPY ./core_utilities/fastapi /assets/core_utilities
-COPY ./src ./src
 
 RUN pip3 install poetry &&\
   poetry config virtualenvs.create false &&\
   poetry install
 
+COPY ./src ./src
+
+RUN poetry install
```
dsinghvi, line 16: this was done to speed up docker builds and prevent reinstall of poetry deps again and again

armandobelardo: How does that do this if it was already doing 
copy -> install

dsinghvi: you dont have to reinstall dependencies if you change source code (thats the difference) and now we install twice

PR #3917, opened 2024-06-24
fix: validate files to be uploaded

trevorblades:
Reverts fern-api/fern#3904


dsinghvi (approved):

packages/cli/yaml/yaml-schema/src/docsAst/visitDocsConfigFileAst.ts
```diff
@@ -140,7 +177,8 @@ export async function visitDocsConfigFileYamlAst(
                     absoluteFilepathToConfiguration,
                     rawUnresolvedFilepath: version.path,
                     visitor,
-                    nodePath: ["versions", `${idx}`]
+                    nodePath: ["versions", `${idx}`],
+                    willBeUploaded: false
```
abvthecity, line 181: background image will be uploaded

packages/cli/yaml/yaml-schema/src/docsAst/visitDocsConfigFileAst.ts
```diff
@@ -18,6 +23,38 @@ export async function visitDocsConfigFileYamlAst(
...
+            absolutePathToFernFolder,
+            absolutePathToMdx: resolve(absolutePathToFernFolder, relativePath)
+        });
+
+        for (const filepath of filepaths) {
+            await visitor.filepath?.(
+                {
+                    absoluteFilepath: filepath,
+                    value: path.relative(absolutePathToFernFolder, filepath),
+                    willBeUploaded: true
```
abvthecity, line 51: markdown files don't get uploaded right to s3

packages/cli/yaml/docs-validator/src/rules/valid-file-types/valid-file-types.ts
```diff
@@ -0,0 +1,77 @@
...
+                    }
+                }
+
+                return [];
+            }
+        };
+    }
+};
+
+export const isValidFileType = async (file: Buffer): Promise<boolean> => {
```
abvthecity, line 63: we're still missing javascript and css files

abvthecity:
can you extend the ete test to include more test cases?

- css files
- js files
- files stored inside of markdown
- icons, logos, background images

packages/cli/yaml/yaml-schema/src/docsAst/visitDocsConfigFileAst.ts
trevorblades: This part of the code is looping over media files referenced in markdown, and running checks on their filepaths. It's my understanding that those are being uploaded to S3, so I've marked them as such. The markdown files themselves are handled elsewhere in this file.

packages/cli/yaml/docs-validator/src/rules/valid-file-types/valid-file-types.ts
trevorblades: Are we uploading JS and CSS files directly to S3?

Those files can't be checked the same way that we're checking binary files in this function, since they're just text. One solution would be to just say "if the file type checker can't derive a mime type from the file, let it pass". This would allow people to upload arbitrary PHP or other text-based files as well, but I can't think of a better solution at the moment.

dsinghvi: @trevorblades can we fallback to extension checking if it can't derive a mime type? not 100% fullproof but still better

packages/cli/yaml/yaml-schema/src/docsAst/visitDocsConfigFileAst.ts
abvthecity: sorry i think the code collapsing got me confused. this makes sense!

packages/cli/yaml/docs-validator/src/rules/valid-file-types/valid-file-types.ts
```diff
@@ -0,0 +1,83 @@
...
+    // otherwise, check the file type
+    const fileType = await fileTypeFromBuffer(file);
+    if (fileType) {
+        return ALLOWED_FILE_TYPES.includes(fileType.mime);
+    }
+
+    // if `fileType` is undefined, its type can't be parsed because it's likely a text file
+    if (path.extname(absoluteFilepath) === "js") {
+        // let JS files through
+        return true;
```
abvthecity, line 78: should we check if it's utf-8? https://github.com/hcodes/isutf8/blob/master/src/index.ts#L25

abvthecity (approved):

PR #3919, opened 2024-06-24
(fix, cli): examples dont print out where the missing property is

dsinghvi:
Previously if a nested property example was missing you would not get any information about where the property lives. Now you do. 

Before
Screenshot 2024-06-24 at 2 35 41 PM

After
Screenshot 2024-06-24 at 2 36 34 PM

### DIFF
--- a/packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
+++ b/packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
@@ -141,7 +141,7 @@ export function convertTypeExample({
                     exampleResolver,
                     file: fileContainingType,
                     workspace,
-                    breadcrumbs: [],
+                    breadcrumbs: []
                 });
                 if (violationsForMember.length === 0) {
                     return ExampleTypeShape.undiscriminatedUnion({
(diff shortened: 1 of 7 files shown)


packages/cli/yaml/validator/src/rules/valid-example-endpoint-call/__test__/valid-example-endpoint-call.test.ts
```diff
@@ -278,7 +278,7 @@ describe("valid-example-endpoint-call", () => {
                 severity: "error"
             },
             {
-                message: 'Example is missing required property "bar"',
+                message: 'Example is missing required property "request.bar"',
```
dsinghvi, line 281: Example of the change

packages/cli/yaml/validator/src/rules/valid-example-type/__test__/valid-example-type.test.ts
```diff
@@ -198,15 +198,13 @@ describe("valid-example-type", () => {
             {
                 severity: "error",
                 relativeFilepath: RelativeFilePath.of("object.yml"),
-                message:
-                    'Example is missing required property "foo". Object1 -> (extends) commons.ObjectWithFooAndBar -> foo',
```
armandobelardo, line 202: QQ: do we want to keep this second sentence? Could be nice, since it's pointing out the extended object the property is on

armandobelardo (approved):
lgtm, just the one question

packages/cli/yaml/validator/src/rules/valid-example-type/__test__/valid-example-type.test.ts
dsinghvi: Down to add it back if you think thats better

PR #3938, opened 2024-06-26
java, improvement: improve java exception naming

dcb6:
Changes:
* Make base exception names configurable.
* Postfix base api exception and base exception class names with `Exception` instead of `Error`
* No longer prefix fern-defined error types with `<organization><workspace>`; use the defined name exactly
* Use `client-class-name` if configured in place of `<organation><workspace` in our base exception types

### DIFF
--- a/generators/java/sdk/CHANGELOG.md
+++ b/generators/java/sdk/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - name: fernapi/fern-java-sdk
     config:
       base-api-exception-class-name: ApiError
+      base-exception-class-name: CompanyException # Optional: This should only be set if default naming is undesirable
   ```
 - Improvement: We now generate Exception types for all errors that are defined in the IR. Generated clients with an
   error discrimination strategy of "status code" will throw one of these typed Exceptions based on the status code of


generators/java/model/CHANGELOG.md
```diff
@@ -10,19 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Feature: The generator now supports BigInteger types.
 - Chore: Bump intermediate representation to v46
 
-## [0.8.2] - 2024-06-04
```
dcb6, line 13: we couldn't find the sdk changelog for the previous exception release because i put it here in `model` 🤦

seed/java-sdk/alias/src/main/java/com/seed/alias/core/SeedAliasApiException.java
```diff
@@ -1,12 +1,12 @@
 /**
  * This file was auto-generated by Fern from our API Definition.
  */
-package com.seed.plainText.core;
+package com.seed.alias.core;
 
 /**
  * This exception type will be thrown for any non-2XX API responses.
  */
-public class SeedPlainTextApiError extends SeedPlainTextError {
```
dsinghvi, line 9: are we doing something fancy here to rename `Error` suffixes to `Exception` ?

dsinghvi: if someone wants to upgrade to 1.x is there a way for them to keep their exception names from breaking ?

dsinghvi (approved):

PR #3944, opened 2024-06-27
fix: api update command now works with unioned + nested APIs

armandobelardo:
The command now uses the absolute path to the workspace, as opposed to the fern directory to know where to download the file
This should work since the gen.yml lives at the workspace path, and the origin is relative to the gen.yml file


packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
```diff
@@ -63,7 +63,11 @@ export async function updateApiSpec({
                 for (const api of apis) {
                     if (typeof api !== "string" && api.origin != null) {
                         cliContext.logger.info(`Origin found, fetching spec from ${api.origin}`);
-                        await fetchAndWriteFile(api.origin, path.join(fernDirectory, api.path), cliContext.logger);
+                        await fetchAndWriteFile(
+                            api.origin,
+                            path.join(workspace.absoluteFilepath, api.path),
```
dsinghvi, line 68: any reason we aren't using the `join` function in `fs-utils` ?

PR #3947, opened 2024-06-27
(fix, csharp): json serialize enums before sending over the wire

dsinghvi:
Enums need to be json serialized before sending over the wire in C#.

### DIFF
--- a/seed/csharp-sdk/enum/src/SeedEnum/QueryParam/QueryParamClient.cs
+++ b/seed/csharp-sdk/enum/src/SeedEnum/QueryParam/QueryParamClient.cs
@@ -1,4 +1,3 @@
-using System.Test.Json;
 using System.Text.Json;
 using SeedEnum;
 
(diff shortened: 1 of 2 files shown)


seed/csharp-sdk/enum/src/SeedEnum/QueryParam/QueryParamClient.cs
```diff
@@ -42,7 +43,7 @@ public async void SendListAsync(SendEnumListAsQueryParamRequest request)
     {
         var _query = new Dictionary<string, object>()
         {
-            { "operand", request.Operand.ToString() },
+            { "operand", JsonSerializer.Serialize(request.Operand) },
```
dsinghvi, line 46: This is the relevant change

armandobelardo (commented):
The code looks reasonable, 2 questions:
- Is there a fixture i can look to to see the change? I didn't see one combing through
- Is there a similar fix needed for unions? I think they also have a custom JSON serializer (not 100% sure when `stringify` is called so may be moot)

armandobelardo (approved):

dsinghvi:
> The code looks reasonable, 1 question:
> 
> * Is there a similar fix needed for unions? I think they also have a custom JSON serializer (not 100% sure when `stringify` is called so may be moot)

You're right that another fix is needed for unions -- I think we need to modify `RawClient.cs` as well to accept a `Map<str, unknown>`

PR #3948, opened 2024-06-27
(fix, cli): remove out of range number validations in `openapi-ir-to-fern` + remove husky

dcb6:
### Core Changes
Before this PR, it was possible for us to fail to deserialize the IR in our generators due to number validations in OpenAPI specs that are outside of the range of values for that type. Ex:
```
age:
  type: integer
  minimum: -2147483649 # outside of min int, would fail to deserialize
  maximum: 2147483648 # outside of max int, would fail to deserialize
```
This PR fixes the issue by setting any such values as `undefined` in `openapi-ir-to-fern`.

### Husky Removal
I also removed husky because we don't use it and it was messing me up.


dsinghvi (approved):

dsinghvi:
@dcb6 to fix CI you'll need to run

```sh
yarn format:fix
yarn test:ete -u
```

PR #3951, opened 2024-06-28
fix: generator upgrade cli upgrades in place

armandobelardo:
Prior to this, the generator upgrade command would mangle the yaml file, even if nothing was updated because parsing the file with Zod orders the schema by the property names in the zod schema

Now this manually parses through the yaml and updates in place using the `yaml` lib

This also skips upgrading if the upgrade is through major versions

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -6495,7 +6495,6 @@ function $$SETUP_STATE(hydrateRuntimeState, basePath) {
             ["@types/jest", "npm:29.0.3"],\
             ["@types/js-yaml", "npm:4.0.8"],\
             ["@types/node", "npm:18.7.18"],\
-            ["@types/semver", "npm:7.5.8"],\
             ["@types/tinycolor2", "npm:1.4.6"],\
             ["depcheck", "npm:1.4.6"],\
             ["dockerode", "npm:4.0.2"],\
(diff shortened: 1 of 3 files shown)


packages/cli/cli/src/commands/upgrade/upgradeGenerator.ts
```diff
@@ -1,30 +1,104 @@
...
+                continue;
+            }
+
+            // check if the versions have the same major version, if not fail, if yes do the upgrade
+            const currentMajor = semver.major(currentVersion);
+            const latestMajor = semver.major(latestVersion);
+            if (currentMajor === latestMajor) {
+                generator.set("version", latestVersion);
+                context.logger.info(chalk.green(`${generatorName} has been upgraded to latest in group: ${groupName}`));
+            }
```
dsinghvi, line 97: should we try to upgrade to the latest patch/minor?

dsinghvi (approved):
looks good to me -- i think a test that makes sure its editing in place so that if we ever mess with the logic we can be sure it works. Upgrading to the latest non major would be sick too

packages/cli/cli/src/commands/upgrade/upgradeGenerator.ts
armandobelardo: I'm not sure how we'd get it, do you have thoughts?
Basically right now we add a "latest" tag to any non-RC to be able to get the latest version
But to get the latest patch/minor we'd need to pull all the images, then grab the version tag, then parse and sort them, which seems rough?

PR #3953, opened 2024-06-28
improvement: add x-fern-base-path to asyncapi extensions

armandobelardo:


packages/cli/openapi-parser/src/parse.ts
```diff
@@ -109,6 +109,9 @@ export async function parse({
                     ...parsedAsyncAPI.schemas
                 };
             }
+            if (parsedAsyncAPI.basePath != null) {
+                ir.basePath = parsedAsyncAPI.basePath;
```
dsinghvi, line 113: what happens when both an async api and openapi have a base path?

dsinghvi: i think we probably want it such that all base paths get handled in `openapi parser` and no base paths makes it to `openapi-ir-to-fern`

dsinghvi (approved):

packages/cli/openapi-parser/src/parse.ts
armandobelardo: Chatted in slack, since this isn't a regression we're going to merge to unblock hume
But will address in the near future, ticket here: https://linear.app/buildwithfern/issue/FER-2202/x-fern-base-path-ends-up-being-treated-globally-across-merged-specs

PR #3956, opened 2024-06-30
Use all FormData headers and don't stringify stream.Readable

williamluer:
Elevenlabs reported issue with their TS SDK where requests to `.dubAVideoOrAnAudioFile()` always returned a 400 saying that `target_lang` was not set. This was fixed by:
* Using `encoder.headers` rather than `{Content-Length: encoder.length}`
* Updating `Fetcher` to no longer stringify `stream.Readable` data type


dcb6 (approved):
💯

PR #3960, opened 2024-07-01
improvement: allow adding extra dependencies to Ruby SDK

armandobelardo:


generators/ruby/sdk/src/CustomConfig.ts
```diff
@@ -1,15 +1,17 @@
-import { BaseGeneratorConfigSchema } from "@fern-api/generator-commons";
+import { BaseGeneratorConfigSchema, ExtraDependenciesSchema } from "@fern-api/generator-commons";
 import { z } from "zod";
 
 export type RubySdkCustomConfig = z.infer<typeof RubySdkCustomConfigSchema>;
 export const RubySdkCustomConfigSchema = BaseGeneratorConfigSchema.extend({
     defaultTimeoutInSeconds: z.optional(z.union([z.literal("infinity"), z.number()])),
+    extraDevDependencies: z.optional(ExtraDependenciesSchema),
```
dsinghvi, line 7: should we do `extraDependencies` as well while we are here?

generators/ruby/sdk/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.8.0-rc0] - 2024-07-01
+
+- Improvement: allow users to specify additional dependencies and dev dependencies for Ruby SDKs.
```
dsinghvi, line 12: would be nice to add an example

dsinghvi (approved):

generators/ruby/sdk/src/CustomConfig.ts
armandobelardo: It's in the base object, so it's already there!

generators/ruby/sdk/CHANGELOG.md
armandobelardo: added!

PR #3961, opened 2024-07-01
improvement: python async snippets now leverage asyncio run

armandobelardo:


dsinghvi (approved):
```python
asyncio.run(
    main(),
)
```

any chance we can disable the trailing comma on this function invocation? i think there must be some boolean we should be able to set

armandobelardo:
> ```python
> asyncio.run(
>     main(),
> )
> ```
> 
> any chance we can disable the trailing comma on this function invocation? i think there must be some boolean we should be able to set

There's not a flag today since it's on the node-writer level it looks like, so you'd have to mode the entire snippet as `write_as_snippet=False` which is odd
Will add one for this though

PR #3963, opened 2024-07-01
(fix, cli): make sure `js` file checking works

dsinghvi:
`fern check` fails when you have css files which causes users to not be able to take upgrades. This is a patch to fix.


abvthecity:
Note: CSS is no longer necessary to check because it is not marked as uploaded to S3: https://github.com/fern-api/fern/pull/3964/files

PR #3971, opened 2024-07-01
(feat): customize status code for typescript express generator 

dsinghvi:
The Express generator now supports a `requestValidationStatusCode` configuration 
where a user can specify the status code that should be thrown when the request doesn't 
match schema validation. 

```yaml
- name: fernapi/fern-typescript-express
  version: 0.16.3
  config: 
    requestValidationStatusCode: 400 # Defaults to 422
```

### DIFF
--- a/generators/typescript/express/generator/src/ExpressGenerator.ts
+++ b/generators/typescript/express/generator/src/ExpressGenerator.ts
@@ -210,7 +210,7 @@ export class ExpressGenerator {
             includeSerdeLayer: config.includeSerdeLayer,
             skipRequestValidation: config.skipRequestValidation,
             skipResponseValidation: config.skipResponseValidation,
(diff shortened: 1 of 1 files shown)


seed/ts-express/imdb/validation-status-code/api/resources/imdb/service/ImdbService.ts
```diff
@@ -0,0 +1,136 @@
...
+                                " the endpoint's errors list in your Fern Definition."
+                        );
+                        await error.send(res);
+                    } else {
+                        res.status(500).json("Internal Server Error");
+                    }
+                    next(error);
+                }
+            } else {
+                res.status(400).json({
```
dsinghvi, line 91: `400` instead of `422`

PR #3975, opened 2024-07-02
(charp, fix): Empty Root Client Methods + `.Core` namespace issue

dcb6:
- Fix: Base client methods are now implemented rather than being empty.
- Fix: All  `Core` files are now generated to the corresponding `Core` namespace.

### DIFF
--- a/seed/csharp-model/alias/src/SeedAlias/Core/CollectionItemSerializer.cs
+++ b/seed/csharp-model/alias/src/SeedAlias/Core/CollectionItemSerializer.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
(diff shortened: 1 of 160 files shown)


seed/csharp-sdk/package-yml/src/SeedPackageYml/SeedPackageYmlClient.cs
```diff
@@ -19,7 +21,23 @@ public SeedPackageYmlClient(ClientOptions clientOptions = null)
 
     public ServiceClient Service { get; }
 
-    public async void EchoAsync() { }
+    public async Task<string> EchoAsync(string id, string request)
```
dcb6, line 24: root client fix shown here

dsinghvi (approved):
looks great, thank you!

PR #3976, opened 2024-07-02
(csharp, fix): base client requests not generated

dcb6:
Base client requests are now generated, whereas previously they were skipped


seed/csharp-sdk/package-yml/src/SeedPackageYml/SeedPackageYmlClient.cs
```diff
@@ -21,7 +21,7 @@ public SeedPackageYmlClient(ClientOptions clientOptions = null)
 
     public ServiceClient Service { get; }
 
-    public async Task<string> EchoAsync(string id, string request)
+    public async Task<string> EchoAsync(string id, EchoRequest request)
```
dcb6, line 24: here's the changed test to cover this, apparently it wasn't covered before (having request objects in base client methods)

generators/csharp/sdk/src/SdkGeneratorCli.ts
```diff
@@ -80,6 +80,23 @@ export class SdkGeneratorCLI extends AbstractCsharpGeneratorCli<SdkCustomConfigS
...
+                        wrapper: endpoint.sdkRequest.shape,
+                        context,
+                        endpoint,
+                        serviceId: rootServiceId
+                    });
+                    const wrappedRequest = wrappedRequestGenerator.generate();
+                    context.project.addSourceFiles(wrappedRequest);
+                }
+            }
+        }
```
dsinghvi, line 98: i wonder if there is a better way to share this functionality 😓

dsinghvi (approved):

PR #3977, opened 2024-07-02
(feat, typescript): refactor `Fetcher` and add unit tests

williamluer:
* Add unit tests to Fetcher
* Refactors Fetcher by splitting up code into functions and reorganizing file structure

Typescript tests could not run, need to make these changes. Other core-utilities packages are updated here: https://github.com/fern-api/fern/pull/4022 
* Fix babel.config.json path
* Add `@babel/present-env` and `@babel/preset-typescript` as dependencies

In follow up PRs:
* end to end fetchImpl tests. There are 6 return options but only 1 is tested.
* FormDataWrapper tests

Run tests with:
* `yarn workspace @fern-typescript/fetcher test`
* `yarn workspace @fern-typescript/fetcher test --env=jsdom`


generators/typescript/utils/core-utilities/fetcher/src/fetcher/requestWithRetries.ts
```diff
@@ -0,0 +1,20 @@
+const INITIAL_RETRY_DELAY = 1;
+const MAX_RETRY_DELAY = 60;
+const DEFAULT_MAX_RETRIES = 2;
+
+export async function requestWithRetries(
+    requestFn: () => Promise<Response>,
+    maxRetries: number = DEFAULT_MAX_RETRIES
+): Promise<Response> {
+    let response: Response;
```
armandobelardo, line 9: Nit: should we make an initial request here? Would also allow us to nix the non-null assertion

I think it also makes it true to the number of retries, I think right now it does one less since the first is part of the loop

But I could be doing the math wrong

williamluer: Good catch! I added an initial request to `requestWithRetries` and updated tests to reflect the additional call: https://github.com/fern-api/fern/pull/3977/commits/bf1ab2ae898e5ea2bf67cb87b8b5514b75296398

dsinghvi (approved):

PR #3978, opened 2024-07-02
(java, improvement): change default `JsonInclude` behavior 

dcb6:
The SDK generator now adds a class-level `@JsonInclude(JsonInclude.Include.NON_ABSENT)` annotation to 
  each generated type in place of the previous `@JsonInclude(JsonInclude.Include.NON_EMPTY)` by default. This ensures 
  that required empty collection fields are not removed from request or response json. This is configurable in the 
  `generators.yml` file:

minor: java model seed source mode is also configured / enabled in this PR


seed/java-sdk/seed.yml
```diff
@@ -24,6 +24,9 @@ fixtures:
         custom-dependencies:
           - "implementation jakarta.websocket:jakarta.websocket-api:2.1.0"
       outputFolder: custom-dependency
+    - customConfig:
+        json-include: non-empty
```
dcb6, line 28: added test here for configurability

dsinghvi (approved):
there are some `gradle-wrapper` changes in the seed diff that may be taking a second look at

PR #3980, opened 2024-07-03
chore: plumb through ruby snippets config to FDR

armandobelardo:


dsinghvi (approved):
lgtm, theres no good way of testing besides merging and checking on dev i think for this

PR #3981, opened 2024-07-03
improvement: allow boolean defaults within IR

armandobelardo:


packages/cli/generation/ir-generator/src/examples/validateTypeReferenceExample.ts
```diff
@@ -239,6 +239,9 @@ function validatePrimitiveExample({
                     example,
                     rules: v2.validation
                 }),
+            bigInteger: () => [],
+            long: () => [],
+            boolean: () => [],
```
dsinghvi, line 244: can we update the validation here ?

armandobelardo: They don't have validation rules though, just defaults

dsinghvi (approved):
lgtm besides the validation rule

PR #3982, opened 2024-07-03
improvement: python respects ir50, inserts defaults

armandobelardo:
This adds in long and bool defaults into the openapi parsing, as well as has python starting to input defaults to the model


seed/python-sdk/validation/with-defaults/src/seed/client.py
```diff
@@ -0,0 +1,325 @@
...
+        except JSONDecodeError:
+            raise ApiError(status_code=_response.status_code, body=_response.text)
+        raise ApiError(status_code=_response.status_code, body=_response_json)
+
+    def get(
+        self,
+        *,
+        decimal: float = 1,
+        even: int = 42,
+        name: str = "fern",
```
dsinghvi, line 121: let's say the default for a field typed as optional<integer> is 4, can the user still send `None` ? Is it worth adding an explicit test for that in the test definition

armandobelardo: I'll add a test
But yeah pydantic will prefer user provided over default

packages/cli/openapi-parser/src/schema/convertNumber.ts
```diff
@@ -80,6 +80,7 @@ export function convertNumber({
             nameOverride,
             generatedName,
             primitive: PrimitiveSchemaValueWithExample.int64({
+                default: _default,
```
dsinghvi, line 83: would be great to add a fixture (or update existing) to check that the parsing for longs and booleans work?

dsinghvi (approved):
looks great would be sick if we were able to include the `object_with_defaults.py` test in the generated SDK (if the customer had objects with deafults)

but ik this is asking a lot

PR #3990, opened 2024-07-05
(fix, openapi): Fix allOf object filtering

amckinney:
This updates the OpenAPI parser's `allOf` object filtering so that we can handle more `allOf` object variants. Previously, we would only check if a `type` property was set, but this includes a new `isValidAllOfObject` helper, which is more accurate. 

Note that we also filter down the potential `allOf` values in two phases. The first phase permits non-object types (like arrays), so that we correctly handle properties like the following (which are sometimes seen in practice):

```yaml
# The following can simply be mapped to an array of files,
# as defined by the second, valid element of the `allOf`.
fileArray:
  allOf:
    - {}
    - type: array
      items:
        type: string
        format: binary
```

### DIFF
--- a/packages/cli/openapi-parser/src/schema/convertSchemas.ts
+++ b/packages/cli/openapi-parser/src/schema/convertSchemas.ts
@@ -643,7 +643,11 @@ export function convertSchemaObject(
         // Now that we've handled the single-element allOf case, filter the
         // allOfs down to just the objects.
         const filteredAllOfObjects = filteredAllOfs.filter((allOf) => {
-            return isValidAllOfObject(allOf);
+            const valid = isValidAllOfObject(allOf);
+            if (!valid) {
+                context.logger.warn(`Skipping non-object allOf element: ${JSON.stringify(allOf)}`);
+            }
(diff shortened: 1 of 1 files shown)


packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -641,7 +640,28 @@ export function convertSchemaObject(
             return maybeInjectDescriptionOrGroupName(convertedSchema, description, groupName);
         }
 
-        // otherwise convert as an object
+        // Now that we've handled the single-element allOf case, filter the
+        // allOfs down to just the objects.
```
dsinghvi, line 644: should we log which `allOfs` we are ignoring that are not objects? (maybe as a warn)

dsinghvi (approved):

packages/cli/openapi-parser/src/schema/convertSchemas.ts
amckinney: Great call, I'll add that in

PR #4002, opened 2024-07-08
Script for Creating New Generators

mikemilla:


dsinghvi:
@mikemilla for a template, it probably doesn't make sense to have the entire C# AST pasted. Thoughts on pairing it down to just a `Writer` or something?

PR #4007, opened 2024-07-08
(feat, typescript): support jsr publish

dsinghvi:
Feature: Support JSR publishing. If you would like your SDK to be published to JSR, there
  is now a configuration option called `publishToJsr: true`. When enabled, the generator will
  generate a `jsr.json` as well as a GitHub workflow to publish to JSR.

  ```yaml
  - name: fernapi/fern-typescript-sdk
    version: 0.27.1
    config:
      publishToJsr: true
  ```

### DIFF
--- a/generators/typescript/express/generator/src/ExpressGenerator.ts
+++ b/generators/typescript/express/generator/src/ExpressGenerator.ts
@@ -258,7 +258,8 @@ export class ExpressGenerator {
             resolutions: {
                 "@types/mime": "3.0.4"
             },
-            extraConfigs: undefined
+            extraConfigs: undefined,
+            outputJsr: false
         });
     }
 


amckinney (approved):

generators/typescript/utils/commons/src/typescript-project/BundledTypescriptProject.ts
```diff
@@ -313,6 +320,17 @@ export * from "./${BundledTypescriptProject.TYPES_DIRECTORY}/${folder}";
         await this.writeFileToVolume(RelativeFilePath.of("package.json"), JSON.stringify(packageJson, undefined, 4));
     }
 
+    private async generateJsrJson(): Promise<void> {
+        if (this.npmPackage != null) {
+            const jsr: JSR = {
+                name: this.npmPackage?.packageName,
+                version: this.npmPackage.version,
+                exports: "src/index.ts"
```
armandobelardo, line 328: Dumb q: do we not output index.ts if we're in esm mode? Like instead we do index.js + index.d.ts?

dsinghvi: We actually did for Hookdeck when we were in `esmMode`

dsinghvi: I think we dont output it on local mode

PR #4008, opened 2024-07-08
(feature, typescript): Add offset pagination

amckinney:
This adds support for offset pagination. The same `Page` and `Pageable` abstractions are used, with a mild adjustment to the `loadPage` and `hasNextPage` helper functions.

With offset pagination, we define the stopping condition (via `hasNextPage`) simply when the server responds with no remaining elements.
* There isn't a "next" response property for us to act upon like in cursor pagination, so this is the best we can do (and is consistent with the other SDKs).
* The generated code is equivalent to the cursor pagination strategy excluding the final `core.Pageable` instantiation shown below:

```typescript
let _offset = request.page != null ? request.page : 1;
    hasNextPage: (response) => (response?.data ?? []).length > 0,
    getItems: (response) => response?.data ?? [],
    loadPage: (_response) => {
        _offset += 1;
        return list({
            ...request,
            page: _offset,
        });
    },
});
```


dsinghvi:
> By default, the value of the step is equal to 1.

is this a reasonable default? or should we choose something more like 10?

dsinghvi:
@amckinney overall this looks great, and builds on the existing primitives!

amckinney:
> > By default, the value of the step is equal to 1.
> 
> is this a reasonable default? or should we choose something more like 10?

I actually had an oversight in the original offset+step configuration - if the step property is optional, then it's unclear what the step size should default to. We don't know how many elements the server will respond with by default, plus the server's behavior can change at any time.
* For example, if the server responds with 50 elements, defaulting to any value NOT equal to `50` will yield invalid results. Some elements might be returned more than once, or if the default is too large some will be skipped on every page.

But fortunately we actually have access to exactly the number of results that we're returned - we should just use that to increment the offset. So the new `step` configuration is valuable to distinguish `offset` from `offset` w/ `step`, but it's not actually critical - we have all the information without an additional request property. We could use the configurable step size for the final request that breaks the boundary, but we could also reasonably just always increment by 1):

```typescript
const _stepSize = request.limit != null ? request.limit : 1;
let _offset = request.page != null ? request.page : 1;
    hasNextPage: (response) => (response?.data ?? []).length > 0,
    getItems: (response) => response?.data ?? [],
    loadPage: (response) => {
        _offset += (response?.data ?? []).length > 0 ? response.data.length + 1 : _stepSize;
        return list({
            ...request,
            page: _offset,
        });
    },
});
```

or simply:

```typescript
let _offset = request.page != null ? request.page : 1;
    hasNextPage: (response) => (response?.data ?? []).length > 0,
    getItems: (response) => response?.data ?? [],
    loadPage: (response) => {
        _offset += (response?.data ?? []).length > 0 ? response.data.length + 1 : 1;
        return list({
            ...request,
            page: _offset,
        });
    },
});
```

Moving forward, I imagine we could simplify the offset configuration to just the following:

```yaml
service:
  endpoints:
    ...
    listUsers:
      ...
      pagination:
        offset: $request.offset
        results: $response.users
        step: true
```

But configuring an arbitrary request property will still suffice for now.

amckinney:
Note that this PR was scoped down to just support basic offset pagination. We'll need to update the TypeScript generator to consume the latest IR for offset+step pagination, so that will be handled in a follow-up.

dsinghvi (approved):

PR #4010, opened 2024-07-08
Template for creating a new SDK generator

mikemilla:
Updates to generating a new SDK generator from templates.


generators/template/codegen/src/ast/__test__/Lang.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    // TODO: 👋 Add your other language tests here
+
+    // Note: You can change the language's indentation size with FULL_LANGUAGE_NAME.indentSize
+
+    it("makes function", () => {
+        const output = FULL_LANGUAGE_NAME.makeFunction({
+            name: "doSomething",
+        });
+        // eslint-disable-next-line no-console
+        console.log(output.toString());
```
dsinghvi, line 16: talked offline: we chatted about `expect(output.ToString()).toMatchSnapshot();`

generators/commons/src/Writer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    public openIndent(): void {
+        this.indentLevel++;
+    }
+
+    public closeIndent(): void {
+        if (this.indentLevel > 0) {
+            this.indentLevel--;
+        }
+    }
```
dsinghvi, line 50: would be nice to have a 



```ts
type CodeWriter = (writer: Writer): void;

public indent(func: CodeWriter): void {

}
```

dsinghvi (approved):

PR #4018, opened 2024-07-09
(fix, internal): `template/codegen` repo plays nicely with mrlint

dsinghvi:
This makes sure that the new template repo adheres to mrlint rules.

### DIFF
--- a/generators/template/codegen/package.json
+++ b/generators/template/codegen/package.json
@@ -1,10 +1,10 @@
 {
-  "name": "@fern-api/LANGUAGE_NAME-codegen",
+  "name": "@fern-api/template-codegen",
   "version": "0.0.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/fern-api/fern.git",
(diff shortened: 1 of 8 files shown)


generators/template/codegen/package.json
```diff
@@ -30,9 +30,7 @@
     "@fern-api/fs-utils": "workspace:*",
     "@fern-api/generator-commons": "workspace:*",
     "@fern-api/logging-execa": "workspace:*",
-    "@fern-fern/ir-sdk": "^32",
```
dsinghvi, line 33: didnt catch this in the review, but we shouldnt be depending on a specific version of IR in the template

yarn.lock
```diff
@@ -3312,29 +3312,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@fern-api/LANGUAGE_NAME-codegen@workspace:generators/template/codegen":
```
dsinghvi, line 3315: renamed

amckinney (approved):

PR #4020, opened 2024-07-09
Basic Swift AST CodeGen (Still work in progress)

mikemilla:


generators/commons/src/Writer.ts
```diff
@@ -28,20 +28,38 @@ export class Writer {
...
     }
 
-    public openIndent(): void {
-        this.indentLevel++;
+    public openBlock(
+        titles: (string | undefined)[],
+        openingCharacter: string | undefined = "{",
+        callback: () => void,
+        closingCharacter: string | undefined = "}"
+    ): void {
```
amckinney, line 43: nit: Something we've become fans of is in-lined request objects. That way, each positional parameter is named which reduces the amount of time you need to jump back on forth between a signature to know that you've ordered it correctly. This is particularly relevant for functions that accept two parameters of the same type.

I'd refactor this like so:
```suggestion
    public openBlock({
        titles,
        openingCharacter,
        callback,
        closingCharacter,
    }: {
        titles: (string | undefined)[];
        openingCharacter: string | undefined = "{";
        callback: () => void;
        closingCharacter: string | undefined = "}";
    }): void {
```

This will require updating every call-site, too.

amckinney: Now that I've reviewed all the call-sites more, is this function even worth it? It's trying to cram in a lot of functionality into one place. It's a generic block writer for classes, function definitions, and structs all in one place (presumably for multiple languages).

Should we instead just explicitly write out each definition explicitly so it's easier to read? We could still capture an indentation helper with a separate `writer.indent` method (or something similar). I imagine you could include something like the following for the `class` definition you have below:

```typescript

        // e.g. public class ClassName {
        if (this.accessLevel != null) {
          writer.write(this.accessLevel);
          writer.write(" ");
        }
        writer.writeLine(`class ${this.name} {`);
 
        writer.indent(() => {
          // TODO: Add fields
          // Add functions
          ...
        });

        writer.writeLine("}")
    }
```

What are your thoughts? Not married to anything here, but trying to consider the simplest / easiest to read solution.

generators/swift/codegen/src/ast/Param.ts
```diff
@@ -0,0 +1,41 @@
...
+        title: string;
+        type: string; // TODO
+        defaultValue?: string; // TODO
+    }
+}
+
+export class Param extends AstNode {
+
+    public readonly title: string;
+    public readonly type: string;
```
dsinghvi, line 15: @mikemilla i don't think you are going to want to model `type` as a string -- you are going to want a way to model a swift type (primitive + custom class).

generators/swift/codegen/src/ast/Func.ts
```diff
@@ -0,0 +1,63 @@
...
+import Swift, { AccessLevel, FunctionModifier, Param } from "../swift";
+
+export declare namespace Func {
+    interface Args {
+        accessLevel?: AccessLevel,
+        modifier?: FunctionModifier,
+        name: string,
+        params?: Param[];
+        async?: "async";
+        throws?: "throws";
```
dsinghvi, line 11: whats the thought behind making these literals ?

generators/swift/codegen/src/ast/Func.ts
```diff
@@ -0,0 +1,63 @@
...
+        returnObject?: string;
```
dsinghvi, line 12: Same question as above, if you have a function that returns a class from a different module, you'll want to be able to represent that in the AST.

generators/swift/codegen/src/ast/Func.ts
```diff
@@ -0,0 +1,63 @@
+import { AstNode, Writer } from "@fern-api/generator-commons";
+import Swift, { AccessLevel, FunctionModifier, Param } from "../swift";
+
+export declare namespace Func {
```
dsinghvi, line 4: (not a blocker): was really valuable to add doc strings in the ast at the function level, class level, parameter level

generators/swift/codegen/src/ast/Param.ts
dsinghvi: In C# you probably saw `csharp.Type.int()` or `csharp.Type.class({ name: "MyClass", namespace: "MyNamespac" });`

This adds some type safety to the `type` field, making it easier for non swift folks to contribute to the generator. Open to other ideas for modeling type as well but wanted to share what works in C#!

mikemilla: Yep, that was just a placeholder as I was building things out. It'll be some form of type.

PR #4022, opened 2024-07-09
Fix core-utilities typescript tests

williamluer:
`zurg`, `callback-queue`, `auth`, and `fetcher` tests could not be run with the command: `yarn workspace @fern-typescript/{zurg|callback-queue|auth|fetcher} test`. These changes (with yarn installs) allow the tests to successfully run. Fetcher is fixed here: https://github.com/fern-api/fern/pull/3977

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -891,17 +891,17 @@ function $$SETUP_STATE(hydrateRuntimeState, basePath) {
           ],\
           "linkType": "HARD"\
         }],\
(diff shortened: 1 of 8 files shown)


armandobelardo (approved):
Should fetcher be here too or is that in the other PR?

williamluer:
> Should fetcher be here too or is that in the other PR?

Yea I added the fetcher fixes to the other PR

PR #4028, opened 2024-07-09
(feature, typescript): Add offset step pagination with IRv48

amckinney:
This upgrades the TypeScript generators to IRv48, and implements the following features:
* Add support for pagination endpoints that require request body properties.
* Add support for pagination with an offset step. This is useful for endpoints that page based on the element index rather than a page index (i.e. the 100th element vs. the 10th page). This feature shares the same UX as both the `offset` and `cursor` pagination variants.

### DIFF
--- a/seed/ts-express/seed.yml
+++ b/seed/ts-express/seed.yml
@@ -1,4 +1,4 @@
-irVersion: v46
+irVersion: v48
 docker: fernapi/fern-typescript-express:local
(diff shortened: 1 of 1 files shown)


dsinghvi (approved):

amckinney:
Discussed offline - build failures are related to https://github.com/fern-api/fern/pull/4024, which will be handled in a follow-up.

PR #4029, opened 2024-07-09
fix: update unchecked base model to not coerce none

armandobelardo:


generators/python/core_utilities/pydantic/unchecked_base_model.py
```diff
@@ -130,6 +130,10 @@ def construct_type(*, type_: typing.Type[typing.Any], object_: typing.Any) -> ty
     Pydantic models.
     The idea is to essentially attempt to coerce object_ to type_ (recursively)
     """
+    # Short circuit when dealing with optionals, don't try to coerces None to a type
+    if object_ is None:
+        return None
```
dsinghvi, line 135: can we add an assertion for this case?

dsinghvi:
@armandobelardo merging to unblock, but can you follow this up with a unit test?

PR #4032, opened 2024-07-10
(feature, typescript): Add setObjectProperty core utility

amckinney:
This adds a new `setObjectProperty` core utility which includes the `setObjectProperty` function. This function acts similar to [Lodash.set](https://lodash.com/docs/4.17.15#set), but is rewritten and simplified to accommodate our use case without bringing in a large dependency.

This function is useful for setting arbitrarily nested values on objects within the SDK generator like so:

```typescript
const person = {
  age: 42
}
const john = setObjectProperty(person, "name", "John Doe");
```

Note that this fixes an issue for pagination endpoints that define nested offset/cursor properties. The generated pagination code now looks like the following:

```typescript
let _offset = request?.pagination?.page != null ? request?.pagination?.page : 1;
  hasNextPage: (response) => (response?.data ?? []).length > 0,
  getItems: (response) => response?.data ?? [],
  loadPage: (_response) => {
    _offset += 1;
    return list(core.setObjectProperty(request, "pagination.page", _offset));
  }
});
```

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -9375,8 +9375,6 @@ function $$SETUP_STATE(hydrateRuntimeState, basePath) {
           "packageLocation": "./generators/typescript/utils/core-utilities/utils/",\
           "packageDependencies": [\
             ["@fern-typescript/utils", "workspace:generators/typescript/utils/core-utilities/utils"],\
-            ["@babel/preset-env", "virtual:2faa50efc97dbc1bcf8362bab30ec9c2420177b3ec5d7b26e8049bc5b8b1e5a679c5ee38acb51b25948e2fd46f810b2882a9f7bdc3d8e6618ffd386357806c31#npm:7.24.3"],\
-            ["@babel/preset-typescript", "virtual:2faa50efc97dbc1bcf8362bab30ec9c2420177b3ec5d7b26e8049bc5b8b1e5a679c5ee38acb51b25948e2fd46f810b2882a9f7bdc3d8e6618ffd386357806c31#npm:7.24.1"],\
             ["@types/jest", "npm:29.0.3"],\
             ["@types/node", "npm:18.7.18"],\
             ["depcheck", "npm:1.4.6"],\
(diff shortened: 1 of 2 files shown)


generators/typescript/utils/core-utilities/reflect/__test__/setObjectProperty.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+interface TestCase {
+    description: string;
+    giveObject: object;
+    givePath: string;
+    giveValue: any;
+    wantObject: object;
+}
+
+describe("Test setObjectProperty", () => {
```
dsinghvi, line 11: @amckinney i think `reflect` is probably a little weird in TS (not going to be easy for someone poking in to understand. 

Do we have a `utils` folder in core? I could see this just going in there as one of the helper functions ? or a `helpers` folder?

generators/typescript/utils/core-utilities/reflect/setObjectProperty.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Sets the value at path of object. If a portion of path doesn’t exist it’s created.
+ *
+ * @param object The object to modify.
+ * @param path The path of the property to set.
+ * @param value The value to set.
+ * @return Returns object.
+ */
+export function setObjectProperty<T extends object>(object: T, path: string, value: any): T {
```
dsinghvi, line 9: lets credit lodash if we took it from there ?

amckinney: Oops, totally meant to do that - will add that in.

generators/typescript/utils/core-utilities/reflect/__test__/setObjectProperty.test.ts
amckinney: Sounds good, we don't have a good place for it already, but happy to name it something more generic like `utils`.

dsinghvi (approved):

PR #4033, opened 2024-07-10
c#, fix: increase supported union size + handle double optionals

dcb6:
- Improvement: Enable generating unions with up to 32 types by adding the OneOf.Extended package.
- Fix: The generator now handles double optional fields properly and only adds a single `?` to the type.

### DIFF
--- a/generators/csharp/codegen/src/ast/Type.ts
+++ b/generators/csharp/codegen/src/ast/Type.ts
@@ -106,7 +106,7 @@ export class Type extends AstNode {
         super();
     }
 
-    public write(writer: Writer): void {
+    public write(writer: Writer, parentType: Type | undefined = undefined): void {
         switch (this.internalType.type) {
             case "integer":
                 writer.write("int");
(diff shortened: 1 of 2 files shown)


test-definitions/fern/apis/exhaustive/definition/types/object.yml
```diff
@@ -33,6 +33,12 @@ types:
       string: string
       NestedObject: ObjectWithOptionalField
 
+  DoubleOptional:
+    properties:
+      optionalAlias: optional<OptionalAlias>
```
dcb6, line 38: Added this test def for double optionals

generators/csharp/codegen/src/ast/Type.ts
```diff
@@ -154,7 +154,9 @@ export class Type extends AstNode {
                 break;
             case "optional":
                 this.internalType.value.write(writer);
-                writer.write("?");
+                if (writer.getLastCharacter() !== "?") {
```
dcb6, line 157: curious what you think of this maybe hacky approach @dsinghvi

dsinghvi: interesting -- can we pass an optional parent type into this func instead? Feels like this will break pretty soon

dcb6: worked!

dsinghvi (approved):

PR #4034, opened 2024-07-10
add in swift to seed runner

armandobelardo:


packages/cli/configuration/src/generators-yml/generatorInvocations.ts
```diff
@@ -133,5 +133,19 @@ export const GENERATOR_INVOCATIONS: Record<GeneratorName, Omit<GeneratorInvocati
...
             path: "../sdks/csharp"
         }
+    },
+    [GeneratorName.SWIFT_MODEL]: {
+        version: "0.0.0",
+        output: {
+            location: "local-file-system",
+            path: "../sdks/swift/model"
+        }
+    },
```
amckinney, line 143: It looks like we omit the `output` configuration for the other _model_ generators above. Let's keep this consistent for now.

packages/cli/generation/ir-migrations/src/migrations/v50-to-v49/migrateFromV50ToV49.ts
```diff
@@ -33,7 +37,9 @@ export const V50_TO_V49_MIGRATION: IrMigration<
         [GeneratorName.RUBY_MODEL]: GeneratorWasNeverUpdatedToConsumeNewIR,
         [GeneratorName.RUBY_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR,
         [GeneratorName.CSHARP_MODEL]: GeneratorWasNeverUpdatedToConsumeNewIR,
-        [GeneratorName.CSHARP_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR
+        [GeneratorName.CSHARP_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR,
+        [GeneratorName.SWIFT_MODEL]: GeneratorWasNotCreatedYet,
```
amckinney, line 41: Let's tag `0.0.0` for these since this is the latest IR (the first one the generator will consume). You can see this in the other migration files when the generator was first introduced (e.g. IRv32 for Ruby).

armandobelardo: cool only doing for model since there might be another version of the IR before we build that

packages/cli/configuration/src/generators-yml/generatorInvocations.ts
armandobelardo: donezo

packages/seed/src/utils/getGeneratorInvocation.ts
```diff
@@ -163,6 +165,8 @@ function getPublishInfo({
             throw new Error("Seed doesn't support publish mode in Ruby!");
         case "csharp":
             throw new Error("Seed doesn't support publish mode in C#!");
+        case "swift":
+            throw new Error("Seed doesn't support publish mode in C#!");
```
amckinney, line 169:
```suggestion
            throw new Error("Seed doesn't support publish mode in Swift");
```

amckinney (approved):

PR #4035, opened 2024-07-10
Initial Swift Codegen

armandobelardo:


generators/swift/codegen/src/ast/Type.ts
```diff
@@ -0,0 +1,127 @@
...
+        /* The field variables in the class */
+        fields?: Field[];
+        /* The functions associated with this type */
+        functions?: Func[];
+        /* The inheritance hierarchy of this type */
+        inheritance?: Type[];
+    }
+}
+
+export class Type extends AstNode {
```
dsinghvi, line 55: nit: as a follow up i think its worth calling this `Class` for clarity (already talked about)

generators/swift/codegen/src/ast/Param.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export declare namespace Param {
+    interface Args {
+        /* param name */
+        title: string;
+        /* Type of the value */
+        type: Type;
+        /* Original value. Optional because it may not exist */
+        defaultValue?: Type;
+    }
```
dsinghvi, line 27: follow up: we'll want to see how to do parameter level docs

generators/swift/codegen/src/ast/Enum.ts
```diff
@@ -0,0 +1,59 @@
+import { AstNode, Writer } from "@fern-api/generator-commons";
+import Swift, { AccessLevel, EnumCase, Type } from "..";
+
+export declare namespace Enum {
+    interface Args {
+        accessLevel?: AccessLevel;
+        name: string;
+        inheritance?: Type[],
+        enumCases: EnumCase[]
```
dsinghvi, line 9: very small nit + follow up: the ir calls this `enumValues` so i think easier to align

generators/swift/codegen/src/ast/Import.ts
```diff
@@ -0,0 +1,25 @@
+import { AstNode, Writer } from "@fern-api/generator-commons";
+import Swift from "..";
+
+export declare namespace Import {
+  interface Args {
+    packageName: string;
+  }
+}
+
+export class Import extends AstNode {
```
dsinghvi, line 10: follow up-- would love to understand how `imports` are tracked in the future

dsinghvi (approved):

PR #4041, opened 2024-07-11
(feature, typescript): Support global header overrides on client options

amckinney:
This builds upon the changes introduced in https://github.com/fern-api/fern/pull/3945 and adds support for client-level overrides for global headers.

For example, users can now override the value of their version header for all method calls by constructing their client like so:

```typescript

const client = new AcmeClient({ version: "2024-07-11" });

// The X-API-Version header is equal to "2024-07-11" for every method call like the following:
await client.users.create({
  firstName: "john",
  lastName: "doe"
});
```

Note that request options take precedence over client-level options like so:

```typescript
                "X-API-Version": requestOptions?.version ?? this._options?.version ?? "02-02-2024",
                "X-API-Enable-Audit-Logging": (
                    requestOptions?.auditLogging ??
                    this._options?.auditLogging ??
                    true
                ).toString(),
```

### DIFF
--- a/generators/typescript/sdk/client-class-generator/src/GeneratedSdkClientClassImpl.ts
+++ b/generators/typescript/sdk/client-class-generator/src/GeneratedSdkClientClassImpl.ts
@@ -704,6 +704,7 @@ export class GeneratedSdkClientClassImpl implements GeneratedSdkClientClass {
                     let value: ts.Expression;
                     if (literalValue != null) {
                         if (typeof literalValue === "boolean") {
+                            const booleanLiteral = literalValue ? ts.factory.createTrue() : ts.factory.createFalse();
                             value = ts.factory.createCallExpression(
                                 ts.factory.createPropertyAccessExpression(
                                     ts.factory.createParenthesizedExpression(
@@ -725,7 +726,7 @@ export class GeneratedSdkClientClassImpl implements GeneratedSdkClientClass {
                                                 )
                                             ),
                                             ts.factory.createToken(ts.SyntaxKind.QuestionQuestionToken),
-                                            ts.factory.createTrue()
+                                            booleanLiteral
                                         )
                                     ),
                                     ts.factory.createIdentifier("toString")


generators/typescript/sdk/client-class-generator/src/GeneratedSdkClientClassImpl.ts
```diff
@@ -1234,7 +1284,7 @@ export class GeneratedSdkClientClassImpl implements GeneratedSdkClientClass {
         return ts.factory.createPropertyAccessExpression(this.getReferenceToOptions(), option);
     }
 
-    private getOptionKeyForNonLiteralHeader(header: HttpHeader): string {
+    private getOptionKeyForHeader(header: HttpHeader): string {
```
amckinney, line 1287: Heads up @dsinghvi, noticed that we were using a mixture of `safeName` and `unsafeName`, so I unified everything on `unsafeName` in this helper (for consistency with other parameter types). 

This should be OK, and none of the test fixtures are updated with this change but wanted to flag it in case there's some context I'm missing.

dsinghvi:
```typescript
(requestOptions?.auditLogging ?? this._options?.auditLogging ?? true).toString()
```

generators/typescript/sdk/client-class-generator/src/GeneratedSdkClientClassImpl.ts
```diff
@@ -698,51 +698,67 @@ export class GeneratedSdkClientClassImpl implements GeneratedSdkClientClass {
...
+                                                    ts.factory.createPropertyAccessExpression(
+                                                        ts.factory.createThis(),
+                                                        GeneratedSdkClientClassImpl.OPTIONS_PRIVATE_MEMBER
+                                                    ),
+                                                    ts.factory.createToken(ts.SyntaxKind.QuestionDotToken),
+                                                    ts.factory.createIdentifier(headerName)
+                                                )
+                                            ),
+                                            ts.factory.createToken(ts.SyntaxKind.QuestionQuestionToken),
+                                            ts.factory.createTrue()
```
dsinghvi, line 728: lmao the nesting tho

dsinghvi (approved):

PR #4047, opened 2024-07-12
feat, csharp: Unit Test Generation + IR Bump 

dcb6:
- Chore: Bump IR to 51.
- Feature: Generate serialization unit tests for models, as well as the CI to run them. These tests still exclude instantiated object equality assertions, which should be included at a later time.

### DIFF
--- a/generators/csharp/sdk/tsconfig.json
+++ b/generators/csharp/sdk/tsconfig.json
@@ -5,7 +5,6 @@
   "references": [
     { "path": "../../../packages/commons/core-utils" },
     { "path": "../../../packages/commons/fs-utils" },
-    { "path": "../../../packages/ir-sdk" },
     { "path": "../../commons" },
     { "path": "../codegen" },
     { "path": "../model" }


generators/csharp/model/src/snippets/ExampleGenerator.ts
```diff
@@ -0,0 +1,173 @@
...
+                    }));
+                    return csharp.dictionary({
+                        keyType: csharp.Type.object(),
+                        valueType: csharp.Type.optional(csharp.Type.object()),
+                        entries
+                    });
+                }
+                break;
+        }
+        return csharp.InstantiatedPrimitive.null();
```
dsinghvi, line 60: cc @amckinney ideally the IR supports `unknown` examples in a first class way instead of the generator author having to parse an `unknown` themselves

generators/csharp/model/src/snippets/ExampleGenerator.ts
```diff
@@ -0,0 +1,173 @@
...
+    ExampleObjectType,
+    ExamplePrimitive,
+    ExampleTypeReference,
+    ExampleUndiscriminatedUnionType,
+    ExampleUnionType
+} from "@fern-fern/ir-sdk/api";
+import { ModelGeneratorContext } from "../ModelGeneratorContext";
+import { ObjectGenerator } from "../object/ObjectGenerator";
+
+export class ExampleGenerator {
```
dsinghvi, line 14: nit: I think we should call this `ModelSnippetGenerator` ?

dsinghvi:
@dcb6 going to merge cause things look great and had very small nits around naming that we can follow up with

PR #4049, opened 2024-07-12
(fix, openapi): Resolve 'refs' specified in overrides

amckinney:
This fixes an issue where `$ref` elements were not resolved in OpenAPI override files. We now resolve any override references by passing the merged document back through `redocly`.

### DIFF
--- a/packages/cli/openapi-parser/src/loadOpenAPI.ts
+++ b/packages/cli/openapi-parser/src/loadOpenAPI.ts
@@ -1,7 +1,8 @@
 import SwaggerParser from "@apidevtools/swagger-parser";
 import { AbsoluteFilePath, dirname, join, RelativeFilePath } from "@fern-api/fs-utils";
 import { TaskContext } from "@fern-api/task-context";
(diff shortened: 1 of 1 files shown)


dsinghvi (approved):

packages/cli/openapi-parser/src/loadOpenAPI.ts
```diff
@@ -79,11 +64,57 @@ export async function loadOpenAPI({
...
+        dereference: false,
+        removeUnusedComponents: false,
+        keepUrlRefs: true
+    };
+    const result =
+        parsed != null
+            ? await bundle({
+                  ...options,
+                  doc: {
+                      source: new Source(absolutePathToOpenAPI, "<openapi>"),
```
amckinney, line 111: Will release this as an RC and come back to this - this fixes the issue at hand, but it's not immediately clear how `Source` is used and/or if it's necessary to serialize the parsed document as the body (instead of a `<openapi>` placeholder).

PR #4051, opened 2024-07-12
[WIP] Address Form-Data Issues

RohinBhargava:
## What this does:

- Removes `formdata-node` dependency that causes unprocessable entity errors (422)
- Effectively routes all form-data requests through `form-data` package

## Considerations:

- Eschews all benefits of the more modern, `formdata-node` package
- Probably lightens dependency tree a little bit

## Risks:

- Potential breaking changes -- any foundational change may cause this

## How was this tested:

Using elevenlabs SDK on multipart form data APIs on following environments:
   - Node 16
   - Node 17
   - Node 18
   - Node 20

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -8969,7 +8969,6 @@ function $$SETUP_STATE(hydrateRuntimeState, basePath) {
             ["eslint", "npm:8.56.0"],\
             ["form-data", "npm:4.0.0"],\
             ["form-data-encoder", "npm:4.0.2"],\
-            ["formdata-node", "npm:6.0.3"],\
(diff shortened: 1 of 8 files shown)


dsinghvi (commented):
@RohinBhargava i think the easiest way to really test this would be to add a dev dependency on express, create a dummy endpoint that serves files, and then make sure that we can use `FormDataWrapper` to send a proper request. Thoughts?

Would be great to fix ci to run jest + jest --env jsdom so that we test this in browser + node as well (https://stackoverflow.com/questions/51953713/what-is-env-jsdom)

RohinBhargava:
Closing due to new approach to fix.

PR #4060, opened 2024-07-15
chore: add back x-readme code samples

armandobelardo:
Just adds back x-readme parsing for code samples since we now validate them correctly


packages/cli/openapi-parser/src/openapi/v3/converters/operation/convertHttpOperation.ts
```diff
@@ -91,21 +91,7 @@ export function convertHttpOperation({
     });
 
     const availability = getFernAvailability(operation);
-    const examples = [...getExamplesFromExtension(operationContext, operation, context)];
-    // Validation on readme examples is wrong, but we're changing this data model so it's a wontfix for now
```
dsinghvi, line 95: did we test that this is now fixed (i would have expected us to fix validation as part of this PR)

armandobelardo: Yeah I can generate cohere's IR, which was the original reason this is removed
I think Myles fixed the issue a while back, but I can confirm

armandobelardo: Yeah looks like it was fixed last month by you: https://github.com/fern-api/fern/pull/3849

dsinghvi (approved):

PR #4067, opened 2024-07-16
(fix): support audiences on query parameters

dsinghvi:
Previously audiences were only respected on inlined request parameters, but not query parameters.

### DIFF
--- a/packages/cli/generation/ir-generator/src/filtered-ir/FilteredIr.ts
+++ b/packages/cli/generation/ir-generator/src/filtered-ir/FilteredIr.ts
@@ -20,14 +20,14 @@ export interface FilteredIr {
 
 export class FilteredIrImpl implements FilteredIr {
     private types: Set<TypeId> = new Set();
-    private properties: Record<TypeId, Set<string>>;
+    private properties: Record<TypeId, Set<string> | undefined>;
     private errors: Set<ErrorId> = new Set();
     private services: Set<ServiceId> = new Set();
     private endpoints: Set<EndpointId> = new Set();
-    private requestProperties: Record<EndpointId, Set<string>>;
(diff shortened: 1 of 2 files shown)


packages/cli/generation/ir-generator/src/filtered-ir/IrGraph.ts
```diff
@@ -388,6 +398,24 @@ export class IrGraph {
...
 
+            for (const [endpointId, queryParametersNode] of Object.entries(this.queryParameters)) {
+                if (!this.endpointsNeededForAudience.has(endpointId)) {
+                    continue;
+                }
+                const parametersForEndpoint = new Set<string>();
+                for (const audience of this.audiences.audiences) {
+                    const parametersByAudience = queryParametersNode.parametersByAudience[audience];
+                    if (parametersByAudience != null) {
+                        parametersForEndpoint.forEach((parameter) => {
```
amckinney, line 409: Should this be the following?
```suggestion
                        parametersByAudience.forEach((parameter) => {
```

packages/cli/generation/ir-generator/src/filtered-ir/FilteredIr.ts
```diff
@@ -140,8 +140,8 @@ export class FilteredIrImpl implements FilteredIr {
 
     public hasWebhookPayloadProperty(webhookId: string, property: string): boolean {
         const properties = this.webhookPayloadProperties[webhookId];
-        // no properties were filtered for inlined request
         if (properties == null) {
+            // No audiences were configured.
```
dsinghvi, line 144: @amckinney how does this check differ from if the value is `undefined` vs. the value is not present in the map?

dsinghvi:
Going to merge for now

PR #4068, opened 2024-07-16
(feature): Add ApiVersionSchema type

amckinney:
## Feature

This proposes a new `ApiVersionSchema` type which is used to configure an API-wide version header enum. Users can configure this with the following:

```yaml
# api.yml
version:
  header: X-API-Version
  default: "1.0.0"
  values:
    - "1.0.0-alpha"
    - "1.0.0-beta"
    - "1.0.0"
```

## Implementation

Given that the version header has a few more restrictions than general HTTP headers and enum type declarations, this proposes a few new YAML schema types for the `VersionDeclarationSchema`. Regardless, these schemas are still capable of mapping into the same IR types (i.e. `http.HttpHeader` and `types.EnumTypeDeclaration`). 

We also plan to support `default` values on enum type declarations and type references, so this defines the `default` value there.

## Behavior

In TypeScript, the user will receive a special `Version` type like so:

```typescript
// version.ts
type Version = "1.0.0-alpha" | "1.0.0-beta" | "1.0.0";
```

If a default value is configured, that `X-API-Version` header is sent on every request. Otherwise, the version header is treated as a required global header, which is specified on the client constructor.

### DIFF
--- a/fern/pages/api-definition/fern-definition/api-yml.mdx
+++ b/fern/pages/api-definition/fern-definition/api-yml.mdx
@@ -39,6 +39,22 @@ docs: |  # <---
 ```
 </CodeBlock>
 
+## API version
+
+You can define your API version scheme, such as a `X-API-Version` header. The supported versions and default value are specified like so:
+
+<CodeBlock title="api.yml">
+```yaml
+version:
+  header: X-API-Version
+  default: "2.0.0"
+  values:
+    - "1.0.0"
+    - "2.0.0"
+    - "latest"
+```
+</CodeBlock>
+
 ## Authentication
 
 You can specify the authentication scheme used by your API.
--- a/fern/pages/api-definition/openapi/extensions.mdx
+++ b/fern/pages/api-definition/openapi/extensions.mdx
@@ -49,6 +49,23 @@ paths:
       x-fern-sdk-method-name: create
 ```
 
+## API version
+
+You can define your API version scheme, such as a `X-API-Version` header. The supported versions and default value are specified like so:
+
+```yaml title="openapi.yaml"
+x-fern-version:
+  version:
+    header: X-API-Version
+    default: "2.0.0"
+    values:
+      - "1.0.0"
+      - "2.0.0"
+      - "latest"
+paths:
+  ...
+```
+
 ## Authentication overrides
 
 For authentication within your SDK, you may want to customize the:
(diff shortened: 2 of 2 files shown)


packages/ir-sdk/fern/apis/ir-types-latest/definition/ir.yml
```diff
@@ -77,11 +77,24 @@ types:
       hasPaginatedEndpoints: boolean
       hasFileDownloadEndpoints: boolean
       platformHeaders: PlatformHeaders
+      version: optional<VersionDeclaration>
```
dsinghvi, line 80: The `SDKConfig` is typically for SDK specific things, but I think API Version is something that may be important for non SDK generators too -- thoughts on putting it on the top level object instead? 

Something like: 
```
IntermediateRepresentation: 
  apiVersion: optional<APIVersionScheme>
```

packages/ir-sdk/fern/apis/ir-types-latest/definition/ir.yml
```diff
@@ -77,11 +77,24 @@ types:
...
+      The available set of versions for the API. This is used to generate a special
+      enum that can be used to specify the version of the API to use. This information
+      is sent as an HTTP header (e.g. X-API-Version) on every request.
+
+      If the enum does _not_ define a default value, the version should be treated like
+      a required global header parameter. The version header should also support any
+      environment variable scanning specified by the header.
+    properties:
+      header: http.HttpHeader
+      value: types.EnumTypeDeclaration
```
dsinghvi, line 97: I would type this as a union to make things easier for the future: 

```
ApiVersionScheme: 
  union: 
     headerBased: HeaderBasedAPIVersioningScheme
     
HeaderBasedAPIVersioningScheme: 
  properties: 
    header: http.HttpHeader
```

amckinney: Sweet, I had this beforehand but thought it might be overkill because we can always migrate to it. But given that you suggested the exact same, I'm happy to go back to this.

packages/ir-sdk/fern/apis/ir-types-latest/definition/ir.yml
amckinney: This was also what I started with, but figured I'd tuck it inside the SDK config because platform headers seemed somewhat related. Let's go with your suggestion.

dsinghvi (approved):

PR #4070, opened 2024-07-16
fix: python sdk serializes bytes within JSON

armandobelardo:


dsinghvi (approved):
Thank you for doing this!! -- would be sick to add a unit test 🚀

PR #4071, opened 2024-07-17
(feature, typescript): Generarte API version scheme

amckinney:
This adds support for generating an API version scheme in `version.ts`. Consider the following `api.yml` configuration:

```yaml
version:
  header: X-API-Version
  default: "1.0.0"
  values:
    - "1.0.0-alpha"
    - "1.0.0-beta"
    - "1.0.0"
```

The following `version.ts` file is generated:

```typescript
/**
 */

/** The version of the API, sent as the X-API-Version header. */
export type AcmeVersion = "1.0.0" | "2.0.0" | "latest";
```

If a default value is specified, it is set on every request but can be overridden in either the client-level `Options` or call-specific `RequestOptions`. If a default value is _not_ specified, the value of the header is required on the generated `Options`.

An example call is shown below:

```typescript

const client = new AcmeClient({ apiKey: "YOUR_API_KEY", xApiVersion: "2.0.0" });
await client.users.create({
  firstName: "john",
  lastName: "doe"
});
```

### DIFF
--- a/packages/cli/generation/ir-migrations/src/migrations/v51-to-v50/__test__/__snapshots__/migrateFromV51ToV50.test.ts.snap
+++ b/packages/cli/generation/ir-migrations/src/migrations/v51-to-v50/__test__/__snapshots__/migrateFromV51ToV50.test.ts.snap
@@ -23,6 +23,7 @@ exports[`migrateFromV51ToV50 simple 1`] = `
       "unsafeName": "simple_api",
     },
   },
+  "apiVersion": null,
   "auth": {
     "docs": null,
     "requirement": "ALL",


seed/ts-sdk/version/src/Client.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    interface RequestOptions {
+        /** The maximum time to wait for a response in seconds. */
+        timeoutInSeconds?: number;
+        /** The number of times to retry the request. Defaults to 2. */
+        maxRetries?: number;
+        /** A hook to abort the request. */
+        abortSignal?: AbortSignal;
+        /** Override the X-API-Version header */
+        xApiVersion?: version.SeedVersionVersion;
```
dsinghvi, line 24: can we inline the version right here ?

dsinghvi: `"1.0.0" | "2.0.0" | "latest"`

amckinney: Discussed offline - I originally defined a separate type to keep things DRY in the generated code, but this ends up hindering the autocomplete experience. With this, I have to jump to definition to see the available set of versions rather than just seeing them on hover.

We'll go with the in-lined version approach, but continue to generate the `version.ts` file at the top level so users can easily see what versions are available (similar to our story with `environments.ts` and default environment).

dsinghvi (approved):

PR #4072, opened 2024-07-17
chore: implement stream-parameter IR change

armandobelardo:
Allows Python SDK to generate `streaming` functions as overloaded functions or separate functions. Concretely, you can now specifiy whether or not you'd prefer the generation of your streaming functions as:

  ```
  def chat(...) -> ChatResponse:
    ...
  # and
  def chat_stream(...) -> typing.Iterator[StreamedChatResponse]:
    ...
  ```

  or:

  ```
  @overload
  def chat(...) -> ChatResponse:
    ...
  # and
  @overload
  def chat(..., stream=True) -> typing.Iterator[StreamedChatResponse]:
    ...

  # where a user sees the below, with the appropriate type hinting
  client.chat(..., stream=True)
  ```

### DIFF
--- a/seed/ts-sdk/seed.yml
+++ b/seed/ts-sdk/seed.yml
@@ -192,3 +192,4 @@ allowedFailures:
   - exhaustive:dev-dependencies
   - audiences
   - enum # throws b/c of undiscriminated union examples
+  - streaming-parameter


amckinney (approved):
👏

PR #4079, opened 2024-07-18
Add Availability to OpenApi Parser and OpenApi IR to Fern IR

RohinBhargava:
## What this does:

- Adds ability to set availability on PathParams, QueryParams, Headers, and Object Schemas (and as previous, endpoints)
- Can set with either `deprecated=true` or `x-fern-availability` in { `beta`, `deprecated`, `generally-available` }

## What's limited (open questions):

- Currently OpenAPI does not allow deprecations on refs:
image. We might want to support this.
- Individual Enum Values are not allowed to be marked as available or not.
- Parent objects, like objects that have `$ref` schema children do not inherit availability upward. If there is wrapping of objects, availability is set on the outer-most object.


generators/typescript/playground/browser/jest.config.ts
RohinBhargava, line 1: removing this path

fern/fern.config.json
```diff
@@ -1,4 +1,4 @@
 {
   "organization": "fern",
-  "version": "0.29.0"
+  "version": "0.32.0"
```
amckinney, line 3: We'll want to revert this - there's actually a broken link in our docs that we can fix in a separate PR. Your change here won't require this version to be upgraded. This is the cause of [this CI failure](https://github.com/fern-api/fern/actions/runs/9994086972/job/27623041862).

packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -127,6 +128,11 @@ export function convertSchemaObject(
     const groupName = typeof mixedGroupName === "string" ? [mixedGroupName] : mixedGroupName;
     const generatedName = getGeneratedTypeName(breadcrumbs);
     const description = schema.description;
+    // TODO: style, map more availability types
```
amckinney, line 131: Are we planning to do this here, or in a follow-up?

amckinney (commented):
Looks great so far! Discussed offline about a couple planned changes, but this is on the right track.

packages/cli/openapi-parser/src/schema/convertSchemas.ts
RohinBhargava: Thank you! Missed that one

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/flexport.test.ts.snap
```diff
@@ -2170,10 +2170,8 @@ exports[`open api parser flexport docs 1`] = `
                   },
                   "line_items": "list<root.CreateCommercialInvoiceLineItem>",
                   "manufacturer_refs": {
-                    "default": undefined,
-                    "docs": "[DEPRECATED] Array of your custom strings used to refer to a manufacturer location.",
```
dsinghvi, line 2174: did we lose transmitting docs in some cases?

packages/cli/openapi-ir-to-fern/src/buildEndpoint.ts
```diff
@@ -405,11 +404,20 @@ function getRequest({
...
                         ];
                     }
-                    return [property.key, propertyTypeReference];
+                    return [
+                        property.key,
+                        availability
+                            ? {
+                                  type: getTypeFromTypeReference(propertyTypeReference),
+                                  availability
+                              }
```
dsinghvi, line 418: we should share this code with lines 404-408 since it looks like we may already be dropping things like `docs` so its error prone

packages/cli/openapi-ir-to-fern/src/buildQueryParameter.ts
```diff
@@ -298,6 +311,8 @@ function getQueryParameterTypeReference({
                         generatedName: schema.generatedName,
                         value: oneOfSchema,
                         description: undefined,
+                        // TODO (rohin): not sure about this:
+                        availability: undefined,
```
dsinghvi, line 315: why not `schema.availability` ?

packages/cli/openapi-ir-to-fern/src/buildTypeDeclaration.ts
```diff
@@ -519,3 +519,21 @@ function getSchemaIdOfResolvedType({
     }
     return schema;
 }
+
+function convertPropertyTypeReferenceToTypeDefinition(
+    typeReference: RawSchemas.TypeReferenceWithDocsSchema,
+    audiences: string[],
+    name?: string | undefined,
+    availability?: RawSchemas.DeclarationWithoutDocsSchema["availability"]
+): RawSchemas.ObjectPropertySchema {
```
dsinghvi, line 528: nit: we basically prefer objects as arguments in typescript because you are forced to name each parameter (e.g. kwargs) so its more readable

packages/cli/openapi-ir-to-fern/src/buildQueryParameter.ts
RohinBhargava: Schema is arbitrary -- some of the schemas don't have availability. Coercion might lead to a runtime error, but might be able to do a property check here

PR #4080, opened 2024-07-18
fix: global headers are only required if the underlying type is required

armandobelardo:
This PR makes it so that:
1. if a header is required on some endpoints and optional on others then it cannot be a global header
2. if a header is optional on every endpoint, it's an optional global header (previously it would be considered required, regardless of the underlying type)
3. a global header is only required if it's required on every endpoint


dsinghvi:
>if a header is required on some endpoints and optional on others then it cannot be a global header

This could be a breaking change so possibly worth reverting? 

The other 2 are great, so less risk of merging.

PR #4083, opened 2024-07-19
improvement: update 'any object' examples to be flatter

armandobelardo:
Previously, we'd create nested objects if we received an `any object` e.g. an unknown value we're binding to an object (since we make those `map<string, unknown>`).
Those generated examples would become: `"object": {"object": {"key": "value"}}`, now we short circuit the example generation to special case `map<string, unknown>` to just return `{"key": "value"}`, making the above now `"object": {"key": "value"}`

As a bonus I also changed the map key and value generation to fall back to "key" and "value" as opposed to the field name which led to some awkward examples like `{"custom_fields": "custom_fields" }` seen below

### DIFF
--- a/packages/cli/ete-tests/src/tests/dependencies/__snapshots__/dependencies.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/dependencies/__snapshots__/dependencies.test.ts.snap
@@ -846,4 +846,4 @@ exports[`dependencies correctly incorporates dependencies 1`] = `
 }"
 `;
 
-exports[`dependencies file dependencies 1`] = `2968379`;
+exports[`dependencies file dependencies 1`] = `3022709`;


packages/cli/openapi-parser/src/__test__/__snapshots__/vellum.test.ts.snap
```diff
@@ -596,25 +588,17 @@ Creates a new document index.",
                   {
                     "key": {
                       "type": "string",
-                      "value": "indexing_config",
+                      "value": "key",
```
dsinghvi, line 591: any thoughts on why the `value` here is now `key` ?

packages/cli/generation/ir-generator/src/__test__/__snapshots__/openapi/enum-casing.txt
```diff
@@ -974,11 +974,11 @@ exports[`generate IR from OpenAPI 1`] = `
                                           "map": [
                                             {
                                               "key": {
-                                                "jsonExample": "custom_fields",
+                                                "jsonExample": "key",
```
dsinghvi, line 977: similar issue here with `custom_fields` -> `key`

armandobelardo: This one's intentional, it's outlined at the end of the pr body, but since it was `{"custom_fields": "custom_fields"}` I updated to `{"key": "value"}`

packages/cli/openapi-parser/src/__test__/__snapshots__/vellum.test.ts.snap
armandobelardo: Chatted offline, but this was similarly intentional for the same reason as the other comment! And since these are the example TRs, this is what we wanna see

amckinney (approved):

PR #4090, opened 2024-07-22
generate StrEnum for python 3.11+

jochs:
(str, Enum) mixins have undergone a breaking change in python 3.11 - see https://github.com/python/cpython/issues/100458

This PR adds poetry core to the dependencies to allow for version constraint parsing, and lists an explicit set of python minor versions to allow.

The target version is the lowest version that satisfies the constraint.

Also I don't have a good way to generate the lockfiles with poetry 1.7, so if somebody could regenerate them for me that'd be appreciated

### DIFF
--- a/generators/python/src/fern_python/generators/context/pydantic_generator_context_impl.py
+++ b/generators/python/src/fern_python/generators/context/pydantic_generator_context_impl.py
@@ -14,7 +14,16 @@ from ..sdk.custom_config import ALLOWED_PYTHON_VERSIONS
 from ..sdk.custom_config import DEFAULT_PYTHON_VERSION_CONSTRAINT
 from ..sdk.custom_config import SDKCustomConfig
 
+def solve_python_constraint(constraint_str: str) -> tuple[int, int]:
+    constraint = parse_constraint(constraint_str)
 
+    # For the sake of demonstration, we'll use a fixed version that satisfies the constraint
+    # Normally, you'd have a list of available versions and find the highest that satisfies the constraint
+    for version in ALLOWED_PYTHON_VERSIONS:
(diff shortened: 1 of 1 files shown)


generators/python/src/fern_python/generators/pydantic_model/type_declaration_handler/enum_generator.py
```diff
@@ -44,17 +45,24 @@ def generate(self) -> None:
                 should_export=True,
             )
         else:
-            enum_class = AST.ClassDeclaration(
-                name=self._get_class_name(),
-                extends=[
+            # If using python 3.11+, generate classes with StrEnum instead of (str, Enum)
+            if sys.version_info >= (3, 11):
```
dsinghvi, line 49: @jochs I think we'll want to discriminate based on this [config](https://github.com/fern-api/fern/blob/9274b21748ccb50183f654e12c86241382305ffb/generators/python/src/fern_python/generators/sdk/custom_config.py#L67)

jochs: why? that already sets up the pyproject.toml -- does that not determine the env that the generator runs in? Happy to plumb it but it doesn't quite feel right to me since that config is a version string for a solver, not true version info

jochs: ugh I see that's just for the generated project... will this even work if the generator is running via 3.9? I'm not sure how the reference imports work (if it's just straight up string concat or not) but StrEnum isn't in the stdlib until 3.11 so if it relies on the presence of the thing we're kinda SOL

PR #4095, opened 2024-07-22
[FER-2339] Pass OpenAPI request parameter examples through Fern IR Schema examples

RohinBhargava:
## What this does:

- When defining OpenAPI specs with examples in request parameters, this change propagates the example when converting to fern IR, so that example is correctly applied.

## What may need to be fixed:

- Noticed some snapshots in IR that have improperly escaped characters (quotes), which could fail upstream parsing.

### Example of before/after:
Screenshot 2024-07-22 at 4 15 54 PM
Screenshot 2024-07-22 at 4 12 23 PM

### DIFF
--- a/packages/cli/openapi-parser/src/openapi/v3/converters/endpoint/convertParameters.ts
+++ b/packages/cli/openapi-parser/src/openapi/v3/converters/endpoint/convertParameters.ts
@@ -52,6 +52,12 @@ export function convertParameters({
         const parameterBreadcrumbs = [...requestBreadcrumbs, resolvedParameter.name];
         const generatedName = getGeneratedTypeName(parameterBreadcrumbs);
 
+        if (getExamplesString({ schema: resolvedParameter, logger: context.logger })?.includes(" ")) {
+            context.logger.warn(
+                "Parameter example contains a space, which is ambiguous. Consider using enums for multiple examples, or use an encoding if a space is part of the parameter."
+            );
+        }
+
         let schema =
             resolvedParameter.schema != null
                 ? convertSchema(
--- a/packages/cli/openapi-ir-to-fern/src/utils/convertFullExample.ts
+++ b/packages/cli/openapi-ir-to-fern/src/utils/convertFullExample.ts
@@ -50,7 +50,7 @@ function convertPrimitive(primitiveExample: PrimitiveExample): RawSchemas.Exampl
         case "datetime":
             try {
                 // remove milliseconds from the datetime
-                if (!primitiveExample.value.endsWith("Z")) {
+                if (!primitiveExample.value.endsWith("Z") && !primitiveExample.value.includes("+")) {
                     primitiveExample.value += "Z";
                 }
                 return new Date(primitiveExample.value).toISOString().replace(/\.\d{3}Z$/, "Z");


packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/intercom.test.ts.snap
```diff
@@ -37490,7 +37527,7 @@ If you are redacting a conversation part, it must have a \`body\`. If you are re
               {
                 "name": "User reply",
                 "path-parameters": {
-                  "id": "id",
+                  "id": "123 or "last"",
```
RohinBhargava, line 37530: This doesn't seem right

dsinghvi: i bet thats right -- we should escape string examples. I wonder if there is a single place in `openapi-ir-to-fern` where this could happen

dsinghvi: https://github.com/fern-api/fern/blob/07161c212c5edf340cf5f693d3dda0bf5854bdbf/packages/cli/openapi-ir-to-fern/src/utils/convertFullExample.ts#L44-L48

abvthecity: it's likely because jest isn't configured to escape strings: https://jestjs.io/blog/2022/08/25/jest-29

packages/cli/openapi-parser/src/openapi/v3/converters/endpoint/convertParameters.ts
```diff
@@ -65,7 +73,7 @@ export function convertParameters({
                           format: undefined,
                           minLength: undefined,
                           maxLength: undefined,
-                          example: getExamplesString(resolvedParameter.example)
+                          example: getExamplesString(resolvedParameter as OpenAPIV3.SchemaObject)
```
dsinghvi, line 76: why do we have to use any casting here?

dsinghvi: should we just update the signature of the method to accept `OpenAPIV3.SchemaObject | OpenAPIV3.ParameterObject`

packages/cli/openapi-parser/src/schema/examples/getExample.ts
```diff
@@ -1,29 +1,49 @@
...
 
-export function getExampleAsBoolean(schema: OpenAPIV3.SchemaObject): boolean | undefined {
+export function getExampleAsBoolean(schema: OpenAPIV3.SchemaObject, exampleUnderride?: boolean): boolean | undefined {
     if (schema.example != null && typeof schema.example === "boolean") {
         return schema.example;
     }
-    return undefined;
+    if (exampleUnderride && typeof exampleUnderride !== "boolean") {
+        // eslint-disable-next-line no-console
+        console.warn(`Expected exampleUnderride to be a boolean, but got ${typeof exampleUnderride}`);
```
dsinghvi, line 21: Ideally we should pass in the `context` and use the `context.logger` instead of `console.log` directly. That way you can get rid of the eslint disable rule as well

packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -169,7 +182,7 @@ export function convertSchemaObject(
                     maxLength: schema.maxLength,
                     pattern: schema.pattern,
                     format: schema.format,
-                    example: getExamplesString(schema)
+                    example: getExamplesString(schema, context.logger, exampleUnderride as string)
```
dsinghvi, line 185: would be valuable to understand why we are still casting here and not checking the type re  `exampleUnderride as string`

packages/cli/openapi-parser/src/schema/examples/getExample.ts
```diff
@@ -1,29 +1,62 @@
...
+        return undefined;
+    }
+    return exampleUnderride;
 }
 
-export function getExampleAsBoolean(schema: OpenAPIV3.SchemaObject): boolean | undefined {
+export function getExampleAsBoolean(
+    schema: OpenAPIV3.SchemaObject | OpenAPIV3.ParameterObject,
+    contextLogger: Logger,
+    exampleUnderride?: boolean
```
dsinghvi, line 22: if we are checking the type here then should we make this unknown ?

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/intercom.test.ts.snap
RohinBhargava: seems ok in IR: 
image
Assuming that the reason is the jest reason @abvthecity mentioned

dsinghvi (approved):

PR #4096, opened 2024-07-22
(fix): OpenAPI parser handles generating examples when no request or response required

dsinghvi:
Previously the OpenAPI importer would fail to generate examples if no request or response was required. This would ultimately cause `fern check` to fail which would be very frustrating for users.

### DIFF
--- a/packages/cli/openapi-parser/src/openapi/v3/converters/ExampleEndpointFactory.ts
+++ b/packages/cli/openapi-parser/src/openapi/v3/converters/ExampleEndpointFactory.ts
@@ -76,8 +76,6 @@ export class ExampleEndpointFactory {
                     });
                     if (example != null) {
                         requestExamples.push([exampleId, example]);
-                    } else {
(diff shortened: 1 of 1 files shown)


packages/cli/openapi-parser/src/openapi/v3/converters/ExampleEndpointFactory.ts
```diff
@@ -225,7 +225,18 @@ export class ExampleEndpointFactory {
...
+            requestResponsePairs = consolidateRequestResponseExamples(requestExamples, responseExamples);
+        } else if (endpoint.request != null) {
+            requestResponsePairs = requestExamples.map(([id, example]) => {
+                return { id, request: example, response: undefined };
+            });
+        } else if (endpoint.response != null) {
+            requestResponsePairs = responseExamples.map(([id, example]) => {
+                return { id, request: undefined, response: example };
+            });
+        }
```
dsinghvi, line 239: critical change

armandobelardo (approved):

PR #4102, opened 2024-07-23
(feature): Add IRv52: uint and enum default values

amckinney:
This adds IRv52, which adds support for the new `uint` and `uint64` types, as well as default values on enum type references. For example, consider the following Fern definition that uses both of these features:

```yaml
types:
  UserType:
    enum:
      - ADMIN
      - UNAUTHENTICATED
      - AUTHENTICATED

  User:
    properties:
      userType:
        type: UserType
        default: UNAUTHENTICATED
      name: string
      age: uint
```

Along the way, this also uncovered an issue with our OpenAPI converter for `integer` types - we now recognize integers with the following `format` values, and convert them as needed (e.g. `int64` -> `long`):
* `uint32`
* `uint64`
* `int32`
* `int64`

### DIFF
--- a/packages/commons/core-utils/src/examples.ts
+++ b/packages/commons/core-utils/src/examples.ts
@@ -18,7 +18,7 @@ export const INT64 = 1000000;
 
 export const UINT = 1;
 
-export const UINT64 = 1;
+export const UINT64 = 1000000;
 
 export const UUID = "d5e9c84f-c2b2-4bf4-b4b0-7ffd7a9ffc32";
 
--- a/packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/uint.test.ts.snap
+++ b/packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/uint.test.ts.snap
@@ -17,7 +17,7 @@ exports[`uint uint docs 1`] = `
               "response": {
                 "body": {
                   "id": 1.1,
-                  "value": 1,
+                  "value": 1000000,
                 },
               },
             },
@@ -81,7 +81,7 @@ exports[`uint uint simple 1`] = `
               "response": {
                 "body": {
                   "id": 1.1,
-                  "value": 1,
+                  "value": 1000000,
                 },
               },
             },
--- a/packages/cli/openapi-parser/src/__test__/__snapshots__/uint.test.ts.snap
+++ b/packages/cli/openapi-parser/src/__test__/__snapshots__/uint.test.ts.snap
@@ -38,7 +38,7 @@ exports[`open api parser uint parse open api 1`] = `
                 "type": "primitive",
                 "value": {
                   "type": "uint64",
-                  "value": 1,
+                  "value": 1000000,
                 },
               },
             },
--- a/seed/ts-sdk/audiences/.mock/definition/api.yml
+++ b/seed/ts-sdk/audiences/.mock/definition/api.yml
@@ -2,3 +2,4 @@ name: audiences
 display-name: Audiences API
 audiences:
   - public
+  - private
--- a/seed/ts-sdk/audiences/.mock/definition/foo.yml
+++ b/seed/ts-sdk/audiences/.mock/definition/foo.yml
@@ -10,12 +10,16 @@ service:
     find:
       audiences:
         - public
+        - private
       path: ""
       method: POST
       request:
         name: FindRequest
         query-parameters:
-          optionalString: OptionalString
+          optionalString:
+            type: OptionalString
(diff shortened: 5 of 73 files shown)


packages/commons/core-utils/src/examples.ts
```diff
@@ -16,6 +16,10 @@ export const INT = 1;
 
 export const INT64 = 1000000;
 
+export const UINT = 1;
+
+export const UINT64 = 1;
```
dsinghvi, line 21: should we make these larger ?

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/belvo.test.ts.snap
```diff
@@ -32411,16 +32402,7 @@ In our documentation example, we use \`{endpoint}\` as a placeholder value. In p
               "pattern": "[1-9]\\d{0,9}",
             },
           },
-          {
-            "type": "integer",
-            "validation": {
-              "exclusiveMax": undefined,
-              "exclusiveMin": undefined,
-              "max": undefined,
-              "min": 1,
```
dsinghvi, line 32420: is losing this `min` value in the snapshots intentional ?

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/superagent.test.ts.snap
```diff
@@ -487,7 +487,14 @@ exports[`open api parser superagent docs 1`] = `
...
                   "authorization": "optional<map<string, unknown>>",
-                  "from_page": "optional<integer>",
+                  "from_page": {
+                    "default": [
+                      1,
+                    ],
+                    "docs": undefined,
+                    "type": "optional<integer>",
+                    "validation": undefined,
+                  },
```
dsinghvi, line 497: would be great if these `undefined` keys didnt show up in the generated fern definition to keep it as compact as possible

packages/cli/generation/ir-migrations/src/ir-versions/IrVersions.ts
```diff
@@ -43,8 +43,9 @@ export { FernIrV46 as V46 } from "@fern-fern/ir-v46-sdk";
 export { FernIrV47 as V47 } from "@fern-fern/ir-v47-sdk";
 export { FernIrV48 as V48 } from "@fern-fern/ir-v48-sdk";
 export { FernIrV49 as V49 } from "@fern-fern/ir-v49-sdk";
-export { FernIrV50 as V50 } from "@fern-fern/ir-v50-sdk";
 export * as V5 from "@fern-fern/ir-v5-model";
+export { FernIrV50 as V50 } from "@fern-fern/ir-v50-sdk";
+export { FernIrV51 as V51 } from "@fern-fern/ir-v51-sdk";
```
dsinghvi, line 48: @amckinney i don't see `irSerialization` updated -- is that something we want to add as well ?

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/belvo.test.ts.snap
amckinney: This is a gap in how we handle validation rules on `long` types (and others). Now that we're correctly recognizing that this type is a `long`, we're dropping the rules. I'm planning to open an issue so that we can come back to this in a commit tailored to this.

packages/commons/core-utils/src/examples.ts
amckinney: Yea, good catch - I'll update to the same number we use for `INT64`.

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/superagent.test.ts.snap
amckinney: Nice catch, it looks like we've had this issue for a while, but I'll fix it here ([link](https://github.com/fern-api/fern/blob/9ae2f2c75b659551205812148eaa1c38ec2e8165/packages/cli/openapi-ir-to-fern/src/buildTypeReference.ts#L458)).

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/belvo.test.ts.snap
dsinghvi: awesome -- mind linking the linear ticket once you file

dsinghvi (approved):

PR #4104, opened 2024-07-23
feat: introduce oas extensions for oauth (client credentials)

armandobelardo:
Fixes FER-2444

The schema effectively takes the schema for the Fern Definition, except endpoints are specified as `METHOD URL`.

The idea is you can explicitly specify the schema in your `generators.yml`:
```
auth-schemes:
 MyOAuth: 
   scheme: oauth
   type: "client-credentials"
   client-id-env: MY_CLIENT_ID
   client-secret-env: MY_CLIENT_ID
   token-prefix: Bearer
   get-token:
     endpoint: POST https://example.com/token
     request-properties:
       client-id: $request.client_id
       client-secret: $request.client_secret
       scopes: $request.scope
     response-properties:
       access-token: $response.access_token
       refresh-token: $response.refresh_token
   refresh-token:
      endpoint: POST /refresh
      request-properties:
        refresh-token: $request.refresh_token
      response-properties:
        access-token: $response.access_token
        refresh-token: $response.refresh_token
   scopes: ["scope1", "scope2"]

api:
  path: ./definition/openapi.yml
```

The idea here is to have a very Fern way to be able to back door the definition without having to cover every edge case in OAS.
For example, if there are multiple endpoints of different methods but the same path, there does not seem to be a way to specify that in OAS today (ex: `authorizationUrl: https://api.example.com/oauth2/authorize`).

Alternatively, trying to meet users where they are, they can just add the info we need to the scheme through these targeted extensions
Enrich the access + refresh token endpoints:
```
components:
  securitySchemes:
    oAuthSample:
      type: oauth2
      flows:
        clientCredentials:
        authorizationUrl: https://api.example.com/oauth2/authorize
        x-fern-access-token-endpoint:
          request:
            clientId: $query.client_id
            clientSecret: $query.client_secret
            scopes: $query.scopes
          response:
            accessToken: $response.access_token
            expiresIn: $response.access_token
            refreshToken: $response.refresh_token
        refreshUrl: https://api.example.com/oauth2/refresh
        x-fern-refresh-token-endpoint:
          request:
            clientId: $query.client_id
            clientSecret: $query.client_secret
            scopes: $query.scopes
          response:
            accessToken: $response.access_token
            expiresIn: $response.access_token
            refreshToken: $response.refresh_token

```


packages/cli/openapi-ir-to-fern/src/buildFernDefinition.ts
```diff
@@ -78,5 +77,7 @@ export function buildFernDefinition(context: OpenApiIrConverterContext): FernDef
         }
     }
 
+    buildAuthSchemes(context);
```
armandobelardo, line 80: Moved to the bottom because we need the endpoints to exist first

PR #4109, opened 2024-07-24
[FER-2381] CLI Forbidden Error Message Improvement

RohinBhargava:
## What's added:
 
- Improved some error messages for 403 issues; whether auth check failed or domain collision.

## What remains:

- There is some funkiness with error deserialization that needs to be addressed. For now, this is acceptable, but more coming soon.

## Test Case:
Screenshot 2024-07-23 at 9 56 15 PM


packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -273,6 +273,14 @@ function startDocsRegisterFailed(
             return context.failAndThrow(
                 "Please make sure that none of your custom domains are not overlapping (i.e. one is a substring of another)"
             );
+        case "UnauthorizedError":
+            return context.failAndThrow("Please make sure that your FERN_TOKEN is set.");
+        case "UserNotInOrgError":
+            return context.failAndThrow(
+                "Please verify if you have access to the organization you are trying to publish the docs to. If you are not a member of the organization, please reach out to the organization owner."
```
dsinghvi, line 280: I think linking `support@buildwithfern.com` here would be great

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -273,6 +273,14 @@ function startDocsRegisterFailed(
...
+            );
+        case "UnavailableError":
+            return context.failAndThrow("Failed to publish docs. Please try again later or reach out to Fern support.");
```
dsinghvi, line 283: would be great to notify in posthog or sentry if any of these are hit

dsinghvi (approved):

abvthecity (approved):

PR #4111, opened 2024-07-24
improvement, python: export the root client from the root init file

armandobelardo:
Fixes FER-2576

### DIFF
--- a/generators/python/src/fern_python/generators/sdk/client_generator/root_client_generator.py
+++ b/generators/python/src/fern_python/generators/sdk/client_generator/root_client_generator.py
@@ -178,14 +178,14 @@ class RootClientGenerator:
             source_file.add_arbitrary_code(AST.CodeWriter(self._write_default_param))
         source_file.add_class_declaration(
             declaration=class_declaration,
-            should_export=True,
+            should_export=False,
         )
         source_file.add_class_declaration(
             declaration=self._create_class_declaration(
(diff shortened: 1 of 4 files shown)


generators/python/src/fern_python/generators/sdk/client_generator/root_client_generator.py
```diff
@@ -163,9 +163,8 @@ def __init__(
     def generate(self, source_file: SourceFile) -> GeneratedRootClient:
         exported_client_class_name = self._context.get_class_name_for_exported_root_client()
         builder = RootClientGenerator.GeneratedRootClientBuilder(
-            module_path=self._context.get_module_path_in_project(
-                self._context.get_filepath_for_exported_root_client().to_module().path
-            ),
+            # HACK: This is a hack to get the module path for the root client to be from the root of the project
+            module_path=self._context.get_module_path_in_project(()),
```
dsinghvi, line 167: why is this a hack? what should it be instead ideally?

armandobelardo: Ideally the import resolver handles writing the import correctly given some relative position

So if we're outside the module it'll understand the import given it understands what is exported where

Like it's odd that we need to create a new class reference with a new path, when we've already created an instance above

It's not critical

dsinghvi (approved):

PR #4113, opened 2024-07-24
feat: introduce typeddicts for request objects

armandobelardo:
This PR makes it so that we create TypedDict clones for any object we have, e.g. aliases of objects, objects, unions and undisc unions. These clones are stored in a requests module alongside the other types, and are used in the method signatures for our SDK methods.

Generally the architecture is:
- `typeddict.py` - general construction and utilities for understanding if something can be a typeddict, snippets, etc.
- Whenever we get a `Reference` now, you can specify if that reference is used in a request -- if so, we then check if the SDK is using TypedDicts and handle retrieving the appropriate reference for the type (e.g. if it can be a typeddict, and you're in a request, and have typeddicts enabled, then we get the reference for the type in the `requests` module)
  - This happens through the plumbing of `as_request` and `use_typeddict_requests`
- Constructing TypedDicts side-saddles existing Pydantic model creation, see `object_generator.py`, etc. -- we effectively gather a list of fields and then pass that to the `pydantic_model.py`, as well as `typeddict.py` such that both model forms get created
- Snippets for TypedDicts side-saddle the existing snippet infra, similar to model construction, see `ObjectSnippetGenerator` in `object_generator.py` as an example

Seed run: https://github.com/fern-api/fern/pull/4128


generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
```diff
@@ -1081,6 +1082,18 @@ def write_ternary(writer: AST.NodeWriter) -> None:
 
                 reference = AST.Expression(AST.CodeWriter(write_ternary))
 
+        elif self._context.custom_config.pydantic_config.use_typeddict_requests and FernTypedDict.can_tr_be_typeddict(
+            query_parameter.value_type, self._context.get_types()
+        ):
+            # We don't need any optional wrappings for the coercion here.
```
dsinghvi, line 1088: what exactly is this doing ?

generators/python/src/fern_python/generators/sdk/context/sdk_generator_context.py
```diff
@@ -33,11 +33,16 @@ def __init__(
         self.generator_config = generator_config
         self.pydantic_generator_context = PydanticGeneratorContextImpl(
             ir=ir,
-            type_declaration_referencer=TypeDeclarationReferencer(skip_resources_module=custom_config.improved_imports),
+            type_declaration_referencer=TypeDeclarationReferencer(
+                skip_resources_module=custom_config.improved_imports,
+                use_typeddict_requests=custom_config.pydantic_config.use_typeddict_requests,
```
dsinghvi, line 38: Is it weird for the pydantic config to contain `use_typeddict_requests` -- this feels like it should be an only sdk config and at the top level

generators/python/src/fern_python/generators/sdk/declaration_referencers/environments_enum_declaration_referencer.py
```diff
@@ -8,11 +8,11 @@ def __init__(self, client_class_name: str, skip_resources_module: bool):
...
 
-    def get_filepath(self, *, name: None) -> Filepath:
+    def get_filepath(self, *, name: None, as_request: bool = False) -> Filepath:
         return Filepath(
             directories=(),
             file=Filepath.FilepathPart(module_name="environment"),
         )
 
-    def get_class_name(self, *, name: None) -> str:
+    def get_class_name(self, *, name: None, as_request: bool = False) -> str:
```
dsinghvi, line 17: My preference would be to update every call site so that you know you are not accidentally passing in the wrong value anywhere

generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
armandobelardo: This is unwrapping the optional off the type, so that we pass the underlying type to the serialization method, as opposed to the optional type

generators/python/src/fern_python/generators/sdk/declaration_referencers/environments_enum_declaration_referencer.py
armandobelardo: How do you mean?

generators/python/src/fern_python/generators/sdk/context/sdk_generator_context.py
armandobelardo: I can probably remove it easily, it was model construction related, so I put it there originally
But not wed to that at all

generators/python/src/fern_python/generators/sdk/declaration_referencers/environments_enum_declaration_referencer.py
dsinghvi: b/c we default `as_request: bool = False` it means that we won't have to update every callsite. If we did `as_request: bool` every function invocation would have to make sure it passes stuff in correctly. 

It's less convenient

generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
dsinghvi: Gotcha, will check the seed snapshot

generators/python/src/fern_python/generators/sdk/declaration_referencers/environments_enum_declaration_referencer.py
armandobelardo: Ah I see, yeah I can just not default the value
It's just a bigger diff is the only reason I defaulted

dsinghvi (approved):

PR #4118, opened 2024-07-25
fix: Add Stream Wrappers for use with various environments

RohinBhargava:
What's included:

- Adds three new StreamWrapper implementations which injects the correct deps depending on environment:
1. Node 18+ StreamWrapper that exposes methods for both fs.Writable  and WritableStream
2. Undici StreamWrapper that exposes methods for browser and cloudflare environments
3. Node 16-17 StreamWrapper that exposes methods for fs.Writable
4. Cloudflare CI Script that checks cloudflare can import SeedApiClient and instantiate it.
5. Browser CI Script that checks browser can import SeedApiClient and instantiate it.

I know we discussed removing 16-17, but this would lead to a bigger change.

Testing on all environments using ElevenLabs and Vellum, will include source code for the tests

Notes:

More on CI tests, if there are bad imports, the CI checks expect to fail, when inspecting the stdout logs, e.g. for cloudflare:
Screenshot 2024-07-26 at 1 05 20 PM

FER-2497

### DIFF
--- a/scripts/browser-startup-test.sh
+++ b/scripts/browser-startup-test.sh
@@ -48,7 +48,7 @@ npm install
 output=$(timeout 25s npm run start &)
 sleep 30
 echo "$output"
-if echo "$output" | grep -q "webpack compiled successfully"; then
+if echo "$output" | grep -q "Compiled successfully!"; then
     echo "Compiled successfully"
 else
     echo "Failed to compile"


generators/typescript/utils/core-utilities/fetcher/src/fetcher/getFetchFn.ts
```diff
@@ -20,6 +21,13 @@ export async function getFetchFn(): Promise<any> {
         return fetch;
     }
 
+    // Uncomment this block if you want to support Node.js 16, Node.js 17
     // Defaults to node `node-fetch` if global fetch isn't available
     return (await import("node-fetch")).default as any;
+
```
RohinBhargava, line 27: this is depending on whether or not we truly want to rip out 16 and 17

scripts/cloudflare-startup-test.sh
```diff
@@ -0,0 +1,60 @@
...
+
+export default {
+    async fetch(
+        request: Request,
+        env: Env,
+        ctx: ExecutionContext
+    ): Promise<Response> {
+        return new Response("Hello World!");
+    },
+};' > src/index.ts
```
dsinghvi, line 49: smart!

dsinghvi (approved):
looks great

PR #4127, opened 2024-07-26
java, fix: match java local config to publish config

dcb6:


generators/java/sdk/src/main/java/com/fern/java/client/JavaSdkDownloadFilesCustomConfig.java
```diff
@@ -31,6 +32,15 @@ public interface JavaSdkDownloadFilesCustomConfig extends DownloadFilesCustomCon
     @JsonProperty("client-class-name")
     Optional<String> clientClassName();
 
+    @JsonProperty("base-api-exception-class-name")
+    Optional<String> baseApiExceptionClassName();
+
+    @JsonProperty("base-exception-class-name")
+    Optional<String> baseExceptionClassName();
+
+    @JsonProperty("custom-dependencies")
```
dcb6, line 41: tossed this in here since it seemed like we might as well

dsinghvi (approved):

PR #4144, opened 2024-07-29
[c#, improvement]: add explicit namespaces to custom config

dcb6:
The C# generator now supports configuration to match namespaces directly to the full path of a file.
  This can lead to more imports being required to use the SDK, but can be helpful to avoid collisions.

  ```yml
    version: 0.3.2
    config:
      explicit-namespaces: true
  ```

### DIFF
--- a/seed/csharp-sdk/exhaustive/explicit-namespaces/.mock/definition/types/object.yml
+++ b/seed/csharp-sdk/exhaustive/explicit-namespaces/.mock/definition/types/object.yml
@@ -1,7 +1,9 @@
 types:
   ObjectWithOptionalField: #generic object that supports any type, makes it easier to use when testing
     properties:
-      string: optional<string>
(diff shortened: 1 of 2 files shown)


seed/csharp-sdk/exhaustive/explicit-namespaces/src/SeedExhaustive/NoAuth/NoAuthClient.cs
```diff
@@ -0,0 +1,38 @@
+using System.Net.Http;
+using SeedExhaustive.Core;
+
+#nullable enable
+
+namespace SeedExhaustive.NoAuth;
```
dcb6, line 6: example of diff here
previously would've been just 
`namespace SeedExhaustive;`

dsinghvi (approved):

PR #4147, opened 2024-07-29
fix: OSS workspace settings propogate to APIs with dependencies

armandobelardo:


packages/cli/workspace-loader/src/workspaces/FernWorkspace.ts
```diff
@@ -48,3 +50,113 @@ export class FernWorkspace extends AbstractAPIWorkspace<void> {
...
+    }
+
+    public async getDefinition(
+        { context }: { context?: TaskContext },
+        settings?: OSSWorkspace.Settings
+    ): Promise<FernDefinition> {
+        return (await this.toFernWorkspace({ context }, settings)).definition;
+    }
+
+    public async toFernWorkspace(
```
dsinghvi, line 102: @armandobelardo can we store a private member on this to see if this function was already called? if already called, would be great to not log anything

dsinghvi:
@armandobelardo can we add a `ete-test` with a fixture that loads from dependencies and snapshots the `LazyFernWorkspace` ?

packages/cli/workspace-loader/src/workspaces/FernWorkspace.ts
armandobelardo: Yeah def!

amckinney (approved):

PR #4151, opened 2024-07-30
c#, improvement: `set` instead of `init` field accessors in types

dcb6:
Types are now generated with `set` accessors rather than `init` accessors to improve flexibility of object construction.


dsinghvi (approved):
@dcb6 any downside of the `set` method ?

dcb6:
@dsinghvi no -- all `required` fields must still be provided at initial object initialization, so users must explicitly choose to put `null` or some other temp value if they wish to set it later user `set`. This feels ideal

PR #4156, opened 2024-07-30
fix: add images from frontmatter as well

RohinBhargava:
Adds images from frontmatter, with tags `image`, `og:image`, `og:logo`, `twitter:logo` where they are specified like:
```
image: "some_path.ext" or "https://someurl.com"
```

Screenshot 2024-07-30 at 5 33 26 PM

### DIFF
--- a/packages/cli/docs-markdown-utils/src/parseImagePaths.ts
+++ b/packages/cli/docs-markdown-utils/src/parseImagePaths.ts
@@ -382,12 +382,6 @@ function visitFrontmatterImages(
         if (value != null) {
             // realtime validation, this also assumes there can be other stuff in the object, but we only care about the valid keys
             if (typeof value === "object") {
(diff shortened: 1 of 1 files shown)


packages/cli/docs-markdown-utils/src/parseImagePaths.ts
```diff
@@ -35,6 +35,13 @@ export function parseImagePaths(
 
     const filepaths = new Set<AbsoluteFilePath>();
 
+    if (data.image) {
+        const resolvedPath = resolvePath(data.image, metadata);
+        if (resolvedPath != null) {
+            filepaths.add(resolvedPath);
+        }
+    }
```
dsinghvi, line 43: @RohinBhargava worth adding a test so that this functionality isnt dropped in the future?

abvthecity (approved):

PR #4157, opened 2024-07-30
(feature, csharp): Add support for allow-multiple query params

amckinney:
This adds support for `allow-multiple` query parameters, which are serialized in `explode` format (e.g. `&foo=one&foo=two`).

The change takes place in two files:
* The `RawClient.cs` core utility now supports `IEnumerable` collection types in the `BuildUrl` helper.
    * The `string` type requires special handling since it is the only primitive that implements the `IEnumerable` interface.
* The query parameters are attached to the `_query` parameter with simple LINQ expressions like the following:

```csharp
var _query = new Dictionary<string, object>() { };
_query["operand"] = request
    .Operand.Select(_value => JsonSerializer.Serialize(_value))
    .ToList();
```

> There's still some missing functionality here. As it stands, we don't support `deepObject` query parameters, so we'll likely need to revisit the `BuildUrl` implementation.

### DIFF
--- a/generators/csharp/sdk/src/wrapped-request/WrappedRequestGenerator.ts
+++ b/generators/csharp/sdk/src/wrapped-request/WrappedRequestGenerator.ts
@@ -142,6 +142,10 @@ export class WrappedRequestGenerator extends FileGenerator<CSharpFile, SdkCustom
             orderedFields.push({ name: exampleQueryParameter.name.name, value: exampleQueryParameter.value });
         }
 
+        for (const header of example.endpointHeaders) {
+            orderedFields.push({ name: header.name.name, value: header.value });
+        }
+
         example.request?._visit({
             reference: (reference) => {
                 orderedFields.push({ name: this.wrapper.bodyKey, value: reference });


seed/csharp-sdk/examples/src/SeedExamples/Service/ServiceClient.cs
```diff
@@ -55,14 +55,14 @@ public async Task<string> CreateMovieAsync(Movie request)
     public async Task<object> GetMetadataAsync(GetMetadataRequest request)
     {
         var _query = new Dictionary<string, object>() { };
+        _query["tag"] = request
+            .Tag.Where(_value => _value != null)
+            .Select(_value => _value)
+            .ToList();
```
dsinghvi, line 61: @amckinney is the idea here that we are filtering out null values?

dsinghvi: I see

dsinghvi (approved):

seed/csharp-sdk/examples/src/SeedExamples/Service/ServiceClient.cs
amckinney: Exactly - otherwise there are certain value formats that could throw exceptions like the one below:

```csharp
        // Without the .Where(...), the _value.Value access is unsafe.
        _query["maybeOperand"] = request
            .MaybeOperand.Where(_value => _value != null)
            .Select(_value => JsonSerializer.Serialize(_value.Value))
            .ToList();
```

I imagine we'll handle this in a separate component responsible for query parameter serialization like `qs` once we actually support `deepObject` query parameters.

seed/csharp-sdk/examples/src/SeedExamples/Service/Requests/GetMetadataRequest.cs
```diff
@@ -4,7 +4,7 @@ public record GetMetadataRequest
 {
     public bool? Shallow { get; set; }
 
-    public string? Tag { get; set; }
+    public IEnumerable<string?> Tag { get; set; } = new List<string?>();
```
dcb6, line 7: is there any reason to keep the `string`s optional?

seed/csharp-sdk/file-upload/src/SeedFileUpload/Service/ServiceClient.cs
```diff
@@ -41,11 +41,13 @@ public async Task JustFileAsync(JustFileRequet request)
...
     public async Task JustFileWithQueryParamsAsync(JustFileWithQueryParamsRequet request)
     {
-        var _query = new Dictionary<string, object>()
-        {
-            { "integer", request.Integer.ToString() },
-            { "listOfStrings", request.ListOfStrings },
-        };
+        var _query = new Dictionary<string, object>() { };
+        _query["integer"] = request.Integer.ToString();
+        _query["listOfStrings"] = request.ListOfStrings.Select(_value => _value).ToList();
```
dcb6, line 46: whats this bit doing?
`request.ListOfStrings.Select(_value => _value).ToList()`

seed/csharp-sdk/examples/src/SeedExamples/Service/Requests/GetMetadataRequest.cs
amckinney: Great catch - dropping the optional would be consistent with both [TypeScript](https://github.com/fern-api/fern/blob/173e2d44467772d8cddcd2726839b29da43abf4f/seed/ts-sdk/examples/examples-with-api-reference/src/api/resources/service/client/requests/GetMetadataRequest.ts#L15) and [Python](https://github.com/fern-api/fern/blob/3e5494da22fc7b77233ba8d36516fbc73a63128f/seed/python-sdk/examples/no-custom-config/src/seed/service/client.py#L155). I'll make that change here.

seed/csharp-sdk/file-upload/src/SeedFileUpload/Service/ServiceClient.cs
amckinney: Nice catch, we should just use the list directly in this case. The implementation of the `Select` call depends on the result of the stringify method, which could yield different expressions like so:

```csharp
_query["operand"] = request
    .Operand.Select(_value => JsonSerializer.Serialize(_value))
    .ToList();
```

We'll just need to add in some context so we can recognize whether or not we actually need the `Select` call at all. I'll add that in here, too.

dcb6: will leave it up to you if this feels worth the lift!

dcb6 (approved):
nice! 🚀

seed/csharp-sdk/file-upload/src/SeedFileUpload/Service/ServiceClient.cs
amckinney: Definitely worth the lift - this actually ended up cleaning up quite a bit!

PR #4161, opened 2024-07-31
internal: update IR to have the FDR API definition ID

armandobelardo:


packages/cli/docs-preview/src/previewDocs.ts
```diff
@@ -104,7 +104,7 @@ class ReferencedAPICollector {
 
         const dbApiDefinition = convertAPIDefinitionToDb(
             apiDefinition,
-            id,
+            ir.apiDefinitionId ?? id,
```
dsinghvi, line 107: hmm I think we'll want to keep this `id` otherwise its possible local preview may break

armandobelardo: Will remove, I wasn't positive if we'd need it but can def nix

PR #4166, opened 2024-07-31
(feature, csharp): Add RequestOptions

amckinney:
This introduces the `RequestOptions` type, which exposes the same set of properties as the `ClientOptions`, but does _not_ include any default values. That way, any option explicitly set by the user at the request call-site will take precedence over the options set on the client constructor.

For example,
```csharp
var user = client.GetUserAsync(
  new GetUserRequest {
    Username = "john.doe"
  },
  new RequestOptions {
      BaseUrl = "https://localhost:3000"
  }).Result;
```

This also updates the `TimeoutInSeconds` option (currently unused) to be a `TimeSpan`. This is more idiomatic and gives the user more control over exactly how they want to construct the time span before it's passed into the HTTP client.

### DIFF
--- a/seed/csharp-sdk/audiences/src/SeedAudiences/Foo/FooClient.cs
+++ b/seed/csharp-sdk/audiences/src/SeedAudiences/Foo/FooClient.cs
@@ -15,7 +15,7 @@ public class FooClient
         _client = client;
     }
 
-    public async Task<ImportingType> FindAsync(FindRequest request, RequestOptions? options)
+    public async Task<ImportingType> FindAsync(FindRequest request, RequestOptions? options = null)
     {
         var _query = new Dictionary<string, object>() { };
         if (request.OptionalString != null)
--- a/seed/csharp-sdk/basic-auth/src/SeedBasicAuth/BasicAuth/BasicAuthClient.cs
+++ b/seed/csharp-sdk/basic-auth/src/SeedBasicAuth/BasicAuth/BasicAuthClient.cs
@@ -17,7 +17,7 @@ public class BasicAuthClient
     /// <summary>
     /// GET request with basic auth scheme
     /// </summary>
-    public async Task<bool> GetWithBasicAuthAsync(RequestOptions? options)
+    public async Task<bool> GetWithBasicAuthAsync(RequestOptions? options = null)
     {
         var response = await _client.MakeRequestAsync(
             new RawClient.JsonApiRequest
@@ -39,7 +39,7 @@ public class BasicAuthClient
     /// <summary>
     /// POST request with basic auth scheme
     /// </summary>
-    public async Task<bool> PostWithBasicAuthAsync(object request, RequestOptions? options)
+    public async Task<bool> PostWithBasicAuthAsync(object request, RequestOptions? options = null)
     {
         var response = await _client.MakeRequestAsync(
             new RawClient.JsonApiRequest
--- a/generators/csharp/sdk/src/endpoint/EndpointGenerator.ts
+++ b/generators/csharp/sdk/src/endpoint/EndpointGenerator.ts
@@ -54,7 +54,8 @@ export class EndpointGenerator {
         parameters.push(
             csharp.parameter({
                 type: csharp.Type.optional(csharp.Type.reference(this.context.getRequestOptionsClassReference())),
-                name: this.context.getRequestOptionsParameterName()
+                name: this.context.getRequestOptionsParameterName(),
+                initializer: "null"
             })
         );
         const return_ = this.getEndpointReturnType({ endpoint });
--- a/seed/csharp-sdk/alias/src/SeedAlias/SeedAliasClient.cs
+++ b/seed/csharp-sdk/alias/src/SeedAlias/SeedAliasClient.cs
@@ -19,7 +19,7 @@ public partial class SeedAliasClient
         );
     }
 
-    public async Task GetAsync(string typeId, RequestOptions? options)
+    public async Task GetAsync(string typeId, RequestOptions? options = null)
     {
         await _client.MakeRequestAsync(
             new RawClient.JsonApiRequest
--- a/seed/csharp-sdk/api-wide-base-path/src/SeedApiWideBasePath/Service/ServiceClient.cs
+++ b/seed/csharp-sdk/api-wide-base-path/src/SeedApiWideBasePath/Service/ServiceClient.cs
@@ -19,7 +19,7 @@ public class ServiceClient
         string serviceParam,
         string resourceParam,
         int endpointParam,
-        RequestOptions? options
+        RequestOptions? options = null
     )
     {
         await _client.MakeRequestAsync(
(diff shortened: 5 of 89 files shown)


generators/csharp/sdk/src/options/RequestOptionsGenerator.ts
```diff
@@ -0,0 +1,132 @@
...
+});
+
+export const TIMEOUT_IN_SECONDS = csharp.field({
+    access: "public",
+    name: "TimeoutInSeconds",
+    get: true,
+    init: true,
+    type: csharp.Type.optional(csharp.Type.integer()),
+    summary: "The timeout for the request in seconds."
+});
```
dsinghvi, line 42: I think its worth sharing some of these fields directly with [ClientOptionsGenerator](https://github.com/fern-api/fern/blob/main/generators/csharp/sdk/src/client-options/ClientOptionsGenerator.ts)

We probably don't want the doc strings to get out of sync for example

generators/csharp/sdk/src/options/RequestOptionsGenerator.ts
```diff
@@ -0,0 +1,132 @@
...
+    name: "MaxRetries",
+    get: true,
+    init: true,
+    type: csharp.Type.optional(csharp.Type.integer()),
+    summary: "The http client used to make requests."
+});
+
+export const TIMEOUT_IN_SECONDS = csharp.field({
+    access: "public",
+    name: "TimeoutInSeconds",
```
dsinghvi, line 37: I didn't really put much thought in the `timeout` property when initially building, but probably makes more sense to align with the underlying http client? 

https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=net-8.0

generators/csharp/sdk/src/options/RequestOptionsGenerator.ts
```diff
@@ -0,0 +1,132 @@
...
+                        name: ENVIRONMENT_FIELD_NAME,
+                        get: true,
+                        init: true,
+                        useRequired: defaultEnvironment != null,
+                        type: csharp.Type.optional(csharp.Type.reference(this.context.getEnvironmentsClassReference())),
+                        summary: "The Environment for the API."
+                    });
+                },
+                _other: () => undefined
+            });
```
dsinghvi, line 116: sorry in advance for the repetitive comment, but this logic should definitely be shared with the RequestOptionsGenerator

amckinney: Sounds good - main reason I copied was because the types were different and so they could evolve independently (i.e. `RequestOptions` will have `CancellationToken`, but `ClientOptions` will not). Happy to add in a common file that both import from wherever possible.

generators/csharp/sdk/src/options/RequestOptionsGenerator.ts
amckinney: Great call here, especially because the option hasn't been used yet. Updated to use `System.TimeSpan`.

generators/csharp/sdk/src/options/RequestOptionsGenerator.ts
amckinney: Refactored this with a `BaseOptionsGenerator`, so that we DRY up as much as possible.

seed/csharp-sdk/alias/src/SeedAlias/Core/ClientOptions.cs
```diff
@@ -22,7 +23,7 @@ public partial class ClientOptions
     public int MaxRetries { get; init; } = 2;
 
     /// <summary>
-    /// The timeout for the request in seconds.
+    /// The timeout for the request.
     /// </summary>
-    public int TimeoutInSeconds { get; init; } = 30;
+    public TimeSpan Timeout { get; init; } = new TimeSpan(0, 0, 30);
```
dcb6, line 28: nit: thoughts on using `TimeSpan.FromSeconds(30);` syntax? bit more legible at a glance

amckinney: Great call, will do 👍

seed/csharp-sdk/audiences/src/SeedAudiences/Foo/FooClient.cs
```diff
@@ -15,7 +15,7 @@ public FooClient(RawClient client)
         _client = client;
     }
 
-    public async Task<ImportingType> FindAsync(FindRequest request)
+    public async Task<ImportingType> FindAsync(FindRequest request, RequestOptions? options)
```
dcb6, line 18: i think we wanna do `RequestOptions? options = null` so that it can be omitted completely

seed/csharp-sdk/basic-auth/src/SeedBasicAuth/BasicAuth/BasicAuthClient.cs
```diff
@@ -17,14 +17,15 @@ public BasicAuthClient(RawClient client)
...
     /// GET request with basic auth scheme
     /// </summary>
-    public async Task<bool> GetWithBasicAuthAsync()
+    public async Task<bool> GetWithBasicAuthAsync(RequestOptions? options)
     {
         var response = await _client.MakeRequestAsync(
             new RawClient.JsonApiRequest
             {
-                BaseUrl = _client.Options.BaseUrl,
+                BaseUrl = options?.BaseUrl ?? _client.Options.BaseUrl,
```
dcb6, line 25: if we're passing in the entire `requestOptions`, can we move this logic into `rawclient`?

seed/csharp-sdk/multi-url-environment-no-default/src/SeedMultiUrlEnvironmentNoDefault/Core/RequestOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+#nullable enable
+
+namespace SeedMultiUrlEnvironmentNoDefault.Core;
+
+public partial class RequestOptions
+{
+    /// <summary>
+    /// The Environment for the API.
+    /// </summary>
+    public SeedMultiUrlEnvironmentNoDefaultEnvironment? Environment { get; init; }
```
dcb6, line 14: This should be baseurl

seed/csharp-sdk/basic-auth/src/SeedBasicAuth/BasicAuth/BasicAuthClient.cs
amckinney: Discussed offline - 

Now that we're refactoring the environment-based option as a simple BaseURL, we can actually apply all the options within the `RawClient` (we previously weren't able to without exposing the environment on the `RequestOptions` type).

seed/csharp-sdk/audiences/src/SeedAudiences/Foo/FooClient.cs
amckinney: I think we already get that without the `= null` since it's specified as an optional (via `?`) - I wrote up a test case to confirm.

amckinney: I stand corrected - we should add that in here!

dcb6 (approved):
🙌

PR #4172, opened 2024-08-01
c# improvement: text responses + inlined request body inheritance

dcb6:


seed/csharp-sdk/extends/src/SeedExtends/Requests/Inlined.cs
```diff
@@ -0,0 +1,17 @@
+using System.Text.Json.Serialization;
+
+#nullable enable
+
+namespace SeedExtends;
+
+public record Inlined
```
dcb6, line 7: inlined request body inheritance example

seed/csharp-sdk/plain-text/src/SeedPlainText/Service/ServiceClient.cs
```diff
@@ -26,6 +26,10 @@ public async Task GetTextAsync(RequestOptions? options = null)
             }
         );
         var responseBody = await response.Raw.Content.ReadAsStringAsync();
+        if (response.StatusCode is >= 200 and < 400)
```
dcb6, line 29: text response example

dsinghvi (approved):

PR #4173, opened 2024-08-01
fix: add literal properties back to typeddict snippets

armandobelardo:


generators/python/core_utilities/sdk/base_test_utilities.py
```diff
@@ -100,9 +102,12 @@ def validate_response(response: typing.Any, json_expectation: typing.Any, type_e
         assert len(response) == len(json_expectation), "Length mismatch, expected: {0}, Actual: {1}".format(
             len(response), len(json_expectation)
         )
+        content_expectation = type_expectations
+        if isinstance(type_expectations, tuple):
+            content_expectation = type_expectations[1]
         for idx, item in enumerate(response):
             validate_response(
-                response=item, json_expectation=json_expectation[idx], type_expectations=type_expectations[idx]
+                response=item, json_expectation=json_expectation[idx], type_expectations=content_expectation[idx]
```
dsinghvi, line 110: what is changing in this test? hard to grok by just looking at the code

dsinghvi (approved):
+1 to unblock, but would be valuable to create a ticket if the ete tests are not catching this

PR #4176, opened 2024-08-01
(fix): add tests for alias extends

dsinghvi:
This PR adds a test definition for extending alias types. This works in C#, TypeScript, Ruby and Go Model. However the other generators such as Go SDK, Python SDK and Java SDK do not handle this edge case.


dsinghvi:
@dcb6 @armandobelardo @amckinney I've created linear tickets to handle this edge case in the generators

PR #4183, opened 2024-08-01
(fix, csharp): Support circular references for undiscriminated unions

amckinney:


amckinney:
Closing this for now - we'll need to revisit this once we generate type aliases for undiscriminated unions in general.

PR #4184, opened 2024-08-01
fix, python: Optional and aliased literals are populated in snippets

armandobelardo:
Fixes FER-2371 and FER-2374

Also fixes an issue found with a customer where httpx returns `''` instead of None when no content is received, so the tests have been updated to reflect that


generators/python/src/fern_python/generators/context/pydantic_generator_context_impl.py
```diff
@@ -207,3 +230,26 @@ def maybe_get_type_ids_for_type_reference(
...
+                alias=lambda alias: self.unwrap_example_type_reference(alias.value),
+                enum=lambda _: example_type_reference,
+                object=lambda _: example_type_reference,
+                union=lambda _: example_type_reference,
+                undiscriminated_union=lambda _: example_type_reference,
+            ),
+            container=lambda container: container.visit(
+                list_=lambda _: example_type_reference,
+                set_=lambda _: example_type_reference,
+                optional=lambda optional: self.unwrap_example_type_reference(optional.optional) if optional.optional is not None else example_type_reference,
```
amckinney, line 250: Weird that we need the `None` check at all - when would `optional.optional` be `None` when the container is represented as `optional`? 

Not actionable here since the IR is typed this way, but I wonder if we should have made this required?

```yaml
  ExampleOptionalContainer:
    properties:
      optional: optional<ExampleTypeReference>
      valueType: TypeReference
```

amckinney (approved):

generators/python/src/fern_python/generators/context/pydantic_generator_context_impl.py
armandobelardo: I guess it makes sense cause you want to make it easy to say you have an optional type, but no example? Agree that it is a bit odd though

PR #4188, opened 2024-08-02
(feat, fastapi): introduce endpoint specific async handlers in fastapi

dsinghvi:
**Feature**: The FastAPI `async_handlers` configuration now accepts a list of string or a boolean so that users
  can turn on async handlers for specific endpoints. For example, to turn on async handlers for a `users.create`
  endpoint, use the configuration below:

  ```yaml
  - name: fernapi/fern-fastapi-server
    version: 1.1.1
    config:
      async_handlers:
        - users.create
  ```


generators/python/fastapi/VERSION
```diff
@@ -1 +1 @@
-1.1.0-rc0
+1.1.1
```
armandobelardo, line 1: Nit: 1.2.0? Just cause it's a feature not a fix

armandobelardo (approved):
LGTM! 
The endpoint list isn't the easiest to maintain as your endpoints grow and change, but not sure how else we'd do it without modifying the fern def which I know we don't want do do here

PR #4198, opened 2024-08-04
fix, python + ts: additional template bugs

armandobelardo:
Before this PR:
- The union templates had the incorrect location
- The certain fields in python were missing a location 
- We were not making literals as union members
- TS was not correctly making object all the time

Of note:
The undiscriminated union template takes a location directly on it, the idea here is the resolver grabs that payload and passes it into the members to score and resolve those templates.
That means that the members themselves need to be relative templates (relative to the payload grabbed from the union template itself), and the union itself needs to pass in the payload at it's given location (e.g. what JSON object represents all the union members)


seed/python-sdk/enum/strenum/snippet-templates.json
```diff
@@ -159,15 +160,16 @@
                                             "less_than": "\"less_than\""
                                         },
                                         "templateInput": {
-                                            "location": "BODY",
-                                            "path": "maybeOperandOrColor"
+                                            "location": "RELATIVE",
+                                            "path": null
```
dsinghvi, line 164: Is this correct?

seed/python-sdk/undiscriminated-unions/snippet-templates.json
```diff
@@ -46,7 +46,7 @@
                                         "templateString": "$FERN_INPUT",
                                         "templateInputs": [
                                             {
-                                                "location": "BODY",
+                                                "location": "RELATIVE",
```
dsinghvi, line 49: Whats the difference between body and relative ?

seed/python-sdk/enum/strenum/snippet-templates.json
armandobelardo: Yep!

seed/python-sdk/undiscriminated-unions/snippet-templates.json
armandobelardo: This should be documented in FDR but basically it's for convenience to declare something as relative with a passed in payload, vs giving the full path within the object

It's mostly used for iterable types to say this is the path within this chunk

But for unions it's nice because you can debug the specific payload easier

dsinghvi (approved):

seed/python-sdk/enum/strenum/snippet-templates.json
armandobelardo: Added a comment to the body to explain why

PR #4201, opened 2024-08-05
(feat): cli caches api dependencies

dsinghvi:
Rather than reoloading dependencies again and again, they are now cached in the `~/.fern` folder as well as the `LazyFernWorkspace`.

### DIFF
--- a/packages/cli/workspace-loader/src/loadDependency.ts
+++ b/packages/cli/workspace-loader/src/loadDependency.ts
@@ -157,7 +157,6 @@ async function validateVersionedDependencyAndGetDefinition({
     settings?: OSSWorkspace.Settings;
 }): Promise<FernDefinition | undefined> {
     const pathToDependency: AbsoluteFilePath = getPathToLocalStorageDependency(dependency);
-    console.log("pathToDependency", pathToDependency);
     const pathToDefinition = join(pathToDependency, RelativeFilePath.of(DEPENDENCIES_FOLDER_NAME));
     const pathToMetadata = join(pathToDependency, RelativeFilePath.of(METADATA_RESPONSE_FILENAME));
 


packages/cli/workspace-loader/src/loadDependency.ts
```diff
@@ -1,14 +1,17 @@
 import { dependenciesYml } from "@fern-api/configuration";
+import { VersionedDependency } from "@fern-api/configuration/src/dependencies-yml";
```
armandobelardo, line 2: Mega nit - looks like this might not be exported from the top level index.ts

packages/cli/workspace-loader/src/loadDependency.ts
```diff
@@ -135,83 +157,98 @@ async function validateVersionedDependencyAndGetDefinition({
...
-            s3PreSignedReadUrl: response.body.definitionS3DownloadUrl,
-            absolutePathToLocalOutput: pathToDependency
-        });
-    } catch (error) {
-        context.failWithoutThrowing("Failed to download API", error);
-        return undefined;
+        metadata = response.body;
+        await writeFile(pathToMetadata, JSON.stringify(metadata));
+    } else {
+        metadata = JSON.parse((await readFile(pathToMetadata)).toString());
```
armandobelardo, line 241: Is there any validation, similar to the above that needs to be done after reading in the dir? I'd imagine no since it's version specific, so should be used immediately

armandobelardo (commented):
Looks good overall! Only question really is: do we need some sort of clean up mechanism? I might've missed this, but maybe in loadDependency if you're on a different version than what's on disk the cli can clear the dir?

packages/cli/workspace-loader/src/loadDependency.ts
dsinghvi: Yup no validation needs to be done

armandobelardo (approved):

PR #4202, opened 2024-08-05
wip, improvement: improve python .dict perf

armandobelardo:
Creating as a draft since I need to test on Candid to see if this actually impacts perf

Have not run seed yet as I imagine this code will change


armandobelardo:
Replaced by https://github.com/fern-api/fern/pull/4302

PR #4219, opened 2024-08-06
(feat, python): move to ruff for formatting

dsinghvi:
This PR: 
1. All `SourceFileImpls` that are created in a project are no longer formatted via black. They are formatted via ruff at the end in one big go. 
2. Everytime `SourceFileFactory.create_snippet()` is called the user has to pass in `should_format`. Now we ensure that respect the `skip_formatting` flag. 

This does not modify `WriterImpl` to format via ruff since there is work required to make that functionality accessible via python.


generators/python/src/fern_python/cli/publisher.py
```diff
@@ -13,12 +13,20 @@ class Publisher:
...
+        should_format: bool,
         generator_exec_wrapper: GeneratorExecWrapper,
         generator_config: GeneratorConfig,
     ):
+        self._should_format = should_format
         self._generator_exec_wrapper = generator_exec_wrapper
         self._generator_config = generator_config
 
+    def run_ruff_format(self) -> None:
+        self._run_command(
```
armandobelardo, line 25: Does this always get called after `poetry install` if not you might want to call that before to make sure ruff is there to be run

generators/python/src/fern_python/generator_cli/readme_snippet_builder.py
```diff
@@ -99,7 +99,7 @@ def _expression_to_snippet_str(
         self,
         expr: AST.Expression,
     ) -> str:
-        snippet = SourceFileFactory.create_snippet()
+        snippet = SourceFileFactory.create_snippet(should_format=True)
```
armandobelardo, line 102: I assume the answer is no, but should we respect the format config here?

generators/python/src/fern_python/snippet/snippet_template_factory.py
```diff
@@ -100,7 +100,7 @@ def _expression_to_snippet_str(
         self,
         expr: AST.Expression,
     ) -> str:
-        snippet = SourceFileFactory.create_snippet()
+        snippet = SourceFileFactory.create_snippet(should_format=True)
```
armandobelardo, line 103: These low key could respect the flag too since we have the resolve with formatting call within the template resolver lib

armandobelardo (approved):

PR #4221, opened 2024-08-06
(fix): support base properties when filtering for audiences

dsinghvi:
This PR makes sure that when we compute dependent types for unions, we also visit base properties.

### DIFF
--- a/packages/cli/generation/ir-generator/src/converters/type-declarations/getReferencedTypesFromRawDeclaration.ts
+++ b/packages/cli/generation/ir-generator/src/converters/type-declarations/getReferencedTypesFromRawDeclaration.ts
@@ -60,7 +60,7 @@ export function getReferencedTypesFromRawDeclaration({
                     return types;
                 }, [])
             );
-            return types;;
+            return types;
         },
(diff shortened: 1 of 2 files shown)


packages/cli/generation/ir-generator/src/converters/type-declarations/getReferencedTypesFromRawDeclaration.ts
```diff
@@ -43,13 +43,24 @@ export function getReferencedTypesFromRawDeclaration({
...
         discriminatedUnion: (unionDeclaration) => {
-            return Object.values(unionDeclaration.union).reduce<string[]>((types, singleUnionType) => {
-                const rawType = typeof singleUnionType === "string" ? singleUnionType : singleUnionType.type;
-                if (typeof rawType === "string") {
-                    types.push(rawType);
-                }
-                return types;
-            }, []);
+            const types: string[] = [];
+            if (unionDeclaration["base-properties"] != null) {
```
dsinghvi, line 47: we previously didn't visit `base-properties`

RohinBhargava (approved):

PR #4222, opened 2024-08-06
ir: add `shape` to `ExampleQueryParameter`

dcb6:
This PR introduces `shape` to `ExampleQueryParameter` to make the type self-sufficient for snippet generation.

### DIFF
--- a/packages/cli/docs-resolver/src/ApiReferenceNodeConverter.ts
+++ b/packages/cli/docs-resolver/src/ApiReferenceNodeConverter.ts
@@ -656,16 +656,19 @@ export class ApiReferenceNodeConverter {
         if (playgroundSettings) {
             return {
                 environments:
-                    playgroundSettings.environments &&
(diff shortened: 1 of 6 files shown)


packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
```diff
@@ -399,6 +399,20 @@ types:
     properties:
       name: commons.NameAndWireValue
       value: types.ExampleTypeReference
+      shape: optional<ExampleQueryParameterShape>
+
+  ExampleQueryParameterShape:
+    union:
+      single: SingleExampleQueryParameter
+      exploded: ExplodedExampleQueryParameter
+      commaSeparated: CommaSeparatedExampleQueryParameter
```
amckinney, line 408: This is great - for now we could even just use `{}` for each of these:

```yaml
  ExampleQueryParameterShape:
    union:
      single: {}
      exploded: {}
      commaSeparated: {}
```

amckinney (approved):

PR #4223, opened 2024-08-07
(feature): Add support for `.proto` inputs

amckinney:
This adds support for `.proto` files as input. This is handled via a new `APIDefinitionSchema` type which can be any of the following:

* OpenAPI
* AsyncAPI
* Protobuf

For now, we compile the Protobuf files locally, but we eventually plan to handle this an endpoint call, so we include a `local-generation` flag to switch on so we reserve the default behavior for remote generation (supported in a later change).

Note that this also updates all of the `seed` workflows by splitting out the setup stage into a callable `setup-seed.yml` workflow. This file is unused for now since it needs to be available on `main`, but its setup for a simple refactor later on.

### DIFF
--- a/packages/cli/workspace-loader/src/protobuf/ProtobufOpenAPIGenerator.ts
+++ b/packages/cli/workspace-loader/src/protobuf/ProtobufOpenAPIGenerator.ts
@@ -78,13 +78,13 @@ export class ProtobufOpenAPIGenerator {
         cwd: AbsoluteFilePath;
         target: RelativeFilePath;
     }): Promise<AbsoluteFilePath> {
-        const command = createLoggingExecutable("command", {
+        const which = createLoggingExecutable("which", {
             cwd,
             logger: this.context.logger
         });
 
         try {
-            await command(["-v", "buf"]);
+            await which(["buf"]);
         } catch (err) {
             this.context.failAndThrow(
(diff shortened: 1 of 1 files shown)


packages/cli/workspace-loader/src/protobuf/ProtobufOpenAPIGenerator.ts
```diff
@@ -0,0 +1,103 @@
...
+            cwd,
+            logger: this.context.logger
+        });
+        await buf(["config", "init"]);
+        await buf(["generate", target]);
+        return join(cwd, RelativeFilePath.of(PROTOBUF_GENERATOR_OUTPUT_FILEPATH));
+    }
+
+    private async generateRemote(): Promise<AbsoluteFilePath> {
+        throw new Error("Remote Protobuf generation is unimplemented.");
```
dsinghvi, line 101: can we throw a loggable fern cli error here ?

dsinghvi: or you need to fail the task using `context`

dsinghvi: contex.failAndThrow("....")

dsinghvi (approved):

Issue #4135, opened 2024-07-28
[Bug] `fern check` does not detect recursive examples


dsinghvi:
### Describe the bug

`fern check` does not detect infinitely recursive types. 

### To reproduce

Create the following yaml file and run `fern check`. 

```typescript
types: 
  Node: 
   properties: 
    child: Node
```
Here is a fern definition with the repro [recursive.zip](https://github.com/user-attachments/files/16405121/recursive.zip)

### Expected behavior

The CLI should emit an error along the lines of `Node is a cyclic type` and then printing out the cycle `Node -> node`. 

### Additional context

This should be added as an additional validation rule in the [validation](https://github.com/fern-api/fern/tree/main/packages/cli/yaml/validator/src/rules) directory.

Schnides123:
Hey, mind if I hop on this one? I think I have an idea of how you could detect cycles for this

dannysheridan:
> Hey, mind if I hop on this one? I think I have an idea of how you could detect cycles for this

Absolutely & welcome to Fern! We review contributor PRs within 5 business days and are always up for chatting through questions you have here in the comments.

Schnides123:
Draft PR is open [here](https://github.com/fern-api/fern/pull/4224). One thing I'm noticing in the E2E testing is that there are some recursive types already defined in test files, [like this one.](https://github.com/fern-api/fern/blob/802920876bfac24780eefb265102248a91f9749e/packages/cli/ete-tests/src/tests/ir/fixtures/simple/fern/definition/imdb.yml#L62) How do we want to update this?

PR #4224, opened 2024-08-07
fix(cli): Add validation rule for recursively-defined types (Issue #4135)

Schnides123:
Adds rule and related utilities for detecting & marking cyclicly-defined types.


packages/cli/generation/ir-generator/src/utils/getCyclicTypes.ts
```diff
@@ -0,0 +1,203 @@
...
+
+interface TypeParams {
+    keys: string[];
+    name: TypeName;
+    type: ResolvedType;
+    filePath: RelativeFilePath;
+}
+
+/**
+ * This function will return the first found path to a named child type from a given type. It will only search within the file of the given type.
```
dsinghvi, line 48: > It will only search within the file of the given type.

Seems like a big limitation, no?

dsinghvi: https://github.com/fern-api/fern/blob/main/packages/cli/yaml/validator/src/rules/no-duplicate-declarations/no-duplicate-declarations.ts#L15 Is an example of a rule that is able to check for name conflicts across files (admittedly much easier)

packages/cli/generation/ir-generator/src/utils/getCyclicTypes.ts
```diff
@@ -0,0 +1,203 @@
...
+        : unboxObjectType(args, resolvedType);
+
+    return isUnion
+        ? checkUnionChildTypes(args, typeProperties.flatMap(getTypeParams))
+        : checkObjectChildTypes(args, typeProperties.flatMap(getTypeParams));
+}
+
+function unboxObjectType(args: CyclicTypeArgs, t: ResolvedType.Named): TypeOrProperty[] {
+    if (isRawObjectDefinition(t.declaration)) {
+        return getAllPropertiesForType(args) as TypeOrProperty[];
```
dsinghvi, line 87: why do we have to use a cast?

packages/cli/generation/ir-generator/src/utils/getCyclicTypes.ts
```diff
@@ -0,0 +1,203 @@
...
+        return values.flatMap(getUnionResolvedTypes);
+    } else if (isRawUndiscriminatedUnionDefinition(union.declaration)) {
+        const declaration = union.declaration as RawSchemas.UndiscriminatedUnionSchema;
+        return declaration.union.flatMap(getUnionResolvedTypes);
+    }
+    return [];
+}
+
+function getTypeParams(t: TypeOrProperty): TypeParams | null {
+    const isObjectProperty = (t as ObjectPropertyWithPath).propertyType !== undefined;
```
dsinghvi, line 123: nit: would prefer a type guard instead of doing the cast twice

packages/cli/generation/ir-generator/src/utils/getCyclicTypes.ts
Schnides123: > > It will only search within the file of the given type.
> 
> 
> 
> Seems like a big limitation, no? 

It's not necessary for this use-case, since the existing rule for cyclic imports would already prevent any cyclic types across multiple files. 

Since that narrows our search to only cycles defined within a single file, we can skip walking to any other files and allow the rule to catch them when the main visitor for the rules reaches them normally.

packages/cli/generation/ir-generator/src/utils/getCyclicTypes.ts
Schnides123: Not really necessary, but the linter on my editor likes casting to the union object rather than returning an array of elements which are either the type object or the properties object. I think it makes it fairly easy to parse, but it should run fine without it if we want to remove the cast here.

packages/cli/generation/ir-generator/src/utils/getCyclicTypes.ts
Schnides123: Makes sense, I'll clean that up on the next version.

packages/cli/generation/ir-generator/src/utils/getCyclicTypes.ts
Schnides123: took it out in the latest revision, should be all good now

PR #4228, opened 2024-08-07
improvement: allow naming for asyncapi messages to pull message name

armandobelardo:


packages/cli/openapi-parser/src/options.ts
```diff
@@ -9,11 +9,14 @@ export interface ParseOpenAPIOptions {
     useTitlesAsName: boolean;
     /* The audiences that the OpenAPI must be filtered down into */
     audiences: string[] | undefined;
+    /* TODO: pick a better name for this flag, and maybe move it to AsyncAPI-only config */
+    useImprovedMessageNaming: boolean;
```
dsinghvi, line 13: yup would be great to put this in a `ParseAsyncAPIOptions` and expose that as well

dsinghvi: `naming: v2` or something ?

packages/cli/openapi-parser/src/asyncapi/parse.ts
```diff
@@ -229,22 +234,37 @@ function convertMessageToSchema({
...
 }: {
     breadcrumbs: string[];
     generatedName: string;
     event: AsyncAPIV2.PublishEvent | AsyncAPIV2.SubscribeEvent;
     context: AsyncAPIV2ParserContext;
+    options: ParseOpenAPIOptions;
 }): SchemaWithExample | undefined {
     if (event.message.oneOf != null) {
         const subtypes: (OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject)[] = [];
+        const prefixes: (string | undefined)[] = [];
```
dsinghvi, line 248: would be simpler to type this as `string[]` and just prevent code paths from appending undefined values i _think_ otherwise we are at risk of a code path that goes `arr.join()` and having undefined show up in the name

dsinghvi (approved):

PR #4237, opened 2024-08-08
feat(cli): openapi support assigning servers to an environment

armandobelardo:
Exposes an `x-fern-server` to add both name and env, and any additional config we want in the future
```yaml
 servers:
   - url: https://api.example.com
     x-fern-server:
        name: api
        environment: Production
```

### DIFF
--- a/packages/cli/openapi-ir-to-fern/src/__test__/fixtures/x-fern-servers/openapi.yml
+++ b/packages/cli/openapi-ir-to-fern/src/__test__/fixtures/x-fern-servers/openapi.yml
@@ -1,0 +1,25 @@
+openapi: 3.1.0
+info:
+  title: Test extension `x-fern-enum` for enums
+  version: 1.0.0
+paths: {}
+servers:
+  - url: https://api.example.com
+    x-fern-server-name: api
+    x-fern-environment: Production
+  - url: https://auth.example.com
+    x-fern-server-name: auth
+    x-fern-environment: Production
+  - url: https://api-sandbox.example.com
+    x-fern-server-name: api
+    x-fern-environment: Sandbox
+  - url: https://auth-sandbox.example.com
+    x-fern-server-name: auth
+    x-fern-environment: Sandbox
+
+components: 
+  schemas: 
+    SpeechModel:
+      type: string
+      description: The speech model to use for the transcription.
+      x-fern-sdk-group-name: transcripts
--- a/packages/cli/openapi-ir-to-fern/src/__test__/fixtures/x-fern-servers/openapi.yml
+++ b/packages/cli/openapi-ir-to-fern/src/__test__/fixtures/x-fern-servers/openapi.yml
@@ -5,17 +5,21 @@ info:
 paths: {}
 servers:
   - url: https://api.example.com
-    x-fern-server-name: api
-    x-fern-environment: Production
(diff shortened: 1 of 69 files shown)


packages/cli/openapi-ir-to-fern/src/__test__/fixtures/x-fern-servers/openapi.yml
```diff
@@ -0,0 +1,25 @@
...
+    x-fern-environment: Production
+  - url: https://auth.example.com
+    x-fern-server-name: auth
+    x-fern-environment: Production
+  - url: https://api-sandbox.example.com
+    x-fern-server-name: api
+    x-fern-environment: Sandbox
+  - url: https://auth-sandbox.example.com
+    x-fern-server-name: auth
+    x-fern-environment: Sandbox
```
dsinghvi, line 18: Do these top level server urls also support `x-fern-server` ? 
```
servers:
  - url: https://api.example.com
    x-fern-server: 
      name: api # should be optional 
      environment: Production
```

packages/cli/openapi-parser/src/openapi/v3/converters/convertServer.ts
```diff
@@ -3,14 +3,44 @@ import { OpenAPIV3 } from "openapi-types";
 import { getExtension } from "../../../getExtension";
 import { FernOpenAPIExtension } from "../extensions/fernExtensions";
 
+interface ServerConfig {
+    name: string;
+    environment: string | undefined;
+}
```
dsinghvi, line 9: are you up to build this as a zod schema and then instead of `getExtension` we can call `getExtensionAndValidate` and pass the schema in?

dsinghvi: And then we can make `environment` required (Production/Staging) and `name` optional (`Auth` vs `AI`). This breaks the pattern with `x-fern-server-name` but I think you are moving it in the right direction.

packages/cli/openapi-parser/src/openapi/v3/extensions/fernExtensions.ts
```diff
@@ -11,7 +11,38 @@ export const FernOpenAPIExtension = {
...
+     *     x-fern-server-name: api
+     *     x-fern-environment: Production
+     *   - url: https://auth.example.com
+     *     x-fern-server-name: auth
+     *     x-fern-environment: Production
+     * paths:
+     *   /path/to/my/endpoint:
+     */
     SERVER_NAME_V2: "x-fern-server-name",
+    SERVER_ENVIRONMENT: "x-fern-environment",
```
dsinghvi, line 31: would be down to delete and have everyone move to: 

```
servers: 
  - url: https://api.example.com
    x-fern-server: 
       environment: Production
       // can eventually add more stuff like docs
```

packages/cli/openapi-ir-to-fern/src/__test__/fixtures/x-fern-servers/openapi.yml
armandobelardo: Yeah, will update the test def to include this one! Good catch

dsinghvi:
@armandobelardo going to close this out since we now support this in `generators.yml`

PR #4238, opened 2024-08-08
custom segment write key

abarrell:


packages/cli/configuration/fern/definition/docs.yml
```diff
@@ -20,13 +20,26 @@ types:
...
+    properties:
+      writeKey: string
       
   DocsConfiguration:
     properties:
       instances: list<DocsInstance>
       title:
         type: optional<string>
         docs: used as tab bar title, and in the navbar if no logo is defined
+      analyticsConfig:
```
dsinghvi, line 38: nit: can we call this `analytics` so that users in the `docs.yml` just have to write out `analytics` 

```
analytics:
  segment: 
     write-key: "..."
```

packages/cli/configuration/fern/definition/docs.yml
```diff
@@ -20,13 +20,26 @@ types:
...
       - ts
       - js
+
+  AnalyticsConfig:
+    properties:
+      segment: optional<SegmentConfig>
+
+  SegmentConfig:
+    properties:
+      writeKey: string
```
dsinghvi, line 30: every user facing config we have is typically kebab-case

abvthecity (approved):

dsinghvi (changes requested):
Before we get this in would be great to add some docs about this too that we can send out! (https://buildwithfern.com/learn/docs/getting-started/quickstart maybe a new analytics page)

dsinghvi (approved):
thank you!

PR #4241, opened 2024-08-09
(feat, cli): add `has-next-page` property to IR

dsinghvi:
This PR introduces a `has-next-page` property to IR for offset pagination. If the response from an offset paginated contains a boolean about whether or not the next page exists, then this will allow the SDK to compute if the next page exists without making an RPC call.


packages/cli/generation/ir-generator/src/converters/services/convertOffsetPagination.ts
```diff
@@ -29,6 +29,14 @@ export async function convertOffsetPagination({
...
         step:
+            paginationPropertyComponents.step != null
+                ? await propertyResolver.resolveRequestPropertyOrThrow({
+                      file,
+                      endpoint: endpointName,
+                      propertyComponents: paginationPropertyComponents.step
+                  })
+                : undefined,
+        hasNextPage:
             paginationPropertyComponents.step != null
```
amckinney, line 40: This should be `hasNextPage`, right?

packages/ir-sdk/fern/apis/ir-types-latest/changelog/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v53.5.0] - 2024-08-05 \*\* (TODO: Make required in next major)
+
+- Feature: Support a `hasOffset
```
amckinney, line 10: nit: `hasNextPage`

packages/cli/yaml/validator/src/rules/valid-pagination/validateOffsetPagination.ts
```diff
@@ -124,3 +134,40 @@ function isValidOffsetType({ resolvedType }: { resolvedType: ResolvedType | unde
...
+        }
+    });
+}
+
+function isValidHasNextPageType({ resolvedType }: { resolvedType: ResolvedType | undefined }): boolean {
+    const primitiveType = maybePrimitiveType(resolvedType);
+    if (primitiveType == null) {
+        return false;
+    }
+    return primitiveType === "BOOLEAN";
```
amckinney, line 172: I could see us eventually supporting `string` so that it can be used similar to an empty cursor token, but let's start with this for now, especially because there aren't any IR changes required.

amckinney (approved):
Looks great overall! The change in `convertOffsetPagination.ts` is the only blocker, but approving in the meantime.

packages/cli/generation/ir-generator/src/converters/services/convertOffsetPagination.ts
dsinghvi: thank you yup

PR #4242, opened 2024-08-09
(fix, docs): filter referenced subpackages appropriately

dsinghvi:
Previously we were confusing subpackage locators + ids when traversing through `referencedSubpackages`. Now, we don't confuse the 2 and have renamed some methods to make the difference between the two identifiers more transparent.


packages/cli/docs-resolver/src/ApiReferenceNodeConverter.ts
```diff
@@ -258,17 +260,21 @@ export class ApiReferenceNodeConverter {
...
-            if (subpackage == null) {
-                this.taskContext.logger.error(`Subpackage ${subpackageId} not found in ${this.apiDefinitionId}`);
-                return false;
-            }
-            return true;
-        });
+        const subpackageIds = section.referencedSubpackages
+            .map((locator) => {
+                const subpackage = this.#holder.getSubpackageByIdOrLocator(locator);
+                return subpackage != null ? ApiDefinitionHolder.getSubpackageId(subpackage) : undefined;
```
dsinghvi, line 266: critical change

PR #4245, opened 2024-08-09
feat (wip): add playground settings for API playground

RohinBhargava:
Adds ability to add playground settings in docs.yaml:
```
playground:
  allowedEnvironments:
    - Prod
    - Dev
```

These can be nested under API reference, packages, and endpoints.

### DIFF
--- a/packages/cli/docs-markdown-utils/package.json
+++ b/packages/cli/docs-markdown-utils/package.json
@@ -1,51 +1,51 @@
 {
- "name": "@fern-api/docs-markdown-utils",
- "version": "0.0.0",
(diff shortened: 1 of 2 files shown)


abvthecity (approved):

dsinghvi:
Hey @RohinBhargava --  what if an environment id is used that doesn't exist?

RohinBhargava:
In the FE, the fallback is to use the first environment in the list of environments by endpoint. In Candid's case, this would be the production URL, so the right fix might be to block CI if environmentid is not specified in api.yml?

PR #4260, opened 2024-08-11
c#, improvements: breaking change with several small improvements

dcb6:
## Changelog
- Break: The C# SDK is now on major version 1. To take this upgrade without any breaks, please add the following configuration to your `generators.yml` file:

```yml
generators:
    config:
      root-namespace-for-core-classes: false
      pascal-case-environments: false
      simplify-object-dictionaries: false
```

- Improvement: Core classes that are exposed publicly are now in the root namespace rather than the Core namespace.
- Improvement: Types that were previously generated as `Dictionary<string, object?>` are now simplified to just `object`.
- Improvement: Environment names are now Pascal-cased.
- Feature: Generating specific error types can now be turned off via the following config:

```yml
generators:
    config:
      generate-error-types: false
```


seed/csharp-model/examples/src/SeedExamples/Types/ExtendedMovie.cs
```diff
@@ -37,5 +37,5 @@ public record ExtendedMovie
     public string? Book { get; set; }
 
     [JsonPropertyName("metadata")]
-    public Dictionary<string, object?> Metadata { get; set; } = new Dictionary<string, object?>();
+    public object Metadata { get; set; } = new Dictionary<string, object?>();
```
dcb6, line 40: example of Dictionary simplification

seed/csharp-sdk/api-wide-base-path/src/SeedApiWideBasePath/Core/Public/ClientOptions.cs
```diff
@@ -3,7 +3,7 @@
 
 #nullable enable
 
-namespace SeedBasicAuth.Core;
+namespace SeedApiWideBasePath;
```
dcb6, line 6: example of public `Core` classes now in the root namespace

they're also in `Core/Public`, inspired by [Stripe](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Constants/ApiVersion.cs)

seed/csharp-sdk/multi-url-environment/no-pascal-case-environments/src/SeedMultiUrlEnvironment/Core/Public/SeedMultiUrlEnvironmentEnvironment.cs
```diff
@@ -0,0 +1,28 @@
+namespace SeedMultiUrlEnvironment;
+
+public class SeedMultiUrlEnvironmentEnvironment
+{
+    public static SeedMultiUrlEnvironmentEnvironment PRODUCTION =
```
dcb6, line 5: example of break-avoiding environment casing configuration applied

seed/csharp-sdk/exhaustive/no-root-namespace-for-core-classes/src/SeedExhaustive/Core/Public/ClientOptions.cs
dcb6, line 1: example of avoiding core namespace break via config

seed/csharp-sdk/optional/no-simplify-object-dictionaries/src/SeedObjectsWithImports/Optional/OptionalClient.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    private RawClient _client;
+
+    internal OptionalClient(RawClient client)
+    {
+        _client = client;
+    }
+
+    public async Task<string> SendOptionalBodyAsync(
+        Dictionary<string, object?>? request,
```
dcb6, line 19: example of dictionary break avoided via config

dsinghvi (approved):

PR #4269, opened 2024-08-12
(chore): Remove all yarn files

amckinney:
Following up on the `pnpm` migration.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -29,8 +29,8 @@ jobs:
           POSTHOG_API_KEY: ${{ secrets.POSTHOG_PROJECT_API_KEY }}
         run: |
           pnpm --filter @fern-api/cli dist:cli:prod
-          cli_path="$(pnpm --filter @fern-api/cli bin fern:prod)"
-          node packages/cli/cli/dist/prod/bundle.cjs --docs --log-level debug --instance fern.docs.buildwithfern.com/learn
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-docs.yml
```diff
@@ -26,10 +26,10 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 18
-          cache: "yarn"
+          cache: "pnpm"
 
       - name: Install
-        run: yarn install
+        run: pnpm install
```
dsinghvi, line 32: can we use the install action now

.github/workflows/version-check.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 18
-          cache: "yarn"
+          cache: "pnpm"
```
dsinghvi, line 22: can we use the install workflow

.github/workflows/publish-docs.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           POSTHOG_API_KEY: ${{ secrets.POSTHOG_PROJECT_API_KEY }}
         run: |
           pnpm --filter @fern-api/cli dist:cli:prod
-          cli_path="$(yarn workspace @fern-api/cli bin fern:prod)"
+          cli_path="$(pnpm --filter @fern-api/cli bin fern:prod)"
```
dsinghvi, line 32: we can delete this line

PR #4286, opened 2024-08-12
(feature, cli): Upload source files to S3

amckinney:
This adds the `SourceUploader` type which is used to upload source files to S3. This is the opposite end of the `SourceFetcher` introduced in https://github.com/fern-api/fern/pull/4270, which can be used to download the sources within the generator.

### DIFF
--- a/packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
+++ b/packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
@@ -183,7 +183,9 @@ async function computeSemanticVersion({
     return response.body.version;
 }
 
-function convertToFdrApiDefinitionSources(sources: IdentifiableSource[]): Record<FdrAPI.api.v1.register.SourceId, FdrAPI.api.v1.register.Source> {
+function convertToFdrApiDefinitionSources(
+    sources: IdentifiableSource[]
+): Record<FdrAPI.api.v1.register.SourceId, FdrAPI.api.v1.register.Source> {
     return Object.fromEntries(
         Object.values(sources).map((source) => [
             source.id,


packages/cli/generation/remote-generation/remote-workspace-runner/src/SourceUploader.ts
```diff
@@ -0,0 +1,141 @@
...
+import { loggingExeca } from "@fern-api/logging-execa";
+import { InteractiveTaskContext } from "@fern-api/task-context";
+import { FernWorkspace, IdentifiableSource } from "@fern-api/workspace-loader";
+import { FernRegistry as FdrAPI } from "@fern-fern/fdr-cjs-sdk";
+import { readFile, unlink } from "fs/promises";
+import tmp from "tmp-promise";
+
+const PROTOBUF_ZIP_FILENAME = "proto.zip";
+
+export type SourceType = "asyncapi" | "openapi" | "protobuf";
```
dsinghvi, line 12: does this work for a fern definition?

packages/cli/generation/remote-generation/remote-workspace-runner/src/SourceUploader.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    constructor(context: InteractiveTaskContext, workspace: FernWorkspace) {
+        this.context = context;
+        this.sources = Object.fromEntries(workspace.getSources().map((source) => [source.id, source]));
+        this.sourceTypes = new Set<SourceType>(Object.values(this.sources).map((source) => source.type));
+    }
+
+    public getSourceTypes(): Set<SourceType> {
+        return this.sourceTypes;
+    }
```
dsinghvi, line 27: nit: ive noticed that typescript classes often dont define a getter and just make the member public `public sourceTypes: Set<SourceType>` (i think its more idiomatic)

packages/cli/generation/remote-generation/remote-workspace-runner/src/SourceUploader.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    public getFdrApiDefinitionSources():
```
dsinghvi, line 29: can we make only the primary method that is used externally public, otherwise private?

packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
```diff
@@ -161,3 +177,24 @@ async function computeSemanticVersion({
...
+async function uploadSources({
+    context,
+    sourceUploader,
+    sourceUploads
+}: {
+    context: InteractiveTaskContext;
+    sourceUploader: SourceUploader;
+    sourceUploads: Record<FdrAPI.api.v1.register.SourceId, FdrAPI.api.v1.register.SourceUpload>;
+}): Promise<SourceConfig> {
+    if (sourceUploads == null && sourceUploader.getSourceTypes().has("protobuf")) {
```
dsinghvi, line 190: can `sourceUploads` even be null here? based on the type signature looks like not?

packages/cli/generation/remote-generation/remote-workspace-runner/src/SourceUploader.ts
```diff
@@ -0,0 +1,141 @@
...
+            return undefined;
+        }
+        return Object.fromEntries(
+            Object.entries(this.sources).map(([id, source]) => [
+                id,
+                {
+                    type: source.type === "protobuf" ? "proto" : source.type
+                }
+            ])
+        );
```
dsinghvi, line 42: can we annotate the return type as part of this function

packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
```diff
@@ -161,3 +177,24 @@ async function computeSemanticVersion({
...
+        // We only fail hard if we need to upload Protobuf source files. Unlike OpenAPI, these
+        // files are required for successful code generation.
+        context.failAndThrow("Did not successfully upload Protobuf source files.");
+    }
+
+    context.logger.debug(`Uploading source files ...`);
+    await sourceUploader.uploadSources(sourceUploads);
```
dsinghvi, line 197: thoughts on having this function just return the converted Fdr Source uploads?

packages/cli/generation/remote-generation/remote-workspace-runner/src/SourceUploader.ts
amckinney: No - the original `Source` is only present whenever we map from some OSS source into a Fern definition ([link](https://github.com/fern-api/fern/blob/d0a717de8a4776ddb949d93e8c31f603eca5ace0/packages/cli/workspace-loader/src/types/Workspace.ts#L25)). Otherwise we just assume the Fern definition file is the source.

The idea here was that we would only be uploading OpenAPI and Protobuf sources alongside the Fern definition files already included in the same register call.

If we want to separately upload the original Fern definition source files alongside the register call, we can follow-up with that change and plumb it through as another `Source`.

packages/cli/generation/remote-generation/remote-workspace-runner/src/SourceUploader.ts
amckinney: That's already the case / this is used publicly - it maps the sources into the format that FDR expects in the register call:

```typescript
const sourceUploader = new SourceUploader(interactiveTaskContext, workspace);
const sources = sourceUploader.getFdrApiDefinitionSources();

const apiDefinition = convertIrToFdrApi({ ir, snippetsConfig: {} });
const response = await fdr.api.v1.register.registerApiDefinition({
    orgId: organization,
    apiId: ir.apiName.originalName,
    definition: apiDefinition,
    sources
});
```

packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
amckinney: Sweet works for me 👍

packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
amckinney: Great catch - this was from before a refactor. Will update 👍

packages/cli/generation/remote-generation/remote-workspace-runner/src/SourceUploader.ts
amckinney: It's actually already annotated, it just ends up being formatted across multiple lines since it's a long type. An alternative would be to create a private alias for the type like so:

```typescript
type FdrSourceMap = Record<FdrAPI.api.v1.register.SourceId, FdrAPI.api.v1.register.Source>;

...

public getFdrApiDefinitionSources(): FdrSourceMap | undefined {
    if (Object.keys(this.sources).length === 0) {
        return undefined;
    }
    return Object.fromEntries(
        Object.entries(this.sources).map(([id, source]) => [
            id,
            {
                type: source.type === "protobuf" ? "proto" : source.type
            }
        ])
    );
}
```

Not sure it's worth it though. I like that this function handles the `undefined` case seamlessly within the function, but I'll see if I can refactor things with a `!= null` guard to prevent such a large signature in the first place.

amckinney: I was able to get rid of this entirely, so feel free to ignore 👍

dsinghvi: my b on this

packages/cli/generation/remote-generation/remote-workspace-runner/src/SourceUploader.ts
dsinghvi: That makes sense, would you be up to file a linear ticket for this on `Fern CLI` ?

dsinghvi (approved):

PR #4287, opened 2024-08-12
fix: read templated env vars in the docs generator config

pujitm:
Allows the cli docs generator (`generate --docs`) to read from environment variables templated with `${}`

- Refactors environment variable substitution into a core-util called `replaceEnvVariables`

### DIFF
--- a/fern/pages/docs/analytics/segment.mdx
+++ b/fern/pages/docs/analytics/segment.mdx
@@ -28,6 +28,6 @@ In your `docs.yml` file, add the Segment writeKey:
 ```yaml
 analytics:
   segment:
-    write-key: YOUR_SEGMENT_WRITE_KEY
+    write-key: ${SEGMENT_WRITE_KEY} # scans environment variable
 ```
 </CodeBlock>
--- a/packages/commons/core-utils/src/__tests__/replaceEnvVars.test.ts
+++ b/packages/commons/core-utils/src/__tests__/replaceEnvVars.test.ts
@@ -1,4 +1,4 @@
-import { vi } from 'vitest';
+import { vi } from "vitest";
 import { replaceEnvVariables } from "../replaceEnvVars";
 
 describe("replaceEnvVariables", () => {


packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForDocsWorkspace.ts
```diff
@@ -22,6 +23,26 @@ export async function runRemoteGenerationForDocsWorkspace({
...
+    // If the run is a preview, we'll substitute ALL environment variables as empty strings
+    //
+    // Otherwise, we'll attempt to read and replace the templated environment variable. Will
+    // bubble up an error if the env var isn't found.
+    //
+    // Although this logic is separate from generating a remote, placing it here helps us
+    // avoid making cascading changes to other workflows.
+    // docsWorkspace = substituteEnvVariables(docsWorkspace, context, { substituteAsEmpty: preview });
+    docsWorkspace = replaceEnvVariables(
+        docsWorkspace,
```
dsinghvi, line 38: i think you only want to do this on `docsWorkspace.config` (which is the contents of `docs.yml`)

packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForDocsWorkspace.ts
```diff
@@ -22,6 +23,26 @@ export async function runRemoteGenerationForDocsWorkspace({
...
+        // Wrap in a closure for correct binding of `this` downstream
+        { onError: (e) => context.failAndThrow(e) },
+        { substituteAsEmpty: preview }
+    );
+
+    console.log(docsWorkspace.config.analytics);
```
dsinghvi, line 44: we'll want to delete these console statements

packages/commons/core-utils/src/replaceEnvVars.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * If `substituteAsEmpty` is true, the variable is always replaced with an empty string, even if it is defined.
+ *
+ * `context.onError` is called when the environment variable is not defined and `substituteAsEmpty` is false.
+ * @param content
+ * @param context
+ * @param options
+ * @returns `content` with the templated variables substituted.
+ */
+export function replaceEnvVariables<T>(
```
dsinghvi, line 26: up to add a unit test for this?

pujitm:
I think I resolved everything @dsinghvi!

packages/commons/core-utils/jest.config.cjs
```diff
@@ -1,7 +1,7 @@
 /** @type {import('ts-jest').JestConfigWithTsJest} **/
 module.exports = {
   testEnvironment: "node",
-  testMatch: [ "**/__tests__/**/*.ts?(x)", "**/?(*.)+(spec|test).ts?(x)" ],
+  testMatch: [ "**/?(*.)+(spec|test).ts?(x)" ],
```
pujitm, line 4: Context: The unit test was failing because jest was trying to run the test's `*.d.ts` file from the compiled output in `lib`. I saw that `cli/workspace-runner` was using vitest and the vitest config was `include: ["**/*.{test,spec}.ts"]`, which obv doesn't include `.d.ts` files. 

Thought about using the same setup with vitest here in core-utils, but didn't want to add a dependency and then update pnpm-lock.yaml for something this minor.

PR #4288, opened 2024-08-12
(feat, typescript): support `hasNextPage`  property for offset pagination

dsinghvi:
This PR upgrades the TypeScript IR by a minor version and adds support for looking at a `hasNextPage` property for pagination.

### DIFF
--- a/generators/typescript/sdk/generator/src/contexts/SdkContextImpl.ts
+++ b/generators/typescript/sdk/generator/src/contexts/SdkContextImpl.ts
@@ -1,5 +1,5 @@
 import { FernGeneratorExec } from "@fern-fern/generator-exec-sdk";
-import { Constants, IntermediateRepresentation, TypeReference } from "@fern-fern/ir-sdk/api";
+import { Constants, IntermediateRepresentation } from "@fern-fern/ir-sdk/api";
 import {
     CoreUtilitiesManager,
     createExternalDependencies,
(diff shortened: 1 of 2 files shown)


seed/ts-sdk/pagination/src/api/resources/users/client/Client.ts
```diff
@@ -532,7 +532,7 @@ export class Users {
         let _offset = request?.page != null ? request?.page : 1;
         return new core.Pageable<SeedPagination.ListUsersPaginationResponse, SeedPagination.User>({
             response: await list(request),
-            hasNextPage: (response) => (response?.data ?? []).length > 0,
+            hasNextPage: (response) => response?.hasNextPage ?? (response?.data ?? []).length > 0,
```
dsinghvi, line 535: core change

amckinney: Sweet, deferring to the length check is great 👍

generators/typescript/sdk/generator/src/contexts/SdkContextImpl.ts
```diff
@@ -322,4 +322,21 @@ export class SdkContextImpl implements SdkContext {
             timeoutSdkErrorGenerator
         });
     }
+
+    public isRequired(type: TypeReference): boolean {
```
amckinney, line 326: What about required primitive types?

amckinney: Should we always just `return true` after the first `if`?

generators/typescript/sdk/generator/src/contexts/SdkContextImpl.ts
```diff
@@ -322,4 +322,21 @@ export class SdkContextImpl implements SdkContext {
...
+        if (type.type === "container" && type.container.type === "optional") {
+            return true;
```
amckinney, line 328: This should be `false` right?

dsinghvi: yup thats my b

amckinney (approved):

PR #4290, opened 2024-08-12
feat: allow namespacing an API from generators.yml

armandobelardo:
Fixes FER-2904

Here we add a `namespace` option to `api` to allow for specifying a record of APIs that will be namespaced (by it's key).
We add the `namespace` key to easily distinguish between the new option and the previous top level options (such as path, overrides, settings, etc.)

### DIFF
--- a/packages/cli/ete-tests/src/tests/write-definition/fixtures/namespaced/fern/generators.yml
+++ b/packages/cli/ete-tests/src/tests/write-definition/fixtures/namespaced/fern/generators.yml
@@ -1,4 +1,4 @@
 api:
-  namespaced:
+  namespaces:
     trains: ./openapi/trains/openapi.yml
     neopets: ./openapi/petstore/openapi.yml


fern/pages/sdks/features/merging-multiple-apis.mdx
```diff
@@ -56,18 +56,19 @@ If you have multiple API Definitions, you can specify them in `generators.yml` a
...
-    ```yaml {2,5}
+    ```yaml {3,6}
     api: 
-      v1: 
-        - path/to/petstore.yml
-        - path/to/plantstore.yml
-      v2: 
-        - path/to/petstore-v2.yml
-        - path/to/plantstore-v2.yml
+      namespaced:
```
dsinghvi, line 65: nit: can we call this `namespaces`

packages/cli/openapi-parser/src/openapi/v3/converters/operation/convertHttpOperation.ts
```diff
@@ -110,7 +114,7 @@ export function convertHttpOperation({
             operation.operationId != null && suffix != null
                 ? operation.operationId + "_" + suffix
                 : operation.operationId,
-        tags: operation.tags ?? [],
+        tags: tags.concat(operation.tags ?? []),
```
dsinghvi, line 117: should we use the `context.resolveTags` method here?

dsinghvi:
Talked offline -- should we add a test for websocket / couldn't find the code for taking the namespace into account

dsinghvi (approved):

PR #4297, opened 2024-08-13
(chore): add any-auth test definition

dsinghvi:
This PR adds an `any-auth` test definition to see what the compatibility is like in existing generators.


amckinney (approved):

test-definitions/fern/apis/any-auth/definition/auth.yml
```diff
@@ -0,0 +1,26 @@
...
+  TokenResponse:
+    docs: |
+      An OAuth token response.
+    properties:
+      access_token: string
+      expires_in: integer
+      refresh_token: optional<string>
+
+service:
+  auth: false
```
amckinney, line 11: Should we set at least one to `true`?

dsinghvi: Yup can do

PR #4298, opened 2024-08-13
(feature, csharp): Generate gRPC core utilities

amckinney:
This adds support for gRPC core utility generation for the C# SDK generator, which includes the `RawGrpcClient` and `GrpcRequestOptions` types.

This also makes a few other improvements along the way, including:
1. Add a `Ternary` AST node.
2. Export the `ClassInstantiation`'s `Argument` type as an independent type (so it can be used elsewhere).
3. Update the `AsIs` file writer to recognize files that belong in the new `Core/Public` package name.
4. Add a `skipDefaultInitializer` option to prevent specific collection fields from having a default initializer.

### DIFF
--- a/packages/cli/generation/ir-generator/src/__test__/test-definitions/grpc-proto.json
+++ b/packages/cli/generation/ir-generator/src/__test__/test-definitions/grpc-proto.json
@@ -473,54 +473,54 @@
     },
     "errors": {},
     "services": {
-        "service_user": {
+        "service_userservice": {
             "availability": null,
             "name": {
                 "fernFilepath": {
                     "allParts": [
                         {
-                            "originalName": "user",
+                            "originalName": "userservice",
                             "camelCase": {
-                                "unsafeName": "user",
-                                "safeName": "user"
+                                "unsafeName": "userservice",
+                                "safeName": "userservice"
                             },
                             "snakeCase": {
-                                "unsafeName": "user",
-                                "safeName": "user"
+                                "unsafeName": "userservice",
+                                "safeName": "userservice"
                             },
                             "screamingSnakeCase": {
-                                "unsafeName": "USER",
-                                "safeName": "USER"
+                                "unsafeName": "USERSERVICE",
+                                "safeName": "USERSERVICE"
                             },
                             "pascalCase": {
-                                "unsafeName": "User",
-                                "safeName": "User"
+                                "unsafeName": "Userservice",
+                                "safeName": "Userservice"
                             }
                         }
                     ],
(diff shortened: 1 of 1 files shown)


generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
```diff
@@ -118,6 +124,16 @@ export abstract class AbstractCsharpGeneratorContext<
         return [this.getNamespace(), ...this.getChildNamespaceSegments(fernFilepath)];
     }
 
+    public getNamespaceFromProtobufFileOrThrow(protobufFile: ProtobufFile): string {
+        const namespace = protobufFile.options?.csharp?.namespace ?? undefined;
```
dcb6, line 128: nit (and am curious): would behavior be any differernt if `?? undefined` was omitted?

generators/csharp/codegen/src/project/CsharpProject.ts
```diff
@@ -75,8 +80,22 @@ export class CsharpProject {
...
+        for (const filename of this.context.getCoreAsIsFiles()) {
+            this.coreFiles.push(
+                await this.createAsIsFile({
+                    filename,
+                    namespace: this.context.getCoreNamespace()
+                })
+            );
+        }
+
+        for (const filename of this.context.getPublicCoreAsIsFiles()) {
```
dcb6, line 92: thanks for this!

generators/csharp/sdk/src/SdkGeneratorContext.ts
```diff
@@ -87,6 +88,10 @@ export class SdkGeneratorContext extends AbstractCsharpGeneratorContext<SdkCusto
         ];
     }
 
+    public getPublicCoreAsIsFiles(): string[] {
+        return [AsIsFiles.GrpcRequestOptions];
```
dcb6, line 92: should this also be conditional on `context.hasGrpcEndpoints()`?

dcb6 (approved):
beautiful 🤩

generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
amckinney: d'oh! Great catch - no difference in behavior. I'll update!

generators/csharp/sdk/src/SdkGeneratorContext.ts
amckinney: Yes definitely, great catch 👍

PR #4302, opened 2024-08-14
improvement: improve `.dict` speed by limiting dict calls

armandobelardo:
Fixes FER-2799

This PR removes the second call to .dict we'd made in order to get "exclude unsets without defaults", which was to union the dicts from exclude_unset and exclude_none in Pydantic.
For larger models, the second .dict call would take a significant amount of time, and so here during `.dict` we:
- Get the list of fields set already in the model
- Traverse the models to see if there are fields not in the field set that have non-None defaults
- If there are, add them to the list
- Pass this list as `includes` within the single pydantic `.dict` call


dsinghvi (approved):
Looks like there is a single seed fixture failing on the python sdk?

PR #4319, opened 2024-08-16
(feature, csharp): Generate well-known types

amckinney:
## Overview

This adds support for the well-known Protobuf types (i.e. the `google.protobuf.Struct` and `google.protobuf.Value`). This requires a few new files and abstractions, including:

1. The `Core/ProtoConverter` is used to map the well-known types to their SDK-equivalent.
2. The `google.protobuf.Value` is generated as a `OneOf` for the possible set of values.
3. The `google.protobuf.Struct` is referenced as a `Dictionary<string, <google.protobuf.Value>>`

For example, suppose the `MetadataValue` type declares itself as a `google.protobuf.Value`:

```csharp
public class MetadataValue(
    OneOf<
        string,
        double,
        bool,
        IEnumerable<MetadataValue?>,
        Dictionary<string, MetadataValue?>
    > value
)
    : OneOfBase<
        string,
        double,
        bool,
        IEnumerable<MetadataValue?>,
        Dictionary<string, MetadataValue?>
    >(value)
{
    public static implicit operator MetadataValue(string value) => new(value);

    ...
}
```

```csharp
internal class ProtoConverter
{
    public static Struct ToProtoStruct(Dictionary<string, MetadataValue?> value)
    {
        var result = new Struct();
        foreach (var kvp in value)
        {
            result.Fields[kvp.Key] = ToProtoValue(kvp.Value);
        }
        return result;
    }
    ...
}
```

## Implementation

Along the way, this introduces a few new code generation utilities including:

1. The `Class.Operator` node (used to specify `implicit` / `explicit` class operators, which is useful for the `google.protobuf.Value`.
2. The `Switch` node which generates an idiomatic C# `switch` / `case`.
3. Added support for `Array` (e.g. `int[]`) and `ListType` (e.g. `List<int>`).
5. The `ProtobufResolver` type which exists on the `AbstractCsharpGeneratorContext` alongside the `CsharpTypeMapper`, so that it can be used across the entire generator.

### DIFF
--- a/generators/csharp/codegen/src/context/ProtobufResolver.ts
+++ b/generators/csharp/codegen/src/context/ProtobufResolver.ts
@@ -30,7 +30,7 @@ export class ProtobufResolver {
 
     public getProtobufStructTypeOrThrow(): csharp.Type {
         const resolvedType = this.getProtobufStructType();
-        if (resolvedType === undefined) {
+        if (resolvedType == null) {
             throw new Error(
                 "Well-known google.protobuf.Struct type could not be found, which is required by google.protobuf.Value."
             );
@@ -135,7 +135,7 @@ export class ProtobufResolver {
 
     private getProtobufTypeForTypeId(typeId: TypeId): ProtobufType | undefined {
         const typeDeclaration = this.ir.types[typeId];
-        if (typeDeclaration == null || typeDeclaration.source == null) {
+        if (typeDeclaration?.source == null) {
             return undefined;
         }
         return typeDeclaration.source.type === "proto" ? typeDeclaration.source.value : undefined;
--- a/seed/csharp-model/grpc-proto/src/SeedApi/Core/ProtoConverter.cs
+++ b/seed/csharp-model/grpc-proto/src/SeedApi/Core/ProtoConverter.cs
@@ -33,7 +33,7 @@ internal class ProtoConverter
     {
         if (value == null)
         {
-            Value.ForNull();
+            return Value.ForNull();
         }
         return value.Match<Value>(
             Value.ForString,
@@ -51,13 +51,11 @@ internal class ProtoConverter
     {
         return value.KindCase switch
         {
-            Value.KindOneofCase.StringValue => new MetadataValue(value.StringValue),
-            Value.KindOneofCase.NumberValue => new MetadataValue(value.NumberValue),
-            Value.KindOneofCase.BoolValue => new MetadataValue(value.BoolValue),
-            Value.KindOneofCase.ListValue
-                => new MetadataValue(value.ListValue.Values.Select(FromProtoValue).ToList()),
-            Value.KindOneofCase.StructValue
-                => new MetadataValue(FromProtoStruct(value.StructValue)),
+            Value.KindOneofCase.StringValue => value.StringValue,
+            Value.KindOneofCase.NumberValue => value.NumberValue,
+            Value.KindOneofCase.BoolValue => value.BoolValue,
+            Value.KindOneofCase.ListValue => value.ListValue.Values.Select(FromProtoValue).ToList(),
+            Value.KindOneofCase.StructValue => FromProtoStruct(value.StructValue),
             _ => null,
         };
     }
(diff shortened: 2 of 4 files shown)


generators/csharp/codegen/src/context/ProtobufResolver.ts
```diff
@@ -0,0 +1,143 @@
...
+            throw new Error(
+                `The 'csharp_namespace' file option must be declared in Protobuf file ${protobufFile.filepath}`
+            );
+        }
+        return namespace;
+    }
+
+    public getProtobufStructTypeOrThrow(): csharp.Type {
+        const resolvedType = this.getProtobufStructType();
+        if (resolvedType === undefined) {
```
dcb6, line 33: why `=== undefined` here instead of `== null`?

generators/csharp/codegen/src/context/ProtobufResolver.ts
```diff
@@ -0,0 +1,143 @@
...
+        const protobufType = this.getProtobufTypeForTypeId(typeId);
+        if (protobufType == null) {
+            return false;
+        }
+        return protobufType.type === "wellKnown" && protobufType.value.type === wellKnownProtobufType.type;
+    }
+
+    private getProtobufTypeForTypeId(typeId: TypeId): ProtobufType | undefined {
+        const typeDeclaration = this.ir.types[typeId];
+        if (typeDeclaration == null || typeDeclaration.source == null) {
```
dcb6, line 138: nit:
```suggestion
        if (typeDeclaration?.source == null) {
```

package.json
```diff
@@ -8,6 +8,7 @@
     "clean": "pnpm -r clean",
     "compile": "pnpm -r --no-bail compile",
     "test": "pnpm -r --filter='!@fern-api/ete-tests' --no-bail test",
+    "test:update": "pnpm -r --filter='!@fern-api/ete-tests' --no-bail test -- -u",
```
dcb6, line 11: nice, thanks!

seed/csharp-model/grpc-proto/src/SeedApi/Core/ProtoConverter.cs
```diff
@@ -0,0 +1,64 @@
...
+            result[kvp.Key] = FromProtoValue(kvp.Value);
+        }
+        return result;
+    }
+
+    public static Value ToProtoValue(MetadataValue? value)
+    {
+        if (value == null)
+        {
+            Value.ForNull();
```
dcb6, line 36: should it be:
```suggestion
            return Value.ForNull();
```

seed/csharp-model/grpc-proto/src/SeedApi/Core/ProtoConverter.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        return value.KindCase switch
+        {
+            Value.KindOneofCase.StringValue => new MetadataValue(value.StringValue),
+            Value.KindOneofCase.NumberValue => new MetadataValue(value.NumberValue),
+            Value.KindOneofCase.BoolValue => new MetadataValue(value.BoolValue),
+            Value.KindOneofCase.ListValue
+                => new MetadataValue(value.ListValue.Values.Select(FromProtoValue).ToList()),
+            Value.KindOneofCase.StructValue
+                => new MetadataValue(FromProtoStruct(value.StructValue)),
```
dcb6, line 60: I think we can simplify slightly to this now that we have the implicit operators!
```suggestion
            Value.KindOneofCase.StringValue => value.StringValue,
            Value.KindOneofCase.NumberValue => value.NumberValue,
            Value.KindOneofCase.BoolValue => value.BoolValue,
            Value.KindOneofCase.ListValue => value.ListValue.Values.Select(FromProtoValue).ToList(),
            Value.KindOneofCase.StructValue => FromProtoStruct(value.StructValue),
```

dcb6 (approved):
awesome 🚀🚀

generators/csharp/codegen/src/context/ProtobufResolver.ts
amckinney: Good catch, we should definitely use `== null` 👍

seed/csharp-model/grpc-proto/src/SeedApi/Core/ProtoConverter.cs
amckinney: Definitely, great catch - thanks for the keen eye

seed/csharp-model/grpc-proto/src/SeedApi/Core/ProtoConverter.cs
amckinney: Nice!! I didn't even think to use that here. The updated code looks great 👍

PR #4324, opened 2024-08-16
fix: fix seed, move unit test to right CoreUtility

RohinBhargava:
Previously, added a unit test to the wrong CoreUtility, which caused seed to break.

Moved `formDataWrapper.test.ts` to the FormDataUtil CoreUtility for proper exporting.

### DIFF
--- a/generators/typescript/sdk/CHANGELOG.md
+++ b/generators/typescript/sdk/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.39.3] - 2024-08-16
+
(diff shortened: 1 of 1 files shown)


seed/ts-sdk/file-upload/no-custom-config/src/core/form-data-utils/__private_test__/multipart.upload.test.ts
```diff
@@ -0,0 +1,101 @@
+import { Blob } from "buffer";
```
dsinghvi, line 1: @RohinBhargava dont think this should be getting copied?

RohinBhargava: yeah, ty for the walkthrough earlier: found this way more quickly: https://github.com/fern-api/fern/blob/b1018059a0ef6f5452cbcac1b962e1c71b64341c/generators/typescript/utils/commons/src/core-utilities/CoreUtilitiesManager.ts#L87

PR #4326, opened 2024-08-17
(feature, csharp): Generate .ToProto method

amckinney:
## Overview

This generates the `.ToProto` method for all the types that support Protobuf encoding. With this, types can now seamlessly map to their Protobuf-equivalent type, e.g.

```csharp
using Proto = Data.V1.Grpc;

public record UpdateRequest
{
    ...   

    internal Proto.UpdateRequest ToProto()
    {
        var result = new Proto.UpdateRequest();
        result.Id = Id;
        if (Values != null && Values.Any())
        {
            result.Values.AddRange(Values);
        }
        if (SetMetadata != null)
        {
            result.SetMetadata = ProtoConverter.ToProtoStruct(SetMetadata);
        }
        if (Namespace != null)
        {
            result.Namespace = Namespace ?? "";
        }
        if (IndexedData != null)
        {
            result.IndexedData = IndexedData.ToProto();
        }
        return result;
    }
}
```

## Implementation

The implementation is primarily handled in the new `CsharpProtobufTypeMapper` abstraction, which is included alongside the other utilities on the `AbstractCsharpGeneratorContext`.

Along the way, this also introduces and/or edits a few C# code generation utilities including:
1. Added support for namespace aliases (e.g. `Proto = Data.V1.Grpc` in the example above).
2. Added support for the `And` and `Or` nodes, which can be used to chain together multiple expressions into a single boolean condition.
3. Updated the `controlFlow` helper so that it's comatible with arbitrary nodes (e.g. the node created by (2)).
4. Added support for writing `internal` methods on the generated classes (i.e.. the `ToProto` method is only meant to be called by the generated code).

## Testing

This replaces the `grpc` test definition with `grpc-proto-exhaustive`, which exercises a significant number of the cases we'll encounter (if not all).

### DIFF
--- a/seed/csharp-model/grpc-proto-exhaustive/src/SeedApi/DescribeResponse.cs
+++ b/seed/csharp-model/grpc-proto-exhaustive/src/SeedApi/DescribeResponse.cs
@@ -19,6 +19,9 @@ public record DescribeResponse
     [JsonPropertyName("totalCount")]
     public uint? TotalCount { get; set; }
 
+    /// <summary>
+    /// Maps the DescribeResponse type into its Protobuf-equivalent representation.
+    /// </summary>
     internal Proto.DescribeResponse ToProto()
     {
         var result = new Proto.DescribeResponse();
--- a/generators/csharp/codegen/src/context/CsharpProtobufTypeMapper.ts
+++ b/generators/csharp/codegen/src/context/CsharpProtobufTypeMapper.ts
@@ -37,8 +37,9 @@ export class CsharpProtobufTypeMapper {
             name: "ToProto",
             access: "internal",
             isAsync: false,
-            return_: csharp.Type.reference(protobufClassReference),
+            summary: `Maps the ${protobufClassReference.name} type into its Protobuf-equivalent representation.`,
             parameters: [],
+            return_: csharp.Type.reference(protobufClassReference),
             body: csharp.codeblock((writer) => {
                 if (properties.length === 0) {
                     writer.write("return ");
--- a/seed/csharp-model/grpc-proto-exhaustive/src/SeedApi/Column.cs
+++ b/seed/csharp-model/grpc-proto-exhaustive/src/SeedApi/Column.cs
@@ -20,6 +20,9 @@ public record Column
     [JsonPropertyName("indexedData")]
     public IndexedData? IndexedData { get; set; }
 
+    /// <summary>
+    /// Maps the Column type into its Protobuf-equivalent representation.
+    /// </summary>
     internal Proto.Column ToProto()
     {
         var result = new Proto.Column();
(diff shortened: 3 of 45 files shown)


generators/csharp/model/src/object/ObjectGenerator.ts
```diff
@@ -58,6 +60,26 @@ export class ObjectGenerator extends FileGenerator<CSharpFile, ModelCustomConfig
...
+                            propertyName: this.getPropertyName({
+                                className: this.classReference.name,
+                                objectProperty: property.name
+                            }),
+                            typeReference: property.valueType
+                        };
+                    })
+                })
+            );
+        }
```
dsinghvi, line 81: nit: can we put this into a helper method

seed/csharp-model/grpc-proto-exhaustive/src/SeedApi/DescribeResponse.cs
```diff
@@ -0,0 +1,47 @@
...
+    [JsonPropertyName("dimension")]
+    public uint? Dimension { get; set; }
+
+    [JsonPropertyName("fullness")]
+    public float? Fullness { get; set; }
+
+    [JsonPropertyName("totalCount")]
+    public uint? TotalCount { get; set; }
+
+    internal Proto.DescribeResponse ToProto()
```
dsinghvi, line 22: another small nit: should we add a quick doc string at the top of this method

generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
```diff
@@ -266,6 +278,43 @@ export abstract class AbstractCsharpGeneratorContext<
...
+            boolean: () => csharp.codeblock("false"),
+            string: () => csharp.codeblock('""'),
+            date: () => csharp.codeblock("DateOnly.MinValue"),
+            dateTime: () => csharp.codeblock("DateTime.MinValue"),
+            uuid: () => csharp.codeblock('""'),
+            base64: () => csharp.codeblock('""'),
+            bigInteger: () => csharp.codeblock('""'),
+            _other: () => csharp.codeblock("null")
+        });
+    }
```
dsinghvi, line 316: is this proto specific?

dsinghvi (approved):

generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
amckinney: No, this actually represents the default value for all of these types. We just happen to only use it for Protobuf so far. The alternative is to simply use the `default` helper ([link](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values#default-value-expressions)), but I figured this looked a little more hand-written / natural. 

The only exception is `string` - technically the default value is `null`, but we actually need to use `""` for protobuf.

PR #4331, opened 2024-08-18
feat(csharp): Generate .FromProto method

amckinney:
Re: https://github.com/fern-api/fern/pull/4326

This generates the `.FromProto` method for all the types that support Protobuf encoding. With this, types can now seamlessly map _from_ their Protobuf-equivalent type, e.g.

```csharp
internal static UpdateResponse FromProto(Proto.UpdateResponse value)
{
    return new UpdateResponse
    {
        Columns = value.Columns?.ToDictionary(
            kvp => kvp.Key,
            kvp => Column.FromProto(kvp.Value)
        ),
        Namespace = value.Namespace,
        Usage = value.Usage != null ? Usage.FromProto(value.Usage) : null
    };
}
```

Note that this also upgrades the Dockerfile dependency to .NET 8 so that we're using the latest version (plus the same one used in CI).

### DIFF
--- a/generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
+++ b/generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
@@ -301,6 +301,10 @@ export abstract class AbstractCsharpGeneratorContext<
             case "container":
                 return typeReference.container.type === "optional";
             case "named": {
+                const typeDeclaration = this.getTypeDeclarationOrThrow(typeReference.typeId);
+                if (typeDeclaration.shape.type === "alias") {
+                    return this.isPrimitive(typeDeclaration.shape.aliasOf);
+                }
                 return false;
             }
             case "unknown":


generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
```diff
@@ -296,6 +296,20 @@ export abstract class AbstractCsharpGeneratorContext<
         }
     }
 
+    public isPrimitive(typeReference: TypeReference): boolean {
+        switch (typeReference.type) {
+            case "container":
+                return typeReference.container.type === "optional";
+            case "named": {
+                return false;
```
dsinghvi, line 304: should we be handling aliases to primitives here?

dsinghvi (approved):

generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
amckinney: Yes, great call - I should have noticed that in the `isOptional` helper above.

PR #4337, opened 2024-08-19
fix(cli): do not throw if the source could not be resolved

armandobelardo:


packages/cli/generation/ir-generator/src/resolvers/SourceResolver.ts
```diff
@@ -10,14 +11,19 @@ export interface SourceResolver {
...
     }) => Promise<ResolvedSource | undefined>;
-    resolveSourceOrThrow: (args: { source: RawSchemas.SourceSchema; file: FernFileContext }) => Promise<ResolvedSource>;
+    resolveSourceOrThrow: (args: {
+        source: RawSchemas.SourceSchema;
+        file: FernFileContext;
+    }) => Promise<ResolvedSource | undefined>;
 }
 
 export class SourceResolverImpl implements SourceResolver {
+    private readonly taskContext: TaskContext;
```
amckinney, line 21: nit: Can we name this property `context` for consistency with elsewhere? That way we just say:

```typescript
this.context.logger.warn(`Cannot resolve source ${source.openapi} from file ${file.relativeFilepath}`);
```

amckinney (approved):

packages/cli/generation/remote-generation/remote-workspace-runner/src/createAndStartJob.ts
```diff
@@ -148,6 +149,21 @@ async function createJob({
                 RelativeFilePath.of(PROJECT_CONFIG_FILENAME)
             );
             await writeFile(absolutePathToFernConfigJson, JSON.stringify(projectConfig.rawConfig, undefined, 2));
+            // write sources
+            const sources = workspace.getSources();
+            for (const source of sources) {
+                context.logger.debug(`[TEST] Writing source ${source.absoluteFilePath}`);
```
dsinghvi, line 155: nit: lets delete this

packages/cli/generation/remote-generation/remote-workspace-runner/src/createAndStartJob.ts
```diff
@@ -148,6 +149,21 @@ async function createJob({
...
+                const sourceContents = await readFile(source.absoluteFilePath);
+                const relativeLocation = relative(workspace.absoluteFilepath, source.absoluteFilePath);
```
dsinghvi, line 157: What does `relative` return if the source absolute filepath is outside of the fern folder?

dsinghvi: I wonder if we need to check the returned result to make sure lines 158-162 dont throw

armandobelardo: I've wrapped it all in a try catch

PR #4340, opened 2024-08-19
feat(csharp): Refactor RawGrpcClient as as-is file

amckinney:
This improves and adapts the `RawGrpcClient` generation in a few different ways:
1. Remove the generator in favor of an easier to maintain as-is file (i.e. `RawGrpcClient.Template.cs`).
2. Update the `Grpc()` method to be a `public readonly Grpc` property.
3. Refactor the `CreateCallOptions` method so that the `CancellationToken` is the last argument.
    * Both [Microsoft](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068) and [Stripe](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Infrastructure/Public/IStripeClient.cs#L47) do this, even if there is a `RequestOptions` type.
4. Remove the `CancellationToken` from the `GrpcRequestOptions` type (re: (2)).


dsinghvi (approved):
Blind +1

PR #4346, opened 2024-08-19
feat(internal): allow seed cli to register versions of generators and CLIs

armandobelardo:
Fixes FER-2914
Introduces version files that declare the release types consumed by FDR

This PR seeds examples for the latest versions of the CLI and python-sdk
```
packages/cli/cli/versions.yml
generators/python/sdk/versions.yml
```


.github/workflows/seed.yml
armandobelardo, line 1: The changes in this file are just adding the call to the `seed register` command once the seed checks have succeeded

packages/seed/fern/definition/config.yml
```diff
@@ -1,7 +1,15 @@
 imports:
   features: features.yml
 types:
+  MinimalSeedWorkspaceConfiguration:
```
armandobelardo, line 4: This is to allow the CLI to have it's own seed folder and `seed.yml` without any of the other metadata we currently track for seed

armandobelardo: Allowing us to add in additional configuration down the line, as well as specify a location for the changelog within another project

packages/seed/src/cli.ts
```diff
@@ -192,6 +197,60 @@ function addRunCommand(cli: Argv) {
...
+                .option("generator", {
+                    type: "array",
+                    string: true,
+                    demandOption: false,
+                    description: "Generator(s) to register"
+                })
+                .option("cli", {
+                    boolean: true,
+                    default: false,
+                    description: "Initialize a docs website."
```
dsinghvi, line 215: is this the right description?

packages/seed/fern/definition/config.yml
dsinghvi: Up to call this `CliSeedWorkspaceConfiguration` ? dont foresee us needing to add more so down to make the naming here hyperspecific

dsinghvi (approved):

PR #4347, opened 2024-08-19
chore(cli): Update the getLatestGeneratorVersion command to leverage the FDR API as opposed to Docker and dockerode

armandobelardo:
Fixes FER-2913

**What is this?**
Updates the underlying logic for `getLatestGeneratorVersion` which in turn impacts `fern generator upgrade`, `fern add <generator>` and `fern init`

**Why do this?**
The existing workflow (as outlined below) is flakey and doesn't even work in Lambdas, so fern-bot cannot execute `fern generator upgrade`
This PR moves to the new endpoints and DB we've added to FDR!

**Why is this safe?**
Note that this replaces the existing workflow which:
- relies on docker being installed
- relies on a tag on the docker container, which was not on every generator for a number of reasons

So the existing workflow is already a bit broken, this should be an improvement, even while the generators migrate to the V2 publishing + registering workflow, as I've also updated the fallback versions we provide when we cannot find one within FDR

### DIFF
--- a/.github/workflows/create-cli-release.yml
+++ b/.github/workflows/create-cli-release.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     paths:
-      - "packages/cli/cli/VERSION"
+      - "packages/cli/cli/versions.yml"
 
 # Add this permissions block
 permissions:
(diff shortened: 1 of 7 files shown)


dsinghvi:
Do we have an ETE test that runs  generator upgrade? (would hope that something is failing here until all of the generator versions are registered)

packages/cli/configuration/src/generators-yml/getGeneratorVersions.ts
```diff
@@ -1,38 +1,140 @@
...
+    //     case "fern-ruby-sdk":
+    //         return { language: "ruby", generatorType: { type: "sdk" } };
+    //     case "fern-ruby-model":
+    //         return { language: "ruby", generatorType: { type: "model" } };
+    //     // Misc.
+    //     case "fern-postman":
+    //         return { language: "postman", generatorType: "other" };
+    //     case "fern-openapi":
+    //         return { language: "openapi", generatorType: "other" };
+    // }
```
dsinghvi, line 139: should we uncomment and leave in a separate file? that way if someone adds php they'll be forced to update this function as well and we can swap during cli v2

packages/cli/init/src/createWorkspace.ts
```diff
@@ -63,17 +71,26 @@ async function getDefaultGeneratorsConfiguration(): Promise<generatorsYml.Genera
                         name: defaultGeneratorName,
                         ...fallbackInvocation,
                         version:
-                            (await generatorsYml.getLatestGeneratorVersion(defaultGeneratorName)) ??
-                            fallbackInvocation.version
+                            (await generatorsYml.getLatestGeneratorVersion({
+                                cliVersion,
+                                generatorName: defaultGeneratorName,
+                                includeRc: false
+                            })) ?? fallbackInvocation.version
```
dsinghvi, line 78: I think its worth emitting a log if we are falling back to another version

packages/cli/configuration/src/generators-yml/getGeneratorVersions.ts
armandobelardo: I removed this since it was legacy! But the above switch should achieve the same thing
The `generatorName` is just a string in Fern today, ideally it'd be an enum or string options to be able to fail when someone adds more languages, but that'd require a few changes throughout the generator yml setup

dsinghvi (commented):
Chatted through adding an ETE test: 
- Setup a Fern project with generators.yml set to `0.0.0`
- Run generator upgrade
- Run generator get 
- Check that versions are different for each generator

dsinghvi (approved):

packages/cli/cli/src/cliV2.ts
```diff
@@ -121,6 +121,11 @@ export function addGeneratorCommands(cli: Argv<GlobalCliOptions>, cliContext: Cl
                             default: false,
                             description:
                                 "Whether or not to include major versions within the upgrade. Defaults to false, meaning major versions will be skipped."
+                        })
+                        .option("rc", {
+                            boolean: true,
+                            hidden: true,
+                            default: false
```
dsinghvi, line 128: nit: believe we chatted about `channel`

packages/cli/cli/src/commands/upgrade/upgradeGenerator.ts
```diff
@@ -81,22 +83,22 @@ export async function loadAndUpdateGenerators({
...
             }
-
-            // check if the versions have the same major version, if not fail, if yes do the upgrade
-            const currentMajor = semver.major(currentVersion);
-            const latestMajor = semver.major(latestVersion);
-            if (currentMajor === latestMajor || includeMajor) {
-                generator.set("version", latestVersion);
-                context.logger.info(chalk.green(`${generatorName} has been upgraded to latest in group: ${groupName}`));
-            }
+            context.logger.debug(`Upgrading ${generatorName} from ${currentGeneratorVersion} to ${latestVersion}`);
```
dsinghvi, line 100: `context.logger.info(chalk.green(`${generatorName} has been upgraded to latest in group: ${groupName}`));`

Let's keep the chalk.green ?

packages/cli/cli/src/cliV2.ts
armandobelardo: You're right! I forgot

packages/cli/configuration/src/generators-yml/addGenerator.ts
```diff
@@ -67,7 +33,13 @@ export async function addGenerator({
...
                 ...invocation,
                 // Fall back to the hardcoded version if a "latest" does not yet exist
-                version: (await getLatestGeneratorVersion(normalizedGeneratorName, context)) ?? invocation.version
+                version:
+                    (await getLatestGeneratorVersion({
+                        cliVersion,
+                        generatorName: normalizedGeneratorName,
+                        context,
+                        includeRc: false
+                    })) ?? invocation.version
```
dsinghvi, line 42: thoughts on getting rid of `invocation.version` ?

dsinghvi: https://linear.app/buildwithfern/issue/FER-3131/remove-generatorinvocationsts for future

armandobelardo: chatted offline, but it's needed/updated while we migrate! so right now this would only really be expected to work for Python and Ruby SDKs, since Java, TS, etc aren't migrated and registered

packages/cli/cli/src/commands/upgrade/upgradeGenerator.ts
armandobelardo: sg! not sure why I removed it, likely a copy paste from elsewhere

PR #4349, opened 2024-08-20
fix(cli): support `x-fern-parameter-name` on path parameters

dsinghvi:
Previously `x-fern-parameter-name` was only respected on query parameters and headers. After this PR, it will also be respected on path parameters.


chdeskur (approved):
lgtm!

PR #4356, opened 2024-08-20
chore(cli): introduce generator list and organization commands

armandobelardo:
This PR introduces a few utility commands for the `fern-bot` app
1. `fern generator list` gets a structured output, which is effectively a simplified generators.yml which we should not break. the Idea here is that fern bot should divvy up PRs per generator invocation, so this spits out api > generator group > generator
2. `fern organization` for fern bot to know what feature flags an org has it first needs to know the org! This simply grabs it and spits it out to stdout

Both optionally write outputs to a file on disk

Fixes FER-2915 and FER-3012


packages/cli/cli/src/cli.ts
```diff
@@ -152,6 +154,8 @@ async function tryRunCli(cliContext: CliContext) {
     addTestCommand(cli, cliContext);
     addUpdateApiSpecCommand(cli, cliContext);
     addUpgradeGeneratorCommand(cli, cliContext);
+    addGetOrganizationCommand(cli, cliContext);
+    addGeneratorListCommand(cli, cliContext);
```
dsinghvi, line 158: @armandobelardo are you cool to start a `cliV2.ts` file and move these into that ?

dsinghvi: no functional difference but will allow us to know to keep these commands / they are agreed upon for v2 already

armandobelardo: Moved them over and referenced them here

PR #4358, opened 2024-08-20
fix(cli): enforce date validation on date-type examples

neha-goswami:
Validate that `date` types are well-formed in Fern Definition `example`s

& a sneaky fix to use consistent uuid validation for v1 and v2

### DIFF
--- a/packages/cli/cli/CHANGELOG.md
+++ b/packages/cli/cli/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.39.14] - 2024-08-21
 
-- Fix: Format validation is enforced on `date` fields that are specified in examples in fern definitions.
+- Fix: Format validation is enforced on `date` fields that are specified in examples specified in an api defintion.
 
 ## [0.39.13] - 2024-08-19
 


packages/cli/generation/ir-generator/src/examples/validateTypeReferenceExample.ts
```diff
@@ -421,6 +421,10 @@ const validateDateTime = createValidator(
     (example) => typeof example === "string" && ISO_8601_REGEX.test(example),
     "an ISO 8601 timestamp"
 );
+const validateDate = createValidator(
+    (example) => typeof example === "string" && example.length === 10 && ISO_8601_REGEX.test(example),
```
neha-goswami, line 425: turns out that dates in the format of yyyy-mm-dd are actually valid date times according to iso 8601 too, so we can leverage the same logic. TIL

packages/cli/yaml/validator/src/rules/valid-example-type/__test__/valid-example-type.test.ts
```diff
@@ -52,6 +52,24 @@ describe("valid-example-type", () => {
                 message: "Expected example to be a boolean. Example is: 100",
                 nodePath: ["types", "BooleanAlias", { key: "examples", arrayIndex: 1 }]
             },
+            {
+                severity: "error",
+                relativeFilepath: RelativeFilePath.of("alias.yml"),
+                message: 'Expected example to be a date. Example is: "2024-01-35"',
```
dannysheridan, line 58: Given that `2024-01-35` isn't a real date, want to swap to `2024-01-25`?

neha-goswami: that's part of the validation! we are erroring on invalid dates. a real date is 2024-01-01 (see in alias.yml L37) that notably does not have an expected error

dannysheridan: An _effective_ test case

packages/cli/generation/ir-generator/src/examples/validateTypeReferenceExample.ts
dsinghvi: @neha-goswami so the datetimes are technically iso8601 rfc 3339, would that format still be valid? (we basically follow OpenAPI) 

https://swagger.io/docs/specification/data-models/data-types/
Screenshot 2024-08-21 at 10 47 45 AM

dsinghvi (commented):
Overall looking great. To release this, you'll also need to: 
- Bump the CLI Version (https://github.com/fern-api/fern/blob/main/packages/cli/cli/VERSION) 
- Add a changelog entry (https://github.com/fern-api/fern/blob/main/packages/cli/cli/CHANGELOG.md)

packages/cli/generation/ir-generator/src/examples/validateTypeReferenceExample.ts
neha-goswami: oh, didn't catch it but yeah the fern docs do specify rfc 3339 - though notably they don't specify iso 8601. are the docs outdated here? https://buildwithfern.com/learn/api-definition/fern/types

this will be a little more lenient than rfc 3339 so i'll make a quick update for it

neha-goswami:
> Overall looking great. To release this, you'll also need to:
> 
> * Bump the CLI Version (https://github.com/fern-api/fern/blob/main/packages/cli/cli/VERSION)
> * Add a changelog entry (https://github.com/fern-api/fern/blob/main/packages/cli/cli/CHANGELOG.md)

should these happen in a separate PR or in the same one? inclined to put them all in one but not sure if there is extra release process i'm missing

pnpm-lock.yaml
```diff
@@ -3179,8 +3179,6 @@ importers:
 
   packages/cli/cli/dist/dev: {}
```
neha-goswami, line 3181: is this an expected change?

dsinghvi:
> > Overall looking great. To release this, you'll also need to:
> > 
> > * Bump the CLI Version (https://github.com/fern-api/fern/blob/main/packages/cli/cli/VERSION)
> > * Add a changelog entry (https://github.com/fern-api/fern/blob/main/packages/cli/cli/CHANGELOG.md)
> 
> should these happen in a separate PR or in the same one? inclined to put them all in one but not sure if there is extra release process i'm missing

same one is great!

pnpm-lock.yaml
dsinghvi: weird, i dont think so but it shouldn't matter

packages/cli/cli/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.39.14] - 2024-08-21
+
+- Fix: Format validation is enforced on `date` fields that are specified in examples in fern definitions.
```
dsinghvi, line 10: this will technically be true for OpenAPI as well (since we go OpenAPI -> FD)

dsinghvi (approved):
🚀

packages/cli/cli/CHANGELOG.md
neha-goswami: updated

PR #4359, opened 2024-08-20
feat(csharp): Generate type for google.protobuf.Struct

amckinney:
This refactors the types generated for `google.protobuf.Struct` and `google.protobuf.Value` so that they implement the `ToProto` and `FromProto` methods just like the other classes. This simplifies the Protobuf mapping code and removes the `ProtoConverter` utility entirely. 

Most importantly, the user-experience is significantly improved - users don't have to declare a long series of `Dictionary<string, Value?>` type declarations anymore - they can just refer to the type declaration directly (e.g. `Metadata`):

```csharp
var response = await client.GetUserAsync(new GetUserRequest {
       Name = "john.doe",
       Filter = new Metadata
       {
           ["state"] =
               new Metadata
               {
                   ["$in"] = new[] { "CA", "NY" }
               }
       }
});
```


dsinghvi (approved):

generators/csharp/codegen/src/asIs/RawGrpcClient.Template.cs
```diff
@@ -47,7 +47,7 @@ public CallOptions CreateCallOptions(
         CancellationToken cancellationToken = default
     )
     {
-        var metadata = new Metadata();
+        var metadata = new global::Grpc.Core.Metadata();
```
dcb6, line 50: what's this part? `global::Grpc.Core.`

dcb6 (approved):

generators/csharp/codegen/src/asIs/RawGrpcClient.Template.cs
amckinney: This helps guard against the edge case where users specify their own `Metadata` type (which will be fairly common). We don't use the global scope for the other gRPC utilities since it's far less likely that they conflict (e.g. `GrpcChannelOptions`).

We normally handle this well when we generate the files dynamically, but we don't have that auto-resolution given that this is an as-is file.

PR #4360, opened 2024-08-20
feat(csharp): Add CancellationToken parameter

amckinney:
This updates the C# endpoint methods to include a `CancellationToken` parameter like so:

```csharp
  /// <summary>
  /// Add a movie to the database
  /// </summary>
  public async Task<string> CreateMovieAsync(
      CreateMovieRequest request,
      RequestOptions? options = null,
  )
  {
      var response = await _client.MakeRequestAsync(
          new RawClient.JsonApiRequest
          {
              BaseUrl = _client.Options.BaseUrl,
              Method = HttpMethod.Post,
              Path = "/movies/create-movie",
              Body = request,
              Options = options,
          },
          cancellationToken
      );
      ...
  }
```

As described in the `CHANGELOG.md`, this aligns with the code analysis quality rule `CA1068` outlined in the official [.NET documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068).

### DIFF
--- a/seed/csharp-sdk/alias-extends/src/SeedAliasExtends/SeedAliasExtendsClient.cs
+++ b/seed/csharp-sdk/alias-extends/src/SeedAliasExtends/SeedAliasExtendsClient.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using SeedAliasExtends.Core;
 
@@ -22,7 +23,8 @@ public partial class SeedAliasExtendsClient
 
     public async Task ExtendedInlineRequestBodyAsync(
         InlinedChildRequest request,
-        RequestOptions? options = null
+        RequestOptions? options = null,
+        CancellationToken cancellationToken = default
     )
     {
         var response = await _client.MakeRequestAsync(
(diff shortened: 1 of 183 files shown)


seed/csharp-sdk/alias-extends/src/SeedAliasExtends/SeedAliasExtendsClient.cs
```diff
@@ -22,7 +23,8 @@ public SeedAliasExtendsClient(ClientOptions? clientOptions = null)
 
     public async Task ExtendedInlineRequestBodyAsync(
         InlinedChildRequest request,
-        RequestOptions? options = null
+        RequestOptions? options = null,
+        CancellationToken cancellationToken = default
```
dcb6, line 27: we could've just the same had `= null` here right? (checking understanding, not requesting a change)

amckinney: Not in this case - it's idiomatic for the `CancellationToken` to be a required parameter, so following the same style actually makes it a lot easier to propagate the token at every call-site.

For example, consider the HTTP client call where this is used:

```csharp
var response = await httpClient.SendAsync(httpRequest, cancellationToken);
```

You'll actually notice that the signature similarly expects a required value like so:

```csharp
public override Task<HttpResponseMessage> SendAsync(
  HttpRequestMessage request,
  CancellationToken cancellationToken);
```

If we typed this as optional, we'd have to include a ternary at every call-site. For some extra justification, [here](https://stackoverflow.com/questions/22359706/default-parameter-for-cancellationtoken) is a relevant discussion.

dcb6 (approved):

PR #4362, opened 2024-08-20
feat(seed): publish generators through seed CLI

armandobelardo:
Fixes FER-3005

**Why?**
Want to be able to use a single source of truth (the seed versions file), as opposed to maintaining a VERSION + CHANGELOG file, as well as the seed versions file (concern there is that things would drift, there could be conflicts etc.), now the same mechanism registers artifacts to FDR and publishes the containers to Docker

**What this does**
1. Removes the publishing logic from the `publish-generators.yml` github action
2. Puts the publishing metadata into each generators seed.yml
3. Introduces a `seed publish` command which can take 2 files (meant to be main and branch merging into main), diffs those files, finds the new generator and publishes that leveraging the publish metadata
   a.  That looks something like `seed publish --generator python-sdk --versions-file file.yml --previous-versions-file file.yml`
4. The `publish-generators.yml` github action is updated to invoke `seed publish`
5. We also add in the `seed register generator` and `seed register cli` commands into our publish github actions

As a follow up, I think we'll want to do the same thing for the CLI, we could add a similar publish block to it's seed.yml and the rest should be pretty much the same as this PR does for the generators


packages/seed/src/commands/register/registerGenerator.ts
```diff
@@ -43,25 +41,17 @@ export async function registerGenerator({
...
-                } catch (e) {
-                    context.logger.error(`Error parsing release: ${e}`);
-                }
+        context.logger.info(`Registering generator ${generatorId} releases from ${absolutePathToChangelogLocation}...`);
+        await parseGeneratorReleasesFile({
+            generatorId,
+            versionsFilePath: absolutePathToChangelogLocation,
+            context,
+            action: async (release) => {
+                context.logger.debug(`Registering generator ${generatorId} release: ${release.version}`);
```
dsinghvi, line 50: lets make this an info log? i think it would be useful to see this on every run

packages/seed/src/utils/convertVersionsFileToReleases.ts
```diff
@@ -0,0 +1,34 @@
...
+    context.logger.debug(`Parsing versions file ${versionsFilePath}`);
+    const changelogs = yaml.load((await readFile(versionsFilePath)).toString());
+    if (Array.isArray(changelogs)) {
+        for (const entry of changelogs) {
+            try {
+                const release = serializers.generators.GeneratorReleaseRequest.parseOrThrow({
+                    generator_id: generatorId,
+                    ...entry
+                });
+                context.logger.debug(`Encountered generator  ${generatorId} release: ${release.version}`);
```
dsinghvi, line 27: i think its okay to delete this debug log (might be too much since looks like the action also logs)

packages/seed/src/utils/convertVersionsFileToReleases.ts
```diff
@@ -0,0 +1,34 @@
...
+                await action(release);
+            } catch (e) {
+                context.logger.error(`Error parsing release: ${e}`);
```
dsinghvi, line 30: nit: `Failed to parse and run action on release: ${(e as Error)?.message}` ?

packages/seed/src/cli.ts
```diff
@@ -197,6 +199,81 @@ function addRunCommand(cli: Argv) {
...
+                    yargs
+                        .positional("generator", {
+                            type: "string",
+                            demandOption: true,
+                            description: "Generator(s) to register"
+                        })
+                        // Version is a reserved option in yargs...
+                        .option("ver", {
+                            type: "string",
+                            demandOption: false
```
dsinghvi, line 219: should this be required?

armandobelardo: It's optional cause you can alternatively pass it paths to 2 versions files
e.g. `seed publish --generator python-sdk --versions-file file.yml --previous-versions-file file.yml` where it will run through and grab the latest version in the newer file that wasn't in the old file

It was easier to include in the command than to write a script that's run in actions

armandobelardo: The CLI then checks that at least one of `ver` or the files is passed in (that's that `.check` addition a little lower down)

packages/seed/src/cli.ts
```diff
@@ -197,6 +199,81 @@ function addRunCommand(cli: Argv) {
...
+                            demandOption: false,
+                            description:
+                                "Path to the latest versions file, used along side `previous-versions-file` to the most recent new version to publish."
+                        })
+                        .option("previous-versions-file", {
+                            type: "string",
+                            demandOption: false,
+                            description:
+                                "Path to the previous versions file, used along side `versions-file` to the most recent new version to publish."
+                        })
```
dsinghvi, line 232: would you be up to rename any references to `versions-file` as `changelog` ? 

`--changelog /path/to/current --previous-changelog /path/to/previous`

packages/seed/src/cli.ts
```diff
@@ -197,6 +199,81 @@ function addRunCommand(cli: Argv) {
...
+                            description: "Generator(s) to register"
+                        })
+                        // Version is a reserved option in yargs...
+                        .option("ver", {
+                            type: "string",
+                            demandOption: false
+                        })
+                        .option("versions-file", {
+                            type: "string",
+                            demandOption: false,
```
dsinghvi, line 223: should this eventually be true? is the primary thing blocking migrating all of the changelog.mds to the new format

packages/seed/src/commands/publish/publishGenerator.ts
```diff
@@ -0,0 +1,181 @@
...
+            versionFilePair: version,
+            context
+        });
+
+        if (maybeNewVersion == null) {
+            context.failWithoutThrowing(
+                "No version to publish! There must not have been a new version since the last publish."
+            );
+            return;
+        }
```
dsinghvi, line 41: @armandobelardo what if i modify the changelog to yank a version or to update a release note?

packages/seed/src/commands/publish/publishGenerator.ts
```diff
@@ -0,0 +1,181 @@
...
+        publishVersion = maybeNewVersion;
+    } else {
+        publishVersion = version;
+    }
+
+    const publishConfig = generator.workspaceConfig.publish;
+    // Instance of PublishDocker configuration, leverage Dockerode here
```
dsinghvi, line 48: nit: update comment to remove `Dockerode`

.github/workflows/seed.yml
```diff
@@ -99,11 +99,17 @@ jobs:
...
+          pnpm seed:local test --generators ruby-model
 
       - name: Ensure no changes to git-tracked files
         run: git --no-pager diff --exit-code  -- ":(exclude)seed/*/.mock/*"
 
+      # Only register if seed is passing
+      - name: Register on push to main
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        run: |
+          pnpm seed:local register --generators ruby-model
```
dsinghvi, line 111: Talked offline, ideally we should have a separate workflow for `register-generator`.

armandobelardo: Stop throwing if no error
Remove register from seed.yml CI job
On merge to main:
- run publish
- run register

armandobelardo: Also update the publish job to use git show instead of clone twice

.github/workflows/publish-generator.yml
```diff
@@ -52,958 +29,112 @@ jobs:
...
-        with:
-          context: .
-          file: ./generators/csharp/sdk/Dockerfile
-          platforms: linux/amd64,linux/arm64
-          cache-from: type=gha
-          cache-to: type=gha,mode=min
-          push: true
-          labels: version=${{env.VERSION}}
-          tags: fernapi/fern-csharp-sdk:${{env.VERSION}}${{ (!contains(env.VERSION, '-rc') && ', fernapi/fern-csharp-sdk:latest') || '' }}
+          pnpm seed:local test --generators ${{ matrix.generator }}
```
dsinghvi, line 138: lets comment this out for now b/c this will mean 20+ min to publish python?

packages/seed/src/commands/publish/publishGenerator.ts
```diff
@@ -0,0 +1,181 @@
...
+        // We were given two version files, so we need to compare them to find if any new
+        // versions have been added since the last publish.
+        const maybeNewVersion = await getNewVersion({
+            generatorId,
+            versionFilePair: version,
+            context
+        });
+
+        if (maybeNewVersion == null) {
+            context.failWithoutThrowing(
```
dsinghvi, line 37: @armandobelardo i think this leads to a non zero exit, lets not use this here?

dsinghvi: reason being that we want publish to noop so that register can run

dsinghvi (approved):

armandobelardo:
Planning to action the above, as well as add back the original publish flows for all except python and start with it + it's version files as a trial before cutting everything else over

PR #4364, opened 2024-08-20
fix(typescript): do not generate environments file if no environments can be used

RohinBhargava:
## Changes

* Adds more robustness around environments config checks to see if the ts-generator should generate an environments file
* Modifies the `audiences` seed definition to accommodate new case, where a `partner` audience can be supplied downstream, that has no associated environment.
* Modifies the `ts-sdk` seed config to accommodate the new test definition structure to prove that the change is valid
* Bumps `Changelog` and `Version`
* Uncovered bugs in python and ts generators, with new `failing-audiences` test-definition and filed here: 
  * https://linear.app/buildwithfern/issue/FER-3045/[everyone]-duplicate-path-definitions-leads-to-bad-generation
  * https://linear.app/buildwithfern/issue/FER-3047/[seed]-failing-audiences-generates-the-wrong-type-in-ir

FER-2927

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,6 +111,8 @@ in the generator.
 
 ### Generator Testing
 
+**Note**: Please make sure that the docker daemon is running before running commands below. 
+
 To test our generators we have built a CLI called seed.
 
 Seed handles building the generators from source and running them against all of the
--- a/fern/docs.yml
+++ b/fern/docs.yml
@@ -223,6 +223,10 @@ navigation:
             slug: development
             path: ./pages/docs/getting-started/development.mdx
             icon: fa-regular fa-terminal
+          - page: Global Configuration 
+            slug: global-configuration
+            path: ./pages/docs/getting-started/global-configuration.mdx   
+            icon: fa-regular fa-wrench               
           - page: Project Structure 
             slug: project-structure
             path: ./pages/docs/getting-started/project-structure.mdx   
--- a/fern/pages/docs/building-your-docs/styling.mdx
+++ b/fern/pages/docs/building-your-docs/styling.mdx
@@ -96,46 +96,6 @@ In `docs.yml`, you can set a single background image or specify different images
     </CodeBlock>
 </CodeBlocks>
 
-### Logo
-
-Specify a logo in `docs.yml` which gets displayed in the top left of your docs.
-
-<CodeBlocks>
-    <CodeBlock title="single color">
-    ```yaml
-    logo: ./images/logo.png
-    ```
-    </CodeBlock>
-    <CodeBlock title="light and dark mode">
-    ```yaml
-    logo: 
-        light: ./images/logo-light.png
-        dark: ./images/logo-dark.png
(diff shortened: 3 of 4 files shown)


seed/ts-sdk/seed.yml
```diff
@@ -31,6 +31,13 @@ fixtures:
         allowCustomFetcher: true
         timeoutInSeconds: 2
       outputFolder: allow-custom-fetcher
+  audiences:
```
RohinBhargava, line 34: relevant change here

test-definitions/fern/apis/audiences/definition/api.yml
```diff
@@ -3,3 +3,12 @@ display-name: Audiences API
 audiences:
   - public
   - private
+environments:
```
RohinBhargava, line 6: and here

test-definitions/fern/apis/audiences/definition/api.yml
```diff
@@ -3,3 +3,12 @@ display-name: Audiences API
...
+  environment-a:
```
RohinBhargava, line 7: and here

generators/typescript/sdk/environments-generator/src/EnvironmentsGenerator.ts
```diff
@@ -20,7 +20,21 @@ export class EnvironmentsGenerator {
         environmentUrlsTypeName,
         environmentsConfig
     }: EnvironmentsGenerator.generateEnvironments.Args): GeneratedEnvironments {
-        if (environmentsConfig == null) {
+        if (
```
RohinBhargava, line 23: critical change

test-definitions/fern/apis/failing-audiences/definition/api.yml
```diff
@@ -0,0 +1,16 @@
+name: audiences
```
dsinghvi, line 1: It isn't easy to understand what edge case you are trying to test with `failing-audiences`. could there be a better name for this definition?

seed/ts-sdk/seed.yml
```diff
@@ -31,6 +31,13 @@ fixtures:
         allowCustomFetcher: true
         timeoutInSeconds: 2
       outputFolder: allow-custom-fetcher
+  failing-audiences:
+    - customConfig: null
+      outputFolder: no-custom-config
+    - customConfig: null
+      audiences:
+        - partner
+      outputFolder: with-partner-audience
```
dsinghvi, line 40: is there a way to test these edge cases using the existing `audiences` fixture?

RohinBhargava: yeah! let me add both like this!

test-definitions/fern/apis/failing-audiences/definition/api.yml
RohinBhargava: This is just a temporary fixture, until the bugs get solved, and then it will be deleted!

dsinghvi (approved):

PR #4366, opened 2024-08-21
fix(cli): example generation takes into account root level path parameters

dsinghvi:
Before this PR if the API had root level path parameters, they were being ignored when generating endpoint examples. This has multiple ramifications such as incorrect cURL examples in the generated docs.


packages/cli/generation/ir-generator/src/converters/services/convertExampleEndpointCall.ts
```diff
@@ -560,17 +560,22 @@ function convertExampleResponseBody({
...
+    file: FernFileContext;
     service: RawSchemas.HttpServiceSchema;
     endpoint: RawSchemas.HttpEndpointSchema;
     example: RawSchemas.ExampleEndpointCallSchema;
     pathParams: Pick<ExampleEndpointCall, "rootPathParameters" | "endpointPathParameters" | "servicePathParameters">;
 }): string {
     let url = service["base-path"] + endpoint.path;
+    if (file.rootApiFile["base-path"] != null) {
+        url = `${file.rootApiFile["base-path"]}${url}`;
+    }
```
dsinghvi, line 578: going to update to safe join these

packages/cli/generation/ir-generator/src/__test__/test-definitions/package-yml.json
```diff
@@ -1082,7 +1082,7 @@
                             "example": {
                                 "id": "7c20f2e4ee05e0f9f96a9d824fd47f991dc6b563d45f4c99c8248c68dc7afdd0",
                                 "name": null,
-                                "url": "//id-219xca8",
+                                "url": "/id-a2ijs82//id-219xca8",
```
dsinghvi, line 1085: example where safe joining is necessary

amckinney, line 1085: Is this right with the two `//` inbetween?

amckinney (approved):

PR #4380, opened 2024-08-21
feat(csharp): Add User-Agent and refactor HTTP header options

amckinney:
This adds support for the `User-Agent` header which is provided by the IR. It's in the `package/version` format.

Additionally, the `RawClient` now supports HTTP headers within the `ClientOptions` and `RequestOptions` types. This is not a user-facing feature, but it simplifies the core utilities. This makes it possible for users that wrap clients to extend and/or modify what headers are sent along with every request.


dcb6 (approved):
🔥 💯 😎

PR #4382, opened 2024-08-22
feat(seed): publish fern CLI through seed CLI

armandobelardo:
Fixes: FER-3067

See https://github.com/fern-api/fern/pull/4362 for a more detailed description
This PR introduces the same change, but for the Fern CLI

Of note, this merges together the original `publish-cli` action with the `publish-dev-cli` action, this makes it so that on every merge to main we:
- publish the dev CLI
- try to publish the prod CLI (which will only happen if there's a new version)
- register the CLI versions with FDR


seed/fern-cli/seed.yml
```diff
@@ -1 +1,25 @@
 changelogLocation: ../../packages/cli/cli/versions.yml
+publishGa:
+  versionSubstitution: $VERSION
```
dsinghvi, line 3: what does this flag do ?

seed/fern-cli/seed.yml
```diff
@@ -1 +1,25 @@
...
+  command:
+    - cd packages/cli/cli
+    - mv package.json package.json.tmp
+    - version_replace="s/0.0.0/$VERSION/"
+    - cat package.json.tmp| sed "$VERSION" > package.json
+    - rm -rf package.json.tmp
+    - pnpm install
+    - pnpm --filter @fern-api/cli compile
+    # Should likely include this within the script itself
+    - pnpm --filter @fern-api/cli publish:cli:prod --tag latest
```
dsinghvi, line 13: do we know this works or does this have to be `pnpm --filter @fern-api/cli publish:cli:prod -- --tag latest`

seed/fern-cli/seed.yml
```diff
@@ -1 +1,25 @@
...
+publishRc:
```
dsinghvi, line 14: should we move over `publishDev` as well since that will allow us to to test this more easily

armandobelardo: chatted offline, needs seed-dev or if too hard, just don't register (but still publish)

armandobelardo: Updated

packages/cli/cli/versions.yml
armandobelardo, line 1: These summaries are not amazing, might wanna just call it on this and just do "Release <VERSION>"

dsinghvi (approved):

PR #4385, opened 2024-08-22
feat(csharp): add snippet examples to endpoint methods

dcb6:
Endpoint method now include docstrings with example usage.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+npx lint-staged
--- a/fern/pages/api-definition/fern-definition/examples.mdx
+++ b/fern/pages/api-definition/fern-definition/examples.mdx
@@ -76,7 +76,7 @@ types:
       - name: EmpireStateBuilding
         value:
           street1: 350 5th Ave
-          street2:
+          street2: Attn: Maintenance Department
           city: New York
           state: NY
           postalCode: "10118"
@@ -102,6 +102,7 @@ types:
  * * The Empire State Building address
  * @example {
  *  street1: "350 5th Ave",
+ *  street2: "Attn: Maintenance Department",
  *  city: "New York",
  *  state: "NY",
  *  postalCode: "10118",
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -1,9 +1,5 @@
 module.exports = {
-    "**/*.ts{,x}": [
-        "eslint --fix --max-warnings 0 --no-eslintrc --config .eslintrc.lint-staged.js --ignore-path .eslintignore",
-        "yarn format:fix",
-    ],
-    "**/*.{js,json,yml,html,css,less,scss,md}": "yarn format",
-    "**": () => "yarn lint:monorepo",
-    "**/package.json": () => "yarn install --immutable",
+    "**/*.ts{,x}": "eslint --fix --max-warnings 0 --no-eslintrc --config .eslintrc.lint-staged.js --ignore-path .eslintignore",
+    "**/*.{ts,js,json,yaml,html,css,less,scss,md}": "pnpm format",
+    "**/package.json": () => "pnpm install --frozen-lockfile",
 };
--- a/package.json
+++ b/package.json
@@ -40,7 +40,8 @@
     "ir:generate": "pnpm --filter @fern-api/ir-sdk generate",
     "openapi-ir:generate": "pnpm --filter @fern-api/openapi-ir-sdk generate",
     "seed-config:generate": "pnpm --filter @fern-api/seed-cli generate",
-    "docs-config:generate": "pnpm --filter @fern-api/configuration generate"
+    "docs-config:generate": "pnpm --filter @fern-api/configuration generate",
+    "prepare": "husky"
   },
   "devDependencies": {
     "@babel/core": "^7.23.2",
@@ -64,10 +65,10 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-tailwindcss": "^3.15.1",
     "execa": "^5.1.1",
+    "husky": "^9.1.5",
     "immer": "^9.0.15",
     "is-ci": "^3.0.1",
     "jest": "^29.7.0",
-    "ts-jest": "^29.2.4",
     "jsonc-parser": "~3.3.0",
     "lint-staged": "^13.0.3",
     "lodash-es": "^4.17.21",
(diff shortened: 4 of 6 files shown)


generators/csharp/codegen/src/ast/InstantiatedPrimitive.ts
```diff
@@ -115,14 +116,18 @@ export class PrimitiveInstantiation extends AstNode {
...
             case "dateTime": {
-                writer.write(`DateTime.Parse("${this.internalType.value.toISOString()}", null, DateTimeStyles.`);
-                writer.writeNode(
-                    csharp.classReference({
-                        name: "AdjustToUniversal",
-                        namespace: "System.Globalization"
-                    })
-                );
-                writer.write(")");
+                if (this.internalType.parsed) {
```
amckinney, line 119: nit: Should we call this `parse` since this is the case where we actually call `DateTime.Parse`? Otherwise it sounds like the string is already parsed and we do _not_ need to call `DateTime.Parse`.

generators/csharp/codegen/src/ast/Method.ts
```diff
@@ -36,6 +36,8 @@ export declare namespace Method {
         classReference?: ClassReference;
         /* Any annotations to add to the method */
         annotations?: Annotation[];
+        /* Any annotations to add to the method */
```
amckinney, line 39:
```suggestion
        /* The code example to add to the method */
```

generators/csharp/codegen/src/ast/Method.ts
```diff
@@ -86,11 +91,34 @@ export class Method extends AstNode {
     public write(writer: Writer): void {
         if (this.summary != null) {
             writer.writeLine("/// <summary>");
-            this.summary.split("\n").forEach((line) => {
+            this.summary?.split("\n").forEach((line) => {
```
amckinney, line 94: Can we drop the `?.` guard if we already have the `this.summary != null` guard above?

generators/csharp/sdk/src/wrapped-request/WrappedRequestGenerator.ts
```diff
@@ -173,12 +173,15 @@ export class WrappedRequestGenerator extends FileGenerator<CSharpFile, SdkCustom
         });
     }
 
-    public doGenerateSnippet(example: ExampleEndpointCall): csharp.CodeBlock {
+    public doGenerateSnippet(example: ExampleEndpointCall, parsedDatetimes: boolean): csharp.CodeBlock {
```
amckinney, line 176: Now that we have a `boolean`, it's probably a good time to use an in-lined request. That way it's easier to see what the function is doing at the call-site. Otherwise it's easy to fall into the trap of a long list of anonymous `boolean` positional parameters. We can also name this `parseDatetimes` (re: my other comment).

```typescript
public doGenerateSnippet({
  example,
  parseDatetimes
}: {
  example: ExampleEndpointCall;
  parseDatetimes?: boolean;     // We can make this optional if that's easier, too.
}: csharp.CodeBlock {
  ...
}
```

Then each call-site is adapted to:

```typescript
const snippet = objectGenerator.doGenerateSnippet({
  example: example.shape,
  parseDatetimes: true,
});
```

The same recommendation applies to the `getSnippetForTypeReference` function.

generators/csharp/codegen/src/ast/core/AstNode.ts
```diff
@@ -27,6 +30,24 @@ export abstract class AstNode {
...
+     */
+    public toFormattedSnippet(code: string): string {
+        code += ";";
+        try {
+            const finalCode = AstNode.formatCSharpCode(code);
+            return finalCode;
+        } catch (e: unknown) {
+            return code;
+        }
+    }
```
amckinney, line 48: Is this still being used?

generators/csharp/codegen/src/ast/core/AstNode.ts
```diff
@@ -17,7 +18,9 @@ export abstract class AstNode {
         allNamespaceSegments: Set<string>,
         allTypeClassReferences: Map<string, Set<Namespace>>,
         rootNamespace: string,
-        customConfig: BaseCsharpCustomConfigSchema
+        customConfig: BaseCsharpCustomConfigSchema,
+        format = false,
```
amckinney, line 22: We might want to update the `toString` function to using an in-lined request too - I think the only reference is in `CSharpFile.ts`, and this will let us use optional parameters.

seed/csharp-sdk/grpc-proto-exhaustive/read-only-memory/src/SeedApi/Dataservice/DataserviceClient.cs
```diff
@@ -22,6 +22,23 @@ internal DataserviceClient(RawClient client)
...
+    /// <code>
+    /// await client.Dataservice.UploadAsync(
+    ///     new UploadRequest
+    ///     {
+    ///         Columns = new List<Column>()
+    ///         {
+    ///             new Column
+    ///             {
+    ///                 Id = "id",
+    ///                 Values = new List<float>() { 1.1 },
```
amckinney, line 35: I would have expected this to be a `ReadOnlyMemory<float>`. https://github.com/fern-api/fern/blob/6bd9d9a2c17d4e30c9e96ad354b691fbbd8fba5f/seed/csharp-sdk/grpc-proto-exhaustive/read-only-memory/src/SeedApi/Types/Column.cs#L14

Is there any chance that the snippet generator is using a different context? This should be handled by the writer automatically.

generators/csharp/codegen/src/ast/core/AstNode.ts
dcb6: used [here] right? (https://github.com/fern-api/fern/pull/4385/files/f61c6bd5d5f2294b376c55ee602cd3202ffefdd7#diff-37bfd52f6d61c28fe29c548d34821cbd6a98482408f2357c94f901688d260e84R34)

generators/csharp/codegen/src/ast/Method.ts
dcb6: good catch!

generators/csharp/sdk/src/wrapped-request/WrappedRequestGenerator.ts
dcb6: yeah great call, changing!

seed/csharp-sdk/grpc-proto-exhaustive/read-only-memory/src/SeedApi/Dataservice/DataserviceClient.cs
dcb6: fixed, great catch!

amckinney (approved):

PR #4386, opened 2024-08-22
feat(typescript): Include .js file extensions on imports when `outputEsm` is true

zachkirsch:
This PR updates the behavior of the typescript generators when `outputEsm` is enabled: we append `.js` when importing from a file and `/index.js` when importing from a directory.

## Motivation

On the most recent versions of node/esm/typescript, it's actually recommended practice to include `.js` extensions in your typescript imports. (I know this is insane). See https://github.com/microsoft/TypeScript/issues/49083#issuecomment-1435399267.

Screenshot 2024-08-22 at 7 13 13 AM

### DIFF
--- a/generators/typescript/sdk/CHANGELOG.md
+++ b/generators/typescript/sdk/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.40.0-rc0] - 2024-08-22
+
+- Feature: When `outputEsm` is enabled, all import statements include `.js` extensions.
+  This may be necessary in projects where `moduleResolution` in tsconfig.json is `node16` or `nodenext`.
+
 ## [0.39.4] - 2024-08-20
 
 - Fix: Ensure that environment files don't generate, unless there is a valid environment available.


seed/ts-express/seed.yml
```diff
@@ -50,6 +50,9 @@ fixtures:
         includeUtilsOnUnionMembers: true
         includeOtherInUnionTypes: true
       outputFolder: union-utils
+    - customConfig:
+        useJsFileExtensionsOnImports: true
+      outputFolder: js-file-extensions
```
zachkirsch, line 55: should I add this test under `exhaustive` or somewhere else?

dsinghvi:
Chatted offline: 
- going to use `outputEsm` instead of a new config flag
- will release an rc for folks to test 
- (bonus) would be nice to wire this up for the SDK generator as well since we are just using `outputEsm`

seed/ts-express/seed.yml
dsinghvi: `exhaustive` works

zachkirsch:
@dsinghvi making sure I don't break https://github.com/fern-api/fern/pull/3397 https://github.com/fern-api/fern/blob/6bd9d9a2c17d4e30c9e96ad354b691fbbd8fba5f/generators/typescript/utils/commons/src/referencing/getReferenceToExportFromRoot.ts#L72-L74 

do we _always_ want `/index` appended or is it okay to only include it when `outputEsm` is true?

generators/typescript/utils/commons/src/referencing/getReferenceToExportFromRoot.ts
```diff
@@ -45,13 +47,16 @@ export function getReferenceToExportFromRoot({
     if (firstDirectory?.exportDeclaration?.defaultExport != null) {
         moduleSpecifier = getRelativePathAsModuleSpecifierTo({
             from: referencedIn,
-            to: convertExportedDirectoryPathToFilePath([])
+            to: convertExportedDirectoryPathToFilePath([]),
+            isDesinationADirectory: true,
+            isOutputtingEsm,
+            forceAppendBarrelFile: true
```
zachkirsch, line 53: This weird override flag is so we don't change the behavior introduced by https://github.com/fern-api/fern/pull/3397 https://github.com/fern-api/fern/blob/6bd9d9a2c17d4e30c9e96ad354b691fbbd8fba5f/generators/typescript/utils/commons/src/referencing/getReferenceToExportFromRoot.ts#L53-L55

zachkirsch:
the failing test is `ete-test` but that also fails when I checkout `main`. PR to fix that here: https://github.com/fern-api/fern/pull/4390

seed/ts-sdk/exhaustive/esm-source/tests/unit/auth/BasicAuth.test.ts
```diff
@@ -0,0 +1,22 @@
+import { BasicAuth } from "../../../src/core/auth/BasicAuth";
```
dsinghvi, line 1: We may need this in the tests as well

dsinghvi (approved):

seed/ts-sdk/exhaustive/esm-source/src/core/auth/index.ts
```diff
@@ -0,0 +1,2 @@
+export { BasicAuth } from "./BasicAuth";
+export { BearerToken } from "./BearerToken";
```
dsinghvi, line 2: Probably want this in core utilities too

dsinghvi:
Because its hard to add this to unit tests + core utilitites, another approach is to write a script that walks the generated code and updates the import to include `.js`. That's probably the most useful way to add this feature because we eventually want to publish a package that is both CJS + ESM compatible

zachkirsch:
@dsinghvi a potentially simpler approach is to just use `.js` extensions in the core-utilities. That's actually the recommended approach for both cjs and esm now, see this comment: https://github.com/microsoft/TypeScript/issues/49083#issuecomment-1145425177

adamsuskin:
@dsinghvi Any chance we can put this behind a flag (`includeExtensions` or something) and release?

PR #4387, opened 2024-08-22
feat(internal): set up husky/lint-staged

zachkirsch:
This PR sets up husky + lint-staged so when you run `git commit`, we format the files with prettier + check that the pnpm lockfile is up-to-date

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+npx lint-staged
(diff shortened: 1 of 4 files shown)


dsinghvi:
We're having some problems with eslint taking too long which is why we removed that -- does the eslint command run fine for you locally?

zachkirsch:
seems fine on my end 

https://github.com/user-attachments/assets/97ff4863-13e5-4079-b410-0af7039bf6d7

PR #4393, opened 2024-08-22
internal(seed): [DNM] command to get IR from generator version

armandobelardo:
Short term helper CLI to take a version of a generator and get the ir version

This adds it to the changelog file right now, so the idea is you make your changelog file without the IR, then run `pnpm seed:local enrich-with-ir <generator>` and then it rights the file back with the IR version

I imagine my IR grabbing logic might be slightly off, but this should still save us a lot of time


dsinghvi:
We're going to be deleting the migrations but down to keep this command for now to help folks

dsinghvi (approved):

PR #4398, opened 2024-08-22
fix(cli): add x-fern-audiences to openapi parsing

RohinBhargava:
Adds `x-fern-audiences` to server specs in OpenApi on `servers`, for use with environment filtering.

Makes changes in:
- openapi-parser (`convertServer.ts`)
- openapi-ir-to-fern (`buildEnvironments.ts`)


packages/cli/openapi-ir-to-fern/src/buildEnvironments.ts
```diff
@@ -1,79 +1,122 @@
+import { RawSchemas } from "@fern-api/yaml-schema";
 import { OpenApiIrConverterContext } from "./OpenApiIrConverterContext";
 
 const PRODUCTION_ENVNIRONMENT_NAME = "Production";
 const DEFAULT_ENVIRONMENT_NAME = "Default";
 
+function isMultiServerSchema(
+    schema: RawSchemas.EnvironmentSchema
+): schema is RawSchemas.MultipleBaseUrlsEnvironmentSchema {
+    return typeof schema === "object" && Object.hasOwn(schema, "urls");
```
amckinney, line 10: nit: First time I've seen `Object.hasOwn` for this - I'm sure this works just fine, but I wonder if it's worth making this consistent with the other `isRaw` helpers like the following:

```typescript
export function isRawObjectDefinition(rawTypeDeclaration: TypeDeclarationSchema): rawTypeDeclaration is ObjectSchema {
    return (
        (rawTypeDeclaration as ObjectSchema).extends != null || (rawTypeDeclaration as ObjectSchema).properties != null
    );
}
```

I imagine we could do something like:

```typescript
function isRawMultipleBaseUrlsEnvironmentSchema(schema: RawSchemas.EnvironmentSchema): schema is RawSchemas.MultipleBaseUrlsEnvironmentSchema {
    return schema as MultipleBaseUrlsEnvironmentSchema).urls != null;
}
```

We could also export this function in a similar place as `isRawObjectDefinition` so that it can be easily referenced by other callers later on. This could be in its own file like [this](https://github.com/fern-api/fern/blob/6f002b525a2e1e512b41fe77dc4acf6fa78a2390/packages/cli/yaml/yaml-schema/src/utils/isRawProtobufSourceSchema.ts#L3).

packages/cli/openapi-ir-to-fern/src/__test__/__snapshots__/x-fern-audiences.test.ts.snap
```diff
@@ -66,7 +66,22 @@ exports[`x-fern-audiences > x-fern-audiences > docs 1`] = `
...
+        ],
+        "url": "https://api.com",
+      },
+      "sandbox": {
+        "audiences": [
+          "private",
+        ],
+        "url": "https://private.api.com",
+      },
+    },
```
amckinney, line 84: Nice 👏

amckinney (approved):
Nicely done, just a structural recommendation 👍

packages/cli/openapi-ir-to-fern/src/buildEnvironments.ts
RohinBhargava: Nice, it already existed, so imported it! Thanks for the tip

PR #4406, opened 2024-08-22
fix(csharp): fix warnings in generated `RawClient`

dcb6:


generators/csharp/codegen/src/asIs/RawClient.Template.cs
```diff
@@ -15,141 +15,141 @@ internal class RawClient(ClientOptions clientOptions)
     /// <summary>
     /// The gRPC client used to make requests.
     /// </summary>
-    public readonly RawGrpcClient Grpc = new RawGrpcClient(clientOptions);
+    public readonly RawGrpcClient Grpc = new(clientOptions);
```
dcb6, line 18: sorry for the horrible diff, prettier insisted on formatting it and i gave up on fixing it after banging my head for a while

dcb6: suggest hiding whitespace

amckinney (approved):

PR #4408, opened 2024-08-23
fix(cli): ir generation accurately generates examples for objects with extends

dsinghvi:
This PR updates the CLI to generate examples when the object type declaration only contains `extends` and no properties. Previously this case was being skipped.


packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
```diff
@@ -478,7 +478,7 @@ function convertObject({
     }
     return ExampleTypeShape.object({
         properties:
-            rawObject.properties != null
+            rawObject.properties != null || rawObject.extends != null
```
dsinghvi, line 481: critical change

amckinney (approved):

PR #4409, opened 2024-08-23
chore(cli): produce ir v57.8.0 which includes raw datetimes

dsinghvi:
This PR updates IR generation to include the raw `datetime` value. This is critical for asserting datetime responses when generating wire tests.

### DIFF
--- a/packages/cli/generation/ir-generator/src/examples/ExampleGenerator.ts
+++ b/packages/cli/generation/ir-generator/src/examples/ExampleGenerator.ts
@@ -905,10 +905,12 @@ export class ExampleGenerator {
                 const exDateTime = example != null && typeof example === "string" ? example : Examples.DATE_TIME;
                 return {
                     jsonExample: exDateTime,
-                    shape: ExampleTypeReferenceShape.primitive(ExamplePrimitive.datetime({
-                        datetime: new Date(exDateTime),
-                        raw: exDateTime,
-                    }))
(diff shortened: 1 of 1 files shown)


packages/ir-sdk/fern/apis/ir-types-latest/definition/types.yml
```diff
@@ -470,6 +468,11 @@ types:
       bigInteger:
         type: bigint
         key: bigInteger
+  
+  ExampleDatetime: 
+    properties: 
+      datetime: datetime
+      raw: optional<string>
```
dsinghvi, line 475: critical change is adding this `raw` field

packages/cli/generation/ir-generator/src/converters/type-declarations/convertExampleType.ts
```diff
@@ -372,7 +372,12 @@ function convertPrimitiveExample({
             if (typeof example !== "string") {
                 throw new Error("Example is not a string");
             }
-            return ExampleTypeReferenceShape.primitive(ExamplePrimitive.datetime(new Date(example)));
+            return ExampleTypeReferenceShape.primitive(
+                ExamplePrimitive.datetime({
+                    datetime: new Date(example),
+                    raw: example
+                })
+            );
```
dsinghvi, line 380: critical change is specifying `raw: example`

PR #4418, opened 2024-08-26
fix(docs): spelling

chdeskur:


dannysheridan:
Why do I suspect that Vale or Grammarly is behind the scenes for this PR?

PR #4432, opened 2024-08-27
chore(cli): add `--version` command to retrieve generator version

armandobelardo:
**What is this?**
This PR adds a command to get the current version for a generator, note that the command is a bit verbose, requiring group and generator name (e.g. `fern generator get --version --group X --generator Y`), and it is nested under a node called `get` to allow for more flags, etc. to be added down the line.

Definitely not wed to the CLI structure for the command

**Why this is needed?**
`fern-bot` needs to know the version range it's upgrading generators through in order to retrieve the changelog.
The idea with this command is it can be run once before the upgrade, then once after, then the bot can take that version range and:
- quit if the version is unchanged
- call FDR to get the changelog given the "from" and "to" versions

**Alternatives**
We could add version to a pre-existing command (ex `fern generator list`), or add before and after as structured output to the `fern generator upgrade` command, but adding a new command felt like the cleanest from a UX perspective and from reading it within fern-bot

### DIFF
--- a/packages/cli/cli/src/cliV2.ts
+++ b/packages/cli/cli/src/cliV2.ts
@@ -159,13 +159,11 @@ export function addGeneratorCommands(cli: Argv<GlobalCliOptions>, cliContext: Cl
                         .option("group", {
                             string: true,
                             demandOption: true,
-                            description:
-                                "The group in which the generator is located."
+                            description: "The group in which the generator is located."
                         })
                         .option("api", {
                             string: true,
-                            description:
-                                "The API in which the generator is located."
+                            description: "The API in which the generator is located."
                         })
                         .option("version", {
                             boolean: true,
@@ -183,17 +181,24 @@ export function addGeneratorCommands(cli: Argv<GlobalCliOptions>, cliContext: Cl
                             includeMajor: argv.includeMajor
                         }
                     });
+                    const generator = await getGeneratorMetadata({
+                        cliContext,
+                        generatorFilter: argv.generator,
+                        groupFilter: argv.group,
+                        apiFilter: argv.api,
+                        project: await loadProjectAndRegisterWorkspacesWithContext(cliContext, {
+                            commandLineApiWorkspace: argv.api,
+                            defaultToAllApiWorkspaces: true
+                        })
+                    });
+                    if (generator == null) {
+                        const maybeApiFilter = argv.api ? ` for API ${argv.api}` : "";
+                        cliContext.failAndThrow(
+                            `Generator ${argv.generator}, in group ${argv.group}${maybeApiFilter} was not found.`
+                        );
+                    }
                     if (argv.version) {
-                        await getGeneratorMetadata({
-                            cliContext,
-                            generatorFilter: argv.generator,
-                            groupFilter: argv.group,
-                            apiFilter: argv.api,
-                            project: await loadProjectAndRegisterWorkspacesWithContext(cliContext, {
-                                commandLineApiWorkspace: argv.api,
-                                defaultToAllApiWorkspaces: true
-                            })
-                        });
+                        process.stdout.write(generator.version);
                     }
                 }
             );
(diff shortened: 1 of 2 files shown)


packages/cli/cli/src/cliV2.ts
```diff
@@ -143,6 +144,58 @@ export function addGeneratorCommands(cli: Argv<GlobalCliOptions>, cliContext: Cl
...
+            .command(
+                // Meant to retrieve metadata about the generator, for now this is hidden and the only option is --version
+                "get",
+                false,
+                (yargs) =>
+                    yargs
+                        .option("generator", {
+                            string: true,
+                            demandOption: true,
+                            description: "The type of generator to upgrade, ex: `fern-typescript-node-sdk`."
```
amckinney, line 157:
```suggestion
                            description: "The name of the generator to get, ex: `fern-typescript-node-sdk`."
```

packages/cli/cli/src/cliV2.ts
```diff
@@ -143,6 +144,58 @@ export function addGeneratorCommands(cli: Argv<GlobalCliOptions>, cliContext: Cl
...
+                        })
+                        .option("group", {
+                            string: true,
+                            demandOption: true,
+                            description:
+                                "The group in which the generator is located, if group is not specified, the all generators of the specified type will be upgraded."
```
amckinney, line 163:
```suggestion
                                "The group in which the generator is located."
```

packages/cli/cli/src/cliV2.ts
```diff
@@ -143,6 +144,58 @@ export function addGeneratorCommands(cli: Argv<GlobalCliOptions>, cliContext: Cl
...
+                        })
+                        .option("api", {
+                            string: true,
+                            description:
+                                "The API to upgrade the generator for. If not specified, the generator will be upgraded for all APIs."
```
amckinney, line 168:
```suggestion
                                "The API in which the generator is located."
```

packages/cli/cli/src/commands/generator-metadata/getGeneratorMetadata.ts
```diff
@@ -0,0 +1,62 @@
+import { generatorsYml } from "@fern-api/configuration";
+import { Project } from "@fern-api/project-loader";
+import { CliContext } from "../../cli-context/CliContext";
+
+export async function getGeneratorMetadata({
```
amckinney, line 5: Should this function return a `GeneratorInfo` type, then the `process.stdout.write` is handled at the CLI command layer? I imagine we'd have something like:

```typescript
// We might be able to just use the zod schema directly instead,
// but this is the idea.
interface GeneratorMetadata {
  name: string;
  version: string;
  ...
}
```

Then the call-site would be adapted like so:

```typescript
const metadata = await getGeneratorMetadata({
    cliContext,
    generatorFilter: argv.generator,
    groupFilter: argv.group,
    apiFilter: argv.api,
    project: await loadProjectAndRegisterWorkspacesWithContext(cliContext, {
        commandLineApiWorkspace: argv.api,
        defaultToAllApiWorkspaces: true
    })
});
if (argv.version) {
  process.stdout.write(metadata.version);
}
```

This would also simplify the `generatorFound` boolean since we could just return the result directly whenever it's found.

armandobelardo: This doesn't mirror the structure of the other commands, but can do

packages/cli/ete-tests/src/tests/generators-get/generators-get.test.ts
```diff
@@ -0,0 +1,18 @@
...
+        const out = await runFernCli(
+            ["generator", "get", "--generator", "fernapi/fern-typescript-node-sdk", "--group", "local", "--version"],
+            {
+                cwd: pathOfDirectory
+            }
+        );
+
+        expect(out.stdout).toMatchSnapshot();
+    }, 60_000);
+});
```
amckinney, line 18: Nice 👏

PR #4435, opened 2024-08-27
chore(seed): migrate fastapi and pydantic generators to v2 publishing

armandobelardo:
See title!

This also adds a validation command to run on PRs to confirm your changelog file is correct

### DIFF
--- a/packages/seed/src/__test__/assets/add-multiple-versions.yml
+++ b/packages/seed/src/__test__/assets/add-multiple-versions.yml
@@ -1,12 +1,15 @@
 - version: 0.39.10
   ir_version: 53
   changelog_entry:
-    type: chore
+    - type: chore
+      summary: asd
 - version: 0.39.11
   ir_version: 53
   changelog_entry:
-    type: chore
+    - type: chore
+      summary: asd
 - version: 0.40.10
   ir_version: 53
   changelog_entry:
-    type: chore
+    - type: chore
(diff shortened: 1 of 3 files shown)


dsinghvi (approved):

generators/python/pydantic/unreleased.yml
```diff
@@ -0,0 +1,6 @@
+# Place to hold unreleased changes before a full release
+# this file is meant to follow the same release format as versions.yml
+# but is not processed, it's simply procedural
+# 
+# Rack up changes here, then copy the block over to versions.yml when ready to release
+- version: unreleased
```
amckinney, line 6: Nice 👏

packages/seed/src/cli.ts
```diff
@@ -342,6 +345,64 @@ function addRegisterCommands(cli: Argv) {
...
+                    const taskContextFactory = new TaskContextFactory(argv["log-level"]);
+                    const context = taskContextFactory.create("Register");
+
+                    await validateCliRelease({
+                        context
+                    });
+                }
+            )
+            .command(
+                "generator <generator>",
```
amckinney, line 369: Do you imagine we'll want an option that can validate _all_ generators? I suppose we can always make the positional parameter optional and roll out a `--all` flag (or somethimg similar).

amckinney (approved):

PR #4437, opened 2024-08-27
feat(csharp): Apply timeout on CancellationToken

amckinney:
This updates the `RawClient.cs` core utility applies the `Timeout` option defined on the generated `ClientOptions` and `RequestOptions` types.

> Note that if the user instruments their own `CancellationToken` with a timeout, it will still be applied. We chain together the `CancellationToken` so that as soon as any of them cancel, they are all cancelled. In other words, the shortest timeout or earliest cancellation will always propagate through linked cancellation tokens (regardless of the default `ClientOptions.Timeout` value).


dcb6 (approved):
awesome! 💯

PR #4438, opened 2024-08-27
feat(csharp): snippet.json generation

dcb6:
We now generate a `snippet.json` for each endpoint in C#.

### DIFF
--- a/generators/csharp/sdk/src/SdkGeneratorCli.ts
+++ b/generators/csharp/sdk/src/SdkGeneratorCli.ts
@@ -9,7 +9,7 @@ import { GeneratorNotificationService } from "@fern-api/generator-commons";
 import { FernGeneratorExec } from "@fern-fern/generator-exec-sdk";
 import { HttpService, IntermediateRepresentation } from "@fern-fern/ir-sdk/api";
 import { writeFile } from "fs/promises";
-import { SnippetJsonGenerator } from "./endpoint/snippets/SnippetsJsonGenerator";
+import { SnippetJsonGenerator } from "./endpoint/snippets/SnippetJsonGenerator";
 import { MultiUrlEnvironmentGenerator } from "./environment/MultiUrlEnvironmentGenerator";
 import { SingleUrlEnvironmentGenerator } from "./environment/SingleUrlEnvironmentGenerator";
 import { BaseApiExceptionGenerator } from "./error/BaseApiExceptionGenerator";


generators/csharp/codegen/src/ast/core/Writer.ts
```diff
@@ -204,6 +204,15 @@ export class Writer {
...
     }
 
+    public importsToString(): string | undefined {
+        const imports = this.stringifyImports();
+        return imports.length > 0 ? imports : undefined;
+    }
+
+    public getBuffer(): string {
+        return this.buffer;
+    }
```
amckinney, line 214: Let's just make the `buffer` property public instead - @dsinghvi recently made a comment in a separate PR, so we should be consistent.

That way we can just access `writer.buffer` instead of `writer.getBuffer()`.

generators/csharp/sdk/src/endpoint/snippets/EndpointSnippetsGenerator.ts
```diff
@@ -0,0 +1,111 @@
...
+        isGrpc: boolean;
+        example: ExampleEndpointCall;
+        serviceId: string;
+    }): SingleEndpointSnippet | undefined {
+        const methodInvocation = isGrpc
+            ? this.grpcEndpointGenerator.generateGrpcEndpointSnippet({
+                  example,
+                  endpoint,
+                  serviceId,
+                  clientVariableName: "client",
```
amckinney, line 85: nit: Should this be some `const` somewhere? If it's used elsewhere in the generator, I'd imagine it could be part of the `context`, or if not just as an un-exported `const` at the top of this file.

generators/csharp/sdk/src/endpoint/snippets/SnippetsJsonGenerator.ts
```diff
@@ -0,0 +1,92 @@
...
+    autogenerated: SingleEndpointSnippet[];
+    userSpecified: SingleEndpointSnippet[];
+}
+
+export interface SingleEndpointSnippet {
+    imports: string | undefined;
+    endpointCall: string;
+}
+
+export class SnippetsJsonGenerator {
```
amckinney, line 20: nit: The file it produces is called `snippet.json` so we can update this to match.
```suggestion
export class SnippetJsonGenerator {
```

generators/csharp/sdk/src/endpoint/snippets/SnippetsJsonGenerator.ts
```diff
@@ -0,0 +1,92 @@
...
+                            identifierOverride: httpEndpoint.id
+                        },
+                        snippet: FernGeneratorExec.EndpointSnippet.typescript({
+                            client: finalSnippet
+                        })
+                    };
+                    endpoints.push(endpoint);
+                }
+            }
+        }
```
amckinney, line 70: nit: It might be worth splitting this up into a few helper functions. I'd imagine the final result reading something like:

```typescript
for (const [_, service] of Object.entries(this.context.ir.services)) {
  for (const httpEndpoint of service.endpoints) {
    const endpointSnippets = this.getSnippetsForEndpoint(httpEndpoint.id) // This function handles lines 41-47.
    for (const endpointSnippet of endpointSnippets) {
      const csharpSnippet = this.getCsharpSnippet(endpointSnippet)
      endpoints.push({
         exampleIdentifier: snippet?.exampleIdentifier,
         id: {
           path: FernGeneratorExec.EndpointPath(this.getFullPathForEndpoint(httpEndpoint)),
           method: httpEndpoint.method,
           identifierOverride: httpEndpoint.id
         },
         snippet: csharpSnippet
      });
    }
  }
}
```

At that point, you can go one step further and just consolidate `this.getSnippetsForEndpoint` into the function responsible for mapping this into the `FernGeneratorExec.EndpointSnippet` instance, so you'd have:

```typescript
for (const [_, service] of Object.entries(this.context.ir.services)) {
  for (const httpEndpoint of service.endpoints) {
    for (const csharpSnippet of this.getCsharpSnippets(httpEndpoint.id)) {
      endpoints.push({
         exampleIdentifier: snippet?.exampleIdentifier,
         id: {
           path: FernGeneratorExec.EndpointPath(this.getFullPathForEndpoint(httpEndpoint)),
           method: httpEndpoint.method,
           identifierOverride: httpEndpoint.id
         },
         snippet: csharpSnippet
      });
    }
  }
}
```

generators/csharp/sdk/src/endpoint/snippets/SnippetsJsonGenerator.ts
```diff
@@ -0,0 +1,92 @@
...
+                        snippet.endpointCall
+                    }`;
+                    const endpoint: Endpoint = {
+                        exampleIdentifier: snippet?.exampleIdentifier,
+                        id: {
+                            path: FernGeneratorExec.EndpointPath(this.getFullPathForEndpoint(httpEndpoint)),
+                            method: httpEndpoint.method,
+                            identifierOverride: httpEndpoint.id
+                        },
+                        snippet: FernGeneratorExec.EndpointSnippet.typescript({
```
amckinney, line 63: We should include a TODO here and follow-up with an issue so we remember to update this to `FernGeneratorExec.EndpointSnippet.csharp` when it's available.

amckinney (approved):
Nice! Only nits - the generated `snippet.json` files look great 👍

seed/csharp-sdk/auth-environment-variables/snippet.json
```diff
@@ -0,0 +1,29 @@
...
+        {
+            "example_identifier": null,
+            "id": {
+                "path": "/apiKey",
+                "method": "GET",
+                "identifier_override": "endpoint_service.getWithApiKey"
+            },
+            "snippet": {
+                "type": "typescript",
+                "client": "using SeedAuthEnvironmentVariables;\n\nvar client = new SeedAuthEnvironmentVariablesClient(\"X_ANOTHER_HEADER\", \"API_KEY\");\nawait client.Service.GetWithApiKeyAsync();\n"
```
amckinney, line 13: Should we expect the `API_KEY` value here if environment variables are enabled? This parameter is optional, right?

seed/csharp-sdk/basic-auth-environment-variables/snippet.json
```diff
@@ -0,0 +1,29 @@
...
+        {
+            "example_identifier": null,
+            "id": {
+                "path": "/basic-auth",
+                "method": "GET",
+                "identifier_override": "endpoint_basic-auth.getWithBasicAuth"
+            },
+            "snippet": {
+                "type": "typescript",
+                "client": "using SeedBasicAuthEnvironmentVariables;\n\nvar client = new SeedBasicAuthEnvironmentVariablesClient(\"USERNAME\", \"PASSWORD\");\nawait client.BasicAuth.GetWithBasicAuthAsync();\n"
```
amckinney, line 13: Similar to the other comment for environment variables, but this is for basic auth.

seed/csharp-sdk/multi-url-environment-no-default/snippet.json
```diff
@@ -0,0 +1,29 @@
...
+        {
+            "example_identifier": null,
+            "id": {
+                "path": "/ec2/boot",
+                "method": "POST",
+                "identifier_override": "endpoint_ec2.bootInstance"
+            },
+            "snippet": {
+                "type": "typescript",
+                "client": "using SeedMultiUrlEnvironmentNoDefault;\n\nvar client = new SeedMultiUrlEnvironmentNoDefaultClient(\"TOKEN\");\nawait client.Ec2.BootInstanceAsync(new BootInstanceRequest { Size = \"string\" });\n"
```
amckinney, line 13: Should we expect an environment variable parameter if there is no default environment?

generators/csharp/sdk/src/endpoint/snippets/EndpointSnippetsGenerator.ts
dcb6: yeah agree there's some weirdness here. when i wrote the method, i wasn't sure if this could vary by context and i'd want to be able to pass in the name of the client variable. turns out i've just always used "client" 

i'll move this to the context

generators/csharp/sdk/src/endpoint/snippets/SnippetsJsonGenerator.ts
dcb6: made it! ([link](https://linear.app/buildwithfern/issue/FER-3108/create-and-use-csharp-snippet-type))

seed/csharp-sdk/auth-environment-variables/snippet.json
dcb6: great find, fixing

seed/csharp-sdk/basic-auth-environment-variables/snippet.json
dcb6: fixing this as well

seed/csharp-sdk/multi-url-environment-no-default/snippet.json
dcb6: good catch, i have [this related ticket open](https://linear.app/buildwithfern/issue/FER-2967/improvement-make-client-options-required-if-multi-url-environment-with). maybe we can leave it as a follow up, given the client needs changes? my assumption was that will be pretty rare case, although I could be wrong

would you expect there to be a parameter, or for `ClientOptions` to be required and for `Environment` to be required? i suppose the former, and we remove `Environment` from `ClientOptions`?

generators/csharp/codegen/src/ast/core/Writer.ts
dcb6: 👍 will do

generators/csharp/sdk/src/endpoint/snippets/SnippetsJsonGenerator.ts
dcb6: appreciate the thorough refactor suggestions!

just pushed, ended up implementing it closer to the first example you gave because of trouble with this line:
```
         exampleIdentifier: snippet?.exampleIdentifier,
```

generators/csharp/sdk/src/endpoint/snippets/EndpointSnippetsGenerator.ts
dcb6: actually, in the mock server tests, i pass in `Client` because it's a property there. going to leave this as is unless you have other suggestions

amckinney: Works for me, thanks for taking a closer look!

seed/csharp-sdk/multi-url-environment-no-default/snippet.json
amckinney: Oh nice, didn't realize you were already tracking! Happy to handle this in a follow-up.

Good point on the required parameter - I think I'd expect it to be a required positional parameter, and we remove `Environment` from `ClientOptions` like you suggested.

generators/csharp/sdk/src/endpoint/snippets/SnippetJsonGenerator.ts
```diff
@@ -0,0 +1,91 @@
...
+
+        function getCsharpSnippet(endpointSnippet: SingleEndpointSnippet): string {
+            const snippetImportList = endpointSnippet.imports?.split("\n") ?? [];
+            const uniqueOrderedImports = Array.from(new Set([...rootClientImportList, ...snippetImportList]))
+                .filter((importString) => importString !== "")
+                .sort();
+            return `${uniqueOrderedImports.join("\n")}\n\nvar client = ${rootClientSnippet.body}${
+                endpointSnippet.endpointCall
+            }`;
+        }
```
amckinney, line 38: I would have expected this to be an additional top-level `private` function. If we need access to the `rootClientSnippet` defined above, we can expose another parameter (rather than relying on the closure).

Otherwise, it's a little tough to follow when the function is defined within the `generate` function body.

seed/csharp-sdk/multi-url-environment-no-default/snippet.json
dcb6: great! linked this thread in the ticket as well so we can see these thoughts

PR #4439, opened 2024-08-27
feat(csharp): Add exponential backoff retrier

amckinney:
## Overview

This adds the exponential backoff retry mechanism to the `RawClient` core utility. This is inspired by the same approach we use in the TypeScript SDK ([link](https://github.com/fern-api/fern/blob/d3b8d8ded44402161fb355f883082a7ead6b38de/seed/ts-sdk/code-samples/src/core/fetcher/requestWithRetries.ts#L5)).

With this, the `MaxRetries` option configured on the `ClientOptions` and `RequestOptions` is used to automatically retry requests that are deemed retriable (i.e. status code is 408, 429, or >= 500).

## Improvements

Note that this also makes a variety of other `RawClient` changes, some of which were directly recommended by [Rider](https://www.jetbrains.com/rider/):
1. Adds a private `BuildHttpRequest` method.
2. Adapts the request body transformation to use a `switch` rather than an `if`.
3. Update the `BuildUrl` method to use the more idiomatic slice operator (i.e. `url[..^1]`).
4. The gRPC channel is lazily loaded (which is better for users that might not actually need to call any gRPC endpoints).

## Tests

This also adds the `RawClientTests.cs` file in the generated `.Test` project, so that the retry logic is tested in every SDK. Just like the mock tests, this uses `WireMock` to test against an in-memory HTTP server. We will follow-up with additional tests for all the core utilities in a separate change, but this at least sets the stage.

### DIFF
--- a/seed/csharp-sdk/exhaustive/explicit-namespaces/src/SeedExhaustive/ReqWithHeaders/ReqWithHeadersClient.cs
+++ b/seed/csharp-sdk/exhaustive/explicit-namespaces/src/SeedExhaustive/ReqWithHeaders/ReqWithHeadersClient.cs
@@ -38,8 +38,8 @@ public partial class ReqWithHeadersClient
         var _headers = new Headers(
             new Dictionary<string, string>()
             {
-                { "X-TEST-ENDPOINT-HEADER", request.XTestEndpointHeader },
                 { "X-TEST-SERVICE-HEADER", request.XTestServiceHeader },
+                { "X-TEST-ENDPOINT-HEADER", request.XTestEndpointHeader },
             }
         );
         var response = await _client.MakeRequestAsync(
--- a/seed/csharp-sdk/exhaustive/explicit-namespaces/src/SeedExhaustive/ReqWithHeaders/Requests/ReqWithHeaders.cs
+++ b/seed/csharp-sdk/exhaustive/explicit-namespaces/src/SeedExhaustive/ReqWithHeaders/Requests/ReqWithHeaders.cs
@@ -6,9 +6,9 @@ namespace SeedExhaustive.ReqWithHeaders;
 
 public record ReqWithHeaders
 {
-    public required string XTestEndpointHeader { get; set; }
-
     public required string XTestServiceHeader { get; set; }
+
+    public required string XTestEndpointHeader { get; set; }
 
     public required string Body { get; set; }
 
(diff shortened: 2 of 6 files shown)


seed/csharp-sdk/seed.yml
```diff
@@ -26,6 +26,7 @@ scripts:
   - docker: bitnami/dotnet-sdk:8
     commands:
       - dotnet build src -c Release /p:ContinuousIntegrationBuild=true
+      - dotnet test src --filter "FullyQualifiedName~SeedApi.Test.Core"
```
amckinney, line 29: Note that we're running the `RawClientTests.cs` in CI now.

seed/csharp-sdk/streaming-parameter/src/SeedStreaming.Test/Unit/MockServer/BaseMockServerTest.cs
```diff
@@ -1,39 +0,0 @@
-using NUnit.Framework;
```
amckinney, line 1: Note that we can remove this file since there aren't any mock tests generated for this test definition.

generators/csharp/codegen/src/asIs/RawClient.Template.cs
```diff
@@ -110,23 +86,68 @@ public record ApiResponse
...
+        var maxRetries = request.Options?.MaxRetries ?? Options.MaxRetries;
+        var response = await httpClient.SendAsync(BuildHttpRequest(request), cancellationToken);
+        for (var i = 0; i < maxRetries; i++)
+        {
+            if (!ShouldRetry(response))
+            {
+                break;
+            }
+            var delayMs = Math.Min(InitialRetryDelayMs * (int)Math.Pow(2, i), MaxRetryDelayMs);
+            await System.Threading.Tasks.Task.Delay(delayMs, cancellationToken);
```
amckinney, line 105: Note that this is the fully-qualified function name for `Task.Delay` to handle the case when the user defines a `Task` type (i.e. the `csharp-namespace-conflict` test definition). 

If there are more conflicts later on, we can refactor the `RawClient` so it's not an as-is file, but the benefits of keeping this file as-is outweighs the potential conflict with other types for now (especially because they're very unlikely).

dsinghvi (approved):
The tests look sick

PR #4445, opened 2024-08-28
fix(csharp): fix unions as request and response types

dcb6:
## Changelog
- Fix: Fix a bug where `OneOf` used directly as request or response types fail serialization.

- Feature: Support service-level headers.

## Other
Also fixes wire testing generation to always include auth when initializing the client, even if it's optional due to env vars.

### DIFF
--- a/generators/csharp/sdk/src/endpoint/request/WrappedEndpointRequest.ts
+++ b/generators/csharp/sdk/src/endpoint/request/WrappedEndpointRequest.ts
@@ -109,7 +109,7 @@ export class WrappedEndpointRequest extends EndpointRequest {
 
     public getHeaderParameterCodeBlock(): HeaderParameterCodeBlock | undefined {
         const service = this.context.getHttpServiceOrThrow(this.serviceId);
-        const headers = [...this.endpoint.headers, ...service.headers];
+        const headers = [...service.headers, ...this.endpoint.headers];
         if (headers.length === 0) {
             return undefined;
         }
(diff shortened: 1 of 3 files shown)


dsinghvi (approved):

generators/csharp/sdk/src/endpoint/request/WrappedEndpointRequest.ts
```diff
@@ -107,12 +108,14 @@ export class WrappedEndpointRequest extends EndpointRequest {
     }
 
     public getHeaderParameterCodeBlock(): HeaderParameterCodeBlock | undefined {
-        if (this.endpoint.headers.length === 0) {
+        const service = this.context.getHttpServiceOrThrow(this.serviceId);
+        const headers = [...this.endpoint.headers, ...service.headers];
```
amckinney, line 112: nit: I would put the service headers _before_ the endpoint headers (so that it's from _least_ specific to _most_ specific.

generators/csharp/sdk/src/root-client/RootClientGenerator.ts
```diff
@@ -292,9 +292,13 @@ export class RootClientGenerator extends FileGenerator<CSharpFile, SdkCustomConf
         };
     }
 
-    public generateExampleClientInstantiationSnippet(
-        clientOptionsArgument?: csharp.ClassInstantiation
-    ): csharp.ClassInstantiation {
+    public generateExampleClientInstantiationSnippet({
+        clientOptionsArgument,
+        includeEnvVarArguments = false
```
amckinney, line 297: nit: Based on the optional declaration + usage, I _think_ we can drop the `= false` default value here.

amckinney (approved):
Nice! Only nits (take em or leave em).

generators/csharp/sdk/src/endpoint/request/WrappedEndpointRequest.ts
dcb6: i like that 👍

PR #4447, opened 2024-08-28
fix(python): update CI to run over . vs a specific directory

armandobelardo:
seed doesn't reflect anything since we always generate mock integration tests as it's a generator exec config variable which isn't super controllable from seed.yml right now


amckinney (approved):
Assuming we don't need any extra flags whenever `tests/` doesn't exist, lgtm.

PR #4448, opened 2024-08-28
fix(cli): still create a GH release if a new version is released

armandobelardo:
This PR:
- adds `seed latest cli` + `seed latest generator` commands such that we can have those called to get the new verison in CI
- updates the current action that adds GH releases to use `seed latest cli`, if there's a new release then it adds the release and follows the steps

### DIFF
--- a/generators/python/sdk/versions.yml
+++ b/generators/python/sdk/versions.yml
@@ -1,3 +1,12 @@
+- version: 3.10.7
+  ir_version: 53
+  changelog_entry:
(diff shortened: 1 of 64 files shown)


dsinghvi (approved):

dsinghvi:
It's probably easier to have: 
- seed latest cli
- seed latest generator <id> 
- seed diff-latest cli 
- seed diff-latest generator <id>

PR #4452, opened 2024-08-28
fix(docs): remove hidden from front matter

chdeskur:
This PR removes `hidden: true` from the front matter documentation.


abvthecity:
Shout out to @cwcowellshah for finding this issue in our docs!

PR #4455, opened 2024-08-28
feat(csharp): Generate README.md

amckinney:
## Overview

This adds support for generating C# `README.md` files by interfacing with the `generator-cli` (via the `CsharpGeneratorAgent` type).

The following sections are generated for _all_ README.md files:
* Installation
* Usage
* Exception Handling
* Retries (Advanced)
* Timeouts (Advanced)

## Implementation

This updates the shared `AbstractGeneratorAgent` type so that it consolidates more of the shared logic required to produce `README.md` and `reference.md` files (e.g. reading the `features.yml`, producing the GitHub remote configuration, etc). The generator is expected to implement an interface that looks like the following:

```typescript
export class CsharpGeneratorAgent extends AbstractGeneratorAgent<SdkGeneratorContext> {
    private readmeConfigBuilder: ReadmeConfigBuilder;

    public getReadmeConfig(
        args: AbstractGeneratorAgent.ReadmeConfigArgs<SdkGeneratorContext>
    ): FernGeneratorCli.ReadmeConfig {
        return this.readmeConfigBuilder.build({
            context: args.context,
            remote: args.remote,
            featureConfig: args.featureConfig,
            endpointSnippets: args.endpointSnippets
        });
    }

    public getLanguage(): FernGeneratorCli.Language {
        return FernGeneratorCli.Language.Csharp;
    }
}
```

The `ReadmeConfigBuilder` used here is scoped to each language, but it shares a similar structure between each implementation (i.e. it interacts with a `ReadmeSnippetBuilder` and produces the corresponding `FernGeneratorCli.LanguageInfo`).

## Future improvements

There's additional functionality we can consolidate in the new `AbstractReadmeSnippetBuilder`. 

For now, both of the language-specific implementations have their own representation of the IR's `HttpEndpoint` (which is extended to include a `FernFilepath`). Once the `HttpEndpoint` itself has the `FernFilepath`, we can easily refactor the `AbstractReadmeSnippetBuilder` as a generic type, and DRY up many of the helper functions (e.g. `buildEndpoints`, `getEndpointsForFeature`, etc).


dsinghvi (approved):

generators/csharp/sdk/src/generator-agent/ReadmeSnippetBuilder.ts
```diff
@@ -0,0 +1,180 @@
...
+import { SdkGeneratorContext } from "../SdkGeneratorContext";
+import { AbstractReadmeSnippetBuilder } from "@fern-api/generator-commons";
+import { EndpointSnippetsGenerator } from "../endpoint/snippets/EndpointSnippetsGenerator";
+
+interface EndpointWithFilepath {
+    endpoint: HttpEndpoint;
+    fernFilepath: FernFilepath;
+}
+
+export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
```
dsinghvi, line 13: @amckinney can we make the file structure here 
```
readme/
  ReadmeConfigBuilder.ts
  ReadmeSnippetBuilder.ts
```

dsinghvi: and move `CsharpSdkGeneratorAgent.ts` to the top?

amckinney: Great call, this reads a lot better. Edited both C# and TypeScript.

PR #4463, opened 2024-08-29
feat(docs): add clarification for generators.yml

chdeskur:


dsinghvi:
The long code block titles are pretty brutal, i think an `AccordionGroup` might be better or just including the `api: ` in the first example 
Screenshot 2024-08-29 at 3 22 58 PM

dsinghvi (approved):

PR #4464, opened 2024-08-29
feat(csharp): Generate reference.md

amckinney:
This adds support for generating `reference.md` by interfacing with the `generator-cli` agent. An example `reference.md` file preview is shown below:

image

This also updates a few things along the way:
1. Extract shared logic in the `EndpointGenerator` into a new `EndpointSignatureInfo` interface, which is retrievable from the `AbstractEndpointGenerator`.
    * This refactors both the HTTP and gRPC variants so that they all use the same underlying implementation.
2. Add a `printType` utility function on the `AbstractCsharpGeneratorContext`.
    * This makes it easier to get a `string` representation for `Type` nodes in particular.

### DIFF
--- a/generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
+++ b/generators/csharp/codegen/src/context/AbstractCsharpGeneratorContext.ts
@@ -430,7 +430,8 @@ export abstract class AbstractCsharpGeneratorContext<
             allNamespaceSegments: this.getAllNamespaceSegments(),
             allTypeClassReferences: this.getAllTypeClassReferences(),
             rootNamespace: this.getNamespace(),
-            customConfig: this.customConfig
+            customConfig: this.customConfig,
+            skipImports: true
         });
     }
 
--- a/generators/csharp/codegen/src/project/CsharpProject.ts
+++ b/generators/csharp/codegen/src/project/CsharpProject.ts
@@ -45,6 +45,10 @@ export class CsharpProject {
         });
     }
 
+    public getProjectDirectory(): RelativeFilePath {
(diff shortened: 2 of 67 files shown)


generators/csharp/sdk/src/reference/getReferenceConfigBuilder.ts
```diff
@@ -0,0 +1,128 @@
+import { ReferenceConfigBuilder } from "@fern-api/generator-commons";
+import { FernGeneratorCli } from "@fern-fern/generator-cli-sdk";
+import { HttpEndpoint, HttpService, ServiceId } from "@fern-fern/ir-sdk/api";
+import { EndpointSignatureInfo } from "../endpoint/EndpointSignatureInfo";
+import { SingleEndpointSnippet } from "../endpoint/snippets/EndpointSnippetsGenerator";
+import { SdkGeneratorContext } from "../SdkGeneratorContext";
+
+export function getReferenceConfigBuilder({ context }: { context: SdkGeneratorContext }): ReferenceConfigBuilder {
```
dsinghvi, line 8: nit should we call this method `buildReference()`

dsinghvi (approved):
looks great cc @chdeskur

generators/csharp/sdk/src/reference/getReferenceConfigBuilder.ts
amckinney: Yea good call, originally felt that it would read weird since we're actually just returning the builder, but `buildReference` is still a lot more clear than `getReferenceConfigBuilder`. I'll update.

PR #4465, opened 2024-08-29
fix(docs): improve openapi overrides docs

dsinghvi:
This PR starts to reorganize the OpenAPI definition section (ideally it should be 1:1 with the fern definition but with how to do things in the OpenAPI spec). 

Other than that, we also rewrite the OpenAPI overrides section so that its easier to understand.

### DIFF
--- a/fern/pages/api-definition/openapi/overrides.mdx
+++ b/fern/pages/api-definition/openapi/overrides.mdx
@@ -52,9 +52,6 @@ Follow the steps below to configure your OpenAPI overrides:
 Simply create a yaml file and write down all the overrides you want to add: 
 
 ```yaml overrides.yml 
-x-fern-global-headers:
-  - header: api-key
-    name: api-key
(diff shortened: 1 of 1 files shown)


fern/docs.yml
```diff
@@ -74,36 +88,22 @@ navigation:
               - page: Parameter Names
                 icon: fa-regular fa-input-text
                 path: ./pages/api-definition/openapi/extensions/parameter-names.mdx
-                slug: parameter-names
-              - page: Audiences
-                icon: fa-duotone fa-users
-                path: ./pages/api-definition/openapi/extensions/audiences.mdx
-                slug: audiences              
+                slug: parameter-names
```
dsinghvi, line 91: fyi that my plan is to eventually get rid of the broad `Extension` category and just have docs for each relevant concept (e.g. Audiences, Streaming, Examples, etc.)

PR #4497, opened 2024-09-02
fix(cli): Update filters within generator upgrade to appropriately ta…

armandobelardo:
…ke into account generator group name


packages/cli/cli/versions.yml
```diff
@@ -1,3 +1,10 @@
+- changelog_entry:
+  - summary: Update filters within `generator upgrade` to appropriately take into account generator group name.
```
dsinghvi, line 2: nit: this release note is a little hard to wrap your head around

```
summary: `fern generator upgrade` respects  the `--group` flag and only upgrades generators within a particular group.
```

dsinghvi (approved):

PR #4510, opened 2024-09-02
break(cli): require `generators.yml` to be present if using OpenAPI

dsinghvi:
Previously, it was valid to have the following structure in your fern folder: 

```sh
fern/
  fern.config.json
  openapi/
    opnenapi.yml
```

Fern would magically read the first file within your `openapi` folder. This PR requires that all users using OpenAPI specify their API location in `generators.yml`. 

When you run `fern ugprade` a migration will run and update the fern directory to contain a `generators.yml` that explicitly points to the OpenAPI spec, so that users don't need to take manual action themselves. 
```yml
# generators.yml
api: 
  path: openapi/openapi.yml
```

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,5 +1,22 @@
 - changelog_entry:
   - summary: | 
+      Every fern folder that is using OpenAPI must configure an explicit location to the 
+      OpenAPI spec. The location can be configured in your `generators.yml`: 
+
+      ```yml
+      api: 
+        path: path/to/openapi.yml
+      ```
+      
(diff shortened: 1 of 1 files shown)


mstade:
> Is our structure wrong in any way? Why would this ask to migrate, and then do nothing?

I believe I've figured out the root cause of our issue and posting it here for posterity. Our setup is admittedly a bit esoteric and what we do is we generate our `generators.yml` files, for reason. Now it turns out the generated files were indented, and while this seems to not have affected the actual generation in any negative way, it does seem like this tripped up whatever the migration check is doing, causing the prompt which broke our build.

Removing the indentation seems to have resolved this however. My apologies for the noise.

PR #4512, opened 2024-09-02
feat(cli): support generics in fern definition

RohinBhargava:
This PR introduces generics in the fern definition for reuse for commonly repeated types. For example, the following should now be possible:

```yaml
GenericTest<T>:
    properties:
      value: T
      other-value: Movie
  
  GenericApplication:
    type: GenericTest<CreateMovieRequest>
    docs: This is a generic application type

  Sample:
    type: GenericTest<Movie>
```

Note, we can do this for any number of generics, one recursion layer deep.

The logic to make this happens is as follows:
1. If we match on regex `/(\w+)<([\w,\s]+)>/`, and the first group is not one of the reserved keywords [`set`, `list`, `literal`, `map`, `optional`], then we treat an alias as a generic.
2. We ignore generic object definitions when appending to IR, but allow for all of the yaml-schema add-ons of the object to be inherited by the generics.
3. We use the name of the applied generic (`GenericApplication` and `Sample` above) as the name of an object, with properties superimposed from the generics underneath. There are no limits to generic reuse.

What we do not support:
- Nesting of any kind for container/recursive generics.
- Inlining generics as property values.
- Inlining of generic definitions.

### DIFF
--- a/fern/pages/api-definition/fern-definition/types.mdx
+++ b/fern/pages/api-definition/fern-definition/types.mdx
@@ -209,20 +209,20 @@ Fern supports shallow generic objects, to minimize code duplication. You can
 define a generic for reuse like so:
 
 ```yaml
-Response<Input, Output>:
+MySpecialMapItem<Key, Value>:
   properties:
-    input: Input,
-    output: Output
+    key: Key, 
+    value: Value,
     diagnostics: string
 ```
 
 Now, you can instantiate generic types as a type alias:
 
 ```yaml
-IntegerResponse:
+StringIntegerMapItem:
   type: Response<string, number>
 
-StringResponse:
+StringStringMapItem:
   type: Response<string, string>
 ```
 
@@ -230,15 +230,13 @@ You can now freely use this type as if it were any other type! Note, generated
 code will not use generics. The above example will be generated in typescript as:
 
 ```typescript
-type IntegerResponse = {
-
-  input: string,
-  output: number,
+type StringIntegerMapI = {
+  key: string,
+  value: number,
   diagnostics: string
 }   
 
 type StringResponse = {
-  type: "string",
   input: string,
   output: string,
   diagnostics: string


packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
```diff
@@ -193,6 +194,10 @@ export async function generateIntermediateRepresentation({
                 }
 
                 for (const [typeName, typeDeclaration] of Object.entries(types)) {
+                    if (getGenericDetails(typeName)?.isGeneric) {
```
dsinghvi, line 197: nit: this would read cleaner if there was just a function called `isGeneric`

packages/cli/generation/ir-generator/src/converters/type-declarations/getReferencedTypesFromRawDeclaration.ts
```diff
@@ -26,7 +27,12 @@ export function getReferencedTypesFromRawDeclaration({
 }): DeclaredTypeName[] {
     const rawTypeReferences = visitRawTypeDeclaration<string[]>(typeDeclaration, {
         alias: (aliasDeclaration) => {
-            return [typeof aliasDeclaration === "string" ? aliasDeclaration : aliasDeclaration.type];
+            const aliasStr = typeof aliasDeclaration === "string" ? aliasDeclaration : aliasDeclaration.type;
```
dsinghvi, line 30: nit: `aliasTypeReference` is more descriptive than `aliasStr`

packages/cli/generation/ir-generator/src/converters/type-declarations/getReferencedTypesFromRawDeclaration.ts
```diff
@@ -26,7 +27,12 @@ export function getReferencedTypesFromRawDeclaration({
...
+            const maybeGenericDetails = getGenericDetails(aliasStr);
+            if (maybeGenericDetails && maybeGenericDetails.isGeneric) {
+                return maybeGenericDetails.arguments ?? [];
+            }
```
dsinghvi, line 34: why are we returning `maybeGenericDetails.arguments ?? [];` here? 

imagine the following case: 
```yml

Node<T>: 
  properties: 
    data: T
    other: Metadata
    next: optional<string>
    prev: optional<string>

NumericNode: 
  type: Node<Datum> 
```

then `getReferencedTypesFromRawDeclaration(NumericNode)` should return `Datum` and `Metadata`

dsinghvi: I would expect that this part of the function builds the underlying object type declaration and calls into the object part of the visitor

packages/cli/generation/ir-generator/src/converters/type-declarations/convertGenericTypeDeclaration.ts
```diff
@@ -0,0 +1,54 @@
...
+            properties: await getObjectPropertiesFromRawObjectSchema(
+                { properties: Object.fromEntries(newProperties) },
+                file
+            ),
+            extraProperties: resolvedBaseGeneric.declaration["extra-properties"] ?? false,
+            extendedProperties: undefined
+        });
+    }
+
+    throw new Error(`Generic type definition not supported: ${genericInstantiation}`);
```
dsinghvi, line 53: as you already mentioned you'll want a `fern check` rule in this case which makes sure that generic types are only being referenced in aliases

packages/cli/generation/ir-generator/src/utils/getGenericDetails.ts
```diff
@@ -0,0 +1,15 @@
+export function getGenericDetails(name: string): { isGeneric: true; name?: string; arguments?: string[] } | undefined {
+    const genericMatch = name.match(/([\w.]+)<([\w,\s]+)>/);
+
+    if (genericMatch?.[0] != null) {
+        if (!genericMatch[1] || !new Set(["optional", "set", "list", "map", "literal"]).has(genericMatch[1]?.trim())) {
```
dsinghvi, line 5: where else are these hardcoded? i think this is going to get out of date too fast

packages/cli/generation/ir-generator/src/utils/isGeneric.ts
```diff
@@ -0,0 +1,9 @@
+export function isGeneric(name: string): boolean {
+    const genericMatch = name.match(/([\w.]+)<([\w,\s]+)>/);
+
+    return (
+        genericMatch?.[0] != null &&
+        genericMatch[1] != null &&
+        !new Set(["optional", "set", "list", "map", "literal"]).has(genericMatch[1].trim())
```
dsinghvi, line 7: same here, can we be sharing this with existing logic?

RohinBhargava: Consolidated a bit, where now `isGeneric` now leverages logic in `getGenericDetails`. Ideally there is a place where these utility functions could be shared between CLI packages. Taking a look for that

packages/cli/generation/ir-generator/src/utils/getGenericDetails.ts
RohinBhargava: Only in these utility files now, of which there are 2.

RohinBhargava: Is there a place for shared Regex?

packages/cli/fern-definition/validator/src/rules/valid-generic/__test__/valid-generic.test.ts
```diff
@@ -0,0 +1,80 @@
...
+                severity: "error",
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericApplicationDiscriminatedUnion"],
+                message: 'Generic values in discriminated unions are not supported: "foo: GenericUsedType<string>".'
+            },
+            {
+                severity: "error",
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericApplicationEnum"],
+                message: 'Generic values in enums are not supported: "GenericUsedType<string>".'
```
dsinghvi, line 33: these error messages aren't exactly clear -- what  are we trying to communicate here?

dsinghvi: How about `Cannot reference generic GenericUsedType<string> from enum`

packages/cli/fern-definition/validator/src/rules/valid-generic/__test__/valid-generic.test.ts
```diff
@@ -0,0 +1,80 @@
...
+            {
+                severity: "error",
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericApplicationObject"],
+                message: 'Generic values in object properties are not supported: "foo: GenericUsedType<string>".'
+            },
+            {
+                severity: "error",
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericApplicationDiscriminatedUnion"],
```
dsinghvi, line 26: ideally this is `types, GenericApplicationDiscriminatedUnion, foo` is there any way to make that happen?

packages/cli/fern-definition/validator/src/rules/valid-generic/__test__/valid-generic.test.ts
```diff
@@ -0,0 +1,80 @@
...
+                message: 'Generic values in discriminated unions are not supported: "foo: GenericUsedType<string>".'
```
dsinghvi, line 27: `Cannot reference generic GenericUsedType<string> from property foo`

packages/cli/fern-definition/validator/src/rules/valid-generic/__test__/valid-generic.test.ts
```diff
@@ -0,0 +1,80 @@
...
+                severity: "error",
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericApplicationEnum"],
+                message: 'Generic values in enums are not supported: "GenericUsedType<string>".'
+            },
+            {
+                severity: "error",
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericApplicationUndiscriminatedUnion"],
+                message: 'Generic values in unions are not supported: "GenericUsedType<string>".'
```
dsinghvi, line 39: `Cannot reference generic GenericUsedType<string> from union`

packages/cli/fern-definition/validator/src/rules/valid-generic/__test__/valid-generic.test.ts
```diff
@@ -0,0 +1,80 @@
...
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericUsedDiscriminatedUnion<T>"],
+                message: "Generic declarations are only supported with objects."
+            },
+            {
+                severity: "error",
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericWrongNumberOfArgumentsApplied"],
+                message:
+                    'Generic "GenericWrongNumberOfArguments" expects 6 arguments but was instantiated with 2 arguments.'
```
dsinghvi, line 70: `GenericWrongNumberOfArguments expects 6 arguments but only two supplied`

packages/cli/fern-definition/validator/src/rules/valid-generic/__test__/valid-generic.test.ts
```diff
@@ -0,0 +1,80 @@
...
+            },
+            {
+                severity: "error",
+                relativeFilepath: "1.yml",
+                nodePath: ["types", "GenericWrongNumberOfArgumentsNoArgumentsApplied"],
+                message: "Generic value is supplied, but no arguments are defined."
```
dsinghvi, line 76: `GenericWrongNumberOfArgumentsNoArgumentsApplied accepts no generics, but received <a,b,c>`

packages/cli/generation/ir-generator/src/__test__/fixtures/generics/fern/definition/imdb.yml
```diff
@@ -0,0 +1,89 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/fern-api/fern/main/fern.schema.json
+
+service:
```
dsinghvi, line 3: lets shrink this fixture to the minimal set needed

dsinghvi:
@RohinBhargava sorry for another round of comments, but I think once these are resolved we should be good to go. Can you also merge the docs PR into this one and add a changelog entry as well

dsinghvi (approved):
just need docs + cli changelog

packages/cli/cli/versions.yml
```diff
@@ -1,3 +1,12 @@
+- changelog_entry:
+  - summary: |
+      Adds generic object declarations to the fern definition. Now we can define generics and
+      use them in alias declarations to minimize code duplication.
```
dsinghvi, line 4: can we add a yaml example hear of how to use the feature

dsinghvi: as well as a link to the docs to learn more

PR #4518, opened 2024-09-03
chore(python): remove usage of pydantic aliases

armandobelardo:
Pydantic field aliases are removed and replaced with an internal representation. This allows for more robust handling of field aliases and prevents issues with Pydantic V2 and mypy.

Previously, you'd have for V1 and V2 compatibility in Pydantic, you'd want to conditionally apply the config class within the base model, however this would lead to mypy errors when filling out a model with it's field alias. To solve this, We used the deprecated `class Config`, regardless of the Pydantic version to satisfy mypy, which lead to warnings in the console.

Now, we've removed the field aliases and replaced them with an internal representation, which allows us to avoid pydantic config altogether.

### DIFF
--- a/generators/python/core_utilities/pydantic/unchecked_base_model.py
+++ b/generators/python/core_utilities/pydantic/unchecked_base_model.py
@@ -35,9 +35,7 @@ Model = typing.TypeVar("Model", bound=pydantic.BaseModel)
 
 class UncheckedBaseModel(UniversalBaseModel):
     if IS_PYDANTIC_V2:
-        model_config: typing.ClassVar[pydantic.ConfigDict] = pydantic.ConfigDict(
-            extra="allow"
-        )  # type: ignore # Pydantic v2
+        model_config: typing.ClassVar[pydantic.ConfigDict] = pydantic.ConfigDict(extra="allow")  # type: ignore # Pydantic v2
     else:
 
         class Config:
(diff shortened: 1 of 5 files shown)


seed/python-sdk/imdb/src/seed/imdb/types/create_movie_request.py
```diff
@@ -1,14 +1,16 @@
...
 from ...core.pydantic_utilities import IS_PYDANTIC_V2
 import typing
 import pydantic
 
 
 class CreateMovieRequest(UniversalBaseModel):
-    title: str
-    rating: float
+    title: typing_extensions.Annotated[str, FieldMetadata(alias="title")]
+    rating: typing_extensions.Annotated[float, FieldMetadata(alias="rating")]
```
dsinghvi, line 13: @armandobelardo how hard is it to ignore adding an alias if the alias is the same as the normal value?

armandobelardo: interesting, this shouldn't be happening, will take a look

armandobelardo: image
 🤦

armandobelardo: it's supposed to just be `if is_aliased`

dsinghvi (approved):

PR #4520, opened 2024-09-03
fix(cli): Substitute env vars before IR generation

amckinney:
This fixes an issue where the IR did not have all of the environment variables expanded from the user's `generators.yml`.

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,6 +1,7 @@
 - changelog_entry:
   - summary: |
-      Fix an issue where some environment variables were not substituted before generating the IR.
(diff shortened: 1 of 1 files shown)


dsinghvi (approved):

packages/cli/cli/versions.yml
```diff
@@ -1,3 +1,11 @@
+- changelog_entry:
+  - summary: |
+      Fix an issue where some environment variables were not substituted before generating the IR.
```
dsinghvi, line 3: nit: the user facing change here is probably more like

Fix an issue where some postman environment variables (e.g. `api key`) were not substituted when running `fern generate`.

amckinney: Sweet will update!

PR #4532, opened 2024-09-04
feat(docs): add hubspot web chat

dannysheridan:
Instructions provided by HubSpot

Place the code right before the end of the <body> tag on any page that you want to enable HubSpot on.

```
<!-- Start of HubSpot Embed Code -->
  <script type="text/javascript" id="hs-script-loader" async defer src="//js-na1.hs-scripts.com/44837001.js"></script>
<!-- End of HubSpot Embed Code -->
```


dannysheridan:
image
image

dannysheridan:
@abvthecity this PR is not ready to merge until we can fix the white background in dark mode. Let me know if you have an idea of how to fix this

dsinghvi:
@chdeskur any custom css/js we can do here ?

chdeskur:
@dsinghvi - the issue with the background color here is related to that Cohere issue, where computers with system default set to dark mode cause some properties to not be set properly. I can look into this further later in the week, but it's not a quick CSS fix.

PR #4533, opened 2024-09-04
feat(cli): support environment overrides with api configuration

dsinghvi:
Support the following configuration: 

```yaml
api: 
  default-environment: Production
  environments: 
    Production: 
      ServiceA: https://prod.serviceA.com	      	      
      ServiceB: https://prod.serviceB.com
    Staging: 
       ServiceA: https://prod.serviceA.com	      	      
       ServiceB: https://prod.serviceB.com
  specs: 
    - path: path/to/openapi
      url: ServiceA
    - path: path/to/openapi
      url: ServiceB
```

### DIFF
--- a/packages/cli/openapi-ir-to-fern/src/__test__/testConvertOpenApi.ts
+++ b/packages/cli/openapi-ir-to-fern/src/__test__/testConvertOpenApi.ts
@@ -8,6 +8,7 @@ import { convert } from "../convert";
 
 const FIXTURES_PATH = join(AbsoluteFilePath.of(__dirname), RelativeFilePath.of("fixtures"));
 
+// eslint-disable-next-line jest/no-export
 export declare namespace TestConvertOpenAPI {
     interface Options {
         asyncApiFilename?: string;


packages/cli/configuration/src/generators-yml/schemas/APIConfigurationV2Schema.ts
```diff
@@ -0,0 +1,26 @@
...
+    openapi: z.string(),
+    overrides: z.string().optional(),
+    namespace: z.string().optional()
+});
+
+export type AsyncAPISchema = z.infer<typeof OpenAPISpecSchema>;
+
+export const SpecSchema = z.union([OpenAPISpecSchema, AsyncAPISchema]);
+
+export const APIConfigurationV2Schema = z.strictObject({
```
dsinghvi, line 22: eventually we have to write a migration that only accepts the v2 schema

packages/cli/fern-definition/schema/src/schemas/file-schemas/RootApiFileSchema.ts
```diff
@@ -9,16 +9,21 @@ import { PaginationSchema } from "../PaginationSchema";
 import { VariableDeclarationSchema } from "../VariableDeclarationSchema";
 import { VersionDeclarationSchema } from "../VersionDeclarationSchema";
 
-export const RootApiFileSchema = z.strictObject({
+export const WithEnvironmentsSchema = z.strictObject({
+    "default-url": z.optional(z.string()),
+    "default-environment": z.optional(z.string().or(z.null())),
+    environments: z.optional(z.record(z.string(), EnvironmentSchema)),
+});
```
dsinghvi, line 16: moved into a separate schema so that we can reuse this across places

packages/cli/configuration/src/generators-yml/schemas/APIConfigurationV2Schema.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from "zod";
+import { RawSchemas } from "@fern-api/fern-definition-schema";
+
+export const OpenAPISpecSchema = z.strictObject({
+    openapi: z.string(),
```
armandobelardo, line 5: nit should this just be `api` with the `specs` block and such? just since this is "v2"?

packages/cli/configuration/src/generators-yml/schemas/APIConfigurationV2Schema.ts
```diff
@@ -0,0 +1,26 @@
...
+    overrides: z.string().optional(),
+    namespace: z.string().optional()
+});
+
+export type OpenAPISpecSchema = z.infer<typeof OpenAPISpecSchema>;
+
+export const AsyncAPISchema = z.strictObject({
+    openapi: z.string(),
```
armandobelardo, line 13: if not loose `api` should this at least be `asyncapi` so there's distinction between the types?

packages/cli/configuration/src/generators-yml/convertGeneratorsConfiguration.ts
```diff
@@ -242,6 +243,35 @@ async function parseAPIConfiguration(
     rawGeneratorsConfiguration: GeneratorsConfigurationSchema
 ): Promise<APIDefinition> {
     const apiConfiguration = rawGeneratorsConfiguration.api;
+
+    if (apiConfiguration != null && isApiConfigurationV2Schema(apiConfiguration)) {
+        return {
+            type: "singleNamespace",
```
armandobelardo, line 249: do we need to take into account the `namespace` configs to know if it's single or multi here?

packages/cli/openapi-ir-sdk/fern/definition/finalIr.yml
```diff
@@ -175,10 +175,18 @@ types:
         docs: |
           Expected error status codes for this endpoint, and their corresponding schema and examples.
           SDK generators will only read the StatusCodes. Docs generators will read the HttpError schema.
-      server: list<commons.Server>
```
armandobelardo, line 178: Do we need to remove `commons.Server` or anything?

armandobelardo: It does look like the types are pretty similar, any reason for the new type? and why `url` is optional?

armandobelardo (approved):

packages/cli/openapi-ir-sdk/fern/definition/finalIr.yml
dsinghvi: Yup so `url` is now optional b/c the user can just do `x-fern-server-name: auth` on an endpoint instead of having to supply the URL whereas before we always mandated that they specify the URL since we were adhering to OpenAPI servers.

packages/cli/configuration/src/generators-yml/convertGeneratorsConfiguration.ts
dsinghvi: Yes we do, the entire APIDefinition data model needs to change so leaving as a follow up. Cohere needs this so will do shortly.

packages/cli/configuration/src/generators-yml/schemas/APIConfigurationV2Schema.ts
dsinghvi: great catch, will update

PR #4560, opened 2024-09-06
feat(cli): support referencing endpoints for oauth via `Method and Path`

dsinghvi:
Endpoint references in oauth can either be via imports or via method and path directly. 

For example, the following in `api.yml` will work 
```
auth-schemes: 
  OAuth: 
    scheme: oauth
    type: client-credentials
    get-token:
      endpoint: POST /oauth/token
api:
  auth: OAuth
```

If the reference is incorrect, the user will also get a shorter error: 
```yml
[api]: api.yml -> auth-scheme -> OAuth
       Failed to resolve endpoint POST /oauth/token
```

### DIFF
--- a/packages/cli/fern-definition/schema/src/utils/HttpEndpointReferenceParser.ts
+++ b/packages/cli/fern-definition/schema/src/utils/HttpEndpointReferenceParser.ts
@@ -21,7 +21,8 @@ export declare namespace HttpEndpointReferenceParser {
  * Parses an HTTP endpoint reference like `POST /users/get`
  */
 export class HttpEndpointReferenceParser {
-    private REFERENCE_REGEX = /^(GET|POST|PUT|DELETE|PATCH)\s([a-zA-Z0-9]*)$/;
+    //eslint-disable-next-line
+    private REFERENCE_REGEX = /^(GET|POST|PUT|DELETE|PATCH)\s(\/[a-zA-Z0-9\/]*)$/;
 
     public validate(reference: string): HttpEndpointReferenceParser.ValidationResult {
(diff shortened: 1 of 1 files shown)


packages/cli/generation/ir-generator/src/resolvers/EndpointResolver.ts
```diff
@@ -17,21 +25,79 @@ interface RawEndpointInfo {
...
+    }): Promise<ResolvedEndpoint> {
+        const resolvedEndpoint = await this.resolveEndpoint({ endpoint, file });
         if (resolvedEndpoint == null) {
             throw new Error("Cannot resolve endpoint: " + endpoint + " in file " + file.relativeFilepath);
         }
         return resolvedEndpoint;
     }
 
-    public resolveEndpoint({
+    public async resolveEndpointByMethodAndPath({
```
amckinney, line 42: Not necessarily actionable here (can do this in a follow-up if you'd prefer), but we should reuse this when we convert the ReadmeConfig. 

Previously we made a [ReadmeEndpointCache](https://github.com/fern-api/fern/blob/af5350c74836fc2310121ad66b649e8d66555fb5/packages/cli/generation/ir-generator/src/converters/convertReadmeConfig.ts#L40) type since this format was specific to `README.md` generation, but we should be able to use the `EndpointResolver` instead now.

amckinney (approved):

packages/cli/generation/ir-generator/src/resolvers/EndpointResolver.ts
dsinghvi: The readme code path wasn't entirely clear to me, will do a separate PR for this

PR #4565, opened 2024-09-06
fix(cli): handle parsing json schema type arrays from openapi specs

dsinghvi:
The Fern OpenAPI importer now handles importing an array for the `type` key. 

```
  User: 
    properties: 
      name: 
        type: ["string"]
      id: 
        type: ["string", "number"]
```

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -15,7 +15,7 @@
   ir_version: 53
   version: 0.41.4
 
-- changelog_entry:
+- changelogEntry:
   - summary: |
       Allow referencing by method and path. For example, when configuring an 
       oauth scheme you can now do: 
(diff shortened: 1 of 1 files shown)


packages/cli/openapi-parser/src/schema/convertSchemas.ts
```diff
@@ -546,6 +546,15 @@ export function convertSchemaObject(
...
     }
 
+    // handle type array
+    if (Array.isArray(schema.type)) {
+        if (schema.oneOf == null) {
+            schema.oneOf = schema.type;
+        } else {
+            schema.oneOf.push(...schema.type);
+        }
+    }
```
dsinghvi, line 556: core change

armandobelardo (approved):

PR #4572, opened 2024-09-06
fix(go): Unknown values aren't dereferenced

amckinney:
This fixes an issue where optional `unknown` values (typed as `interface{}`) were mistakenly dereferenced.


generators/go/sdk/versions.yml
```diff
@@ -1,3 +1,11 @@
+- version: 0.23.7
+  createdAt: '2024-09-06'
```
armandobelardo, line 2: This isn't needed as a heads up

armandobelardo (approved):

PR #4575, opened 2024-09-07
fix(fastapi): default on pydantic aliases to fix serialization

dsinghvi:
The previous version of the FastAPI generator defaulted on a setting which removed pydantic aliases. Because FastAPI calls into the pydantic serialization itself, the generated models must have aliases.


dsinghvi:
fyi @armandobelardo

PR #4578, opened 2024-09-07
chore(csharp): refactor writer, ast node, and code block into commons

dsinghvi:
In this PR, we update the generator commons package to export an `AbstractWriter`, `AbstractAstNode` and `CodeBlock`. These core building blocks will help hit the ground running as new generators are added.

### DIFF
--- a/generators/csharp/codegen/src/csharp.ts
+++ b/generators/csharp/codegen/src/csharp.ts
@@ -19,8 +19,7 @@ import {
     Set,
     Switch,
     Ternary,
-    TestClass,
-    Type
+    TestClass
 } from "./ast";
 import { ReadOnlyMemory } from "./ast/ReadOnlymemory";
 
@@ -112,7 +111,6 @@ export function enumInstantiation(args: EnumInstantiation.Args): EnumInstantiati
     return new EnumInstantiation(args);
 }
 
-export const Types = Type;
 export {
     Annotation,
     Class,
@@ -128,6 +126,7 @@ export {
     MethodType,
     Parameter,
     Type,
+    Type as Types,
     Writer,
     type Argument,
     type Arguments,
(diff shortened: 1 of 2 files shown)


generators/commons/src/ast/AbstractWriter.ts
```diff
@@ -0,0 +1,161 @@
...
+        node.write(this);
+    }
+
+    /**
+     * Writes a node but then suffixes with a `;` and new line
+     * @param node
+     */
+    public writeNodeStatement(node: AbstractAstNode): void {
+        node.write(this);
+        this.write(";");
```
amckinney, line 50: Details like this might need to be left out of the abstract classes - not every language ends statements with a `;`. I know most languages do use this syntax though, so we have a couple options:

1. Leave this as-is and rely on method overrides in language-specific writers.
2. Refactor the `AbstractWriter` to be fairly similar to the original `generators/commons.Writer`.

You might have already had (1) as the desired solution, but figured it was worth calling out.

generators/commons/src/ast/AbstractWriter.ts
```diff
@@ -0,0 +1,161 @@
...
+        codeBlock.write(this);
+        this.write(";");
+        this.writeNewLineIfLastLineNot();
+    }
+
+    /**
+     * Writes text but then suffixes with a `;`
+     * @param node
+     */
+    public controlFlow(prefix: string, statement: AbstractAstNode): void {
```
amckinney, line 69: Similar to above - control flow and blocks will differ based on the language, so the decision applies here too.

generators/commons/src/AbstractGeneratorAgent.ts
```diff
@@ -5,7 +5,7 @@ import { FernGeneratorCli } from "@fern-fern/generator-cli-sdk";
 import { readFile } from "fs/promises";
 import yaml from "js-yaml";
 import path from "path";
-import { ReferenceConfigBuilder } from "./ReferenceConfigBuilder";
+import { ReferenceConfigBuilder } from "./reference";
```
amckinney, line 8: Nice 👏

PR #4595, opened 2024-09-09
fix(cli): Validate union base properties in examples

amckinney:
This fixes an issue where union base properties weren't recognized in user-specified examples, so `fern check` would yield an error and prevent generation. 

Now that `fern check` supports this use case, this also fixes an issue in the IR generator, which prevented unions with base properties to have the base properties' original type declaration resolved.

Note that this also refactors the internal packages to better support this use case. We reuse the same object property validator functions with union base properties, so each of the relevant functions accept properties and extends rather than the object declaration itself. For the union use case, `extends` is always set to `undefined` (as expected).

### DIFF
--- a/packages/cli/generation/ir-generator/src/examples/validateUnionExample.ts
+++ b/packages/cli/generation/ir-generator/src/examples/validateUnionExample.ts
@@ -9,6 +9,9 @@ import { ExampleViolation } from "./exampleViolation";
 import { getViolationsForMisshapenExample } from "./getViolationsForMisshapenExample";
 import { validateObjectExample } from "./validateObjectExample";
 import { validateTypeReferenceExample } from "./validateTypeReferenceExample";
+import { keyBy } from "lodash-es";
+import { getAllPropertiesForObject } from "../utils/getAllPropertiesForObject";
+import { validateObjectProperties } from "./validateObjectProperties";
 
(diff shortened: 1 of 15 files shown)


packages/cli/generation/ir-generator/src/examples/validateUnionExample.ts
```diff
@@ -61,11 +64,24 @@ export function validateUnionExample({
...
+            typeName,
+            unionBaseProperties,
+            nonDiscriminantPropertyExamples,
+            typeResolver,
+            exampleResolver,
+            file,
+            workspace,
+            breadcrumbs
+        });
     }
```
dsinghvi, line 85: is this diff right? `getRuleViolationForExtraProperties` -> `validateUnionBaseProperties` ?

amckinney: Yes, but I think we could make this a little more clear. The `validateUnionBaseProperties` function calls `getRuleViolationForExtraProperties` when `unionBaseProperties` is empty which preserves the intended behavior. 

We should hoist that logic out of the function though because it's otherwise a little difficult to follow. I'll update.

amckinney:
This is superseded by https://github.com/fern-api/fern/pull/5032

PR #4596, opened 2024-09-09
[DNM, Dupe] fix(cli): namespacing no longer breaks when objects conflict across namespaces

armandobelardo:
For now, this just adds a repro of the namespace break, you can see in the ETE test, `trains` imports from `neopets` (a different namespace) when the objects share the same name


packages/cli/openapi-ir-to-fern/src/buildFernDefinition.ts
```diff
@@ -46,26 +46,31 @@ export function buildFernDefinition(context: OpenApiIrConverterContext): FernDef
...
+    for (const [namespace, schemas] of Object.entries(context.ir.schemas)) {
+        for (const [id, schema] of Object.entries(schemas)) {
+            if (schemaIdsToExclude.includes(id)) {
+                continue;
+            }
 
-        const declarationFile = getDeclarationFileForSchema(schema);
-        const typeDeclaration = buildTypeDeclaration({ schema, context, declarationFile });
+            const declarationFile = getDeclarationFileForSchema(schema);
+            const typeDeclaration = buildTypeDeclaration({ schema, context, declarationFile, namespace });
```
armandobelardo, line 56: This is really only nesting the for loop into another for loop, the only actual change is passing in namespace here

packages/cli/ete-tests/src/tests/write-overrides/__snapshots__/writeOverrides.test.ts.snap
```diff
@@ -18,11 +18,7 @@ exports[`overrides > petstore 1`] = `
       x-fern-sdk-method-name: showPetById
 components:
   schemas:
-    Pet:
-      x-fern-type-name: Pet
-    Pets:
-      x-fern-type-name: Pets
-    Error:
-      x-fern-type-name: Error
+    __no_namespace__:
```
armandobelardo, line 21: I'm not positive what would cause this

packages/cli/openapi-ir-sdk/fern/definition/finalIr.yml
```diff
@@ -71,6 +77,7 @@ types:
     extends:
       - commons.WithDescription
       - commons.WithSource
+      - commons.WithNamespace
```
dsinghvi, line 80: We need to make sure that all `Webhook`, `Endpoint`, `Channel` all have `commons.WithNamespace` across both the parseIr and finalIr (i believe we are missing in a bunch of places right now)

packages/cli/openapi-parser/src/openapi/v3/generateIr.ts
```diff
@@ -286,6 +288,11 @@ export function generateIr({
 
     const groupInfo = getFernGroups({ document: openApi, context });
 
+    // TODO: It'd be better if I could reference the sentinel directly, but cannot as it's just a type
+    const NoNamespaceSentinelValue: NoNamespaceSentinel = "__no_namespace__";
```
dsinghvi, line 292: would creating a enum with a single value work here?

packages/cli/openapi-ir-sdk/fern/definition/finalIr.yml
armandobelardo: I think these are covered but let me know if you see any specific missing!

packages/cli/openapi-ir-sdk/fern/definition/finalIr.yml
```diff
@@ -33,6 +33,17 @@ types:
...
+  Schemas:
+    properties:
+      rootSchemas: map<commons.SchemaId, Schema>
+      namespacedSchemas:
+          type: map<commons.NamespaceId, map<commons.SchemaId, Schema>>
+          docs: |
+            The schemas are organized by namespace and schema name.
+            The schema name is the key in the inner map.
+            Note that this also includes un-namespaced schemas with a static sentinel key of `__no_namespace__`,
+            as even when namespacing we allow for root, unnamespaced APIs, and this allows for the most flexibility.
```
dsinghvi, line 45: 🚀

dsinghvi (approved):

PR #4611, opened 2024-09-11
fix(python): all pydantic v2 warnings are now resolved

armandobelardo:
This PR also adds in function overrides when leveraging in-house aliases to allow users to leverage `construct` and `model_construct` as normal

This PR introduces a number of changes to the `UniversalBaseModel` class in the `pydantic_utilities.py` file.

- The `json_encoders` attribute has been removed from the `Config` class and replaced with a `model_serializer` method.
- Additional model construction functions have been added to allow users to construct a model from JSON without the need for dealiasing the object themselves.
- The `model_construct` and `construct` methods have been added to the `UniversalBaseModel` class.

## Changes
- The `json_encoders` attribute has been removed from the `Config` class and replaced with a `model_serializer` method.
- Additional model construction functions have been added to allow users to construct a model from JSON without the need for dealiasing the object themselves.
- The `model_construct` and `construct` methods have been added to the `UniversalBaseModel` class.


dsinghvi (approved):
+1ing so you can merge post fixing aliases

PR #4612, opened 2024-09-11
feat(php): Generate project

amckinney:
## Overview

This sets up the `PhpProject` (which depends on the new `AbstractProject` also used by the `CsharpProject`). With this, we now have a lot of the core infra setup to:
1. Generate source files
2. Define as-is files copied directly into the project
3. Generate all the base project files (i.e. `composer.json`, `phpunit.xml`, `phpstan.neon`, etc).
4. Build, test, and lint all files in `seed` and in CI (via the generated `.github/workflows/ci.yml`).

## Misc. notes

1. A simple `Placeholder.php` file is generated for now so that we demonstrate the full directory tree.
2. The TypeScript generator context was updated to include a `version` so that it could be consolidated on the `AbstractGeneratorContext`.
3. The `namespace` configuration option was added to both the model and SDK generators since this will be inevitably be required by a variety of users.

### DIFF
--- a/generators/commons/src/project/AbstractProject.ts
+++ b/generators/commons/src/project/AbstractProject.ts
@@ -15,9 +15,7 @@ export abstract class AbstractProject<GeneratorContext extends AbstractGenerator
     }
 
     public async writeRawFiles(): Promise<void> {
-        for (const file of this.rawFiles) {
-            await file.write(this.absolutePathToOutputDirectory);
-        }
+        await Promise.all(this.rawFiles.map(async (file) => await file.write(this.absolutePathToOutputDirectory)));
     }
 
     /**
--- a/generators/php/codegen/src/project/PhpProject.ts
+++ b/generators/php/codegen/src/project/PhpProject.ts
@@ -167,9 +167,7 @@ export class PhpProject extends AbstractProject<AbstractPhpGeneratorContext<Base
         files: File[];
     }): Promise<AbsoluteFilePath> {
         await this.mkdir(absolutePathToDirectory);
-        for (const file of files) {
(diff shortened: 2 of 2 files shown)


seed/php-sdk/seed.yml
```diff
@@ -9,10 +9,12 @@ test:
...
-
 language: php
 generatorType: SDK
 defaultOutputMode: github
 fixtures: {}
-scripts: []
+scripts:
+  - docker: composer:2.7.9
+    commands:
+      - composer install && composer build && composer analyze && composer test
```
amckinney, line 19: Note that we're running _all_ of the scripts defined in the generated `composer.json` file, so we'll know whenever anything is wrong with the generated code. 

We do this in a single line to reduce the number of containers spun up.

generators/commons/src/project/AbstractProject.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public addRawFiles(file: File): void {
+        this.rawFiles.push(file);
+    }
+
+    public async writeRawFiles(): Promise<void> {
+        for (const file of this.rawFiles) {
+            await file.write(this.absolutePathToOutputDirectory);
+        }
```
dsinghvi, line 20: can we make this ```
await Promise.all(this.rawFiles.map((rawFile) => await file.write(this.absolutePathToOutputDirectory)); 
```

WIll make sure we are writing files in parallel as fast as we can

amckinney: Good call, will update!

dcb6 (approved):
🚀

PR #4618, opened 2024-09-11
feat(cli): add generator versions to the fern upgrade message

armandobelardo:
Fixes FER-3127

There are tests, but here's what it looks like 
image

This PR adds:
- generator upgrades to the traditional CLI upgrade nudge message (capped at 2 upgrades communicated)
- if there are more than 2 upgrades available then the prompt recommends using `fern generator upgrade --list` to view the full list
- this PR also adds `fern generator upgrade --list` which is the same logic as the original nudge but not truncated at all

### DIFF
--- a/packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
+++ b/packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
@@ -10,7 +10,10 @@ function hasGeneratorUpgrade(generatorUpgradeInfo: FernGeneratorUpgradeInfo[]): 
 }
 
 function hasUpgrade(upgradeInfo: FernUpgradeInfo): boolean {
-    return upgradeInfo.cliUpgradeInfo?.isUpgradeAvailable || hasGeneratorUpgrade(upgradeInfo.generatorUpgradeInfo);
+    return (
+        (upgradeInfo.cliUpgradeInfo?.isUpgradeAvailable ?? false) ||
+        hasGeneratorUpgrade(upgradeInfo.generatorUpgradeInfo)
+    );
 }
 
 export async function getFernUpgradeMessage({
--- a/packages/cli/ete-tests/src/tests/upgrade-generator/upgrade-generator.test.ts
+++ b/packages/cli/ete-tests/src/tests/upgrade-generator/upgrade-generator.test.ts
@@ -207,7 +207,7 @@ describe("fern generator upgrade", () => {
         ).toMatchSnapshot();
     }, 60_000);
 
-    it("fern generator upgrade message", async () => {
+    it.skip("fern generator upgrade message", async () => {
         const tmpDir = await tmp.dir();
         const directory = AbsoluteFilePath.of(tmpDir.path);
 


packages/cli/cli/src/cli-context/upgrade-utils/getFernCliUpgradeMessage.ts
armandobelardo, line 1: This file was just renamed to `getFernUpgradeMessage.ts`, bad commit here

packages/cli/ete-tests/src/tests/upgrade-generator/__snapshots__/upgrade-generator.test.ts.snap
```diff
@@ -32,5 +32,34 @@ Options:
...
+         │                                                            │
+         │                     Upgrades available                     │
+         │                                                            │
+         │                                                            │
+         │           Java SDK (Group: local) 0.0.1 → 0.10.1           │
+         │       Python SDK (Group: python-sdk) 3.0.0 → 3.10.8        │
+         │                                                            │
+         │   Run fern generator upgrade to upgrade your generators.   │
+         │                                                            │
+         ╰────────────────────────────────────────────────────────────╯
```
dsinghvi, line 63: sick!!

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
```diff
@@ -0,0 +1,162 @@
...
+            return "OpenAPI";
+
+        default: {
+            if (generatorName.startsWith("fern-")) {
+                return generatorName.replace("fern-", "");
+            }
+            return generatorName;
+        }
+    }
+}
```
dsinghvi, line 162: I know you recently added display name so would be nice to delete this and leverage that here

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
```diff
@@ -0,0 +1,162 @@
...
+    let message = header ?? "";
+    let generatorUpgradeNeeded = false;
+
+    let generatorsNeedingUpgrades = generatorUpgradeInfo.filter((gui) => gui.isUpgradeAvailable);
+    if (limit != null) {
+        generatorsNeedingUpgrades = generatorsNeedingUpgrades.slice(0, limit + 1);
+    }
+    for (const generatorUpgrade of generatorsNeedingUpgrades) {
+        // Format => Generator 1.0.0 → 1.1.0 (API: API Name, Group: Group Name)
+        // ex: "Python SDK 1.0.0 → 1.1.0 (API: myApi, Group: myGroup)"
```
dsinghvi, line 78: thoughts on excluding api if only one is present?

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
```diff
@@ -0,0 +1,162 @@
+import boxen from "boxen";
+import chalk from "chalk";
+import { FernUpgradeInfo } from "../CliContext";
+import { CliEnvironment } from "../CliEnvironment";
+import { FernGeneratorUpgradeInfo } from "./getGeneratorVersions";
+
+export function getFernUpgradeMessage({
```
amckinney, line 7: nit: Not functionally any different, but more of a code style suggestion -

It looks like we do a fair amount of work and compose a `message` that isn't always used (i.e. we return `undefined` at the end and the `message` is dropped altogether).

We could rearrange this a bit so that we first detect whether or not we need to do anything, and return early if so. That way we can drop the `upgradeAvailable` variable entirely.

```typescript
if (!hasUpgrade(upgradeInfo)) {
  return undefined;
}
...

function hasUpgrade(upgradeInfo: FernUpgradeInfo): boolean {
  return upgradeInfo.cliUpgradeInfo?.isUpgradeAvailable || (upgradeInfo.generatorUpgradeInfo.filter((gui) => gui.isUpgradeAvailable)).length > 0
}
```

---

Alternatively, if we find ourselves repeating the same functions to determine whether or not there is an upgrade for CLI vs. generators, we could make `FernUpgradeInfo` a class and give it a method like:

```typescript
if (!upgradeInfo.hasUpgrade()) {
  ...
}
if (upgradeInfo.hasGeneratorUpgrade()) {
  ...
}
...
```

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
armandobelardo: I think it is in theory! See the snapshotted test which doesn't have API name, which is omitted if undefined, which I think happens if there's only one API

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
armandobelardo: Yup can do!

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
```diff
@@ -0,0 +1,162 @@
...
+    limit,
+    includeBoxen = true
+}: {
+    generatorUpgradeInfo: FernGeneratorUpgradeInfo[];
+    header?: string;
+    limit?: number;
+    includeBoxen?: boolean;
+}): string | undefined {
+    let message = header ?? "";
+    let generatorUpgradeNeeded = false;
```
amckinney, line 70: Similar thought with the suggestion above - any time I see a `let ...` for some boolean sentinel value, I usually stop and wonder how we can refactor it.

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
```diff
@@ -0,0 +1,162 @@
...
+                  textAlignment: "center",
+                  borderColor: "yellow",
+                  borderStyle: "round"
+              })
+            : message;
+    }
+    return;
+}
+
+function normalizeGeneratorName(generatorName: string): string {
```
amckinney, line 106: Not actionable here, but it'd be great if we could define this generate this from somewhere as an enum so that we know exactly when we need to update it (e.g. when PHP is available, it'd be great if this was a compile error).

packages/cli/cli/src/cli-context/upgrade-utils/getGeneratorVersions.ts
```diff
@@ -0,0 +1,243 @@
...
+                        previousVersion: generator.version,
+                        latestVersion
+                    };
+                }
+            },
+            generatorFilter,
+            groupFilter
+        });
+        return versions;
+    } else {
```
amckinney, line 85: We return in the `if`, so we can drop the `else`.

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
```diff
@@ -0,0 +1,162 @@
...
+    includeBoxen?: boolean;
+}): string | undefined {
+    let message = header ?? "";
+    let generatorUpgradeNeeded = false;
+
+    let generatorsNeedingUpgrades = generatorUpgradeInfo.filter((gui) => gui.isUpgradeAvailable);
+    if (limit != null) {
+        generatorsNeedingUpgrades = generatorsNeedingUpgrades.slice(0, limit + 1);
+    }
+    for (const generatorUpgrade of generatorsNeedingUpgrades) {
```
amckinney, line 76: Based on the async implementation to get the latest generator versions, I think this list order will be non-deterministic. Not the end of the world since there's only a few entries, but it might be worth sorting the entries so that users always see the same result.

But if the test you added isn't flaky, then I could be wrong.

amckinney (approved):
Only suggestions - output looks great and as long as it's stable for tests then this looks good to me! Feel free to take the feedback or leave it.

packages/cli/cli/src/cli-context/upgrade-utils/getFernUpgradeMessage.ts
armandobelardo: I added an API to do this (you can update display name in the generators' seed.yml) so will update to that!

PR #4625, opened 2024-09-12
feat(php): Implement RawClient.php

amckinney:
This adds a variety of the `Core` package utilities, all of which are required for the `RawClient` type, which is the primary mechanism used to make HTTP requests.

Several things to note:
1. The `BaseApiRequest` will be used by every flavor of HTTP endpoint call (i.e. JSON, multipart form, streaming, etc).
    * For now, we only implement JSON.
2. The `RawClientTest.php` file is a set of simple tests to confirm the `RawClient` acts as expected.
    * We can add more tests over time, but this at least sets the stage.
3. The `RawClient` currently uses the synchronous method call.
    * This will either be refactored to be exclusively async, or an additional method will be rolled out to support those use cases.
    * It's unclear if we will need to support both options, or simply just `async`.
4. Tidies up a few things w.r.t. `namespace` from the original `PhpProject` PR [here](https://github.com/fern-api/fern/pull/4612).


seed/php-sdk/imdb/src/ClientOptions.php
```diff
@@ -0,0 +1,45 @@
...
+     * @var string
+     */
+    public string $baseUrl = '';
+
+    /**
+     * The HTTP client used to make requests.
+     *
+     * @var Client
+     */
+    public Client $httpClient;
```
dcb6, line 21: I think we can have this be a `Psr\Http\Client\ClientInterface` which lets the user pass in any [PSR-18](https://www.php-fig.org/psr/psr-18/) compliant client! Guzzle's client implements this so I think that's the only change needed

seed/php-sdk/imdb/src/Core/RawClient.php
```diff
@@ -0,0 +1,142 @@
...
+     * The client options used to make requests.
+     *
+     * * @param ClientOptions|null $clientOptions
+     */
+    public function __construct(
+        ?ClientOptions $clientOptions,
+    )
+    {
+        $this->clientOptions = $clientOptions ?? new ClientOptions();
+    }
```
dcb6, line 32: nit: I believe that this is equivalent
```suggestion
    /**
     * @param ClientOptions $clientOptions The client options used to make requests.
     */
    public function __construct(
        private readonly ClientOptions $clientOptions = new ClientOptions(),
    )
    {
    }
```

not important right now, but will comment on these things so we can build our php chops together!

seed/php-sdk/imdb/src/Core/RawClient.php
```diff
@@ -0,0 +1,142 @@
...
+     */
+    public function sendRequest(
+        BaseApiRequest $request,
+    ): BaseApiResponse
+    {
+        $httpRequest = $this->buildRequest($request);
+        $httpResponse = $this->clientOptions->httpClient->send($httpRequest);
+        return new BaseApiResponse(
+            $httpResponse->getStatusCode(),
+            $httpResponse,
```
dcb6, line 47: what's the purpose of pulling out the status code here if they can get it from the `httpResponse` itself?

seed/php-sdk/imdb/src/ClientOptions.php
amckinney: Great call, that's definitely what we should do 👍

seed/php-sdk/imdb/src/Core/RawClient.php
amckinney: Primarily for convenience + consistency with the C# generator ([link](https://github.com/fern-api/fern/blob/d6ccbc91c2a7ffea9a1deba62912ee5175c17538/seed/csharp-sdk/imdb/no-custom-config/src/SeedApi/Core/RawClient.cs#L72)).

Given that that this is entirely internal I'm also fine with just returning the underlying `ResponseInterface` for now. Might as well keep things as simple as we can until there's a more compelling reason for it

seed/php-sdk/imdb/src/Core/HttpMethod.php
```diff
@@ -0,0 +1,12 @@
...
+namespace Seed\Core;
+
+enum HttpMethod: string
+{
+    case GET = 'GET';
+    case POST = 'POST';
+    case PUT = 'PUT';
+    case PATCH = 'PATCH';
+    case DELETE = 'DELETE';
+}
```
dcb6, line 12: any particular reason to make this string-backed rather than just the following?
```suggestion
enum HttpMethod
{
    case GET;
    case POST;
    case PUT;
    case PATCH;
    case DELETE;
}
```

seed/php-sdk/imdb/src/Core/RawClient.php
amckinney: Nice, just verified this works well - let's go with this!

seed/php-sdk/imdb/src/Core/HttpMethod.php
amckinney: Primarily to have a simpler usage at the `buildRequest` call-site:

```php
    private function buildRequest(
        BaseApiRequest $request
    ): Request {
       $url = $this->buildUrl($request);
       $headers = $this->encodeHeaders($request);
       $body = $this->encodeRequestBody($request);
       return new Request(
           $request->method->value,
           $url,
           $headers,
           $body,
       );
    }
```

When the enum is string-backed, we can just use `$request->method->value`. Thoughts?

seed/php-sdk/imdb/src/Core/RawClient.php
```diff
@@ -0,0 +1,142 @@
...
+     */
+    public function __construct(
+        ?ClientOptions $clientOptions,
+    )
+    {
+        $this->clientOptions = $clientOptions ?? new ClientOptions();
+    }
+
+    /**
+     * The client options used to make requests.
```
dcb6, line 35: nit: comment needs updating or removing

seed/php-sdk/imdb/src/Core/RawClient.php
```diff
@@ -0,0 +1,142 @@
...
+       $body = $this->encodeRequestBody($request);
+       return new Request(
+           $request->method->value,
+           $url,
+           $headers,
+           $body,
+       );
+    }
+    /**
+     * The client options used to make requests.
```
dcb6, line 65: same nit: comment needs updating or removing

seed/php-sdk/imdb/src/Core/RawClient.php
amckinney: Sweet, will drop `BaseApiResponse` in favor of `ResponseInterface` 👍

seed/php-sdk/imdb/src/Core/HttpMethod.php
dcb6: i think if it's not string-backed, you can do:
```php
           $request->method->name,
```

amckinney: Oh nice, just tried that and it works well 👍

dcb6 (approved):

PR #4635, opened 2024-09-13
feat(php): Support writing comments (w/ type hints)

amckinney:
This builds upon the changes introduced in https://github.com/fern-api/fern/pull/4634 and adds support for writing comments on arbitrary ASTNodes.

The idea here is simple - a separate `Comment` `ASTNode` sets the `comment` option in the `Writer`, which controls how the `Type` should be formatted (which is relevant for `array` and `map` types).

In the generator implementation, we continue to use the same `php.Type` constructors and don't need to think about when to choose comment rendering or not - it's all handled within the `ASTNode` that is writing it.

### DIFF
--- a/generators/php/codegen/src/ast/Comment.ts
+++ b/generators/php/codegen/src/ast/Comment.ts
@@ -58,7 +58,7 @@ export class Comment extends AstNode {
 
     private writeTag({ writer, tag }: { writer: Writer; tag: Comment.Tag }): void {
         writer.write(` * @${tag.tagType} `);
-        tag.type.write(writer, undefined, true); // Write the tag type as a comment.
+        tag.type.write(writer, { comment: true });
         if (tag.name != null) {
             writer.write(` ${tag.name}`);
         }
--- a/generators/php/codegen/src/ast/Type.ts
+++ b/generators/php/codegen/src/ast/Type.ts
@@ -50,7 +50,7 @@ export class Type extends AstNode {
         super();
     }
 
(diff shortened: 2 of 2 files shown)


dsinghvi (approved):
Nice idea with the `Comment` node 🚀

generators/php/codegen/src/ast/Comment.ts
```diff
@@ -0,0 +1,70 @@
...
+        }
+        for (const tag of this.tags) {
+            this.writeTag({ writer, tag });
+        }
+        writer.writeLine(" */");
+    }
+
+    private writeTag({ writer, tag }: { writer: Writer; tag: Comment.Tag }): void {
+        writer.write(` * @${tag.tagType} `);
+        tag.type.write(writer, undefined, true); // Write the tag type as a comment.
```
dsinghvi, line 61: this would be easier to read if we used dictionaries `tag.type.write(writer, { comment: true })`

amckinney: Nice! I actually thought about this for a bit, and ended up sticking with it as is because I didn't want to sacrifice the positional writer parameter. Splitting out the optional into a separate dictionary is exactly what we should do 👍

dsinghvi: The signature would be something like 

```
public void write(writer: Writer, opts: { comment: boolean } = { comment: false }); 
```

Issue #4639, opened 2024-09-14
[Bug] CLI crashes with `RangeError: Invalid string length` when generating SDKs from large specs


mstade:
### Describe the bug

When trying to generate an SDK from a very large spec definition, the generator crashes with a `RangeError: Invalid string length` message when using `--log-level=debug`.

### To reproduce

1. Create a very large spec definition using thousands of models
2. Generate an SDK of any language
3. Verify that it fails with `RangeError: Invalid string length` after the step `Migrated IR` as reported by `--log-level=debug`

If you're not able to make it fail, your spec isn't large enough. I'm unfortunately not able to share our spec, due to confidentiality. I may be able to present you with the issue in a screen share, but will have to check with management.

### Expected behavior

I expect the generation to work regardless of size of the spec.

### Screenshots

n/a

### CLI Version

0.41.12

### Additional context

I believe this is caused by size limitations set by the node runtime, more context is available here: https://github.com/nodejs/node/issues/35973

The error is not very helpful, as this is really about not having enough memory to process such a large amount of data in one go. From reverse engineering the minified bundle, I believe the root cause of the error is here: https://github.com/fern-api/fern/blob/27fb37831c91a466bb75e952b8c3a886f2efeb64/packages/cli/generation/local-generation/local-workspace-runner/src/runGenerator.ts#L160

I was able to get past this step by implementing a dirty workaround in the minified bundle, by using `Object.entries` to stringify each individual top-level key's value, and then constructing a JSON string from that. Essentially I did this:

```js
const stringified = `{ ${Object.entries(e).map(([key, value]) => `"${key}": ${JSON.stringify(value)}`).join(', ')} }`
await writeFile(absolutePathToIr, stringified); 
```

This does seem to have worked, but I fear it's a band-aid at best. At some point one of those root level key's value may turn out to be too big, causing the error again. A more solid fix would be to use something like a streaming JSON writer, if there is such a thing.

dsinghvi:
Should be fixed in the latest version of the CLI, thanks @mstade for the contribution

PR #4640, opened 2024-09-14
fix(cli): workaround for crash caused by huge specs

mstade:
This is a fix for the bug described in https://github.com/fern-api/fern/issues/4639, where if the spec to generate the SDK from is large enough it will cause the Fern CLI to crash due to insufficient memory when writing the IR JSON.

More context available in https://github.com/fern-api/fern/issues/4639.

This PR introduces a fix for a crash that occurs when generating local SDKs from large spec definitions. It also includes a change to the `runGenerator.ts` file, replacing the `writeFile` function with the `streamObjectToFile` function for writing IR data to a file.

## Changes
- Fixes a crash when generating local SDKs from large spec definitions.
- Replaces `writeFile` with `streamObjectToFile` in `runGenerator.ts` for writing IR data to a file.

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,6 +1,7 @@
 - changelogEntry:
   - summary: |
-      Fixes a crash when generating local SDKs from huge spec definitions. See issue #4639 for more detail.
+      `fern generate --local` no longer crashes on large API Definitions because we 
+      stream the JSON to file instead of calling `JSON.stringify`. See [PR 4640](https://github.com/fern-api/fern/pull/4640). 
     type: fix
   irVersion: 53
   version: 0.41.13


dsinghvi:
@mstade would recommend leveraging https://github.com/fern-api/fern/blob/main/packages/commons/fs-utils/src/streamObjectToFile.ts

dsinghvi:
Additionally if you can add a changelog entry to cli/cli/versions.yml as well as a test

dsinghvi (changes requested):
Excited to get this fix in

mstade:
Thank you @dsinghvi for the swift review! I have made the requested changes now. For the changelog entry I didn't really know what to put there so I just winged it. I hope this helps!

mstade:
FWIW I don't know for a fact that this change actually fixes our issue, I'm not sure how to test this with our spec locally. I just assume the function you pointed me to does what it says on the tin.

dsinghvi:
@mstade you should be able to use https://github.com/fern-api/fern/blob/main/CONTRIBUTING.md#building-the-cli-from-source to test locally. 

TLDR:  Run 

```
yarn fern:build
cd /your/fern/folder
node /path/to/clic.js generate
```

dsinghvi (approved):

mstade:
> @mstade you should be able to use https://github.com/fern-api/fern/blob/main/CONTRIBUTING.md#building-the-cli-from-source to test locally.

Ah I must've missed that section. I tried your tip and it does seem to work with our spec and indeed fix #4639! 🥳

dsinghvi:
> > @mstade you should be able to use [`main`/CONTRIBUTING.md#building-the-cli-from-source](https://github.com/fern-api/fern/blob/main/CONTRIBUTING.md?rgh-link-date=2024-09-14T13%3A15%3A05Z#building-the-cli-from-source) to test locally.
> 
> Ah I must've missed that section. I tried your tip and it does seem to work with our spec and indeed fix #4639! Thank you for the assistance @dsinghvi! 🥳

Awesome, going to merge and you'll see a new release in ~10 minutes

mstade:
Alas, while this does indeed fix #4639 I jumped the gun a bit to proclaim victory. There is a similar issue now at the end of the process. The generation errors out saying there's a non-zero exit code from Docker:

```
Docker exited with a non-zero exit code. Logs here: /var/folders/hj/t83zsbqx74dgj05hyds2bzxh0000gn/T/tmp-77020-270QKtGd4J1H
```

The logs at that path are a bit more descriptive:
```
❯ less /var/folders/hj/t83zsbqx74dgj05hyds2bzxh0000gn/T/tmp-77020-270QKtGd4J1H

Read /fern/config.json
Sent error event to coordinator
node:buffer:786
    return this.utf8Slice(0, this.length);
                ^

Error: Cannot create a string longer than 0x1fffffe8 characters
ESC[90m    at Buffer.toString (node:buffer:786:17)ESC[39m
    at loadIntermediateRepresentation (/nodeCli.cjs:287246:71)
    at async SdkGeneratorCli.run (/nodeCli.cjs:287433:37)
    at async SdkGeneratorCli.runCli (/nodeCli.cjs:287400:5) {
  code: ESC[32m'ERR_STRING_TOO_LONG'ESC[39m
}

Node.js v18.1.0
/var/folders/hj/t83zsbqx74dgj05hyds2bzxh0000gn/T/tmp-77020-270QKtGd4J1H (END)
```

It looks like it's essentially the same root cause, but in a different place this time. I'll see if I can figure it out.

mstade:
Sleuthing around I think the issue is likely to be this:
https://github.com/fern-api/fern/blob/acb1e5880c23be26d815142115fb77c44a330760/generators/typescript/utils/abstract-generator-cli/src/loadIntermediateRepresentation.ts#L7

If it's too big to write, it's likely going to be too big to read as well. I tried looking around but couldn't find an inverse to `streamObjectToFile` in the utils – is there something like that in there @dsinghvi, or is this turning into a bit of a rabbit hole perhaps? 🤔

dsinghvi:
@mstade i bet there is some way to do this, also looking around

mstade:
@dsinghvi I believe I may have found a solution using the [stream-json](https://github.com/uhop/stream-json) package. I implemented a utility like so:

```js
import { createReadStream } from "fs";
import { AbsoluteFilePath } from "./AbsoluteFilePath";
import StreamObject from 'stream-json/streamers/StreamObject.js';

export async function streamObjectFromFile(filePath: AbsoluteFilePath): Promise<unknown> {
    const jsonStream = StreamObject.withParser();
    const accumulator = {};

    jsonStream.on('data', ({ key, value }: { key: string, value: unknown }) => 
        Object.assign(accumulator, { [key]: value })
    );

    return new Promise((resolve, reject) => {
        createReadStream(filePath)
            .pipe(jsonStream)
            .on("error", reject)
            .on("finish", () => {
                resolve(accumulator)
            });
    })
}
```

However, I'm not sure how to test this locally, since this is not in the CLI code anymore but in the generators. I've tried following the instructions in CONTRIBUTING.md but I don't think I'm doing it right, and not sure exactly where I'm going wrong either.

Before I dive deeper into this rabbit hole, do you think adding a utility like the above is the right approach?

dsinghvi:
@mstade this is sick, happy to help push this over the line when i am back online tonight

mstade:
Much obliged @dsinghvi! I'll push a draft PR with the code I have so far, and move the conversation there. Thank you for the help!

dsinghvi:
@mstade this is fixed in the latest TypeScript generator after this PR https://github.com/fern-api/fern/pull/4647

PR #4642, opened 2024-09-14
fix(cli): parse IR json with streaming parser

mstade:
PR #4640 fixed the issue reported in #4639, but sadly this was only part of the solution as the IR JSON is also read and parsed by the generators, causing similar issues as those reported in #4639.

For more context, see:
https://github.com/fern-api/fern/pull/4640#issuecomment-2350992867.

This PR introduces a new function `streamObjectFromFile` to the `fs-utils` package. The function reads a JSON file and returns a Promise that resolves to the parsed JSON object.

## Changes
- Added a new function `streamObjectFromFile` to the `fs-utils` package.
- The function reads a JSON file and returns a Promise that resolves to the parsed JSON object.
- The function uses the `createReadStream` function from the `fs` module to create a read stream for the file.
- The `StreamObject` class from the `stream-json` module is used to parse the JSON data.
- The `stream-json` package is added to the dependencies of the `fs-utils` package.
- The `@types/stream-json` package is added to the devDependencies of the `fs-utils` package.
- The `stream-json` package is added to the dependencies of the `cli` package.
- The `@types/stream-json` package is added to the dependencies of the `cli` package.

### DIFF
--- a/generators/typescript/utils/abstract-generator-cli/src/loadIntermediateRepresentation.ts
+++ b/generators/typescript/utils/abstract-generator-cli/src/loadIntermediateRepresentation.ts
@@ -1,10 +1,9 @@
 import { IntermediateRepresentation } from "@fern-fern/ir-sdk/api";
 import * as IrSerialization from "@fern-fern/ir-sdk/serialization";
-import { readFile } from "fs/promises";
+import { AbsoluteFilePath, streamObjectFromFile } from "@fern-api/fs-utils";
 
 export async function loadIntermediateRepresentation(pathToFile: string): Promise<IntermediateRepresentation> {
-    const irString = (await readFile(pathToFile)).toString();
-    const irJson = JSON.parse(irString);
+    const irJson = await streamObjectFromFile(AbsoluteFilePath.of(pathToFile));
     return IrSerialization.IntermediateRepresentation.parseOrThrow(irJson, {
         unrecognizedObjectKeys: "passthrough",
         allowUnrecognizedEnumValues: true,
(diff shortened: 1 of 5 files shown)


generators/typescript/utils/abstract-generator-cli/src/loadIntermediateRepresentation.ts
```diff
@@ -1,10 +1,9 @@
 import { IntermediateRepresentation } from "@fern-fern/ir-sdk/api";
 import * as IrSerialization from "@fern-fern/ir-sdk/serialization";
-import { readFile } from "fs/promises";
+import { AbsoluteFilePath, streamObjectFromFile } from "@fern-api/fs-utils";
 
 export async function loadIntermediateRepresentation(pathToFile: string): Promise<IntermediateRepresentation> {
-    const irString = (await readFile(pathToFile)).toString();
-    const irJson = JSON.parse(irString);
+    const irJson = await streamObjectFromFile(AbsoluteFilePath.of(pathToFile));
```
mstade, line 6: I've not been able to test this locally. I did try following the instructions in `CONTRIBUTING.md`, however the process output a lot of errors and the docker container does not seem to do anything but open a shell. I'm obviously doing something wrong, but it's not clear to me what.

packages/commons/fs-utils/src/streamObjectFromFile.ts
mstade, line 1: I've only been able to test this locally by importing the library in a node repl, and then using this function to load a json file. It does seem to do what it says on the tin, but I'd like to somehow test this end-to-end with our giant spec definition.

mstade (commented):
## Author Notes

generators/typescript/utils/abstract-generator-cli/src/loadIntermediateRepresentation.ts
mstade: I believe this issue is common to all generators, so other `loadIntermediateRepresentation` functions should probably be updated the same way.

dsinghvi:
@mstade looking into this now!

dsinghvi:
@mstade closing your PR in favor of https://github.com/fern-api/fern/pull/4647

PR #4645, opened 2024-09-15
feat(php): types + enums

dcb6:
## Author Notes
* This PR introduces type generation, enums, and serde for PHP

## Generated

This PR introduces a new feature to the codebase: the ability to generate PHP models from IR types. It also adds support for generating PHP models from IR types that are enums.

## Changes
- **New files:** `Array.ts, `Attribute.ts, `Enum.ts, `EnumGenerator.ts, `ObjectGenerator.ts`
- **Modified files:** `SerializableType.Template.php, `Class.ts, `Field.ts, `Map.ts, `Parameter.ts, `Type.ts, `AbstractPhpGeneratorContext.ts, `PhpTypeMapper.ts, `php.ts, `PhpFile.ts, `generateModels.ts`
- **Deleted files:** `Placeholder.php`

## New files
- `Array.ts: This file defines the `Array` class, which extends `AstNode`. It has a constructor that takes an `entries` argument, which is an array of `AstNode` objects. The `write` method writes the array to a `Writer` object.
- `Attribute.ts: This file defines the `Attribute` class, which extends `AstNode`. It has a constructor that takes a `reference` argument, which is a `ClassReference` object, and an `arguments` argument, which is an array of strings or `AstNode` objects. The `write` method writes the attribute to a `Writer` object.
- `Enum.ts: This file defines the `Enum` class, which extends `AstNode`. It has a constructor that takes a `name`, `namespace`, `backing`, and `docs` argument. The `write` method writes the enum to a `Writer` object.
- `EnumGenerator.ts: This file defines the `EnumGenerator` class, which extends `FileGenerator`. It has a constructor that takes a `context`, `typeDeclaration`, and `enumDeclaration` argument. The `doGenerate` method generates a `PhpFile` object. The `getFilepath` method returns a `RelativeFilePath` object.
- `ObjectGenerator.ts: This file defines the `ObjectGenerator` class, which extends `FileGenerator`. It has a constructor that takes a `context`, `typeDeclaration`, and `objectDeclaration` argument. The `doGenerate` method generates a `PhpFile` object. The `getFilepath` method returns a `RelativeFilePath` object.

## Modified files
- `SerializableType.Template.php: This file has been modified to add a check for a null `$jsonKey` in the `toJson` and `fromArray` methods. It also adds a check for a `mixed` type in the `serializeSingleValue` and `deserializeSingleValue` methods.
- `Class.ts: This file has been modified to add a `parentClassReference` argument to the `Class` constructor. It also adds a check for a non-null `parentClassReference` in the `write` method.
- `Field.ts: This file has been modified to add an `attributes` argument to the `Field` constructor. It also adds a `writeAttributesIfPresent` method, which writes the attributes to a `Writer` object if they are present.
- `Map.ts: This file has been modified to add a check for a null `$jsonKey` in the `write` method. It also changes the way the entries are iterated over.
- `Parameter.ts: This file has been modified to add a check for a non-null `type.isOptional` in the `write` method.
- `Type.ts: This file has been modified to remove the `parentType` argument from the `write` method. It also adds a check for a non-null `comment` in the `write` method.
- `AbstractPhpGeneratorContext.ts: This file has been modified to add a `PrimitiveType` import. It also adds several new methods for getting class references: `getDateTypeAttributeClassReference`, `getConstantClassReference`, `getJsonPropertyAttributeClassReference`, `getSerializableTypeClassReference`, `getUnionClassReference`, and `getArrayTypeClassReference`.
- `PhpTypeMapper.ts: This file has been modified to change the return type of the `_other` function from `php.Type.object` to `php.Type.mixed`.
- `php.ts: This file has been modified to add `Array` and `Enum` imports. It also adds several new functions: `attribute`, `array`, `enum_`, `Attribute`, `Array`, and `Enum`.
- `PhpFile.ts: This file has been modified to add an `Enum` import. It also adds an `Enum` type to the `clazz` argument of the `phpFileContent` function.
- `generateModels.ts: This file has been modified to remove the `Placeholder` class and replace it with a loop that iterates over the `context.ir.types` object. It also adds a new `ObjectGenerator` class and a new <co: 16>`EnumGenerator


seed/php-sdk/imdb/src/Types/WeatherReport.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Seed\Types;
+
+enum WeatherReport: string {
```
dcb6, line 5: I'm back to having this an actual `enum` (rather than string consts) 🙃 

The IDE complains either way if I mark the type as this in the PHPDoc. I think users can figure out to do `.value` if they use the enum, and enum allows easier `.match` authorimg

seed/php-sdk/imdb/src/Types/ExampleType.php
```diff
@@ -0,0 +1,67 @@
...
+use Seed\Core\ArrayType;
+use Seed\Core\DateType;
+use Seed\Core\JsonProperty;
+use Seed\Core\SerializableType;
+use Seed\Core\Union;
+
+class ExampleType extends SerializableType
+{
+    /**
+     * @param string[] $stringList
```
dcb6, line 15: IDE only complains about arrays not having PHPDocs since those are the only ones that aren't strongly typed directly in the declaration

So, I've only included these here to fix warnings, and skipped adding the non-array types.

However, I have no strong feelings here, and could include all params

amckinney: Sounds good - I actually implemented the comment rendering to include the typed equivalent of `array`. Right now we're representing single dimensional arrays as `array<string>` instead of `string[]`, but we can easily change that [here](https://github.com/fern-api/fern/blob/7f623dd4d5699493e0171c0c019d970eec36e93b/generators/php/codegen/src/ast/Type.ts#L78).

amckinney: The implementation will create all params by default (for consistency and simplicity), but don't feel the need to do it in your hand-written equivalent. The comments will show up once we actually generate things.

seed/php-sdk/imdb/src/Types/WeatherReport.php
amckinney: Should we do the same thing for environments, or should we special case that? I was originally thinking we'd have something like:

```php
class Environments {
  public const PRODUCTION = "https://acme.co";
  public const STAGING = "https://acme-staging.co";
}
```

Happy to model this as an enum if we want though, especially because most people won't even interact with it - the default environment will normally be set for them:

```php
enum Environments: string {
  case PRODUCTION = "https://acme.co";
  case STAGING = "https://acme-staging.co";
}
```

seed/php-sdk/imdb/tests/Seed/Core/RawClientTest.php
```diff
@@ -1,17 +1,13 @@
 <?php
 
-namespace Seed\Tests\Core;
```
amckinney, line 3: I assume this was built off an earlier branch? We'll want to revert the changes in this file once we actually generate things.

seed/php-sdk/imdb/tests/Seed/Types/ExampleTypeTest.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+namespace Seed\Types;
```
amckinney, line 3: Heads up, we'll want to use `Seed\Tests\...` in the `tests/` directory here (e.g. `Seed\Tests\Types`). Nothing actionable here, just FYI once you revisit this in the generator. 

This will naturally be fixed for you once you use the [getTestsNamespace](https://github.com/fern-api/fern/blob/7f623dd4d5699493e0171c0c019d970eec36e93b/generators/php/codegen/src/context/AbstractPhpGeneratorContext.ts#L31) function.

seed/php-sdk/imdb/src/Core/SerializableType.php
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    /**
+     * Serializes a single value based on its type.
+     *
+     * @param mixed $data The value to serialize.
+     * @param string $type The expected type.
+     * @return mixed The serialized value.
+     */
+    private static function serializeSingleValue($data, $type)
```
amckinney, line 111: nit: Should we use type declarations in the signature, too?
```suggestion
    private static function serializeSingleValue(mixed $data, string $type): mixed
```

seed/php-sdk/imdb/src/Core/SerializableType.php
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    /**
+     * Deserializes a single value based on its type.
+     *
+     * @param mixed $data The data to deserialize.
+     * @param string $type The expected type.
+     * @return mixed The deserialized value.
+     */
+    private static function deserializeSingleValue(mixed $data, string $type)
```
amckinney, line 328:
```suggestion
    private static function deserializeSingleValue(mixed $data, string $type): mixed
```

seed/php-sdk/imdb/src/Core/SerializableType.php
```diff
@@ -0,0 +1,386 @@
...
+            $jsonKey = self::getJsonKey($property);
+            $value = $property->getValue($this);
+
+            // Handle DateTime properties
+            $dateTypeAttr = $property->getAttributes(DateType::class)[0] ?? null;
+            if ($dateTypeAttr && $value instanceof DateTime) {
+                $dateType = $dateTypeAttr->newInstance()->type;
+                $value = ($dateType === DateType::TYPE_DATE)
+                    ? $value->format('Y-m-d')
+                    : $value->format(DateTime::RFC3339);
```
amckinney, line 49: Given that we reference `Y-m-d` and `DateTime::RFC3339` in so many places, it might be worth adding a separate `Core\Constant.php` file:

```php
class Constant {
  public const DateFormat = 'Y-m-d';
  public const DateTimeFormat = DateTime::RFC3339;
}
```

Alternatively, could create a format class if we imagine other formats:

```php
class Format {
  public const Date = 'Y-m-d';
  public const DateTime = DateTime::RFC3339;
}
```

Up to you though - if this feels overkill, I'm down to leave this as-is.

seed/php-sdk/imdb/src/Core/SerializableType.php
```diff
@@ -0,0 +1,386 @@
...
+        if ($type instanceof Union) {
+            foreach ($type->types as $unionType) {
+                try {
+                    return self::serializeSingleValue($data, $unionType);
+                } catch (Exception $e) {
+                    continue;
+                }
+            }
+            throw new \InvalidArgumentException("Cannot serialize value with any of the union types.");
+        }
```
amckinney, line 99: Does this guarantee that the correct discriminant is set, even if multiple union types have the same underlying form? For example, what if I had a union like:

```yaml
types:
  Name: string
  Union:
     s: string
     name: Name
```

If I specify a `Name`, I want the JSON representation to be:

```json
{
  "type": "name",
  "name": "..."
}
```

But if `Name` is recognizable as a `string`, then would `string` succeed first?

---

Normally we use the discriminant to determine how we should serialize and only attempt that. Maybe that's what is happening here already?

seed/php-sdk/imdb/src/Core/SerializableType.php
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    /**
+     * Serializes a value based on the type definition.
+     *
+     * @param mixed $data The value to serialize.
+     * @param mixed $type The type definition.
+     * @return mixed The serialized value.
+     */
+    private static function serializeValue($data, $type)
```
amckinney, line 88:
```suggestion
    private static function serializeValue(mixed $data, mixed $type): mixed
```

seed/php-sdk/imdb/src/Core/SerializableType.php
amckinney: Also note if this ends up not being the case, we can always come back to unions later - most important thing to focus on in the first pass is objects and enums.

seed/php-sdk/imdb/src/Types/WeatherReport.php
dcb6: No strong feelings! enum seems reasonable to me with the first class support for string-backed, but as you said, not much user interaction!

seed/php-sdk/imdb/tests/Seed/Core/RawClientTest.php
dcb6: PHPstorm suggested i change it as i did, interestingly

that being said, i didn't really mean to check that in, so i'll change it back in this PR! but something to talk about for our final state

seed/php-sdk/imdb/src/Core/SerializableType.php
dcb6: Good call, will do!

seed/php-sdk/imdb/src/Core/SerializableType.php
dcb6: This code will only be reusable for undiscriminated unions, similar to C#

We'll have to figure out how we wanna handle discriminated unions, maybe something similar to Java's

generators/php/codegen/src/context/AbstractPhpGeneratorContext.ts
```diff
@@ -148,11 +181,11 @@ export abstract class AbstractPhpGeneratorContext<
     public abstract getLocationForTypeId(typeId: TypeId): FileLocation;
 
     protected getFileLocation(filepath: FernFilepath, suffix?: string): FileLocation {
-        let parts = [this.getRootNamespace(), ...filepath.allParts.map((path) => path.pascalCase.safeName)];
```
dcb6, line 151: refactored this a bit, but same outcome

generators/php/codegen/src/asIs/SerializableType.Template.php
```diff
@@ -171,19 +176,19 @@ private static function serializeMap(array $data, array $type): array
...
-        return array_map(fn($item) => self::serializeValue($item, $valueType), $data);
+        return array_map(fn ($item) => self::serializeValue($item, $valueType), $data);
     }
 
     /**
      * Helper function to retrieve the JSON key for a property.
      *
      * @param ReflectionProperty $property
-     * @return string
+     * @return string|null
```
amckinney, line 186: nit: Should we use `?string` instead?

generators/php/codegen/src/asIs/SerializableType.Template.php
```diff
@@ -171,19 +176,19 @@ private static function serializeMap(array $data, array $type): array
     private static function serializeList(array $data, array $type): array
     {
         $valueType = $type[0];
-        return array_map(fn($item) => self::serializeValue($item, $valueType), $data);
+        return array_map(fn ($item) => self::serializeValue($item, $valueType), $data);
```
amckinney, line 179: Is this what the formatter produces? I would have expected `fn($item)`

generators/php/codegen/src/ast/Attribute.ts
```diff
@@ -0,0 +1,39 @@
+import { AstNode } from "./core/AstNode";
+import { ClassReference } from "./ClassReference";
+import { Writer } from "./core/Writer";
+
+export declare namespace Attribute {
+    interface Args {
+        /* Reference to the Attribute */
```
amckinney, line 7: nit: 
```suggestion
        /* Reference to the attribute class reference */
```

generators/php/codegen/src/ast/Class.ts
```diff
@@ -59,7 +63,11 @@ export class Class extends AstNode {
         }
         this.writeComment(writer);
         writer.writeLine(`class ${this.name}`);
-        writer.writeLine("{");
+        if (this.parentClassReference != null) {
+            writer.write(" extends ");
+            this.parentClassReference.write(writer);
+        }
+        writer.writeLine(" {");
```
amckinney, line 70: I know it's formatted anyway, but I think this should be:
```suggestion
        writer.writeLine("{");
```

amckinney: Then we can handle the additional space required in the `parentClassReference` case.

generators/php/codegen/src/ast/Enum.ts
```diff
@@ -0,0 +1,77 @@
...
+import { Writer } from "./core/Writer";
+
+export declare namespace Enum {
+    interface Args {
+        /* The name of the PHP enum */
+        name: string;
+        /* The namespace of the PHP enum*/
+        namespace: string;
+        /* If present, specified that the enum is backed by this type */
+        backing?: "string" | "int";
```
amckinney, line 12: Nice 👏

generators/php/codegen/src/ast/Field.ts
```diff
@@ -53,7 +59,7 @@ export class Field extends AstNode {
         }
 
         this.type.write(writer);
-        writer.write(` ${this.name}`);
+        writer.write(` ${this.name.startsWith("$") ? "" : "$"}${this.name}`);
```
amckinney, line 62: nit: It might be worth making a helper function for this.

generators/php/codegen/src/context/AbstractPhpGeneratorContext.ts
amckinney: Much better 👏

generators/php/model/src/generateModels.ts
```diff
@@ -1,25 +1,27 @@
...
-    );
+    for (const [typeId, typeDeclaration] of Object.entries(context.ir.types)) {
+        const file = typeDeclaration.shape._visit<PhpFile | undefined>({
+            alias: () => undefined,
+            enum: (etd: EnumTypeDeclaration) => {
+                return new EnumGenerator(context, typeDeclaration, etd).generate();
+            },
+            object: (otd) => {
+                return new ObjectGenerator(context, typeDeclaration, otd).generate();
+            },
```
amckinney, line 17: nit: Let's spell out the variable name since we do that elsewhere:
```suggestion
            enum: (enumDeclaration: EnumTypeDeclaration) => {
                return new EnumGenerator(context, typeDeclaration, enumDeclaration).generate();
            },
            object: (objectDeclaration) => {
                return new ObjectGenerator(context, typeDeclaration, objectDeclaration).generate();
            },
```

generators/php/model/src/object/ObjectGenerator.ts
```diff
@@ -0,0 +1,179 @@
...
+                return php.codeblock((writer) => {
+                    writer.writeNode(reference);
+                    writer.write("::class");
+                });
+            }
+            default:
+                assertNever(type.internalType);
+        }
+    }
+}
```
amckinney, line 179: Discussed offline - 

These functions might better belong alongside the Type definitions, but the configurable core namespace makes this a little tricky. Given that these attributes always pertain to object generation, we can leave this as-is for now but it might be worth revisiting.

amckinney (approved):

generators/php/codegen/src/ast/Field.ts
dcb6: i think we'll wanna remove this eventually so leaving as is for now

PR #4647, opened 2024-09-15
fix(typescript): read ir through a streaming read as opposed to `JSON.parse`

dsinghvi:
This PR introduces a new function, `parseIR`, which is used to parse intermediate representations (IRs) in various generator files. The function replaces the existing `JSON.parse` method and is designed to handle large IR JSONs, specifically those larger than 500MB.

The `parseIR` function is implemented in the following files:
- `generators/commons/src/utils/parseIR.ts`
- `generators/openapi/src/writeOpenApi.ts`
- `generators/postman/src/writePostmanCollection.ts`
- `generators/typescript/utils/abstract-generator-cli/src/loadIntermediateRepresentation.ts`
- `packages/commons/fs-utils/src/streamObjectFromFile.ts`

Additionally, the PR includes the following changes:
- The `parseIR` function is exported in `generators/commons/src/utils/index.ts`.
- The `parseIR` function is imported and used in `generators/csharp/codegen/src/cli/AbstractCsharpGeneratorCli.ts`, `generators/php/codegen/src/cli/AbstractPhpGeneratorCli.ts`, and `generators/ruby/cli/src/loadIntermediateRepresentation.ts`.
- The `parseIR` function is added to the list of imports in `generators/openapi/package.json`, `generators/postman/package.json`, and `generators/typescript/utils/abstract-generator-cli/src/AbstractGeneratorCli.ts`.
- The `parseIR` function is added to the list of exports in `packages/commons/fs-utils/package.json`.
- The `streamObjectFromFile` function is exported in `packages/commons/fs-utils/src/index.ts`.
- The `streamObjectFromFile` function is added to the list of dependencies in `pnpm-lock.yaml`.

### DIFF
--- a/generators/commons/src/utils/parseIR.ts
+++ b/generators/commons/src/utils/parseIR.ts
@@ -65,7 +65,7 @@ export declare namespace parseIR {
 export async function parseIR<IR>({ absolutePathToIR, parse }: parseIR.Args<IR>): Promise<IR> {
     const irJson = await streamObjectFromFile(absolutePathToIR);
     // eslint-disable-next-line no-console
-    console.log(`Read ${absolutePathToIR}`);
+    console.log(`Parsed ${absolutePathToIR}`);
     const parsedIR = await parse(irJson, {
         unrecognizedObjectKeys: "passthrough",
         allowUnrecognizedEnumValues: true,
@@ -74,7 +74,7 @@ export async function parseIR<IR>({ absolutePathToIR, parse }: parseIR.Args<IR>)
 
     if (!parsedIR.ok) {
         // eslint-disable-next-line no-console
-        console.log(`Failed to read ${absolutePathToIR}`);
+        console.log(`Failed to parse ${absolutePathToIR}`);
         throw new Error(`Failed to parse IR: ${parsedIR.errors}`);
     }
 
--- a/generators/commons/src/utils/parseGeneratorConfig.ts
+++ b/generators/commons/src/utils/parseGeneratorConfig.ts
@@ -5,7 +5,7 @@ import { readFile } from "fs/promises";
 export async function parseGeneratorConfig(pathToConfig: string): Promise<FernGeneratorExec.GeneratorConfig> {
     const configStr = await readFile(pathToConfig);
     // eslint-disable-next-line no-console
-    console.log(`Read ${pathToConfig}`);
+    console.log(`Parsed ${pathToConfig}`);
     const rawConfig = JSON.parse(configStr.toString());
     const parsedConfig = await GeneratorExecParsing.GeneratorConfig.parse(rawConfig, {
         unrecognizedObjectKeys: "passthrough"
@@ -13,7 +13,7 @@ export async function parseGeneratorConfig(pathToConfig: string): Promise<FernGe
 
     if (!parsedConfig.ok) {
         // eslint-disable-next-line no-console
-        console.log(`Failed to read ${pathToConfig}`);
+        console.log(`Failed to parse ${pathToConfig}`);
         throw new Error("Failed to parse the generator configuration");
     }
 


generators/typescript/sdk/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.40.2] - 2024-09-12
+
+- Fix: The generated SDK now handles reading IR JSONs that are larger than 500MB. In order to 
+  to this, the function `streamObjectFromFile` is used instead of `JSON.parse`.
```
dsinghvi, line 11: I only released the TypeScript SDK generator since that is the one a user complained about

generators/commons/src/utils/parseIR.ts
```diff
@@ -0,0 +1,82 @@
...
+         * helpful for zurg's internal debug logging.
+         */
+        breadcrumbsPrefix?: string[];
+    }
+}
+
+export async function parseIR<IR>({ absolutePathToIR, parse }: parseIR.Args<IR>): Promise<IR> {
+    const irJson = await streamObjectFromFile(absolutePathToIR);
+    // eslint-disable-next-line no-console
+    console.log(`Read ${absolutePathToIR}`);
```
amckinney, line 68: Thoughts on the following?

```suggestion
    console.log(`Parsing ${absolutePathToIR} ...`);
```

Otherwise, I feel like we'll forget what stage this got to before we receiving an error. It'll be _slightly_ more clear that we're about to parse the IR.

Totally fine with other options too, just feel like `Read` is not clear enough.

generators/commons/src/utils/parseIR.ts
```diff
@@ -0,0 +1,82 @@
...
+    const parsedIR = await parse(irJson, {
+        unrecognizedObjectKeys: "passthrough",
+        allowUnrecognizedEnumValues: true,
+        allowUnrecognizedUnionMembers: true
+    });
+
+    if (!parsedIR.ok) {
+        // eslint-disable-next-line no-console
+        console.log(`Failed to read ${absolutePathToIR}`);
```
amckinney, line 77: Based on the above, maybe "failed to parse"? That way its consistent with the error message too.

amckinney (approved):
Nice! 👏

generators/commons/src/utils/parseIR.ts
dsinghvi: Sounds great, I basically just kept in line with https://github.com/fern-api/fern/blob/5ec83da6ede3c1b6019c73db4b2e4b793ddf6fca/generators/commons/src/utils/parseGeneratorConfig.ts#L8 but down to change

generators/typescript/sdk/CHANGELOG.md
mstade: "Complained" is a bit of a harsh choice of words perhaps – I reported that your software doesn't work properly with large specs, and then offered up assistance in solving the bugs.

Nevertheless, I can confirm this does work with our spec now. I've also checked the size of the `ir.json` and it's 170 356 624 bytes (170,4 MB on disk) – so significantly smaller than 500 MB. I believe the limit isn't file size, but allocation when parsing using `JSON.parse`, which is obviously not the same as the file size.

Presumably there's some correlation between file size and memory allocation when parsing, but I doubt it's a strict 1:1 mapping. I wouldn't be surprised if parsing could exceed the limit even with smaller files than our case. Likely not a lot smaller, but still. But I digress.

Streaming the parsing and the writing should hopefully be a solid fix, thank you for looking into this. We also use the Python and Go generators, so if those could be updated as well that would be much appreciated – thank you! 🙏

dsinghvi: Apologies, didn't mean to offend you! And tbh totally reasonable to be frustrated by Fern having this bug.

PR #4653, opened 2024-09-16
fix(python): file properties now respect custom content type

dsinghvi:
This PR introduces a new `with_content_type` function to handle file content types. The function is added to the file.py module and is used to set the content type of a file when sending multipart requests.

The function is also added to the core_utilities.py module and is used to set the content type of a file when writing to an AST node.

The function is tested in the test_file.py module with various file types and content types.

The function is used in the client.py module to set the content type of a file when sending a request.

The function is added to the file.rb module and is used to set the content type of a file when sending a request.

The function is also added to the file_utilities.rb module and is used to set the content type of a file when writing to a Faraday multipart file part.

The function is used in the requests.rb module to set the content type of a file when sending a request.

The function is added to the service.rb module and is used to set the content type of a file when sending a request.

The function is also added to the service_client.rb module and is used to set the content type of a file when sending a request.

The function is used in the types_export.rb module to set the content type of a file when exporting types.

The function is added to the service.yml module and is used to set the content type of a file when sending a request.

The function is also added to the service_client.yml module and is used to set the content type of a file when sending a request.

The function is used in the service_client.gemspec module to set the content type of a file when sending a request.

The function is added to the service_client.gemspec module and is used to set the content type of a file when sending a request.

- `with_content_type` function added to `file.py` module
- `with_content_type` function added to `core_utilities.py` module
- `with_content_type` function tested in `test_file.py` module
- `with_content_type` function used in `client.py` module
- `with_content_type` function added to `file.rb` module
- `with_content_type` function added to `file_utilities.rb` module
- `with_content_type` function used in `requests.rb` module
- `with_content_type` function added to `service.rb` module
- `with_content_type` function added to `service_client.rb` module
- `with_content_type` function used in `types_export.rb` module
- `with_content_type` function added to `service.yml` module
- `with_content_type` function added to `service_client.yml` module
- `with_content_type` function used in `service_client.gemspec` module
- `with_content_type` function added to `service_client.gemspec` module

### DIFF
--- a/.github/workflows/seed.yml
+++ b/.github/workflows/seed.yml
@@ -1,6 +1,9 @@
 name: seed
 
 on:
+  push:
+    branches:
+      - main 
   pull_request:
     branches:
       - main
--- a/seed/fastapi/examples/resources/types/types/moment.py
+++ b/seed/fastapi/examples/resources/types/types/moment.py
@@ -25,7 +25,7 @@ class Moment(UniversalBaseModel):
             "1994-01-01",
         ),
         datetime=datetime.datetime.fromisoformat(
-            "1994-01-01 01:01:01+00:00",
+            "datetime=datetime.datetime(1994, 1, 1, 1, 1, 1, tzinfo=TzInfo(UTC)) raw='1994-01-01T01:01:01Z'",
         ),
     )
     """
--- a/seed/pydantic/examples/src/seed/examples/resources/types/moment.py
+++ b/seed/pydantic/examples/src/seed/examples/resources/types/moment.py
@@ -25,7 +25,7 @@ class Moment(UniversalBaseModel):
             "1994-01-01",
         ),
         datetime=datetime.datetime.fromisoformat(
-            "1994-01-01 01:01:01+00:00",
+            "datetime=datetime.datetime(1994, 1, 1, 1, 1, 1, tzinfo=TzInfo(UTC)) raw='1994-01-01T01:01:01Z'",
         ),
     )
     """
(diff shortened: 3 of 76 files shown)


generators/python/core_utilities/shared/file.py
```diff
@@ -39,3 +39,22 @@ def convert_file_dict_to_httpx_tuples(
...
     return httpx_tuples
+
+
+def with_content_type(*, file: File, content_type: str) -> File:
+    """ """
+    if isinstance(file, tuple):
+        if len(file) == 2:
+            filename, content = cast(Tuple[Optional[str], FileContent], file)
+            return (filename, content, content_type)
+        elif len(file) == 3:
```
armandobelardo, line 50: It feels a bit odd that the input to these could be:
```
with_content_type(file=("file.json", <bytes>, "application/json"), content_type="application/json")
```

Since the tuple already allows for overriding the content type


It might make more sense to be something like

```
def construct_file(
    *,
    file_contents: FileContent,
    file_name: Optional[str]=None,    
    content_type: Optional[str]=None,
    headers: Optional[Mapping[str,str]]=None
```
So that you're effectively just passing in a well typed and documented `core.File` constructor

armandobelardo: Unless this is strictly internal

armandobelardo (approved):
One small comment on the usage of with_content_type, but otherwise looks reasonable

generators/python/core_utilities/shared/file.py
dsinghvi: Yeah its a core utility thats strictly internal

dsinghvi: is there a safe way to index into tuples to call into that if i changed the method signature?

dsinghvi: going to merge to get out of incident, but happy to break this since its internal

PR #4655, opened 2024-09-16
feat(php): Outline client call structure

amckinney:
This demonstrates an example of the root client and subpackage client method calls, including a few examples with:
* Request body properties
* Query parameters
* Headers
* Authorization bearer token
* Authorization environment variable

From here, we'll add support for each of these features in a series of PRs, but this outlines the full flow.


dsinghvi:
@amckinney closing this out, but feel free to reopen

PR #4657, opened 2024-09-16
feat(php): introduce `SerializableType` class

dcb6:
This PR introduces new files and classes for serialisation and deserialisation of various data types, including arrays, dates, and unions. It also adds tests for these new classes.

## Changes
- New files added: ```ArrayType.Template.php```, ```Constant.Template.php```, ```DateType.Template.php```, ```JsonProperty.Template.php```, ```SerializableType.Template.php```, ```Union.Template.php```
- New classes added: ```ArrayType```, ```Constant```, ```DateType```, ```JsonProperty```, ```SerializableType```, ```Union```
- New tests added: ```DateArrayTypeTest```, ```EmptyArraysTest```, ```InvalidTypesTest```, ```MixedDateArrayTypeTest```, ```NestedUnionArrayTypeTest```, ```NullableArrayTypeTest```, ```NullPropertyTypeTest```, ```ScalarTypesTest```, ```TestTypeTest```, ```UnionArrayTypeTest```
- The ```ModelGeneratorContext``` and ```SdkGeneratorContext``` classes now return the new serialisation and test files
- The ```AsIsFiles``` enum has been updated to include the new files
- The ```packages/cli/cli/dist/dev``` and ```packages/cli/cli/dist/prod``` directories have been removed


generators/php/codegen/src/asIs/Constant.Template.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace <%= namespace%>;
+
+use DateTimeInterface;
+
+class Constant
+{
+    public const DateFormat = 'Y-m-d';
+    public const DeserializationDateFormat = "!" . self::DateFormat;
+    public const DateTimeFormat = DateTimeInterface::RFC3339;
```
amckinney, line 11: I noticed that we _might_ need to export an additional constant here when we want to serialize the DateTIme as a `string` - it looks like `DateTimeInterface::RFC3339` might not work in all cases (re: https://github.com/fern-api/fern/pull/4655):

```php
            $query['deadline'] = $request->deadline->format('c');
```

Not sure what we should name this given that we need to use this `DateTimeFormat` elsewhere.

generators/php/codegen/src/asIs/Constant.Template.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace <%= namespace%>;
+
+use DateTimeInterface;
+
+class Constant
+{
+    public const DateFormat = 'Y-m-d';
+    public const DeserializationDateFormat = "!" . self::DateFormat;
```
amckinney, line 10: nit: Thoughts on this instead?

```suggestion
    public const DateDeserializationFormat = "!" . self::DateFormat;
```

generators/php/codegen/src/asIs/NullableArrayTypeTest.Template.php
```diff
@@ -0,0 +1,43 @@
...
+
+        $json = json_encode($data, JSON_THROW_ON_ERROR);
+
+        $object = NullableArrayType::fromJson($json);
+
+        $this->assertEquals(['one', null, 'three'], $object->nullableStringArray, 'nullable_string_array should match the original data.');
+
+        $serializedJson = $object->toJson();
+
+        $this->assertJsonStringEqualsJsonString($json, $serializedJson, 'Serialized JSON does not match original JSON for nullable_string_array.');
```
amckinney, line 41: nit: I'd consider grouping some of these statements together for readability:

```php
$json = json_encode(
  [
      'nullable_string_array' => ['one', null, 'three']
  ],
  JSON_THROW_ON_ERROR
);

$object = NullableArrayType::fromJson($json);
$this->assertEquals(['one', null, 'three'], $object->nullableStringArray, 'nullable_string_array should match the original data.');

$serializedJson = $object->toJson();
$this->assertJsonStringEqualsJsonString($json, $serializedJson, 'Serialized JSON does not match original JSON for nullable_string_array.');
```

amckinney: The same applies elsewhere in other tests, but no big deal either way. Can easily change whenever later on.

amckinney (approved):

generators/php/codegen/src/asIs/Constant.Template.php
dcb6: talked offline, it does seem to work as is!

generators/php/codegen/src/asIs/Constant.Template.php
dcb6: done ✅

generators/php/codegen/src/asIs/NullableArrayTypeTest.Template.php
dcb6: as discussed offline, will follow up with this!

PR #4662, opened 2024-09-16
feat(php): Generate subpackage layout

amckinney:
This generates the full subpackage layout, and updates the general root client and subpackage client constructor flow to match the outline demonstrated in https://github.com/fern-api/fern/pull/4655.

Additionally, this includes a new `TypeDict` node, which is used to render a comment for a structured array (i.e. for `ClientOptions`). We still have a variety of files with unused private variables so `phpstan` will complain, but this will be addressed as soon as we generate endpoint methods.

### DIFF
--- a/generators/php/codegen/src/asIs/Constant.Template.php
+++ b/generators/php/codegen/src/asIs/Constant.Template.php
@@ -7,6 +7,6 @@ use DateTimeInterface;
 class Constant
 {
     public const DateFormat = 'Y-m-d';
-    public const DeserializationDateFormat = "!" . self::DateFormat;
+    public const DateDeserializationFormat = "!" . self::DateFormat;
     public const DateTimeFormat = DateTimeInterface::RFC3339;
 }
--- a/seed/php-sdk/any-auth/src/Auth/AuthClient.php
+++ b/seed/php-sdk/any-auth/src/Auth/AuthClient.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Seed\Auth;
+
+use Seed\Core\RawClient;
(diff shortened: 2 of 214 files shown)


generators/php/codegen/src/asIs/Constant.Template.php
```diff
@@ -7,6 +7,6 @@
 class Constant
 {
     public const DateFormat = 'Y-m-d';
-    public const DeserializationDateFormat = "!" . self::DateFormat;
+    public const DateDeserializationFormat = "!" . self::DateFormat;
```
dcb6, line 10: oh lol i only fixed it in seed 😅

seed/php-sdk/any-auth/src/Auth/AuthClient.php
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * @var RawClient $client
+     */
+    private RawClient $client;
+
+    /**
+     * @param RawClient $client
+     */
+    public function __construct(
+        RawClient $client,
```
dcb6, line 18: any reason not to do constructor property promotion for these? (can follow up with it)

dcb6 (approved):

seed/php-sdk/any-auth/src/Auth/AuthClient.php
amckinney: I originally left it out since it's a `private` member, but no good reason - we can easily refactor this to use promotion.

I'll leave this as-is for now, but will revisit in a follow-up when we start implementing the methods!

PR #4663, opened 2024-09-17
fix(cli): improve JSON stream performance

mstade:
These changes significantly improve performance of both reading and writing JSON files using the streaming methods in `@fern-api/fs-utils`. Total runtime, memory and disk utilization are all improved.

This is achieved by:

- Not writing prettified JSON IR files (dramatically reduces on-disk size)
- Upgrading to a newer and *much* faster version of the `json-stream-stringify` package – this upgrade alone improved the performance of serializing a ~176 MB ir.json from about 3 minutes down to 3-4 seconds
- Reimplementing `streamObjectToFile` using `Assembler` which is lower-level and faster than `StreamObject`, and also setting parsing options to avoid emitting unnecessary parsing tokens, for a 30-35% improvement in parsing performance

This pull request updates the `json-stream-stringify` package to version 3.1.4.

- In `packages/commons/fs-utils/package.json`, the `json-stream-stringify` version is updated to 3.1.4.
- In `packages/commons/fs-utils/src/streamObjectFromFile.ts`, the `StreamObject` and `StreamJSON` imports are updated to `stream-json`.
- In `packages/cli/generation/local-generation/local-workspace-runner/src/runGenerator.ts`, the `pretty` option is set to `false` when writing IR to a file.
- In `pnpm-lock.yaml`, the `json-stream-stringify` version is updated to 3.1.4, and the `packages/cli/cli/dist/prod` package is added.

### DIFF
--- a/packages/commons/fs-utils/src/streamObjectFromFile.ts
+++ b/packages/commons/fs-utils/src/streamObjectFromFile.ts
@@ -6,14 +6,12 @@ import Assembler from "stream-json/Assembler.js";
 export async function streamObjectFromFile(filePath: AbsoluteFilePath): Promise<unknown> {
     return new Promise((resolve, reject) => {
         const jsonStream = StreamJSON.parser({
-            streamValues: false,
+            streamValues: false
         });
 
         const asm = Assembler.connectTo(jsonStream);
-        asm.on("done", ({current}) => resolve(current));
+        asm.on("done", ({ current }) => resolve(current));
 
-        createReadStream(filePath)
-            .pipe(jsonStream)
-            .on("error", reject);
+        createReadStream(filePath).pipe(jsonStream).on("error", reject);
     });
 }


packages/cli/generation/local-generation/local-workspace-runner/src/runGenerator.ts
```diff
@@ -157,7 +157,7 @@ async function writeIrToFile({
         tmpdir: workspaceTempDir.path
     });
     const absolutePathToIr = AbsoluteFilePath.of(irFile.path);
-    await streamObjectToFile(absolutePathToIr, ir, { pretty: true });
+    await streamObjectToFile(absolutePathToIr, ir, { pretty: false });
```
mstade, line 160: This reduces the file size of the generated IR JSON significantly. Makes it a bit more tedious to debug, but running a prettifier on the file solves that problem in a jiffy.

packages/commons/fs-utils/package.json
```diff
@@ -30,7 +30,7 @@
     "@fern-api/core-utils": "workspace:*",
     "glob": "^8.0.3",
     "glob-promise": "^5.0.0",
-    "json-stream-stringify": "^3.0.1",
+    "json-stream-stringify": "^3.1.4",
```
mstade, line 33: Upgrading this dependency provides a _huge_ performance increase. Before this, serializing our large spec to the IR would take 3-4 minutes, and after this dependency upgrade that is now down to seconds.

packages/commons/fs-utils/src/streamObjectFromFile.ts
mstade, line 1: The unified diff of the changes in this file makes these changes unreadable. I would suggest using the split mode instead when reviewing this code.

mstade (commented):
## Author Notes

mstade:
@dsinghvi generating SDKs from our large spec took several minutes to complete, and while debugging this I noticed writing and subsequently reading the IR JSON is where a significant portion of time was spent. These changes aim to improve performance of the stream objects to/from file utilities.

Particularly the writing performance was poor, but also reading the file in the docker container is quite slow. Alas, I think the latter is more due to inherent performance issues with Docker bind mounts in MacOS than an underperforming parser. That said, I did discover some options to improve the parsing time by some amount as well.

I offer these improvements for your review.

mstade:
BTW, I was able to sort out my issues with running `seed` locally, and was thus able to test at least the TS generator locally. For some reason running the TS generator this way is _significantly_ faster when it comes to reading the IR JSON, than when using the remote image. 🤷🏻‍♂️

dsinghvi (approved):
@mstade this is epic!

PR #4667, opened 2024-09-17
fix(cli): openapi importer supports a setting for making additional properties optional

dsinghvi:
…onal additional properties

This PR introduces a new feature to allow additional properties in OpenAPI to be optional.

## Changes
- The `optionalAdditionalProperties` field is added to the APIDefinitionSettings interface.
- The `optionalAdditionalProperties` field is added to the SpecImportSettings interface.
- The `optionalAdditionalProperties` field is added to the ParseOpenAPIOptions interface.
- The `optionalAdditionalProperties` field is added to the OSSWorkspace.Settings interface.
- The `optionalAdditionalProperties` field is added to the Workspace interface.
- The `optionalAdditionalProperties` field is added to the OpenAPISettingsSchema and AsyncAPISettingsSchema objects.
- The `optionalAdditionalProperties` field is added to the OpenAPISpecSchema and AsyncAPISchema objects.
- The `optionalAdditionalProperties` field is added to the parseAPIConfigurationToApiLocations function.
- The `optionalAdditionalProperties` field is added to the convertAdditionalProperties function.
- The `optionalAdditionalProperties` field is added to the loadSingleNamespaceAPIWorkspace function.

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,5 +1,23 @@
 - changelogEntry:
   - summary: | 
+      Previously the OpenAPI converter would incorrectly mark 
+      the values of `additionalProperties` as optional. Now, we have 
+      introduced a feature flag to turn this behavior off. 
+
+      The feature flag can be configured in generators.yml: 
+      ```yml
+      api: 
+        specs: 
+          - openapi: /path/to/openapi
+            settings: 
+              optional-additional-properties: false
+      ```
+    type: fix
+  irVersion: 53
+  version: 0.41.16
+
+- changelogEntry:
+  - summary: | 
       Performance improvements for stringifiying large Intermediate Representations. If 
       you have a large OpenAPI spec or Fern Definition, this can potentially shave off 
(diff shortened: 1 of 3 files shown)


packages/cli/openapi-parser/src/options.ts
```diff
@@ -9,11 +9,14 @@ export interface ParseOpenAPIOptions {
     useTitlesAsName: boolean;
     /* The audiences that the OpenAPI must be filtered down into */
     audiences: string[] | undefined;
+    /* Whether or not to make additional proeprties optional */
```
amckinney, line 12:
```suggestion
    /* Whether or not to make additional properties optional */
```

amckinney: Could also make this _slightly_ more clear and say "... property _values_ optional".

amckinney (approved):

packages/cli/openapi-parser/src/options.ts
dsinghvi: updated, thank you!

PR #4668, opened 2024-09-17
fix(cli): release performance improvements around json read/write

dsinghvi:
This pull request introduces performance enhancements for the `fern generate` command, particularly when working with large API specifications and Intermediate Representations.

- The `stream-json` dependency is upgraded, improving performance when reading large API specs.
- The `fern generate` command now benefits from performance improvements when stringifying large Intermediate Representations, potentially reducing execution time by minutes.
- The version number is updated from `0.40.3` to `0.40.4`.


dsinghvi:
@mstade this will release your changes and make them accessible

amckinney (approved):

mstade:
> @mstade this will release your changes and make them accessible

Thanks @dsinghvi and team – this helps _a lot_! I just ran the generation with debug logging for v0.41.15 and v0.40.4 of the CLI and TS generator respectively, and writing the temp IR JSON file took about 2.5 seconds, down from 2-3 minutes, and parsing it back again in the TS generator container took about 20 seconds, down from about a minute or so. I'm surprised the parsing was so improved, in my testing I only saw about a 30-35% improvement but this is more like double that, which is of course wonderful!

May I ask for new releases of the Go and Python generators as well, incorporating these improvements? Thank you so much! 🙏

PR #4671, opened 2024-09-17
feat(python): fastapi server codegen now supports file upload and download

armandobelardo:
Fixes FER-2782

The title says it all, this still needs to be tested, but the codegen looks sane

This supports upload and download, for upload I inlined all the parameters, since I wasn't seeing in FastAPI docs if you could have file upload params within a model, and it felt more sane to do this then to split up the object into inlining file upload properties and referencing body properties as an object

### DIFF
--- a/generators/python/src/fern_python/generators/fastapi/service_generator/endpoint_generator.py
+++ b/generators/python/src/fern_python/generators/fastapi/service_generator/endpoint_generator.py
@@ -199,7 +199,7 @@ class EndpointGenerator:
                 writer.write_line(f'path="{self._get_endpoint_path()}",')
 
                 # Void responses make more sense as response_class, but keeping as response_model to not modify existing users
-                if  not self._get_is_return_type_pydantic_model():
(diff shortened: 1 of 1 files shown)


dsinghvi:
@armandobelardo codegen looks good to me too -- will hold off on approving till after you test

seed/fastapi/plain-text/resources/service/service/service.py
```diff
@@ -67,7 +67,7 @@ def wrapper(*args: typing.Any, **kwargs: typing.Any) -> str:
 
         router.post(
             path="/text",
-            response_model=str,
+            response_class=str,
```
dsinghvi, line 70: know why this changed?

dsinghvi (approved):

seed/fastapi/plain-text/resources/service/service/service.py
armandobelardo: Yeah will update, this is technically correct, but will revert for strings to maintain parity

PR #4675, opened 2024-09-17
fix(cli): try to resolve `Cannot read properties of undefined (reading 'includes')` error

dsinghvi:
This PR introduces a fix to the `rerunFernCliAtVersion` function in the `rerunFernCliAtVersion.ts` file. The fix addresses an issue where the function fails to check for optionality when parsing `stdout` and `stderr` in the CLI, leading to an error: `Cannot read properties of undefined (reading 'includes')`.

## Changes:
- The `if` statement in the `rerunFernCliAtVersion` function now uses optional chaining (`stdout?.includes`) to check for the presence of the string "code EEXIST" in `stdout` or `stderr`.
- The `changelogEntry` in the `versions.yml` file has been updated to reflect the fix, with a summary describing the issue and the type set to "fix".

### DIFF
--- a/packages/cli/cli/src/rerunFernCliAtVersion.ts
+++ b/packages/cli/cli/src/rerunFernCliAtVersion.ts
@@ -35,7 +35,7 @@ export async function rerunFernCliAtVersion({
             [FERN_CWD_ENV_VAR]: process.env[FERN_CWD_ENV_VAR] ?? process.cwd()
         }
     });
-    if (stdout.includes("code EEXIST") || stderr.includes("code EEXIST")) {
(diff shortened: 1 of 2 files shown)


packages/cli/cli/src/rerunFernCliAtVersion.ts
```diff
@@ -35,7 +35,7 @@ export async function rerunFernCliAtVersion({
             [FERN_CWD_ENV_VAR]: process.env[FERN_CWD_ENV_VAR] ?? process.cwd()
         }
     });
-    if (stdout.includes("code EEXIST") || stderr.includes("code EEXIST")) {
```
dsinghvi, line 38: this feels like the only recent additions to the CLI that use `includes`

PR #4677, opened 2024-09-18
feat(php): Generate endpoint method layout

amckinney:
This generates endpoint methods for all PHP client classes. The general flow of the endpoint implementation is shown below:

```php
/**
 * @returns GetUserResponse
 */
public function get(?array $options): GetUserResponse
{
    try {
        $response = $this->client->sendRequest(...);
            return GetUserResponse::fromJson($response->getBody->getContents());
        }
    } catch (JsonException $e) {
        throw new Exception("Failed to deserialize response", 0, $e);
        throw new Exception($e->getMessage());
    }
    throw new Exception("Error with status code " . $statusCode);
}
```

The implementation here is incomplete, but the general structure of each endpoint is implemented by the `EndpointGenerator` (which is largely inspired by the C#-equivalent).

Future changes will actually generate the in-lined request type, deserialize the JSON response, and get everything working.

### DIFF
--- a/generators/php/sdk/src/endpoint/request/WrappedEndpointRequest.ts
+++ b/generators/php/sdk/src/endpoint/request/WrappedEndpointRequest.ts
@@ -52,7 +52,7 @@ export class WrappedEndpointRequest extends EndpointRequest {
         const requiredQueryParameters: QueryParameter[] = [];
         const optionalQueryParameters: QueryParameter[] = [];
         for (const queryParameter of this.endpoint.queryParameters) {
-            if (!queryParameter.allowMultiple && this.context.isOptional(queryParameter.valueType)) {
+            if (this.context.isOptional(queryParameter.valueType)) {
                 optionalQueryParameters.push(queryParameter);
             } else {
                 requiredQueryParameters.push(queryParameter);
@@ -77,11 +77,6 @@ export class WrappedEndpointRequest extends EndpointRequest {
         };
     }
 
-    private writeQueryParameter(writer: php.Writer, query: QueryParameter): void {
-        writer.write(`${QUERY_PARAMETER_BAG_NAME}['${query.name.wireValue}'] = `);
-        writer.writeNodeStatement(this.stringify({ reference: query.valueType, name: query.name.name }));
-    }
-
     public getHeaderParameterCodeBlock(): HeaderParameterCodeBlock | undefined {
         const service = this.context.getHttpServiceOrThrow(this.serviceId);
         const headers = [...service.headers, ...this.endpoint.headers];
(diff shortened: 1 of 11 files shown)


generators/php/codegen/src/context/AbstractPhpGeneratorContext.ts
```diff
@@ -72,6 +100,14 @@ export abstract class AbstractPhpGeneratorContext<
         return `${this.rootNamespace}\\Tests\\Core`;
     }
 
+    public getParameterName(name: Name): string {
+        return `$${name.camelCase.safeName}`;
```
dcb6, line 104: can do `unsafe` as we found out!

generators/php/codegen/src/context/AbstractPhpGeneratorContext.ts
```diff
@@ -72,6 +100,14 @@ export abstract class AbstractPhpGeneratorContext<
...
+    }
+
+    public getPropertyName(name: Name): string {
+        return name.camelCase.safeName;
```
dcb6, line 108: same here!

generators/php/codegen/src/context/AbstractPhpGeneratorContext.ts
```diff
@@ -94,6 +130,69 @@ export abstract class AbstractPhpGeneratorContext<
         }
     }
 
+    public isEnum(typeReference: TypeReference): boolean {
```
dcb6, line 133: fine for now, but would like to revisit these; i think i have ideas for a bit more code reuse

seed/php-sdk/any-auth/src/User/UserClient.php
```diff
@@ -19,4 +22,25 @@ public function __construct(
     ) {
         $this->client = $client;
     }
+
+    /**
+     * @param ?array{baseUrl?: string} $options
+     * @returns mixed
+     */
+    public function get(?array $options): mixed
```
dcb6, line 30: is the parameter optional even without the `= null`?

generators/php/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -0,0 +1,165 @@
...
+const STATUS_CODE_VARIABLE_NAME = "$statusCode";
+
+export class HttpEndpointGenerator extends AbstractEndpointGenerator {
+    public constructor({ context }: { context: SdkGeneratorContext }) {
+        super({ context });
+    }
+
+    public generate({ serviceId, endpoint }: { serviceId: ServiceId; endpoint: HttpEndpoint }): php.Method {
+        const endpointSignatureInfo = this.getEndpointSignatureInfo({ serviceId, endpoint });
+        const parameters = [...endpointSignatureInfo.baseParameters];
```
dcb6, line 28: are we ordering by required parameters first then optional?

dcb6 (approved):
nice!!

seed/php-sdk/any-auth/src/User/UserClient.php
amckinney: Great catch - I noticed your update in https://github.com/fern-api/fern/pull/4645 actually resolved this as soon as I sync'd 👍

generators/php/sdk/src/endpoint/http/HttpEndpointGenerator.ts
amckinney: Yes, but only because we list them in order of path parameters -> request type -> request options. Path parameters are inherently required so we can get away with this for now. 

If we end up needing to support optional path parameters here (which is still unclear whether or not this should be a valid use case), we will need to reorder.

PR #4689, opened 2024-09-18
feat(php): fix model bugs

dcb6:
Fixes the following issues:
* Recursive types don't need import
* Types in same namespace don't need import
* Attributes go after the comment.
* Invalid property / parameter names

# Generated

This pull request introduces a number of changes to the codebase, primarily focused on updating the `@param` type in the `ArrayType` class and modifying the `ObjectGenerator` class.

## ArrayType
- The `@param` type has been changed from `array` to `mixed[]` in the `ArrayType` class. This change is reflected in the following files:
  - generators/php/codegen/src/asIs/ArrayType.Template.php
  - seed/php-model/alias-extends/src/Core/ArrayType.php
  - seed/php-model/alias/src/Core/ArrayType.php
  - seed/php-model/any-auth/src/Core/ArrayType.php
  - seed/php-model/api-wide-base-path/src/Core/ArrayType.php
  - seed/php-model/audiences/src/Core/ArrayType.php
  - seed/php-model/auth-environment-variables/src/Core/ArrayType.php
  - seed/php-model/basic-auth-environment-variables/src/Core/ArrayType.php
  - seed/php-model/basic-auth/src/Core/ArrayType.php
  - seed/php-model/bearer-token-environment-variable/src/Core/ArrayType.php
  - seed/php-model/bytes/src/Core/ArrayType.php
  - seed/php-model/circular-references-advanced/src/Core/ArrayType.php

## ObjectGenerator
- The `ObjectGenerator` class has been modified to use the `properties` array to create a new array of objects with the `property`, `type`, `name`, and `docs` properties.
- The `requiredProperties` and `optionalProperties` arrays are now created by filtering the `properties` array based on the `type.internalType.type` property.
- The `orderedProperties` array is created by concatenating the `requiredProperties` and `optionalProperties` arrays.
- The `parameters` array is now created by mapping over the `orderedProperties` array and using the `php.parameter` function.
- The constructor of the `clazz` object is updated to use the `orderedProperties` array to set the properties of the object.

## Other Changes
- The `AbstractPhpGeneratorContext` class has been updated to use the `prependUnderscoreIfNeeded` function to prepend an underscore to the property name if it does not start with a letter or underscore.
- The `Writer` class has been modified to remove the type name from the reference and filter out the current namespace.
- The `Child`, `Parent_`, `Type`, `TokenResponse`, `User`, `Response`, `Foo`, `FolderCFoo`, `Response`, `FilteredType`, `ImportingType`, `UnauthorizedRequestErrorBody` classes have been updated to use the `JsonProperty` annotation.
- The `ObjectFieldValue` class has been updated to use the `JsonProperty` annotation for the `name` and `value` properties.

### DIFF
--- a/.github/workflows/seed.yml
+++ b/.github/workflows/seed.yml
@@ -3,7 +3,7 @@ name: seed
 on:
   push:
     branches:
-      - main 
+      - main
   pull_request:
     branches:
       - main
@@ -651,7 +651,7 @@ jobs:
         env:
           FORCE_COLOR: "2"
         run: |
-          pnpm seed:local test --generator php-model --parallel 16
+          pnpm seed:local test --generator php-model --parallel 8
 
       - name: Ensure no changes to git-tracked files
         run: git --no-pager diff --exit-code  -- ":(exclude)seed/*/.mock/*"
@@ -682,7 +682,7 @@ jobs:
         env:
           FORCE_COLOR: "2"
         run: |
-          pnpm seed:local test --generator php-sdk --parallel 16
+          pnpm seed:local test --generator php-sdk --parallel 8
 
       - name: Ensure no changes to git-tracked files
         run: git --no-pager diff --exit-code  -- ":(exclude)seed/*/.mock/*"


generators/php/model/src/object/ObjectGenerator.ts
```diff
@@ -35,38 +35,35 @@ export class ObjectGenerator extends FileGenerator<PhpFile, ModelCustomConfigSch
             parentClassReference: this.context.getSerializableTypeClassReference()
         });
         // todo: handle extended properties
-        const requiredProperties = this.objectDeclaration.properties.filter(
```
dcb6, line 38: changes in this file are just using the new `getPropertyName` implementation, then some refactoring

amckinney (approved):

PR #4694, opened 2024-09-18
fix(python): respect serialization with more types and dump JSON correctly

armandobelardo:
This PR does 3 main things:
1. Updates example generation to no longer add a generic object once the max depth is reached within a container, instead, the container is empty if max depth is reached -- allowing the generated example to remain valid
2. Update the Python test gen to write examples for unknowns as we do within tests, as opposed to using `json.dumps` which produces valid JSON, but not valid python
3. Update Python serialization to respect dicts when serializing without Pydantic aliases

### DIFF
--- a/seed/python-sdk/alias/src/seed/core/serialization.py
+++ b/seed/python-sdk/alias/src/seed/core/serialization.py
@@ -80,12 +80,7 @@ def convert_and_respect_annotation_metadata(
         value_type = typing_extensions.get_args(clean_type)[1]
 
         return {
-            convert_and_respect_annotation_metadata(
-                object_=key,
-                annotation=annotation,
-                inner_type=key_type,
-                direction=direction,
-            ): convert_and_respect_annotation_metadata(
+            key: convert_and_respect_annotation_metadata(
                 object_=value,
                 annotation=annotation,
                 inner_type=value_type,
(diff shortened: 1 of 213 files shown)


seed/python-sdk/alias/src/seed/core/serialization.py
```diff
@@ -71,6 +71,29 @@ def convert_and_respect_annotation_metadata(
...
+        key_type = typing_extensions.get_args(clean_type)[0]
+        value_type = typing_extensions.get_args(clean_type)[1]
+
+        return {
+            convert_and_respect_annotation_metadata(
+                object_=key,
+                annotation=annotation,
+                inner_type=key_type,
+                direction=direction,
+            ): convert_and_respect_annotation_metadata(
```
dsinghvi, line 88: would love to not process the key

dsinghvi (approved):

dsinghvi:
>Update the Python test gen to write generic examples as we do within tests, as opposed to using json.dumps which produces valid JSON, but not valid python

**For unknowns

seed/python-sdk/alias/src/seed/core/serialization.py
armandobelardo: good call updated!

PR #4696, opened 2024-09-18
feat(php): Generate in-lined request types

amckinney:
This generates in-lined request types, which are used to call endpoints with the `RawClient->sendRequest` method.

A few other things were updated to get things working, including:
1. Class and method invocations now include a `multiline` option.
2. Argument writing is consolidated in the `ast/util/writeArguments` helper function.
3. PHP variable name resolution is handled with the `ast/util/convertToPhpVariableName` helper function.
4. The `PhpAttributeMapper` is added to the context so that it can be used across both objects and in-lined requests.

In a follow-up, we will update the endpoint signature to return the typed response, and use the `JsonDecoder` (to be introduced) to handle decoding the response.

### DIFF
--- a/generators/php/sdk/src/endpoint/request/WrappedEndpointRequestGenerator.ts
+++ b/generators/php/sdk/src/endpoint/request/WrappedEndpointRequestGenerator.ts
@@ -34,58 +34,51 @@ export class WrappedEndpointRequestGenerator extends FileGenerator<
     }
 
     protected doGenerate(): PhpFile {
-        const class_ = php.class_(this.classReference);
+        const clazz = php.class_(this.classReference);
+
+        const properties: php.Field.Args[] = [];
 
         const service = this.context.getHttpServiceOrThrow(this.serviceId);
         for (const header of [...service.headers, ...this.endpoint.headers]) {
-            class_.addField(
-                php.field({
-                    name: this.context.getPropertyName(header.name.name),
-                    type: this.context.phpTypeMapper.convert({ reference: header.valueType }),
-                    access: "public",
-                    docs: header.docs
-                })
-            );
+            properties.push({
+                name: this.context.getPropertyName(header.name.name),
+                type: this.context.phpTypeMapper.convert({ reference: header.valueType }),
+                access: "public",
+                docs: header.docs
+            });
         }
 
         for (const query of this.endpoint.queryParameters) {
             const type = query.allowMultiple
                 ? php.Type.array(this.context.phpTypeMapper.convert({ reference: query.valueType }))
                 : this.context.phpTypeMapper.convert({ reference: query.valueType });
-
-            class_.addField(
-                php.field({
-                    name: this.context.getPropertyName(query.name.name),
-                    type,
-                    access: "public",
-                    docs: query.docs
-                })
-            );
+            properties.push({
+                name: this.context.getPropertyName(query.name.name),
+                type,
+                access: "public",
+                docs: query.docs
+            });
         }
 
         this.endpoint.requestBody?._visit({
             reference: (reference) => {
-                class_.addField(
-                    php.field({
-                        name: this.context.getPropertyName(this.wrapper.bodyKey),
-                        type: this.context.phpTypeMapper.convert({ reference: reference.requestBodyType }),
-                        access: "public",
-                        docs: reference.docs
-                    })
-                );
+                properties.push({
+                    name: this.context.getPropertyName(this.wrapper.bodyKey),
+                    type: this.context.phpTypeMapper.convert({ reference: reference.requestBodyType }),
+                    access: "public",
+                    docs: reference.docs
+                });
             },
             inlinedRequestBody: (request) => {
                 for (const property of request.properties) {
                     const type = this.context.phpTypeMapper.convert({ reference: property.valueType });
-                    class_.addField(
-                        php.field({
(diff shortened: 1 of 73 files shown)


generators/php/sdk/src/endpoint/request/EndpointRequest.ts
```diff
@@ -25,7 +25,7 @@ export abstract class EndpointRequest {
     ) {}
 
     public getRequestParameterName(): string {
-        return this.context.getParameterName(this.sdkRequest.requestParameterName);
+        return `$${this.context.getParameterName(this.sdkRequest.requestParameterName)}`;
```
dcb6, line 28: i'd rather keep the convention that `get...Name` does not include `$`

in my mind, the `$` is not part of the name, so with the method name semantics, this makes sense

generators/php/sdk/src/endpoint/request/WrappedEndpointRequestGenerator.ts
```diff
@@ -0,0 +1,113 @@
...
+        this.serviceId = serviceId;
+        this.endpoint = endpoint;
+        this.classReference = this.context.getRequestWrapperReference(this.serviceId, this.wrapper.wrapperName);
+        this.location = this.context.getLocationForWrappedRequest(this.serviceId);
+    }
+
+    protected doGenerate(): PhpFile {
+        const class_ = php.class_({
+            ...this.classReference
+        });
```
dcb6, line 39: this is unimportant and more of a curiosity -- why not do this?
```suggestion
        const class_ = php.class_(this.classReference);
```

generators/php/sdk/src/endpoint/request/WrappedEndpointRequestGenerator.ts
```diff
@@ -0,0 +1,113 @@
...
+                    php.field({
+                        name: this.context.getPropertyName(this.wrapper.bodyKey),
+                        type: this.context.phpTypeMapper.convert({ reference: reference.requestBodyType }),
+                        access: "public",
+                        docs: reference.docs
+                    })
+                );
+            },
+            inlinedRequestBody: (request) => {
+                for (const property of [...request.properties, ...(request.extendedProperties ?? [])]) {
```
dcb6, line 80: i've been ignoring `extendedProperties` for now

PHP doesn't support multiple inheritance, but we could potentially leverage traits for a similar effect, see [ticket](https://linear.app/buildwithfern/issue/FER-3215/[php]-inheritance)

generators/php/sdk/src/endpoint/request/WrappedEndpointRequestGenerator.ts
```diff
@@ -0,0 +1,113 @@
...
+        this.wrapper = wrapper;
+        this.serviceId = serviceId;
+        this.endpoint = endpoint;
+        this.classReference = this.context.getRequestWrapperReference(this.serviceId, this.wrapper.wrapperName);
+        this.location = this.context.getLocationForWrappedRequest(this.serviceId);
+    }
+
+    protected doGenerate(): PhpFile {
+        const class_ = php.class_({
+            ...this.classReference
```
dcb6, line 38: donde esta la constructora

seed/php-sdk/audiences/src/Foo/FooClient.php
```diff
@@ -25,18 +27,26 @@ public function __construct(
...
+     * @param FindRequest $request
      * @param ?array{baseUrl?: string} $options
      * @returns mixed
      */
     public function find(FindRequest $request, ?array $options = null): mixed
     {
         $query = [];
-        if (request->optionalString != null) {
-            $query['optionalString'] = request->optionalString;
+        if ($request->optionalString != null) {
```
dcb6, line 37: i've wondering this in other languages; why not just construct this map all in one go and then filter out any property will null values after the fact?

generators/php/sdk/src/endpoint/request/EndpointRequest.ts
amckinney: Discussed offline - this is the only place where it's required since we manually specify the request variable name (without the AST helpers).

seed/php-sdk/audiences/src/Foo/FooClient.php
dcb6: talked offline and made a `Nits` ticket to consider this improvement down the line

generators/php/sdk/src/endpoint/request/WrappedEndpointRequestGenerator.ts
amckinney: Good catch, will commit your suggestion!

generators/php/sdk/src/endpoint/request/WrappedEndpointRequestGenerator.ts
amckinney: Sounds good - I'll drop them here for now so that we can add them all in at the same time.

generators/php/sdk/src/endpoint/request/WrappedEndpointRequestGenerator.ts
amckinney: d'oh! Will add 👍

dcb6 (approved):
🙌

PR #4699, opened 2024-09-19
feat(php): Generate single url environment

amckinney:
## Environments

This adds support for single URL environments, which are generated with a simple, string-backed enum like so:

```php
<?php

namespace Acme;

enum Environments: string
{
    case Production = "https://production.com/api";
    case Staging = "https://staging.com/api";
}
```

With this, the generated code can access the default environment (if any):

```php
    new JsonApiRequest(
        baseUrl: $this->options['baseUrl'] ?? $this->client->options['baseUrl'] ?? Environments::Production->value,
        path: "/users",
        method: HttpMethod::GET,
    ),
);
```

## Root client

Tthis also improves the root client constructor flow, so that everything is a lot easier to follow and extensible. The constructor now looks like the following:

```php
public function __construct(
    ?array $options = null,
) {
    $defaultHeaders = [
        "X-Fern-Language" => "PHP",
        "X-Fern-SDK-Name" => "Seed",
        "X-Fern-SDK-Version" => "0.0.1",
    ];

    $this->options = $options ?? [];
    $this->options['headers'] = array_merge(
        $defaultHeaders,
        $this->options['headers'] ?? [],
    );

    $this->client = new RawClient(
        options: $this->options,
    );

    $this->imdb = new ImdbClient($this->client);
}
```

### DIFF
--- a/generators/php/codegen/src/asIs/RawClient.Template.php
+++ b/generators/php/codegen/src/asIs/RawClient.Template.php
@@ -2,6 +2,7 @@
 
 namespace <%= namespace%>;
 
+use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Psr7\Utils;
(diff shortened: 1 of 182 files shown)


dcb6 (approved):
💪

PR #4700, opened 2024-09-19
fix(cli): OpenAPI importer handles arbitrary content types instead of skipping

dsinghvi:
This pull request introduces a fix to the OpenAPI importer's handling of content types. Previously, the importer would ignore parsing arbitrary content types "*/*", but now it treats this content type as `application/json`.

- The `changelogEntry` field in `versions.yml` has been updated to reflect this change.
- The `getApplicationJsonSchemaMediaObject` function in `getApplicationJsonSchema.ts` now includes a check for the "*/*" content type.


armandobelardo (approved):
Approved to unblock, but are there any docs you could link in a comment that explain `*/*`? Couldn't find anything on a quick search

dsinghvi:
> Approved to unblock, but are there any docs you could link in a comment that explain `*/*`? Couldn't find anything on a quick search

Yup -- 
https://swagger.io/docs/specification/media-types/
Screenshot 2024-09-19 at 7 50 31 AM

PR #4702, opened 2024-09-19
feat(php): Add auth w/ env support

amckinney:
This adds support for generating auth schemes, including those backed by an environment variable. If an environment variable is defined, the generated root client constructor looks like the following:

```php
public function __construct(
    ?string $apiKey = null,
    ?array $options = null,
) {
    $apiKey ??= $this->getFromEnvOrThrow('API_KEY', 'Please pass in apiKey or set the environment variable API_KEY.');
    $defaultHeaders = [
        'Authorization' => "Bearer $apiKey",
        'X-Fern-Language' => 'PHP',
        'X-Fern-SDK-Name' => 'Acme',
        'X-Fern-SDK-Version' => '0.0.1',
    ];

    $this->options = $options ?? [];
    $this->options['headers'] = array_merge(
        $defaultHeaders,
        $this->options['headers'] ?? [],
    );

    $this->client = new RawClient(
        options: $this->options,
    );

    $this->user = new UserClient($this->client);
}

private function getFromEnvOrThrow(string $env, string $message): string
{
    $value = getenv($env);
    return $value ? (string) $value : throw new Exception($message);
}
```

Again, the implementation is inspired by the C#-equivalent, but it's been adapted to satisfy additional use cases. 

Note that this also updates a variety of the `string` constants to use single quotes (e.g. `'X-Fern-Language'`) whenever we don't need to use string interpolation (which is more idiomatic).

### DIFF
--- a/seed/php-sdk/alias-extends/src/SeedClient.php
+++ b/seed/php-sdk/alias-extends/src/SeedClient.php
@@ -29,9 +29,9 @@ class SeedClient
         ?array $options = null,
     ) {
         $defaultHeaders = [
-            "X-Fern-Language" => "PHP",
-            "X-Fern-SDK-Name" => "Seed",
-            "X-Fern-SDK-Version" => "0.0.1",
+            'X-Fern-Language' => 'PHP',
+            'X-Fern-SDK-Name' => 'Seed',
+            'X-Fern-SDK-Version' => '0.0.1',
         ];
 
         $this->options = $options ?? [];
--- a/generators/php/codegen/src/context/AbstractPhpGeneratorContext.ts
+++ b/generators/php/codegen/src/context/AbstractPhpGeneratorContext.ts
@@ -113,7 +113,7 @@ export abstract class AbstractPhpGeneratorContext<
     }
 
     public getLiteralAsString(literal: Literal): string {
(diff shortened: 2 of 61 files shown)


seed/php-sdk/alias-extends/src/SeedClient.php
```diff
@@ -29,9 +29,9 @@ public function __construct(
         ?array $options = null,
     ) {
         $defaultHeaders = [
-            "X-Fern-Language" => "PHP",
-            "X-Fern-SDK-Name" => "Seed",
-            "X-Fern-SDK-Version" => "0.0.1",
+            'X-Fern-Language' => 'PHP',
```
dcb6, line 32: To confirm, we'll use single quotes globally for strings?

dcb6 (approved):

seed/php-sdk/alias-extends/src/SeedClient.php
amckinney: Yes, that's the plan! I added a note to the summary - we'll only want to use double quotes when we need string interpolation, e.g. when the API key is specified in-line like so:

```php
$defaultHeaders = [
  'Authorization' => "Bearer $apiKey"
]
```

PR #4707, opened 2024-09-19
feat(php): Generate backwards compatible constructors

amckinney:
## Overview

This updates the generated object and in-lined request constructors to use an `array` instead of a flattened set of properties. 

This prevents a wide range of breaking changes (e.g. adding an optional parameter anywhere except the end of the constructor parameter list). This is especially important for in-lined request types since there isn't a global order between headers, query parameters, and request body properties.

With this, constructors now look like the following:

```php
/**
 * @param array{
 *   title: string,
 *   rating: float,
 * } $values
 */
public function __construct(
    array $values,
) {
    $this->title = $values['title'];
    $this->rating = $values['rating'];
}
```

## Implementation

* This is handled with the new `DataClass` AST node, which largely uses the same implementation as the `Class`. Behind the scenes, this creates a `TypeDict`, which is used as the constructor's single argument.

* This also adds a new generic `orderByAccess` function that can be used to order any node with an `Access` property (i.e. fields and methods).

### DIFF
--- a/packages/cli/ete-tests/src/tests/generate/__snapshots__/generate.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/generate/__snapshots__/generate.test.ts.snap
@@ -1,3 +1,7 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`fern generate > missing docs page 1`] = `"Failed to create organization: fern"`;
+exports[`fern generate > missing docs page 1`] = `
+"[docs]: Found 1 errors and 0 warnings. Run fern check --warnings to print out the warnings.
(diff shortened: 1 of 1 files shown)


generators/php/codegen/src/ast/Class.ts
```diff
@@ -6,6 +6,7 @@ import { Access } from "./Access";
 import { Field } from "./Field";
 import { Method } from "./Method";
 import { Comment } from "./Comment";
+import { orderByAccess } from "./utils/orderByAccess";
```
dcb6, line 9: fancy

dcb6 (approved):
🤙

PR #4708, opened 2024-09-20
fix(python): file upload requests respect OMIT as well

armandobelardo:
Fixes FER-3295

### DIFF
--- a/generators/python/src/fern_python/generators/sdk/client_generator/request_body_parameters/file_upload_request_body_parameters.py
+++ b/generators/python/src/fern_python/generators/sdk/client_generator/request_body_parameters/file_upload_request_body_parameters.py
@@ -4,8 +4,9 @@ import fern.ir.resources as ir_types
 
 from fern_python.codegen import AST
 from fern_python.external_dependencies.json import Json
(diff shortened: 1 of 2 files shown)


generators/python/src/fern_python/generators/sdk/client_generator/request_body_parameters/file_upload_request_body_parameters.py
```diff
@@ -25,13 +25,17 @@ def __init__(
...
+            type_hint = self._get_property_type(property)
             parameters.append(
                 AST.NamedFunctionParameter(
                     name=self._get_property_name(property),
-                    type_hint=self._get_property_type(property),
+                    type_hint=type_hint,
                     docs=self._get_docs(property),
                     raw_type=self._get_raw_property_type(property),
                     raw_name=self._get_raw_property_name(property),
+                    initializer=AST.Expression(DEFAULT_BODY_PARAMETER_VALUE)
```
armandobelardo, line 36: If not specified, we actually default optional params to None, which is no good here

dsinghvi (approved):

PR #4710, opened 2024-09-20
feat(php): update serde for backwards compatible constructors

dcb6:
## Author
In addition to what the title describes, this PR adds the following:
* Add serde handling for object type properties + test case (previously were only handling objects in arrays 😓 )
* Small improvements to error-handling / logging

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,38 @@ jobs:
           cli_path="$(pwd)/packages/cli/cli/dist/dev/bundle.cjs" 
           ./scripts/live-test.sh "$cli_path" "$FERN_ORG_TOKEN_DEV"
 
+  live-test-dev-windows:
+    environment: Fern Dev
+    if: github.ref == 'refs/heads/main'
+    runs-on: windows-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Install
+        uses: ./.github/actions/install
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+
+      - name: Check API definition is valid
+        env:
+          FORCE_COLOR: "2"
+          FERN_ORG_TOKEN_DEV: ${{ secrets.FERN_ORG_TOKEN_DEV }}
+          AUTH0_DOMAIN: ${{ secrets.AUTH0_DOMAIN }}
+          AUTH0_CLIENT_ID: ${{ secrets.AUTH0_CLIENT_ID }}
+        run: |
+          pnpm --filter @fern-api/cli dist:cli:dev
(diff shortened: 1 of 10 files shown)


seed/php-sdk/imdb/src/Core/JsonSerializer.php
```diff
@@ -60,11 +60,11 @@ private static function serializeValue(mixed $data, mixed $type): mixed
             foreach ($type->types as $unionType) {
                 try {
                     return self::serializeSingleValue($data, $unionType);
-                } catch (Exception $e) {
+                } catch (Exception) {
                     continue;
                 }
             }
-            throw new \InvalidArgumentException("Cannot serialize value with any of the union types.");
+            throw new JsonException("Cannot serialize value with any of the union types.");
```
amckinney, line 67: Nice change here 👍

seed/php-sdk/imdb/src/Core/Union.php
```diff
@@ -13,4 +13,8 @@ public function __construct(string ...$strings)
     {
         $this->types = $strings;
     }
+
+    public function __toString(): string {
+        return implode(' | ', $this->types);
+    }
```
amckinney, line 19: Nice 👏

seed/php-sdk/imdb/src/Core/Utils.php
```diff
@@ -39,4 +39,23 @@ public static function castKey(mixed $key, string $keyType): mixed
...
+    /**
+     * Returns a human-readable representation of the input's type.
+     *
+     * @param mixed $input The input value to determine the type of.
+     * @return string A readable description of the input type.
+     */
+    public static function getReadableType(mixed $input): string
+    {
+        if (is_object($input)) {
+            return 'object(' . get_class($input) . ')';
```
amckinney, line 52: I wonder if it's be better to just say the class, e.g. `GetUserResponse` vs. `object(GetUserResponse)`? Not actionable here, just thinking out loud.

seed/php-sdk/imdb/src/Core/Utils.php
```diff
@@ -39,4 +39,23 @@ public static function castKey(mixed $key, string $keyType): mixed
...
+        } elseif (is_array($input)) {
+            return 'array(' . count($input) . ' items)';
```
amckinney, line 54: Could we resolve the the of the array from the first element (if any), and include that in the string representation, too? That way it could read like: `array<string>`.

Again, not actionable here since it's probably non-trivial to handle arbitrarily nested `array` types. Just things for us to consider moving forward.

amckinney (approved):

seed/php-sdk/imdb/src/Core/Utils.php
dcb6: yeah agreed this would be great improvement 👍

seed/php-sdk/imdb/src/Core/Utils.php
dcb6: done!

PR #4705, opened 2024-09-19
feat(php): more serde methods + serde refactor

dcb6:
### Generated

This pull request introduces new classes for JSON encoding, decoding, serialization, and deserialization, along with associated tests.

- **JsonDecoder**: Decodes JSON strings into PHP values, handling various data types.
- **JsonDeserializer**: Deserializes JSON strings into PHP values, supporting arrays, maps, and single values.
- **JsonEncoder**: Encodes PHP values into JSON strings.
- **JsonSerializer**: Serializes PHP values into JSON strings, handling arrays, maps, and single values.
- **SerializableType**: Provides generic serialization and deserialization methods, now implementing `\JsonSerializable`.
- **Utils**: Contains utility functions for type checking and key casting.
- **ImdbClient**: Adds methods to create movies and movie lists, handling JSON serialization and deserialization.
- **MixedDateArrayTypeTest**: Tests handling of mixed date and string types in arrays.
- **NestedUnionArrayTypeTest**: Tests handling of nested arrays with union types.
- **NullPropertyTypeTest**: Tests serialization and deserialization of objects with null properties.

PR #4712, opened 2024-09-20
feat(php): Decode JSON response in endpoints

amckinney:
This updates the generated PHP methods to decode the response as JSON using the core utilities introduced in https://github.com/fern-api/fern/pull/4705.

With this, the generated code reads like the following:

```php
/**
 * @param string $movieId
 * @return Movie
 */
public function getMovie(string $movieId, ?array $options = null): Movie
{
    try {
            new JsonApiRequest(
                path: "/movies/$movieId",
                method: HttpMethod::GET,
            ),
        );
            $json = $response->getBody()->getContents();
            return Movie::fromJson($json);
        }
    } catch (JsonException $e) {
        throw new Exception("Failed to deserialize response", 0, $e);
        throw new Exception($e->getMessage());
    }
    throw new Exception("Error with status code " . $statusCode);
}
```

Note that this also includes a few other edits along the way:
1. Update the attribute mapper to use the idiomatic single quote string convention.
2. Update the `@returns` tag to `@return`.
3. Conditionally add the ` // @phpstan-ignore-line` comment whenever we need to decode an `array`.

### DIFF
--- a/generators/php/sdk/src/endpoint/http/HttpEndpointGenerator.ts
+++ b/generators/php/sdk/src/endpoint/http/HttpEndpointGenerator.ts
@@ -17,7 +17,7 @@ export declare namespace EndpointGenerator {
     }
 }
 
-const JSON_VARIALBE_NAME = "$json";
+const JSON_VARIABLE_NAME = "$json";
 const RESPONSE_VARIABLE_NAME = "$response";
 const STATUS_CODE_VARIABLE_NAME = "$statusCode";
 
@@ -148,7 +148,7 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
                     }
                     writer.writeNodeStatement(this.getResponseBodyContent());
                     if (return_.isOptional()) {
-                        writer.controlFlow("if", php.codeblock(`empty(${JSON_VARIALBE_NAME})`));
+                        writer.controlFlow("if", php.codeblock(`empty(${JSON_VARIABLE_NAME})`));
                         writer.writeTextStatement("return null");
                         writer.endControlFlow();
                     }
@@ -182,7 +182,7 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
         if (return_ == null) {
             return php.codeblock("");
         }
-        const arguments_: UnnamedArgument[] = [php.codeblock(JSON_VARIALBE_NAME)];
+        const arguments_: UnnamedArgument[] = [php.codeblock(JSON_VARIABLE_NAME)];
         const internalType = return_.underlyingType().internalType;
         switch (internalType.type) {
             case "reference":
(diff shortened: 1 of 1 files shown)


generators/php/codegen/src/asIs/JsonDecoder.Template.php
```diff
@@ -106,14 +106,25 @@ public static function decodeInt(string $json): int {
...
+    /**
+     * Decodes a JSON string and returns a mixed.
+     *
+     * @param string $json The JSON string to decode.
+     * @return mixed The decoded mixed.
+     * @throws JsonException If the decoded value is not an mixed.
+     */
+    public static function decodeMixed(string $json): mixed {
+        return self::decode($json);
+    }
```
dcb6, line 126: should we just replace the `decode` method?

generators/php/codegen/src/ast/MethodInvocation.ts
```diff
@@ -20,23 +22,29 @@ export class MethodInvocation extends AstNode {
...
         if (this.on != null) {
             this.on.write(writer);
-            writer.write("->");
+            writer.write(this.getMethodAccessor());
         }
         writer.write(this.method);
         writeArguments({ writer, arguments_: this.arguments_, multiline: this.multiline });
     }
+
+    private getMethodAccessor(): string {
```
dcb6, line 47: noice

generators/php/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -130,16 +137,23 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
-                        `return json_decode(${RESPONSE_VARIABLE_NAME}->getBody()->getContents(), true, 512, JSON_THROW_ON_ERROR)`
-                    );
+                    if (return_ == null) {
+                        writer.write("return;");
+                        writer.endControlFlow();
+                        return;
+                    }
+                    writer.writeNodeStatement(this.getResponseBodyContent());
+                    if (return_.isOptional()) {
+                        writer.controlFlow("if", php.codeblock(`empty(${JSON_VARIALBE_NAME})`));
```
dcb6, line 151: nit:
```suggestion
                        writer.controlFlow("if", php.codeblock(`empty(${JSON_VARIABLE_NAME})`));
```

dcb6 (approved):

generators/php/codegen/src/asIs/JsonDecoder.Template.php
amckinney: I actually kind of like having both so that everything else just reads as `decode`, but I could be convinced either way. If you'd rather just have `decodeMixed`, I'm OK with that

generators/php/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -163,4 +177,104 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+        return php.codeblock((writer) => {
+            writer.writeNode(
+                php.invokeMethod({
+                    on: this.context.getJsonDecoderClassReference(),
+                    method: "decodeArray",
+                    arguments_: [...arguments_, this.context.phpAttributeMapper.getArrayTypeAttributeArgument(type)],
+                    static_: true
+                })
+            );
+            writer.writeLine("; // @phpstan-ignore-line");
```
dcb6, line 252: 💯

dcb6 (approved):

PR #4674, opened 2024-09-17
fix(cli): support uploading files for docs deployments on windows

RohinBhargava:
This pull request introduces a new function, `convertToOsPath`, which converts a given path to the operating system's path format. This function is used to ensure that file paths are correctly formatted for the operating system in use.

The `convertToOsPath` function is added to the `fs-utils` package, and is used in the following files:
- `AbsoluteFilePath.ts`
- `RelativeFilePath.ts`
- `resolve.ts`

The function is also exported from `osPathConverter.ts` and used in `previewDocs.ts` and `publishDocs.ts`.

Screenshot 2024-09-17 at 5 27 37 PM
Screenshot 2024-09-17 at 5 27 50 PM

PR #4718, opened 2024-09-22
feat(php): Support serializing all request body types

amckinney:
This wraps up the bulk of the remaining work required for JSON endpoints, which includes the following changes:
1. Support serializing all request body types (including optional requests).
    * This also adds the `Ternary` node to `@fern-api/generator-commons`.
    * A separate type is required in C# because C#'s `AstNode` is extended, but C#'s `Ternary` simply delegates to the shared implementation.
3. Update the method comment to include a newline between the user-defined docs and the PHP doc tags.
4. Stop generating empty subpackage clients.

Note that this also updates the `php-sdk/seed.yml` file so that we run `composer analyze` and add the remaining fixtures to `allowedFailures`. Each added failure includes a comment explaining why the fixture is failing so that they're easy to follow up on.

### DIFF
--- a/packages/cli/ete-tests/src/tests/upgrade-generator/__snapshots__/upgrade-generator.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/upgrade-generator/__snapshots__/upgrade-generator.test.ts.snap
@@ -43,7 +43,7 @@ exports[`fern generator upgrade > fern generator upgrade list 1`] = `
               │                                                  │
               │                Generator Upgrades                │
               │                                                  │
-              │   Python SDK (Group: python-sdk) 3.0.0 → 4.2.5   │
+              │   Python SDK (Group: python-sdk) 3.0.0 → 4.2.6   │
               │                                                  │
               │                                                  │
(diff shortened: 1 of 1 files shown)


generators/csharp/codegen/src/project/CsharpProject.ts
```diff
@@ -609,7 +609,7 @@ ${this.getAdditionalItemGroups().join(`\n${FOUR_SPACES}`)}
         return result;
     }
 
-    private relativePathToWindowsPath(relativePath: RelativeFilePath): RelativeFilePath {
-        return RelativeFilePath.of(path.win32.normalize(relativePath));
+    private relativePathToWindowsPath(relativePath: RelativeFilePath): string {
+        return path.win32.normalize(relativePath);
```
amckinney, line 613: Note that this change was required due to the changes we introduced in https://github.com/fern-api/fern/pull/4674.

That PR is working just as intended - we just shouldn't have been using `RelativeFilePath` in this case.

dcb6 (approved):

PR #4719, opened 2024-09-23
feat(php): Generate base exceptions

amckinney:
This builds upon the endpoint changes in https://github.com/fern-api/fern/pull/4718 and generates the base exceptions in the `src/Exceptions` directory (which are included in every PHP SDK).

With this, endpoints now include the relevant `throws` tags, and all exceptions thrown from each endpoint method are either a `BaseException` or a `BaseApiException`. For example,

```php
/**
 * @param string $movieId
 * @param ?array{
 *   baseUrl?: string,
 * } $options
 * @return Movie
 * @throws SeedException
 * @throws SeedApiException
 */
public function getMovie(string $movieId, ?array $options = null): Movie
{
    try {
            new JsonApiRequest(
                path: "/movies/$movieId",
                method: HttpMethod::GET,
            ),
        );
            $json = $response->getBody()->getContents();
            return Movie::fromJson($json);
        }
    } catch (JsonException $e) {
        throw new SeedException(message: "Failed to deserialize response: {$e->getMessage()}", previous: $e);
        throw new SeedException(message: $e->getMessage(), previous: $e);
    }
    throw new SeedApiException(
        message: 'API request failed',
        statusCode: $statusCode,
        body: $response->getBody()->getContents(),
    );
}
```

A few things to note:
1. PHP's `Exception` type supports the `->getMessage()` and `->getCode()` methods by default. It's common to use the HTTP status code as the general exception code, so we don't need to support a separate `statusCode` field or getter.
    * The response body is supported with a `getBody` getter method instead a public property for consistency.
    * The `string` representation is customized with a `__toString` override that includes the status code and body.
2. The exceptions are deposited in the `Exceptions` directory, which is common amongst other SDKs (e.g. [Stripe](https://github.com/stripe/stripe-php/tree/master/lib/Exception), [Guzzle](https://github.com/guzzle/guzzle/tree/7.9/src/Exception), and [Square](https://github.com/square/square-php-sdk/tree/master/src/Exceptions)).
    * Note that we use the plural equivalent for consistency with the other top-level directories (e.g. `Types` and `Requests`).

### DIFF
--- a/seed/php-sdk/alias-extends/src/Exceptions/SeedApiException.php
+++ b/seed/php-sdk/alias-extends/src/Exceptions/SeedApiException.php
@@ -48,6 +48,6 @@ class SeedApiException extends SeedException
         if (empty($this->body)) {
             return "$this->message; Status Code: $this->code\n";
         }
-        return "$this->message; Status Code: $this->code; Body: " . json_encode($this->body) . "\n";
+        return "$this->message; Status Code: $this->code; Body: " . $this->body . "\n";
     }
 }
--- a/seed/php-sdk/basic-auth/src/Exceptions/SeedApiException.php
+++ b/seed/php-sdk/basic-auth/src/Exceptions/SeedApiException.php
@@ -48,6 +48,6 @@ class SeedApiException extends SeedException
         if (empty($this->body)) {
             return "$this->message; Status Code: $this->code\n";
         }
-        return "$this->message; Status Code: $this->code; Body: " . json_encode($this->body) . "\n";
+        return "$this->message; Status Code: $this->code; Body: " . $this->body . "\n";
     }
 }
--- a/generators/php/sdk/src/error/BaseApiExceptionGenerator.ts
+++ b/generators/php/sdk/src/error/BaseApiExceptionGenerator.ts
@@ -72,7 +72,7 @@ export class BaseApiExceptionGenerator extends FileGenerator<PhpFile, SdkCustomC
                 writer.writeTextStatement('return "$this->message; Status Code: $this->code\\n"');
                 writer.endControlFlow();
                 writer.writeTextStatement(
-                    'return "$this->message; Status Code: $this->code; Body: " . json_encode($this->body) . "\\n"'
+                    'return "$this->message; Status Code: $this->code; Body: " . $this->body . "\\n"'
                 );
             })
         });
--- a/seed/php-sdk/alias/src/Exceptions/SeedApiException.php
+++ b/seed/php-sdk/alias/src/Exceptions/SeedApiException.php
@@ -48,6 +48,6 @@ class SeedApiException extends SeedException
         if (empty($this->body)) {
             return "$this->message; Status Code: $this->code\n";
         }
-        return "$this->message; Status Code: $this->code; Body: " . json_encode($this->body) . "\n";
+        return "$this->message; Status Code: $this->code; Body: " . $this->body . "\n";
     }
 }
--- a/seed/php-sdk/any-auth/src/Exceptions/SeedApiException.php
+++ b/seed/php-sdk/any-auth/src/Exceptions/SeedApiException.php
@@ -48,6 +48,6 @@ class SeedApiException extends SeedException
         if (empty($this->body)) {
             return "$this->message; Status Code: $this->code\n";
         }
-        return "$this->message; Status Code: $this->code; Body: " . json_encode($this->body) . "\n";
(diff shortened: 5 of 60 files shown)


seed/php-sdk/alias-extends/src/Exceptions/SeedApiException.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @param string $message
+     * @param int $statusCode
+     * @param string $body
+     * @param ?Throwable $previous
+     */
+    public function __construct(
+        string $message,
+        int $statusCode,
+        string $body,
```
dcb6, line 26: in c# we made this an `object` (equivalent in php would be `mixed`) so that when we extend the `ApiException` to make the specific error types, we can make the body a specific class

i think we'll want to go ahead and do this here so that people aren't depending on the body being a string, although I haven't played around with this in PHP at all. happy to help with that, but wanted to check in with your thinking here first

seed/php-sdk/any-auth/src/Auth/AuthClient.php
```diff
@@ -51,11 +54,10 @@ public function getToken(GetTokenRequest $request, ?array $options = null): Toke
                 return TokenResponse::fromJson($json);
             }
         } catch (JsonException $e) {
-            throw new Exception("Failed to deserialize response", 0, $e);
+            throw new SeedException("Failed to deserialize response: {$e->getMessage()}");
```
dcb6, line 57: i think we should still the set the `previous` argument to `$e`?

seed/php-sdk/any-auth/src/Auth/AuthClient.php
```diff
@@ -51,11 +54,10 @@ public function getToken(GetTokenRequest $request, ?array $options = null): Toke
...
         } catch (ClientExceptionInterface $e) {
-            throw new Exception($e->getMessage());
+            throw new SeedException($e->getMessage());
```
dcb6, line 59: same here

seed/php-sdk/alias-extends/src/Exceptions/SeedApiException.php
amckinney: Great shout out - I didn't actually realize that's why we modeled this as `object` in C#. We should definitely do the same thing here, so let's go with `mixed`.

seed/php-sdk/basic-auth/src/Exceptions/SeedApiException.php
```diff
@@ -48,6 +48,6 @@ public function __toString(): string
         if (empty($this->body)) {
             return "$this->message; Status Code: $this->code\n";
         }
-        return "$this->message; Status Code: $this->code; Body: $this->body\n";
+        return "$this->message; Status Code: $this->code; Body: " . json_encode($this->body) . "\n";
```
dcb6, line 51: in this case, isn't the body already json-encoded, so this would double encode?
```
        throw new SeedApiException(
            message: 'API request failed',
            statusCode: $statusCode,
            body: $response->getBody()->getContents(),
        );
        ```
In this case, maybe we just want the default `toString()` of whatever the type is? and we'll eventually give our error bodies a good `toString()`?

Comment (amckinney) [reply seed/php-sdk/basic-auth/src/Exceptions/SeedApiException.php:51]: In most cases yes, the body will already be the `string` representation of a JSON object. However, if we type the body as `mixed` (as we should) it's simplest to call `json_encode` to support _all_ types at this point (e.g. other `Exception` classes, etc).

It doesn't look like there's a seamless way to map a `mixed` to a `string` representation, so `json_encode` felt like the most flexible approach, especially because it formats the type in a structured, human-readable way. In the case that the body is already a JSON object, the output will effectively be a no-op.

Thoughts?

Comment (dcb6) [reply seed/php-sdk/basic-auth/src/Exceptions/SeedApiException.php:51]: >  In the case that the body is already a JSON object, the output will effectively be a no-op
I think if we're passing in a string like `$response->getBody()->getContents()`, it's going to re-encode that (ie, add extra quotes around it)

I think our two cases for the type of mixed are `string` (direct json from the response) or some `object`, where the object will implement `toString()`. In that case, according to chatgpt, we should be safe to just do 
```
        return "$this->message; Status Code: $this->code; Body: " . $this->body . "\n";
```

let me know if i'm missing something!

CleanShot 2024-09-23 at 13 59 42

amckinney: Ah good point on the double quoting. I like the simple concatenation approach - let's go with that!

dcb6 (approved):

PR #4735, opened 2024-09-24
chore(python): make all base models (except universal ones) respect pydantic version flag

armandobelardo:
This PR does 2 main things:
- Makes the generator respect the given Pydantic version in every portion of generated code, but not the as-is files
- Makes the generator respect `exclude_unset`, a config option for `.dict` on models

### DIFF
--- a/generators/python/core_utilities/shared/pydantic_utilities.py
+++ b/generators/python/core_utilities/shared/pydantic_utilities.py
@@ -160,9 +160,7 @@ class UniversalBaseModel(pydantic.BaseModel):
                     # If the default values are non-null act like they've been set
                     # This effectively allows exclude_unset to work like exclude_none where
                     # the latter passes through intentionally set none values.
-                    if default is not None or (
-                        "exclude_unset" in kwargs and not kwargs["exclude_unset"]
-                    ):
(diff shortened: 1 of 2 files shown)


generators/python/fastapi/versions.yml
```diff
@@ -1,4 +1,16 @@
...
+  irVersion: 53
+  changelogEntry:
+    - type: fix
+      summary: |
+        Use of `IS_PYDANTIC_V2` flag is now only used within universal base models (`pydantic_utilities.py`) and if pydantic version compatibility is not specified
+        or set to `Both`. Concretely, this means that if your configuration specifies a version for Pydantic, then your generated code will respect that version, and
+        generate the code for that specific version of Pydantic, as opposed to always generating code compatible with both versions of Pydantic.
+      fixes:
+        - |
+          `exclude_unset` is now respected when passed in as an option to a model's `.dict` call.
```
dsinghvi, line 12: do we need to release new versions of pydantic and sdk as well?

dsinghvi: i already asked this; but based on the code it looks like we would not send explicit Nones for all SDKs ?

dsinghvi (approved):

PR #4741, opened 2024-09-24
feat(php): use strings for request enums + other small fixes

dcb6:
changelog:
```yml
    - type: feat
      summary: >-
        Represent enums in objects as strings.
    - type: fix
      summary: >-
        Generated wrapped requests now implement `SerializableType`.
        Represent enums in objects as strings.
    - type: fix
      summary: >-
        Fix a bug where we don't set the request options baseurl properly.
    - type: fix
      summary: >-
        Fix bugs in our numeric type serde and add tests.
```

also:
* some initial support for unions, which is leftover from a previous enum handling implementation

### DIFF
--- a/seed/php-sdk/enum/src/QueryParam/QueryParamClient.php
+++ b/seed/php-sdk/enum/src/QueryParam/QueryParamClient.php
@@ -38,10 +38,10 @@ class QueryParamClient
     public function send(SendEnumAsQueryParamRequest $request, ?array $options = null): void
     {
         $query = [];
-        $query['operand'] = $request->operand->value;
+        $query['operand'] = $request->operand;
         $query['operandOrColor'] = $request->operandOrColor;
         if ($request->maybeOperand != null) {
-            $query['maybeOperand'] = $request->maybeOperand->value;
+            $query['maybeOperand'] = $request->maybeOperand;
         }
         if ($request->maybeOperandOrColor != null) {
             $query['maybeOperandOrColor'] = $request->maybeOperandOrColor;
@@ -80,10 +80,10 @@ class QueryParamClient
     public function sendList(SendEnumListAsQueryParamRequest $request, ?array $options = null): void
     {
         $query = [];
-        $query['operand'] = $request->operand->value;
+        $query['operand'] = $request->operand;
         $query['operandOrColor'] = $request->operandOrColor;
         if ($request->maybeOperand != null) {
-            $query['maybeOperand'] = $request->maybeOperand->value;
+            $query['maybeOperand'] = $request->maybeOperand;
         }
         if ($request->maybeOperandOrColor != null) {
             $query['maybeOperandOrColor'] = $request->maybeOperandOrColor;


generators/php/sdk/versions.yml
```diff
@@ -1,11 +1,26 @@
+- version: 0.1.2
+  changelogEntry:
+    - type: feat
+      summary: >-
+        Represent enums in objects as strings.
+    - type: fix
+      summary: >-
+        Generated wrapped requests now implement `SerializableType`.
+        Represent enums in objects as strings.
```
amckinney, line 9: Let's remove this (I know you meant to based on the entry above).
```suggestion
        Generated wrapped requests now implement `SerializableType`.
```

generators/php/codegen/src/project/PhpProject.ts
```diff
@@ -238,7 +238,7 @@ class ComposerJson {
         return `
 {
   "name": "${composerProjectName}",
-  "version": ${this.context.version != null ? `"${this.context.version}"` : "0.0.0"},
+  "version": ${this.context.version != null ? `"${this.context.version}"` : '"0.0.0"'},
```
amckinney, line 241: Nice 👏

generators/php/codegen/src/context/PhpAttributeMapper.ts
```diff
@@ -103,6 +111,8 @@ export class PhpAttributeMapper {
                     writer.write("::class");
                 });
             }
+            case "enumString":
+                return php.codeblock('"string"');
```
amckinney, line 115: Should we prefer single quotes here?
```suggestion
                return php.codeblock("'string'");
```

generators/php/codegen/src/ast/Enum.ts
```diff
@@ -48,8 +56,16 @@ export class Enum extends AstNode {
         if (this.backing != null) {
             writer.write(` : ${this.backing}`);
         }
+        if (this.serializable) {
+            writer.addReference(
+                new ClassReference({
+                    name: "JsonSerializable",
+                    namespace: ""
+                })
```
amckinney, line 64: nit: We should make this a function on the abstract context, so we can just read:
```
writer.addReference(this.context.getJsonSerializableClassReference())
```

Check out [getJsonExceptionClassReference](https://github.com/fern-api/fern/blob/6d3572ac09858e9883b96fca458463347083a4bc/generators/php/sdk/src/SdkGeneratorContext.ts#L116) for an example.

seed/php-sdk/alias-extends/src/SeedClient.php
```diff
@@ -67,7 +67,7 @@ public function extendedInlineRequestBody(InlinedChildRequest $request, ?array $
         try {
             $response = $this->client->sendRequest(
                 new JsonApiRequest(
-                    baseUrl: $this->options['baseUrl'] ?? $this->client->options['baseUrl'] ?? '',
```
amckinney, line 70: ty 👏

generators/php/codegen/src/context/PhpTypeMapper.ts
```diff
@@ -17,6 +17,11 @@ import { AbstractPhpGeneratorContext } from "./AbstractPhpGeneratorContext";
 export declare namespace PhpTypeMapper {
     interface Args {
         reference: TypeReference;
+        /*
+         * By default, we represent enums as strings, with a phpstan phpdoc referencing the generated enum. If this flag is
+         * is true, then we reference the enum type directly.
+         */
+        preserveEnums?: boolean;
```
amckinney, line 24: Is this still used? I don't see anywhere that sets it to `true`.

amckinney (approved):
Nice, only a few nits + suggestions!

generators/php/sdk/versions.yml
dcb6: oop thanks!

generators/php/codegen/src/ast/Enum.ts
dcb6: we don't have `context` in Ast classes though right?

generators/php/codegen/src/context/PhpTypeMapper.ts
dcb6: nope not anymore! ok if we leave it around in case, or you want it gone?

PR #4745, opened 2024-09-25
chore(openapi): upgrade from ir v23 -> v53

dsinghvi:
This PR upgrades the OpenAPI generator to IR v53 so that we can start improving the OpenAPI generator.

### DIFF
--- a/generators/openapi/src/converters/typeConverter.ts
+++ b/generators/openapi/src/converters/typeConverter.ts
@@ -10,6 +10,7 @@ import {
     IntermediateRepresentation,
     PrimitiveType,
     PrimitiveTypeV1,
+    PrimitiveTypeV2,
     SingleUnionTypeProperties,
     Type,
     TypeDeclaration,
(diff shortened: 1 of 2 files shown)


seed/openapi/streaming/openapi.yml
```diff
@@ -9,7 +9,9 @@ paths:
       tags:
         - Dummy
       parameters: []
-      responses: {}
```
dsinghvi, line 12: responses: {} is invalid

generators/openapi/versions.yml
```diff
@@ -1,3 +1,13 @@
+- version: 0.1.0
+  createdAt: '2024-03-24'
```
armandobelardo, line 2: Nit: could we remove or update this

generators/openapi/versions.yml
```diff
@@ -1,3 +1,13 @@
...
+  changelogEntry:
+    - type: chore
```
armandobelardo, line 4: Thoughts on `internal` as the type here?

seed/openapi/query-parameters/openapi.yml
```diff
@@ -40,16 +40,16 @@ paths:
             format: byte
         - name: user
           in: query
-          required: false
+          required: true
           schema:
-            type: string
-            nullable: true
+            $ref: '#/components/schemas/User'
```
dsinghvi, line 45: we now support objects as query params

seed/openapi/query-parameters/openapi.yml
```diff
@@ -40,16 +40,16 @@ paths:
             format: byte
         - name: user
           in: query
-          required: false
+          required: true
```
dsinghvi, line 43: we now respect required query params

seed/openapi/plain-text/openapi.yml
```diff
@@ -10,11 +10,8 @@ paths:
         - Service
       parameters: []
       responses:
-        '200':
+        '204':
```
dsinghvi, line 13: we dont handle text responses so this is a followup

generators/openapi/src/security.ts
armandobelardo, line 1: Nit: do we want to add the whole `flows` block with oauth into the spec?

generators/openapi/src/converters/typeConverter.ts
```diff
@@ -38,7 +39,7 @@ export function convertType(typeDeclaration: TypeDeclaration, ir: IntermediateRe
             return convertEnum({ enumTypeDeclaration, docs });
         },
         object: (objectTypeDeclaration) => {
-            const exampleType: ExampleType | undefined = typeDeclaration.examples[0];
+            const exampleType: ExampleType | undefined = typeDeclaration.userProvidedExamples[0];
```
armandobelardo, line 42: Should we default to generated if no user provided?

generators/openapi/src/converters/typeConverter.ts
```diff
@@ -312,7 +342,7 @@ function convertContainerType(containerType: ContainerType): OpenApiComponentSch
         map: (mapType) => {
             if (
                 mapType.keyType.type === "primitive" &&
-                mapType.keyType.primitive === "STRING" &&
+                mapType.keyType.primitive.v1 === "STRING" &&
```
armandobelardo, line 345: Do you need to check the v2 type as well?

generators/openapi/src/converters/servicesConverter.ts
```diff
@@ -375,11 +391,15 @@ function convertResponse({
...
 
         responseByStatusCode["200"] = {
-            description: httpResponse.docs ?? "",
+            description: httpResponse.body.value.docs ?? "",
             content: {
                 "application/json": convertedResponse
             }
         };
+    } else {
+        responseByStatusCode["204"] = {
```
armandobelardo, line 400: What's this about? (e.g. why is 204 the fallback)

armandobelardo (approved):

PR #4747, opened 2024-09-25
fix(openapi): support explicit null semantics on openapi config

dsinghvi:
The `customOverrides` configuration flag now supports explict nulls as a way to 
delete keys in the generated OpenAPI spec. For example, if you want to 
exclude the `openapi.info` block in the generated spec, then you can do something 
along the lines: 
```yml generators.yml
- name: fernapi/fern-openapi
  version: 0.1.1
  config: 
    customOverrides: 
      info: null # excludes the info field
```

### DIFF
--- a/packages/commons/core-utils/src/mergeWithOverrides.ts
+++ b/packages/commons/core-utils/src/mergeWithOverrides.ts
@@ -33,10 +33,8 @@ export const omitDeepBy: OmitDeepBy = (object: unknown, cb: any): any => {
 
         if (isPlainObject(object)) {
             const temp: Record<string, unknown> = {};
-            for (const [key, value] of Object.entries<{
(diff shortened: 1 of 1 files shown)


packages/commons/core-utils/src/mergeWithOverrides.ts
```diff
@@ -0,0 +1,47 @@
...
+                ? // nested arrays of objects are merged
+                  undefined
+                : // nested arrays of primitives are replaced
+                  [...src]
+            : undefined
+    ) as T;
+    // Remove any nullified values
+    const filtered = omitDeepBy(merged, isNull) as T;
+    return filtered;
+}
```
amckinney, line 17: Nice 👏

amckinney (approved):

PR #4750, opened 2024-09-25
fix(openapi): handle when custom config for filename is optional

dsinghvi:
The generator now handles `config.filename` being null so that the 
        configuration for the generator can continue to be backwards compatible.


generators/openapi/src/customConfig.ts
```diff
@@ -3,7 +3,7 @@ import { GeneratorConfig } from "@fern-api/generator-commons";
 export interface FernOpenapiCustomConfig {
     format: "yaml" | "json";
     customOverrides: Record<string, unknown>;
-    filename: string;
+    filename?: string;
```
dsinghvi, line 6: key change

PR #4765, opened 2024-09-26
feat(java): allow streaming byte requests

dcb6:

### DIFF
--- a/generators/java/sdk/src/main/java/com/fern/java/client/generators/FileStreamGenerator.java
+++ b/generators/java/sdk/src/main/java/com/fern/java/client/generators/FileStreamGenerator.java
@@ -26,15 +26,12 @@ import java.nio.charset.StandardCharsets;
 public final class FileStreamGenerator extends AbstractFileGenerator {
 
     public FileStreamGenerator(ClientGeneratorContext clientGeneratorContext) {
-        super(
-                clientGeneratorContext.getPoetClassNameFactory().getFileStreamClassName(),
-                clientGeneratorContext);
+        super(clientGeneratorContext.getPoetClassNameFactory().getFileStreamClassName(), clientGeneratorContext);
     }
 
     @Override
     public GeneratedResourcesJavaFile generateFile() {
-        try (InputStream is =
-                FileStreamGenerator.class.getResourceAsStream("/FileStream.java")) {
+        try (InputStream is = FileStreamGenerator.class.getResourceAsStream("/FileStream.java")) {
             String contents = new String(is.readAllBytes(), StandardCharsets.UTF_8);
             return GeneratedResourcesJavaFile.builder()
                     .className(className)


seed/java-sdk/bytes/src/main/java/com/seed/bytes/resources/service/ServiceClient.java
```diff
@@ -58,4 +61,9 @@ public void upload(byte[] request, RequestOptions requestOptions) {
             throw new SeedBytesException("Network error executing HTTP request", e);
         }
     }
+
+    // Overload for backward compatibility
+    public void upload(byte[] request, RequestOptions requestOptions) {
```
dcb6, line 66: thoughts on moving the direct bytes request to this?

Claude gave me these downsides but to me seems worth the cleanliness

> Potential downsides of the upload overload:

Memory Usage:

The original implementation used the byte array directly.
The new overload wraps the byte array in a ByteArrayInputStream, which is then wrapped in a FileStream.
While ByteArrayInputStream is generally efficient, it does create an additional object in memory.


Performance Overhead:

There's a slight performance cost in creating the additional objects (ByteArrayInputStream and FileStream).
The streaming approach might introduce a small amount of overhead compared to sending the byte array directly, especially for small files.

seed/java-sdk/bytes/src/main/java/com/seed/bytes/core/FileStream.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.InputStream;
+import java.util.Objects;
+import okhttp3.MediaType;
+import okhttp3.RequestBody;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Represents a file stream with associated metadata for file uploads.
+ */
+public class FileStream {
```
dcb6, line 12: we don't actually need this for bytes requests, going to delete unless anyone votes I keep it around

seed/java-sdk/bytes/src/main/java/com/seed/bytes/resources/service/ServiceClient.java
dsinghvi: great idea

dsinghvi (approved):

seed/java-sdk/bytes/src/main/java/com/seed/bytes/core/FileStream.java
dsinghvi: awesome, may be useful to keep around to make it better for multipart (just as a uncopied utility)

generators/java/sdk/versions.yml
```diff
@@ -1,10 +1,19 @@
...
+  version: 2.3.0
+
 - changelogEntry:
     - summary: |
         Bump Jackson version to latest (2.17.2)
       type: chore
   createdAt: '2024-09-26'
   irVersion: 46
-  version: 1.2.0
+  version: 2.2.0
```
dcb6, line 16: this was a blunder 😅 

thankfully, we never previously released a `1.2.0` and niels won't be using that until next week
I deleted this erroneous `1.2.0` release from docker hub

PR #4776, opened 2024-09-30
fix(docs): ship up markdown files that contain components

dsinghvi:
If the files that contain components are md files, then we should still ship them up as part of `pages`.

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,5 +1,13 @@
 - changelogEntry:
     - summary: |
+        Any markdown files that have custom components are also pushed up to the Fern Docs 
+        platform.
+      type: fix
+  irVersion: 53
+  version: 0.43.8
+
+- changelogEntry:
+    - summary: |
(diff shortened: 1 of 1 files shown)


RohinBhargava (approved):

packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
```diff
@@ -182,6 +182,25 @@ export class DocsDefinitionResolver {
...
+                                    markdown: await readFile(AbsoluteFilePath.of(file), "utf-8"),
+                                    editThisPageUrl: createEditThisPageUrl(this.editThisPage, relativeFilePath)
+                                };
+                            })
+                        );
+
+                        files.forEach((file) => {
+                            jsFilePaths.add(file);
+                        });
                     } else if (absoluteFilePath.match(/\.(js|ts|jsx|tsx)$/) != null) {
```
abvthecity, line 204: Missing direct file match

abvthecity (approved):

PR #4783, opened 2024-09-30
feat(php): undiscriminated unions

dcb6:
This PR introduces support in PHP for undiscriminated unions, including the following:
- serde for undiscriminated union type properties, with types indicated by a Union annotation (the previous Union class was easily converted into an annotation)
- adding a test for serde of union type properties
- serde for responses unions
- serde for requests unions
- handling union type headers and query parameters


generators/php/codegen/src/asIs/Union.Template.php
```diff
@@ -2,20 +2,61 @@
...
+ * Union type attribute for flexible type declarations.
+ *
+ * This class is used to define a union of multiple types for a property.
+ * It allows a property to accept one of several types, including strings, arrays, or other Union types.
+ * This class supports complex types, nested unions, and arrays, providing a flexible mechanism for property type validation and serialization.
+ *
+ * Example:
+ *
+ * ```php
+ * #[Union('string', 'int', 'null', new Union('boolean', 'float'))]
```
amckinney, line 17: Is this equivalent to the following?

```suggestion
 * #[Union('string', 'int', 'null', 'boolean', 'float')]
```

Is the format you've included here used whenever there is an undiscriminated union that references another undiscriminated union?

The main reason I ask is for the docstring included here, we might want to use the simpler, more common representation.

generators/php/codegen/src/ast/Type.ts
```diff
@@ -176,14 +176,30 @@ export class Type extends AstNode {
...
+                    }
                     if (index > 0) {
                         writer.write("|");
                     }
-                    type.write(writer);
-                });
+                    type.write(writer, { comment });
+                    index++;
+                }
                 break;
```
amckinney, line 201: nit: I wonder if it's simpler to split out a function that retrieves the unique set of type strings, then the implementation here remains as simple as possible. Handling the set amidst the writing is otherwise a little hard to keep track of.

generators/php/codegen/src/context/PhpAttributeMapper.ts
```diff
@@ -41,14 +45,41 @@ export class PhpAttributeMapper {
...
+                        arguments: this.getUnionTypeParameters(underlyingInternalType.types)
+                    })
+                );
+            }
+        }
         return attributes;
     }
 
-    public getArrayTypeAttributeArgument(type: php.Type): php.AstNode {
+    public getUnionTypeClassRepresentation(arguments_: php.AstNode[]): ClassInstantiation {
```
amckinney, line 67: nit: 
```suggestion
    public getUnionTypeClassInstantiation(arguments_: php.AstNode[]): ClassInstantiation {
```

Ideally the argument could be a little more strongly typed (e.g. `php.Type`). Otherwise it's possible to write in arbitrary `php.CodeBlock` here. But I know it's difficult because of the `getTypeAttributeArgument` helper function below - the attributes aren't the same as the type representation (e.g. `int` vs `integer`).

Fine with this for now, but we'll need to be careful to call this with arguments that make sense.

generators/php/codegen/src/context/PhpAttributeMapper.ts
```diff
@@ -92,17 +123,20 @@ export class PhpAttributeMapper {
...
             }
             case "union": {
-                return php.instantiateClass({
-                    classReference: this.context.getUnionClassReference(),
-                    arguments_: type.internalType.types.map((unionType) =>
-                        this.getArrayTypeAttributeArgument(unionType)
-                    )
-                });
+                const unionTypeParameters = this.getUnionTypeParameters(type.internalType.types);
+                // dedupe in getUnionTypeParameters could result in a single value
```
amckinney, line 127: nit: I'd remove this comment since it's already fairly clear from the code. This might apply elsewhere.

generators/php/codegen/src/context/PhpTypeMapper.ts
```diff
@@ -118,7 +119,15 @@ export class PhpTypeMapper {
             case "union":
                 return php.Type.mixed();
             case "undiscriminatedUnion": {
-                return php.Type.mixed();
+                return php.Type.union(
+                    // need to dedupe because lists and sets are both represented as array,
```
amckinney, line 123:
```suggestion
                    // need to dedupe because lists and sets are both represented as array
```

amckinney (approved):
Nice! Only nits 👏

generators/php/codegen/src/asIs/Union.Template.php
dcb6: as discussed offline, leaving this in for comprehensive testing purposes, but hopefully will not be seen in the wild

generators/php/codegen/src/context/PhpAttributeMapper.ts
dcb6: done!

generators/php/codegen/src/ast/Type.ts
dcb6: good call, made the change!

PR #4785, opened 2024-10-01
feat(csharp): XML encode contents of <summary> and <example><code> code comments

Swimburger:
When `<` and `>` are used inside of summary or code example XML docs, the XML doc is invalid.
This PR HTML encodes the contents before putting it in the XML doc.

~~This will also XML encode other characters which aren't really necessary, like `"`, but IDEs continue to show the XML docs correctly.~~


seed/csharp-model/multi-line-docs/src/SeedMultiLineDocs/User/User.cs
```diff
@@ -11,7 +11,7 @@ public record User
     public required string Id { get; set; }
 
     /// <summary>
-    /// The user's name. This name is unique to each user. A few examples are included below:
+    /// The user&#39;s name. This name is unique to each user. A few examples are included below:
```
dcb6, line 14: it's looking like lodash's escape might be a little overkill for us -- from my understanding, lodash is going to escape the following chars: `& < > " '`

But I think that we only need carats to be escaped

Can we roll our own method only escapes carats?

Swimburger: Yes, I was on the fence about that. Let's do it ourselves.

Swimburger: @dcb6 The code has been updated with our own escaping.

generators/csharp/codegen/src/ast/Method.ts
```diff
@@ -90,23 +90,16 @@ export class Method extends AstNode {
...
-            writer.writeLine("/// <example>");
-            writer.writeLine("/// <code>");
-            this.codeExample.split("\n").forEach((line) => {
-                if (line !== "") {
-                    writer.writeLine(`/// ${line}`);
-                }
-            });
-            writer.writeLine("/// </code>");
-            writer.writeLine("/// </example>");
+            writer
```
dcb6, line 96: The `DocXmlWriter ` is awesome! 

and overall i like the ideas that you're bringing here, such as this fluent writer API, but i'd prefer consistency with our existing writer. I'd also like to avoid adding new `write` methods to our existing writer. So, I have a couple changes to propose:

1. remove the `writeDocXml` method from the `Writer` class and instead just initialize a `DocXmlWriter` here (and elsewhere)
2. remove the fluent elements of the `DocXmlWriter` by simply have the methods return `void`, and changing the usages accordingly

I ack this is fairly opinionated, but I think we consider it more important to adhere to a consistent style than to introduce new patterns that bring fairly slight improvements

dcb6:
This part of the description is no longer accurate, correct?
> This will also XML encode other characters which aren't really necessary, like ", but IDEs continue to show the XML docs correctly.

Swimburger:
> This part of the description is no longer accurate, correct?
> 
> > This will also XML encode other characters which aren't really necessary, like ", but IDEs continue to show the XML docs correctly.

Correct!

generators/csharp/codegen/src/ast/Method.ts
Swimburger: Done!

amckinney (approved):
Nice, this is slick

PR #4787, opened 2024-10-01
fix(typescript): file arrays call appendFile

dsinghvi:
Fix: File array uploads now call `request.appendFile` instead of `request.append` which  was causing form data to be in a corrupted state.


RohinBhargava (approved):
Looks great to me; are there any other containers that might need this as well?

PR #4788, opened 2024-10-01
feat(seed): introduce scripts to generators so fern bot can build the SDKs

armandobelardo:
This upgrades the generators sdk dep to expose generator scripts from FDR

Then this exposes a similar configuration from seed config to have authors specify scripts within each generator's `seed.yml` that is then run within the generated artifacts resultant repo

The idea here is that within CI on fern config repos we:
- Create a check per-generator
- That check runs `fern generate --preview` for the specific generator
- On the resultant preview repo, we run the checks outlined here (within the corresponding `seed.yml`)
- We complete the check with the appropriate status (attaching logs)
    - We link to a preview of the SDK change in Github within the check as well


seed/go-sdk/seed.yml
```diff
@@ -2,6 +2,18 @@ irVersion: v40
 displayName: Go SDK
 image: fernapi/fern-go-sdk
 changelogLocation: ../../generators/go/sdk/versions.yml
+
+artifactMetadata:
+  preInstallScript:
+    - sudo yum -y install go
+    - mkdir /tmp/go
+    - echo "export GOPATH=\$HOME/go" >> ~/.bash_profile
+    - source ~/.bash_profile
```
amckinney, line 11: I don't think we'll need to explicitly configure the `GOPATH` here - we're just running the `go build` command to verify the program compiles.

I would try to simplify Go's `preInstallScript` to just:
```yaml
artifactMetadata:
  preInstallScript:
    - sudo yum -y install go
```

If for some reason we actually need to configure a `GOPATH`, I would just export the envrionment variable directly rather than writing to `~/.bash_profile` and `source`'ing from there:

```yaml
artifactMetadata:
  preInstallScript:
    - sudo yum -y install go
    - export GOPATH=\$HOME/go
```

seed/csharp-sdk/seed.yml
```diff
@@ -2,20 +2,29 @@ displayName: C# SDK
 irVersion: v53
 image: fernapi/fern-csharp-sdk
 changelogLocation: ../../generators/csharp/sdk/versions.yml
+
+artifactMetadata:
```
dsinghvi, line 6: could we rename this to `buildScripts` ? think it communicates the meaning a bit more

seed/go-sdk/seed.yml
armandobelardo: Updated!

seed/csharp-sdk/seed.yml
armandobelardo: most definitely!

dsinghvi (approved):

PR #4790, opened 2024-10-02
feat(cli): init the conjure importer

dsinghvi:
This PR configures and sets up the conjure importer.

### DIFF
--- a/packages/cli/api-importers/commons/vitest.config.ts
+++ b/packages/cli/api-importers/commons/vitest.config.ts
@@ -1,1 +1,1 @@
-export { default } from "../../../../../shared/vitest.config";
+export { default } from "../../../../shared/vitest.config";


packages/cli/api-importers/commons/src/APIDefinitionImporter.ts
```diff
@@ -0,0 +1,17 @@
+import { DefinitionFileSchema, PackageMarkerFileSchema, RootApiFileSchema } from "@fern-api/fern-definition-schema";
+import { RelativeFilePath } from "@fern-api/fs-utils";
+
+export declare namespace APIDefinitionImporter {
+
+    interface Return {
```
amckinney, line 6: nit: `Result`?

packages/cli/api-importers/commons/src/FernDefnitionBuilder.ts
```diff
@@ -92,28 +92,25 @@ export class FernDefinitionBuilderImpl implements FernDefinitionBuilder {
...
         this.root = new FernDefinitionDirectory();
         this.rootApiFile = {
             name: "api",
             "error-discrimination": {
                 strategy: "status-code"
             }
         };
-        if (ir.title != null) {
-            this.rootApiFile["display-name"] = ir.title;
-        }
```
amckinney, line 109: Is this intended? Why not include `display-name` anymore? 

It looks like we have the new `setAPIDisplayName`, but that sets the root API file's `name`, not `display-name`.

packages/cli/api-importers/conjure/conjure-sdk/fern/definition/types.yml
```diff
@@ -0,0 +1,62 @@
...
+      docs: optional<string>
+    
+  ConjurePrimitive: 
+    enum: 
+      - rid
+      - string
+      - safelong
+      - integer
+      - double
+      - boolean
```
amckinney, line 62: Should we use the `PrimitiveV2` approach here? We might otherwise pay for this later.

packages/cli/api-importers/commons/src/APIDefinitionImporter.ts
dsinghvi: will follow up with this

packages/cli/api-importers/conjure/conjure-sdk/fern/definition/types.yml
dsinghvi: discussed offline; have to match conjure spec

PR #4792, opened 2024-10-02
feat(php): inheritance

dcb6:
This PR enables inheritance in PHP for types and inlined requests, which is achieved via PHP traits.

### DIFF
--- a/generators/php/codegen/src/AsIs.ts
+++ b/generators/php/codegen/src/AsIs.ts
@@ -25,6 +25,7 @@ export enum AsIsFiles {
     DateArrayTypeTest = "Json/DateArrayTypeTest.Template.php",
     EmptyArraysTest = "Json/EmptyArraysTest.Template.php",
     EnumTest = "Json/EnumTest.Template.php",
+    TraitTest = "Json/TraitTest.Template.php",
     InvalidTypesTest = "Json/InvalidTypesTest.Template.php",
     MixedDateArrayTypeTest = "Json/MixedDateArrayTypeTest.Template.php",
     NestedUnionArrayTypeTest = "Json/NestedUnionArrayTypeTest.Template.php",
--- a/generators/php/codegen/src/asIs/Json/TraitTest.Template.php
+++ b/generators/php/codegen/src/asIs/Json/TraitTest.Template.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace <%= namespace%>;
+
+use <%= coreNamespace%>\Json\SerializableType;
+use <%= coreNamespace%>\Json\JsonProperty;
+use PHPUnit\Framework\TestCase;
+
+trait IntegerPropertyTrait
+{
+    /**
+     * @var int $integerProperty
+     */
+    #[JsonProperty('integer_property')]
+    public int $integerProperty;
+}
+
+class TypeWithTrait extends SerializableType
+{
+    use IntegerPropertyTrait;
+
+    /**
+     * @var string $stringProperty
+     */
+    #[JsonProperty('string_property')]
+    public string $stringProperty;
+
+    /**
+     * @param array{
+     *   integerProperty: int,
+     *   stringProperty: string,
+     * } $values
+     */
+    public function __construct(array $values)
+    {
+        $this->integerProperty = $values['integerProperty'];
+        $this->stringProperty = $values['stringProperty'];
+    }
+}
+
+class TraitTest extends TestCase
+{
+    public function testTraitPropertyAndString(): void
+    {
+        $data = [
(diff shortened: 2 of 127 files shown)


generators/php/model/src/generateTraits.ts
```diff
@@ -0,0 +1,34 @@
+import { ModelGeneratorContext } from "./ModelGeneratorContext";
+import { TraitGenerator } from "./trait/TraitGenerator";
+
+export function generateTraits(context: ModelGeneratorContext): void {
+    const extendedTypeIdsFromTypes: Set<string> = new Set(
```
amckinney, line 5: Easy lint fix here 
```suggestion
    const extendedTypeIdsFromTypes = new Set<string>(
```

generators/php/model/src/generateTraits.ts
```diff
@@ -0,0 +1,34 @@
...
+            return endpoint.requestBody?.type === "inlinedRequestBody"
+                ? endpoint.requestBody.extends.map((declaredTypeName) => declaredTypeName.typeId)
+                : [];
+        });
+    for (const typeId of [...extendedTypeIdsFromTypes, ...extendedTypeIdsFromInlinedRequests]) {
+        const typeDeclaration = context.getTypeDeclarationOrThrow(typeId);
+        const objectTypeDeclaration = context.getUnderlyingObjectTypeDeclarationFromTypeDeclaration(typeDeclaration);
+        if (objectTypeDeclaration == null) {
+            throw new Error("Unexpected no object type declaration");
+        }
```
amckinney, line 30: nit: We could roll out a separate `getUnderlyingObjectTypeDeclarationOrThrow(typeId)` function to handle this block so the call-site reads a little simpler.

generators/php/codegen/src/ast/Class.ts
```diff
@@ -20,6 +22,8 @@ export declare namespace Class {
         docs?: string;
         /* The class to inherit from if any */
         parentClassReference?: AstNode;
+        /* The traits that this class uses, if any */
+        usedTraits?: ClassReference[];
```
amckinney, line 26: nit: I would update this to just be `traits`, then the helper is just `addTrait`. We can assume it's used.

generators/php/codegen/src/ast/Trait.ts
```diff
@@ -0,0 +1,101 @@
...
+export declare namespace Trait {
+    interface Args {
+        /* The name of the PHP trait */
+        name: string;
+        /* The namespace of the PHP trait */
+        namespace: string;
+        /* Docs associated with the trait */
+        docs?: string;
+        /* The traits that this trait uses, if any */
+        usedTraits?: ClassReference[];
```
amckinney, line 18: Same thing with `usedTraits` -> `traits` here.

generators/php/model/src/object/ObjectGenerator.ts
```diff
@@ -23,24 +23,17 @@ export class ObjectGenerator extends FileGenerator<PhpFile, ModelCustomConfigSch
...
-                    name: this.context.getPropertyName(property.name.name),
-                    access: "public",
-                    docs: property.docs,
-                    attributes: this.context.phpAttributeMapper.convert({
-                        type: convertedType,
-                        property
-                    })
-                })
-            );
+            clazz.addField(this.toField({ property, inherited: false }));
```
amckinney, line 33: nit: We can drop the `inherited: false` parameter if we make the parameter optional in `toField`.

amckinney (approved):
Nice, this looks great! It might be worth adding a lightweight test that we can generate into the SDK to confirm the traits are serialized like we expect.

generators/php/model/src/generateTraits.ts
dcb6: ty! looking at this prompted to make a fix and wait to create the set around the concatenated final list of typeids 👍

generators/php/model/src/generateTraits.ts
dcb6: i modified the existing methods to throw instead! let me know if you'd still like changes

i do need the `typeDeclaration` in the `TraitGenerator`, so the method still takes in a `typeDeclaration`

generators/php/codegen/src/ast/Trait.ts
dcb6: yeah good call, i wasn't sure if it was weird for a `Trait` field to have a `traits` field, but with your blessing i'm comfortable with it :)

generators/php/model/src/object/ObjectGenerator.ts
dcb6: nice, done!

PR #4794, opened 2024-10-02
feat(csharp): Add forward compatible enums using a string enum

Swimburger:
- Add IStringEnum as an interface
- Generate string enums for all enums
- Add StringEnumSerializer
- Rename existing StringEnumSerializer to EnumSerializer
- Add feature flag to opt-out of the new enum 'enable-forward-compatible-enums'
- Update C# model and SDK version
- Add tests for old and new serializer


Swimburger:
I assume we want to enable forward-compatible enums by default, so I coded it that way.
Is there a way to change existing users' generator files so we can set it to `false` for them?
Or should I make it `false` by default?

Swimburger:
I did not test whether the old enum serializer still works. 
I did test the new enum serializer manually. Is there a way to include tests for our serializer?

dsinghvi:
> I assume we want to enable forward-compatible enums by default, so I coded it that way. Is there a way to change existing users' generator files so we can set it to `false` for them? Or should I make it `false` by default?

We'll want to release a minor version where it is set to `false` and then a major version where we swap the default. Automating these config upgrades is something we want to do and @armandobelardo is thinking about

seed/csharp-model/enum/forward-compatible-enums/src/SeedEnum/Operand.cs
```diff
@@ -0,0 +1,97 @@
...
+        return Value.GetHashCode();
+    }
+
+    public static bool operator ==(Operand value1, Operand value2) => value1.Equals(value2);
+
+    public static bool operator !=(Operand value1, Operand value2) => !(value1 == value2);
+
+    public static bool operator ==(Operand value1, string value2) => value1.Value.Equals(value2);
+
+    public static bool operator !=(Operand value1, string value2) => !value1.Value.Equals(value2);
```
dcb6, line 96: couldn't we extract quite a bit of this into an abstract class, that perhaps replaces the `IStringEnum` interface?

ex:
```csharp
    public abstract class AbstractStringEnum : IEquatable<string>
    {
        public abstract string Value { get; }

        public override string ToString()
        {
            return Value;
        }

        public bool Equals(string? other)
        {
            return Value.Equals(other);
        }

        public override bool Equals(object? obj)
        {
            if (obj is null)
                return false;
            if (obj is string stringObj)
                return Value.Equals(stringObj);
            if (obj.GetType() != GetType())
                return false;
            return Equals((AbstractStringEnum)obj);
        }

        public override int GetHashCode()
        {
            return Value.GetHashCode();
        }

        public static bool operator ==(AbstractStringEnum value1, AbstractStringEnum value2) => value1.Equals(value2);

        public static bool operator !=(AbstractStringEnum value1, AbstractStringEnum value2) => !(value1 == value2);

        public static bool operator ==(AbstractStringEnum value1, string value2) => value1.Value.Equals(value2);

        public static bool operator !=(AbstractStringEnum value1, string value2) => !value1.Value.Equals(value2);
    }
    ```

Comment (dcb6) [reply seed/csharp-model/enum/forward-compatible-enums/src/SeedEnum/Operand.cs:96]: (might need some fixes, but that's the direction)

File: generators/csharp/codegen/src/ast/Class.ts [hunk 2]
```diff
@@ -21,13 +22,19 @@ export declare namespace Class {
...
         sealed?: boolean;
         /* Defaults to false */
         partial?: boolean;
         /* Defaults to false */
-        record?: boolean;
+        readonly?: boolean;
+        /* Defaults to class */
+        type?: "class" | "record" | "struct";
+        /* Defaults to false */
+        struct?: boolean;
```
dcb6, line 37: is this a leftover?

seed/csharp-model/enum/forward-compatible-enums/src/SeedEnum/Color.cs
```diff
@@ -0,0 +1,85 @@
+using System;
+using System.Text.Json.Serialization;
+using SeedEnum.Core;
+
+#nullable enable
+
+namespace SeedEnum;
+
+[JsonConverter(typeof(StringEnumSerializer<Color>))]
+public readonly struct Color : IStringEnum, IEquatable<Color>
```
dcb6, line 10: would using a `record struct` simplify any of the equality logic?

seed/csharp-model/enum/forward-compatible-enums/src/SeedEnum/Operand.cs
Swimburger: Unfortunately, structs can't inherit from classes, and structs themselves can't be abstract.
Using a class instead of a struct would be easier and more flexible, but it increases memory usage and the amount of garbage collection necessary, slowing perf.

seed/csharp-model/enum/forward-compatible-enums/src/SeedEnum/Color.cs
Swimburger: I'll take a look at that and verify if its backwards compatible to older .NET versions with PolySharp, that would be great

dcb6: sounds good, you may already be aware but we use `record`s for our generated object types and `dotnet build` doesn't complain

(my understanding is that the command will complain if anything we write is not compatible with any of our target frameworks, which happened when i tried to use `record` without PolySharp)

seed/csharp-sdk/enum/plain-enums/src/SeedEnum/Core/StringEnumSerializer.cs
```diff
@@ -0,0 +1,26 @@
+using System.Runtime.Serialization;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace SeedEnum.Core;
+
+internal class StringEnumSerializer<T> : JsonConverter<T>
```
dan-thera, line 7: Why is this serializer getting included here with "plain enums"?

dcb6, line 7: Looks like this serializer is getting erroneously included here with "plain enums"

generators/csharp/csharp.sln
```diff
@@ -0,0 +1,73 @@
+﻿
```
dcb6, line 1: is this intentional?

generators/csharp/model/src/enum/StringEnumGenerator.ts
```diff
@@ -0,0 +1,317 @@
...
+
+    protected doGenerate(): CSharpFile {
+        const serializerAnnotation = csharp.annotation({
+            reference: csharp.classReference({
+                name: "JsonConverter",
+                namespace: "System.Text.Json.Serialization"
+            }),
+            argument: csharp.codeblock((writer) => {
+                writer.write("typeof(");
+                writer.writeNode(csharp.classReference(this.context.getStringEnumSerializerClassReference()));
```
dcb6, line 27: nit: you should be able to use the `generics` field in `ClassReference` to handle the next few lines

generators/csharp/model/src/enum/StringEnumGenerator.ts
```diff
@@ -0,0 +1,317 @@
...
+                        type: csharp.Type.string(),
+                        docs: "Custom enum value"
+                    })
+                ],
+                summary: "Create a string enum with the given value.",
+                return_: csharp.Type.reference(this.classReference),
+                type: csharp.MethodType.STATIC,
+                classReference: csharp.classReference(this.classReference),
+                body: csharp.codeblock((writer) => {
+                    writer.write("return new ");
```
dcb6, line 140: nit: can use `ClassInstantiation` here

seed/csharp-model/enum/forward-compatible-enums/src/SeedEnum/Color.cs
Swimburger: There is potential for some of these features to not work at runtime, but running tests in the runtimes we support should remove any doubt.

generators/csharp/csharp.sln
Swimburger: No, I don't know why this is added.

generators/csharp/model/src/enum/StringEnumGenerator.ts
```diff
@@ -0,0 +1,317 @@
...
+            access: "public",
+            parameters: [
+                csharp.parameter({
+                    name: "value",
+                    type: csharp.Type.string(),
+                    docs: "The string value of the enum."
+                })
+            ],
+            body: csharp.codeblock((writer) => {
+                writer.write("Value = value;");
```
dcb6, line 64: super nit: we try not to write semicolons directly and prefer `writer.write...Statement(...)`, in this case `writer.writerTextStatement(...)`

generators/csharp/csharp.sln
dcb6: Gotcha! Just make sure to remove before merging :)

generators/csharp/model/src/enum/StringEnumGenerator.ts
```diff
@@ -0,0 +1,317 @@
...
+                writer.writeNode(this.classReference);
+                writer.write(">");
+                writer.write(")");
+            })
+        });
+
+        const stringEnum = csharp.class_({
+            ...this.classReference,
+            interfaceReferences: [
+                csharp.classReference({
```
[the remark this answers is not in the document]
Swimburger: Apart from this class, I don't think we'll reuse it, but I could be wrong.

Swimburger:
All feedback should be addressed now.

dsinghvi (approved):

dsinghvi (approved):

Swimburger:
I reseeded with latest from main

generators/csharp/codegen/src/ast/Class.ts
```diff
@@ -353,27 +372,40 @@ export class Class extends AstNode {
...
+                    break;
+                default:
+                    assertNever(operator.type);
+            }
+            writer.write("operator ");
+            writer.write(`${this.name}(`);
+        } else {
+            if (!operator.return) {
+                throw new Error("Operator must have a return type, unless it's implicit or explicit");
+            }
```
amckinney, line 391: We should avoid throwing errors within the AST nodes - I think we can refactor things so that we have a separate `Operator` interface that actually requires the `return` parameter.

Given that this isn't user-facing, we can handle this in a follow-up.

generators/csharp/codegen/src/ast/Class.ts
```diff
@@ -63,26 +68,30 @@ export declare namespace Class {
...
     public readonly abstract: boolean;
     public readonly sealed: boolean;
+    public readonly readonly: boolean;
     public readonly partial: boolean;
     public readonly reference: ClassReference;
     public readonly parentClassReference: AstNode | undefined;
     public readonly interfaceReferences: ClassReference[];
     public readonly isNestedClass: boolean;
-    public readonly record: boolean;
+    public readonly type: "class" | "record" | "struct" | "record struct";
```
amckinney, line 94: Given that we define this enum in two places (in the same file), let's create a type for it and reference it in both places.

generators/csharp/codegen/src/ast/Field.ts
```diff
@@ -137,19 +155,31 @@ export class Field extends AstNode {
...
+            writer.write("readonly ");
+        }
         writer.writeNode(this.type);
         writer.write(` ${this.name}`);
 
         const useExpressionBodiedPropertySyntax = this.get && !this.init && !this.set && this.initializer != null;
         if ((this.get || this.init || this.set) && !useExpressionBodiedPropertySyntax) {
             writer.write(" { ");
             if (this.get) {
+                if (this.get !== true && this.access !== this.get) {
```
amckinney, line 168: I think we can make this a little more clear - it seems like the main goal is to confirm `this.get` is not a boolean right? If not, that's when we write the access.

How about something like this instead?

```suggestion
                if (typeof this.get !== "boolean") {
                  writer.write(`${this.get} `);
                }
```

At that point, we could even include a helper function local to this file (until it's required elsewhere). That way we can just do:

```typescript
if (!isBoolean(this.get)) {
  writer.write(`${this.get} `);
}
```

generators/csharp/model/src/enum/StringEnumGenerator.ts
```diff
@@ -0,0 +1,236 @@
...
+                    name: "IStringEnum",
+                    namespace: this.context.getCoreNamespace()
+                })
+            ],
+            annotations: [serializerAnnotation],
+            access: "public",
+            type: "record struct",
```
amckinney, line 44: nit: If we make this an enum, it might be nice to refer to these types like:

```typescript
type: ClassType.RecordStruct

```

amckinney (approved):
Only nits, but feel free to address in a follow-up (if ever)!

generators/csharp/codegen/src/ast/Class.ts
Swimburger: How does this look to you? I'm not a fan of the `NormalOperator` name.

```ts
    interface CastOperator {
        parameter: Parameter;
        type: "implicit" | "explicit";
        body: CodeBlock;
        useExpressionBody?: boolean;
    }
    interface NormalOperator {
        type: "==" | "!=";
        parameters: Parameter[];
        return: Type;
        body: CodeBlock;
        useExpressionBody?: boolean;
    }
    type Operator = CastOperator | NormalOperator;

    private writeOperator({ writer, operator }: { writer: Writer; operator: Class.Operator }): void {
        writer.write("public static ");
        if (operator.type === "explicit" || operator.type === "implicit") {
            writer.write(`${operator.type} `);
            writer.write("operator ");
            writer.write(`${this.name}(`);
            operator.parameter.write(writer);
        } else {
            const normalOperator = operator as Class.NormalOperator;
            normalOperator.return.write(writer);
            writer.write(" operator ");
            writer.write(`${operator.type}(`);
            normalOperator.parameters.forEach((parameter, idx) => {
                parameter.write(writer);
                if (idx < normalOperator.parameters.length - 1) {
                    writer.write(", ");
                }
            });
        }
        if (operator.useExpressionBody) {
            writer.write(") => ");
            writer.writeNodeStatement(operator.body);
        } else {
            writer.write(") {");
            writer.writeNode(operator.body);
            writer.writeLine("}");
        }
    }
```

generators/csharp/codegen/src/ast/Field.ts
Swimburger: How about this?
```ts
    private hasSameAccess(access: Access | boolean): boolean {
        return access === true || access === this.access;
    }
    
            if (this.get) {
                if (!this.hasSameAccess(this.get)) {
                    writer.write(`${this.get} `);
                }
                writer.write("get; ");
            }
            if (this.init) {
                if (!this.hasSameAccess(this.init)) {
                    writer.write(`${this.init} `);
                }
                writer.write("init; ");
            }
            if (this.set) {
                if (!this.hasSameAccess(this.set)) {
                    writer.write(`${this.set} `);
                }
                writer.write("set; ");
            }
    
    ```

File: generators/csharp/codegen/src/ast/Class.ts [hunk 14]
```diff
@@ -66,19 +66,30 @@ export declare namespace Class {
...
+    interface NormalOperator {
+        type: "==" | "!=";
         parameters: Parameter[];
-        /* The body of the operator */
+        return: Type;
         body: CodeBlock;
-        return?: Type;
         useExpressionBody?: boolean;
     }
+    type Operator = CastOperator | NormalOperator;
```
amckinney, line 82: Nice 👏

amckinney (approved):
🎉 🎉

PR #4798, opened 2024-10-03
fix(php): fix annotations for optional undiscriminated unions

dcb6:
Add `null` to annotations for optional undiscriminated unions.

Also, add a test type that tests this behavior.


seed/php-sdk/undiscriminated-unions/src/Union/Types/TypeWithOptionalUnion.php
```diff
@@ -0,0 +1,27 @@
...
+use Seed\Core\Json\JsonSerializableType;
+use Seed\Core\Json\JsonProperty;
+use Seed\Core\Types\Union;
+
+class TypeWithOptionalUnion extends JsonSerializableType
+{
+    /**
+     * @var string|array<string>|int|array<int>|array<array<int>>|null $myUnion
+     */
+    #[JsonProperty('myUnion'), Union('string', ['string'], 'integer', ['integer'], [['integer']], 'null')]
```
dcb6, line 14: this would not have had `'null'` at the end without the fix in this PR

generators/php/codegen/src/context/PhpAttributeMapper.ts
```diff
@@ -71,12 +74,16 @@ export class PhpAttributeMapper {
         });
     }
 
-    public getUnionTypeParameters(types: php.Type[]): php.AstNode[] {
+    public getUnionTypeParameters({
+        types,
+        isOptional = false
```
amckinney, line 79: I think we can drop this since the fact that it's typed as `isOptional?` the value will be `false`.

generators/php/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -302,7 +302,10 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
         arguments_: UnnamedArgument[];
         types: php.Type[];
     }): php.CodeBlock {
-        const unionTypeParameters = this.context.phpAttributeMapper.getUnionTypeParameters(types);
+        const unionTypeParameters = this.context.phpAttributeMapper.getUnionTypeParameters({
+            types,
+            isOptional: false
```
amckinney, line 307: Think we can drop this since this is the default value.

amckinney (approved):

PR #4808, opened 2024-10-05
fix(cli): the fern cli appropriately generates examples for undiscriminated unions

dsinghvi:
We don't handle max depth properly in the ir examples generator.


packages/cli/generation/ir-generator/src/examples/generator/EndpointExampleFactory.ts
armandobelardo, line 1: Nit: what's this?

dsinghvi: adding this now (currently only type example generation)

packages/cli/generation/ir-generator/src/examples/ExampleGenerator.ts
```diff
@@ -202,87 +209,138 @@ export class ExampleGenerator {
...
+
+        for (const pathParameter of rootPathParameters) {
+            const example = this.generatePathParameterExample({
+                pathParameter,
+                maybePathParameterExample: examples
+                    .flatMap((example) => example.rootPathParameters)
+                    .find((sp) => sp.name === pathParameter.name)
+            });
+            if (example == null) {
+                return undefined;
```
armandobelardo, line 239: Why return here? vs just adding if not null? Presumably you'd want to finish the loop and continue on?

packages/cli/generation/ir-generator/src/examples/ExampleGenerator.ts
```diff
@@ -795,21 +913,28 @@ export class ExampleGenerator {
...
+        }
         return {
             jsonExample: exampleTypeReference.jsonExample,
-            shape: ExampleTypeReferenceShape.container(
-                ExampleContainer.optional({
-                    optional: exampleTypeReference,
-                    valueType: typeReference
-                })
-            )
+            shape: exampleTypeReference.shape
```
armandobelardo, line 926: Do you not need to wrap this like before?

packages/cli/generation/ir-generator/src/examples/ExampleGenerator.ts
dsinghvi: that file is going to completely deleted now

packages/cli/generation/ir-generator/src/examples/ExampleGenerator.ts
dsinghvi: this file is going to completely deleted now

PR #4816, opened 2024-10-07
chore(cli): allow for passing in output mode filters into fern generator list command

armandobelardo:
For PR previews + checks, we cannot run checks for any of our generators in local modes, this is because we do not output project files for them and so checks would always fail.

This allows for filtering them out specifically for fern-bot


packages/cli/cli/src/cliV2.ts
```diff
@@ -79,6 +79,16 @@ export function addGeneratorCommands(cli: Argv<GlobalCliOptions>, cliContext: Cl
...
+                        })
+                        .option("included-modes", {
+                            choices: Object.values(GenerationModeFilter),
+                            type: "array",
+                            description: "The generator output modes to include within the outputted list."
+                        })
+                        .option("excluded-modes", {
+                            choices: Object.values(GenerationModeFilter),
+                            type: "array",
+                            description: "The generator output modes to exclude within the outputted list."
```
amckinney, line 91: This might be pedantic, but in case we want to support other include/exclude filters later, we might actually want to define an internal `Filter` interface and parse the command line flag into the structure before deciding how to filter.

```typescript
interface Filter {
  // The generator output modes we filter on.
  mode?: ModeType // local-file-system, etc.
}
```

Then each flag can be combined into an argument tailored for generator command:

```typescript
interface GeneratorListArgs {
  include?: Filter
  exclude?: Filter
}
```

Finally, the CLI invocation could be:

```sh
$ fern generator list --filter {...}
```

I imagine we'll want to include other filters like generator name someday so it might be worth starting with something more flexible like this, especially because the implementation should largely be the same.

Thoughts?

amckinney: Discussed offline - 

The user-facing documentation experience is tough given that we can use yargs `choices` here (i.e. the user clearly knows what values are valid). The best we could do is type this as a free-form string and parse it into the JSON object we define here, but the `--help` is poor. We would need to include a link to some external reference that contains the structure we expect.

I have a feeling we'll do this in the future, but it's not crazy to start with something simple that can easily be migrated later:

```sh
$ fern generator list --include-mode local-file-system --exclude-mode github
```

packages/cli/cli/src/cliV2.ts
```diff
@@ -79,6 +79,16 @@ export function addGeneratorCommands(cli: Argv<GlobalCliOptions>, cliContext: Cl
                             // which this feels better than.
                             description:
                                 "The APIs to list the generators for. If not specified, the generator will be upgraded for all APIs."
+                        })
+                        .option("included-modes", {
```
amckinney, line 83: Let's use `include-mode` (singular) so that it'll be natural to migrate to `--include` someday (if ever). Same thing with `exclude-mode` below.

packages/cli/ete-tests/src/tests/generators-list/generators-list.test.ts
```diff
@@ -12,4 +12,24 @@ describe("fern generator list", () => {
...
+
+    it("fern generator list with include", async () => {
+        const pathOfDirectory = await init();
+
+        const out = await runFernCli(["generator", "list", "--included-modes", "local-file-system"], {
+            cwd: pathOfDirectory
+        });
+
+        expect(out.stdout).toMatchSnapshot();
+    }, 60_000);
```
amckinney, line 34: Nice 👍

amckinney (approved):
A couple nits on naming, but once those are resolved lgtm!

PR #4820, opened 2024-10-07
feat(python): expose a chunk size config option for file download

armandobelardo:
Fixes FER-3339

This PR specifically puts this option behind `request_options`, given the alternative of exposing this as a parameter would lead to easy ways of breaking the function (e.g. the API is updated to have a `chunk_size` parameter)


dsinghvi (approved):

dsinghvi:
@armandobelardo up to fix python/compile whenever you get a second

PR #4821, opened 2024-10-07
chore(docs): explain instances

dannysheridan:

### DIFF
--- a/fern/pages/docs/getting-started/global-configuration.mdx
+++ b/fern/pages/docs/getting-started/global-configuration.mdx
@@ -68,7 +68,7 @@ favicon: path/to/favicon.ico
 
 ## Instances configuration
 
-An `instance` is the backend of a distinct docs website. Each instance is published to a unique domain using the` --instance flag`. It is most common to use instances to configure staging and production docs which publish to separate URLs.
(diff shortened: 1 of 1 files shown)


fern/pages/docs/getting-started/global-configuration.mdx
```diff
@@ -67,6 +67,9 @@ favicon: path/to/favicon.ico
 </ParamField>
 
 ## Instances configuration
+
+An `instance` is the backend of a distinct docs website. Each instance has a unique set of content (such as a Landing Page, Guides, and API Reference) published to a unique domain. It is most common to use multiple instances on the Enterprise plan.
```
chdeskur, line 71: I would propose: 

An `instance` is the backend of a distinct docs website. Each instance is published to a unique domain using the `--instance` flag. It is most common to use instances to configure staging and production docs which publish to separate URLs.

chdeskur: Would add: 

<Note title="Custom domains">
Adding custom domains for multiple instances is a feature on the Enterprise plan. 
</Note>

chdeskur (approved):

PR #4835, opened 2024-10-08
feat(csharp): add client credentials oauth support

dsinghvi:
This PR adds Oauth client credential support to the C# generator.

### DIFF
--- a/generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
+++ b/generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
@@ -246,7 +246,7 @@ export class OauthTokenProviderGenerator extends FileGenerator<CSharpFile, SdkCu
 
             writer.endControlFlow();
 
-            writer.writeTextStatement(`return $"Bearer ${OauthTokenProviderGenerator.ACCESS_TOKEN_FIELD.name}"`);
+            writer.writeTextStatement(`return $"Bearer \{${OauthTokenProviderGenerator.ACCESS_TOKEN_FIELD.name}\}"`);
         });
     }
 
--- a/seed/csharp-sdk/oauth-client-credentials-default/src/SeedOauthClientCredentialsDefault/Core/OAuthTokenProvider.cs
+++ b/seed/csharp-sdk/oauth-client-credentials-default/src/SeedOauthClientCredentialsDefault/Core/OAuthTokenProvider.cs
@@ -32,6 +32,6 @@ public partial class OAuthTokenProvider
             );
             _accessToken = tokenResponse.AccessToken;
         }
-        return $"Bearer _accessToken";
+        return $"Bearer {_accessToken}";
     }
 }
--- a/seed/csharp-sdk/oauth-client-credentials-environment-variables/src/SeedOauthClientCredentialsEnvironmentVariables/Core/OAuthTokenProvider.cs
+++ b/seed/csharp-sdk/oauth-client-credentials-environment-variables/src/SeedOauthClientCredentialsEnvironmentVariables/Core/OAuthTokenProvider.cs
@@ -37,6 +37,6 @@ public partial class OAuthTokenProvider
                 .UtcNow.AddSeconds(tokenResponse.ExpiresIn)
                 .AddMinutes(-BufferInMinutes);
         }
-        return $"Bearer _accessToken";
+        return $"Bearer {_accessToken}";
     }
 }
--- a/seed/csharp-sdk/oauth-client-credentials-nested-root/src/SeedOauthClientCredentials/Core/OAuthTokenProvider.cs
+++ b/seed/csharp-sdk/oauth-client-credentials-nested-root/src/SeedOauthClientCredentials/Core/OAuthTokenProvider.cs
@@ -37,6 +37,6 @@ public partial class OAuthTokenProvider
                 .UtcNow.AddSeconds(tokenResponse.ExpiresIn)
                 .AddMinutes(-BufferInMinutes);
         }
-        return $"Bearer _accessToken";
+        return $"Bearer {_accessToken}";
     }
 }
--- a/seed/csharp-sdk/oauth-client-credentials/src/SeedOauthClientCredentials/Core/OAuthTokenProvider.cs
+++ b/seed/csharp-sdk/oauth-client-credentials/src/SeedOauthClientCredentials/Core/OAuthTokenProvider.cs
@@ -37,6 +37,6 @@ public partial class OAuthTokenProvider
                 .UtcNow.AddSeconds(tokenResponse.ExpiresIn)
                 .AddMinutes(-BufferInMinutes);
         }
-        return $"Bearer _accessToken";
+        return $"Bearer {_accessToken}";
     }
 }


generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,267 @@
...
+
+                    writer.write(`var tokenResponse = `);
+                    writer.writeNode(
+                        csharp.invokeMethod({
+                            async: true,
+                            method: `client.${this.tokenEndpoint.name.pascalCase.safeName}`,
+                            arguments_: [
+                                csharp.instantiateClass({
+                                    classReference: requestType.internalType.value,
+                                    // TODO(dsinghvi): assumes only top level client id and client secret inputs
```
dsinghvi, line 209: this feels basically impossible with the IR data model as it exists today

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,267 @@
...
+                    });
+                    if (requestType == null || requestType.internalType.type !== "reference") {
+                        throw new Error("Failed to get request class reference");
+                    }
+
+                    writer.write(`var tokenResponse = `);
+                    writer.writeNode(
+                        csharp.invokeMethod({
+                            async: true,
+                            method: `client.${this.tokenEndpoint.name.pascalCase.safeName}`,
```
dsinghvi, line 205: i should be able to call `snippetGenerator` here but that requires an IR Example, which is also impossible to construct at this point

generators/csharp/sdk/src/SdkGeneratorCli.ts
```diff
@@ -145,6 +146,14 @@ export class SdkGeneratorCLI extends AbstractCsharpGeneratorCli<SdkCustomConfigS
         const baseApiException = new BaseApiExceptionGenerator(context);
         context.project.addSourceFiles(baseApiException.generate());
 
+        if (context.ir.auth.schemes[0] != null && context.ir.auth.schemes[0].type === "oauth") {
+            const oauthTokenProvider = new OauthTokenProviderGenerator({
+                context,
+                scheme: context.ir.auth.schemes[0]
+            });
+            context.project.addSourceFiles(oauthTokenProvider.generate());
+        }
```
amckinney, line 155: Should we generate the provider if _any_ of the auth schemes specify `oauth`? I imagine there will be some customers that include multiple, but maybe list `bearer` first.

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,267 @@
...
+        this.tokenEndpointReference = this.scheme.configuration.tokenEndpoint.endpointReference;
+        this.tokenEndpointHttpService = this.context.getHttpServiceOrThrow(this.tokenEndpointReference.serviceId);
+        const httpEndpoint = this.tokenEndpointHttpService.endpoints.find(
+            (endpoint) => endpoint.id === this.tokenEndpointReference.endpointId
+        );
+
+        if (httpEndpoint == null) {
+            throw new Error(`Failed to find token endpoint ${this.tokenEndpointReference.endpointId}`);
+        }
+        this.tokenEndpoint = httpEndpoint;
```
amckinney, line 82: nit: for symmetry with the `getHttpServiceOrThrow` method call above, a private helper function that read a little better.

```suggestion
        this.tokenEndpointReference = this.scheme.configuration.tokenEndpoint.endpointReference;
        this.tokenEndpointHttpService = this.context.getHttpServiceOrThrow(this.tokenEndpointReference.serviceId);
        this.tokenEndpoint = this.resolveTokenEndpointOrThrow(this.tokenEndpointReference);
```

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,267 @@
...
+
+        class_.addField(OauthTokenProviderGenerator.CLIENT_ID_FIELD);
+        class_.addField(OauthTokenProviderGenerator.CLIENT_SECRET_FIELD);
+
+        class_.addMethod(this.getAccessTokenMethod());
+
+        const subpackage = this.context.getSubpackageForServiceId(this.tokenEndpointReference.serviceId);
+        if (subpackage == null) {
+            throw new Error(`Failed to find subpackage for serviceId ${subpackage}`);
+        }
```
amckinney, line 107: Let's add an `OrThrow` variant on the `context` so that we never accidentally reimplement this elsewhere?

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,267 @@
...
+                            async: true,
+                            method: `client.${this.tokenEndpoint.name.pascalCase.safeName}`,
+                            arguments_: [
+                                csharp.instantiateClass({
+                                    classReference: requestType.internalType.value,
+                                    // TODO(dsinghvi): assumes only top level client id and client secret inputs
+                                    arguments_: [
+                                        {
+                                            name: this.scheme.configuration.tokenEndpoint.requestProperties.clientId
+                                                .property.name.name.snakeCase.safeName,
```
amckinney, line 213: Is there a helper we can use for the naming here? I don't think `snakeCase` is right.

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,267 @@
...
+                                            assignment: csharp.codeblock(
+                                                OauthTokenProviderGenerator.CLIENT_ID_FIELD.name
+                                            )
+                                        },
+                                        {
+                                            name: this.scheme.configuration.tokenEndpoint.requestProperties.clientSecret
```
amckinney, line 219: Same as above.

generators/csharp/sdk/src/SdkGeneratorCli.ts
dsinghvi: The behavior is kind of off here across generators -- ie most things only work with the first auth scheme

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
dsinghvi: def

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
dsinghvi: yup definitely

generators/csharp/sdk/src/SdkGeneratorCli.ts
dsinghvi: Decision: Okay to do first auth scheme but there is an IR change that will make this a union forcing generators to handle all, any, or one.

generators/csharp/sdk/src/SdkGeneratorCli.ts
```diff
@@ -145,6 +146,15 @@ export class SdkGeneratorCLI extends AbstractCsharpGeneratorCli<SdkCustomConfigS
...
 
+        const oauth = context.getOauth();
+        if (oauth != null) {
+            const oauthTokenProvider = new OauthTokenProviderGenerator({
+                context,
+                scheme: oauth
+            });
+            context.project.addSourceFiles(oauthTokenProvider.generate());
+        }
+
```
amckinney, line 157: Nice 👍

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,273 @@
...
+                                                OauthTokenProviderGenerator.CLIENT_ID_FIELD.name
+                                            )
+                                        },
+                                        {
+                                            name: this.context.getNameForField(
+                                                this.scheme.configuration.tokenEndpoint.requestProperties.clientSecret
+                                                    .property.name
+                                            ),
+                                            assignment: csharp.codeblock(
+                                                OauthTokenProviderGenerator.CLIENT_ID_FIELD.name
```
amckinney, line 218: I think this should be:

```suggestion
                                                OauthTokenProviderGenerator.CLIENT_SECRET_FIELD.name
```

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,273 @@
...
+                            async: true,
+                            method: `client.${this.tokenEndpoint.name.pascalCase.safeName}`,
+                            arguments_: [
+                                csharp.instantiateClass({
+                                    classReference: requestType.internalType.value,
+                                    // TODO(dsinghvi): assumes only top level client id and client secret inputs
+                                    arguments_: [
+                                        {
+                                            name: this.context.getNameForField(
+                                                this.scheme.configuration.tokenEndpoint.requestProperties.clientSecret
```
amckinney, line 205: I think this should be the following:
```suggestion
                                                this.scheme.configuration.tokenEndpoint.requestProperties.clientId
```

generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    private resolveEndpointOrThrow(service: HttpService, endpointId: EndpointId): HttpEndpoint {
+        const httpEndpoint = service.endpoints.find(
+            (endpoint) => endpoint.id === this.tokenEndpointReference.endpointId
+        );
+        if (httpEndpoint == null) {
+            throw new Error(`Failed to find token endpoint ${this.tokenEndpointReference.endpointId}`);
+        }
+        return httpEndpoint;
+    }
```
amckinney, line 272: nit: We might want to put this on the context instead.

amckinney (commented):
A few last things, but once those are resolved this lgtm!

seed/csharp-sdk/oauth-client-credentials-default/src/SeedOauthClientCredentialsDefault/Core/OAuthTokenProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+    public async Task<string> GetAccessTokenAsync()
+    {
+        if (_accessToken == null)
+        {
+            var tokenResponse = await client.GetToken(
+                new GetTokenRequest { ClientId = _clientId, ClientSecret = _clientSecret }
+            );
+            _accessToken = tokenResponse.AccessToken;
+        }
+        return $"Bearer _accessToken";
```
amckinney, line 35: Do we need to wrap this in the string interpolation syntax?

```suggestion
        return $"Bearer {_accessToken}";
```

I could be wrong, but would be good to confirm.

amckinney (approved):

PR #4836, opened 2024-10-08
feat(typescript): support bytes requests

armandobelardo:


generators/typescript/sdk/client-class-generator/src/endpoint-request/GeneratedBytesEndpointRequest.ts
```diff
@@ -0,0 +1,231 @@
...
+    public getFetcherRequestArgs(
+        context: SdkContext
+    ): Pick<Fetcher.Args, "headers" | "queryParameters" | "body" | "contentType" | "requestType" | "duplex"> {
+        return {
+            headers: this.getHeaders(context),
+            queryParameters: this.queryParams != null ? this.queryParams.getReferenceTo(context) : undefined,
+            contentType: this.requestBody.contentType,
+            requestType: "bytes",
+            body: ts.factory.createIdentifier(GeneratedBytesEndpointRequest.BYTES_VARIABLE_NAME),
+            duplex: ts.factory.createStringLiteral("half")
```
RohinBhargava, line 206: nit: why does this have tsmorph?

armandobelardo: `duplex` is expecting a ts-morph expression! I couldn't handled it within the function, but seemed roughly the same complexity, so just made a ts-morph string literal

PR #4838, opened 2024-10-08
fix(csharp): token provider correctly references underlying endpoint

dsinghvi:
Fixes a bug where the `OauthTokenProvider.cs` was incorrectly referencing the endpoint method, causing code to fail to compile.

## Previous
```cs
  var tokenResponse = await client.GetToken(
      new AuthGetTokenRequest { ClientId = _clientId, ClientSecret = _clientSecret }
  );
```

## After
```cs
  var tokenResponse = await _client.GetTokenAsync(
      new AuthGetTokenRequest { ClientId = _clientId, ClientSecret = _clientSecret }
  );
```

### DIFF
--- a/generators/csharp/sdk/versions.yml
+++ b/generators/csharp/sdk/versions.yml
@@ -1,4 +1,5 @@
 - version: 1.8.1
+  irVersion: 53
   createdAt: '2024-10-08'
   changelogEntry:
     - type: fix


generators/csharp/sdk/src/oauth/OauthTokenProviderGenerator.ts
```diff
@@ -194,7 +195,9 @@ export class OauthTokenProviderGenerator extends FileGenerator<CSharpFile, SdkCu
                     writer.writeNode(
                         csharp.invokeMethod({
                             async: true,
-                            method: `client.${this.tokenEndpoint.name.pascalCase.safeName}`,
+                            method: `${this.clientField.name}.${this.context.getEndpointMethodName(
```
dsinghvi, line 198: key change

amckinney (approved):

PR #4841, opened 2024-10-09
feat(cli): adds streaming examples

franklinharvey:


packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/frank-test.json
```diff
@@ -0,0 +1,284 @@
+{
```
dsinghvi, line 1: nit: can we delete this

PR #4842, opened 2024-10-09
feat(go): Dynamic snippets prototype

amckinney:
This outlines a prototype for an offline dynamic snippets library. With this approach, we generate a lightweight TypeScript SDK that is packaged together with one or more language-specific snippet generators (in this case Go).

The high level file layout of the offline library is shown below:

```sh
.
├── config
│   └── ir.json
├── pacakge.json
└── src
    ├── Language.ts
    ├── Generator.ts
    └── Resolver.ts
```

Instantiate the snippet resolver like so:

```typescript
import { Resolver } from "snippets";

const resolver = new Resolver();
const snippets = resolver.resolve({ language: "go" });

const snippet = snippets.generate({
  endpointID: "POST /users",
  requestBody: {
    username: "john.doe",
    status: "DEACTIVATED"
  },
});

// Returns:
//
// import (
//   context "context"
//   acme "github.com/buildwithfern/acme-go"
//   client "github.com/buildwithfern/acme-go/client"
// )
//
// client := client.NewClient()
// client.User.Create(
//   context.TODO(),
//   &acme.CreateUserRequest{
//     Username: "john.doe",
//     Status:   acme.StatusDeactivated,
//   },
// )
```

## Implementation details

To demonstrate this with Go, we had to roll out the following:
1. Go AST library written in TypeScript
2. A simplified IR focused on creating "dynamic types"
    * Note that none of the primary IR is used in this implementation to reduce the size requirements of the offline library.
4. The DynamicSnippetGenerator and DynamicTypeMapper which use the Go AST nodes to create Go snippets.


amckinney:
The prototype served its purpose - we'll follow up with a variety of separate PRs (e.g. https://github.com/fern-api/fern/pull/4970) that implement this functionality.

PR #4843, opened 2024-10-09
feat(cli): support `extra-properties` in the docs 

RohinBhargava:
This PR does a couple things: 
- Upgrades the FDR SDK (migration to branded aliases, and explicit undefined values)
- Sends `extra-properties` in the IR-to-FDR converter

### DIFF
--- a/packages/cli/ete-tests/src/tests/fdr/__snapshots__/fdr.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/fdr/__snapshots__/fdr.test.ts.snap
@@ -74,7 +74,10 @@ exports[`fdr > {"name":"changelog"} 1`] = `
                             "value": "type_director:Age"
                         }
                     }
-                ]
+                ],
+                "extraProperties": {
+                    "type": "unknown"
+                }
             }
         },
         "type_director:Age": {
@@ -169,7 +172,10 @@ exports[`fdr > {"name":"changelog"} 1`] = `
                             }
                         }
                     }
-                ]
+                ],
+                "extraProperties": {
+                    "type": "unknown"
+                }
             }
         },
         "type_imdb:CreateMovieRequest": {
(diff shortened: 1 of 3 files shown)


packages/cli/configuration/src/docs-yml/ParsedDocsConfiguration.ts
```diff
@@ -49,7 +49,7 @@ export interface AbsoluteJsFileConfig {
 }
 
 export interface JavascriptConfig {
-    remote?: CjsFdrSdk.docs.v1.commons.JsRemoteConfig[];
+    remote?: WithoutQuestionMarks<CjsFdrSdk.docs.v1.commons.JsRemoteConfig>[];
```
dsinghvi, line 52: lets remove the `WithoutQuestionMarkds`

packages/cli/configuration/src/docs-yml/parseDocsConfiguration.ts
```diff
@@ -143,10 +146,43 @@ export async function parseDocsConfiguration({
...
+            amplitude: undefined,
+            mixpanel: undefined,
+            hotjar: undefined,
+            koala: undefined,
+            logrocket: undefined,
+            pirsch: undefined,
+            plausible: undefined,
+            fathom: undefined,
+            clearbit: undefined,
+            heap: undefined
```
dsinghvi, line 178: unclear why we still we have this

packages/cli/docs-resolver/src/ApiReferenceNodeConverter.ts
```diff
@@ -370,13 +371,23 @@ export class ApiReferenceNodeConverter {
...
+        apiDefinitionPackageIdRaw: string | undefined,
+        parentSlug: FernNavigation.V1.SlugGenerator,
         idgen: NodeIdGenerator
-    ): FernNavigation.ApiPackageChild | undefined {
+    ): FernNavigation.V1.ApiPackageChild | undefined {
+        if (!apiDefinitionPackageIdRaw) {
+            this.taskContext.logger.error(
+                "No ApiDefinition ID supplied in the API reference layout: ",
+                endpointItem.endpoint
+            );
```
dsinghvi, line 382: this error message is going to be put in front of a user fyi so we probably want to make it far more friendly and even better `fern check` it

PR #4851, opened 2024-10-11
feat(docs): audience filters in docs.yml

abvthecity:
- includes an `audience` field to every level of the navigation config (except versions, tbd in the next PR)
- include an `all-audiences-must-be-declared` rule indicating that audiences used inside of each level must be declared at the root.

### DIFF
--- a/packages/cli/generation/ir-migrations/src/migrations/v39-to-v38/__test__/__snapshots__/migrateFromV39ToV38.test.ts.snap
+++ b/packages/cli/generation/ir-migrations/src/migrations/v39-to-v38/__test__/__snapshots__/migrateFromV39ToV38.test.ts.snap
@@ -353,7 +353,7 @@ exports[`migrateFromV39ToV38 > snapshot 1`] = `
                   "jsonExample": {
                     "access_token": "access_token",
                     "expires_in": 1,
-                    "refresh_token": undefined,
+                    "refresh_token": "refresh_token",
                   },
                   "shape": {
                     "shape": {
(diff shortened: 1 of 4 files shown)


packages/cli/configuration/fern/definition/docs.yml
```diff
@@ -449,6 +451,14 @@ types:
       - fixed
       - static
 
+  AudienceId: string
+
+  Audience:
+    discriminated: false
+    union:
+      - AudienceId
+      - list<AudienceId>
```
dsinghvi, line 460: do you think there should be a top level place in docs.yml where a user should declare their audiences? in the fern definition they have to list them all out in `api.yml`

dsinghvi: then you can have a `fern check` rule if an audience goes unused etc, instead of trying to compute all the different sets

joshknowles: Agreed!

abvthecity: rule added!

packages/cli/yaml/docs-validator/src/rules/all-audiences-must-be-declared/index.ts
```diff
@@ -0,0 +1,25 @@
...
+                            severity: "error",
+                            // TODO: add a link to the docs
+                            message: `Audience "${usedAudience}" is used but not declared at the top level of the docs.yml file.`
+                        });
+                    }
+                }
+                return violations;
+            }
+        };
+    }
```
dsinghvi, line 24: nice!

dsinghvi (approved):

dsinghvi:
@abvthecity https://github.com/fern-api/fern/blob/main/packages/cli/cli/versions.yml can you add a entry for releasing the CLI here

PR #4911, opened 2024-10-13
chore(csharp): Add Basic Test for AST Class

noanflaherty:
Adds a very basic test for asserting that a Class definition can be written to a string as expected.

## Summary by CodeRabbit

- **Tests**
	- Introduced a new test for validating C# class generation, specifically for a "Car" class within the "Automotive" namespace.

### DIFF
--- a/generators/csharp/codegen/src/ast/__test__/Class.test.ts
+++ b/generators/csharp/codegen/src/ast/__test__/Class.test.ts
@@ -28,12 +28,6 @@ describe("class", () => {
                 rootNamespace: "",
                 customConfig: {}
             })
-        ).toEqual(`\
-namespace Automotive;
-
-public class Car(string make,string model)
-{
-}
-`);
+        ).toMatchSnapshot();
     });
 });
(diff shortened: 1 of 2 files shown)


generators/csharp/codegen/src/ast/__test__/Class.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        });
+        expect(
+            clazz.toString({
+                namespace: "",
+                allNamespaceSegments: new Set<string>(),
+                allTypeClassReferences: new Map<string, Set<string>>(),
+                rootNamespace: "",
+                customConfig: {}
+            })
+        ).toEqual(`\
```
dsinghvi, line 31: @noanflaherty want to update this to be `toMatchSnapshot()` ? that way we'll have a dedicated file with the generated code and itll be easier to track diffs.

noanflaherty: Yeah good call. I'll look for examples of `toMatchSnapshot` elsewhere.

noanflaherty:
@dsinghvi this has been updated to use `toMatchSnapshot`. This is my first time using it – quite nice!

dsinghvi (approved):
🚀

PR #4919, opened 2024-10-15
feat(python): Introduce Type, Annotation, Variable, and Integrate w/ Class

noanflaherty:
This PR introduces everything needed to start adding class attributes to class, with a basic example in the `Car` class introduced previously.

Crucially, introduces the beginnings of three new AST classes:
* Type
* Annotation; and
* Variable

## Release Notes

- **New Features**
  - Enhanced the `.gitignore` file to improve specificity for ignored files and directories.
  - Introduced the `ClassReference` class to encapsulate Python class references with associated module paths.
  - Added the `Field` class for structured representation of fields in the abstract syntax tree.
  - Expanded the type system with a new `InternalType` and added support for `Reference` types.
  - Improved the `Writer` class to manage and format import statements based on class references.

- **Bug Fixes**
  - Improved handling of environment configurations in the `.gitignore` file.

- **Tests**
  - Added a comprehensive suite of unit tests for the `Field` and `Type` classes to validate their functionality and output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ packages/**/generated
 
 **/.idea/*
 vitest.workspace.ts
+**.code-workspace
 
 .python-version
 
--- a/generators/pythonv2/codegen/src/ast/Field.ts
+++ b/generators/pythonv2/codegen/src/ast/Field.ts
@@ -3,19 +3,26 @@ import { Writer } from "./core/Writer";
 import { Type } from "./Type";
 
 export declare namespace Field {
-    export interface Args {
+    export interface BaseArgs {
         /* The name of the field */
         name: string;
-        /* The type annotation of the field */
-        type: Type;
-        /* The initializer for the field */
-        initializer?: string;
     }
+
+    /* At least one of type or initializer must be defined
+     * type: The type annotation of the field
+     * initializer: The initializer for the field
+     */
+    export type Args = BaseArgs &
+        (
+            | { type: Type; initializer?: string }
+            | { type?: Type; initializer: string }
+            | { type: Type; initializer: string }
+        );
 }
 
 export class Field extends AstNode {
     public readonly name: string;
-    public readonly type: Type;
+    public readonly type: Type | undefined;
     public readonly initializer: string | undefined;
 
     constructor({ name, type, initializer }: Field.Args) {
@@ -27,9 +34,13 @@ export class Field extends AstNode {
 
     public write(writer: Writer): void {
         writer.write(this.name);
-        writer.write(": ");
-        this.type.write(writer);
-        if (this.initializer != null) {
+
+        if (this.type !== undefined) {
+            writer.write(": ");
+            this.type.write(writer);
+        }
+
+        if (this.initializer !== undefined) {
             writer.write(` = ${this.initializer}`);
         }
     }
--- a/generators/pythonv2/codegen/src/ast/__test__/Field.test.ts
+++ b/generators/pythonv2/codegen/src/ast/__test__/Field.test.ts
@@ -27,6 +27,15 @@ describe("Field", () => {
         expect(writer.toString()).toMatchSnapshot();
     });
 
+    it("writes a field with a name and value but no type", () => {
+        const field = python.field({
+            name: "my_field",
+            initializer: "'default_value'"
+        });
+        field.write(writer);
+        expect(writer.toString()).toMatchSnapshot();
+    });
+
     it("writes a field with a complex type", () => {
         const field = python.field({
             name: "my_list",
--- a/generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Field.test.ts.snap
+++ b/generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Field.test.ts.snap
@@ -14,6 +14,8 @@ config: Dict[str, Any]"
 
 exports[`Field > writes a field with a name and type 1`] = `"my_field: str"`;
 
+exports[`Field > writes a field with a name and value but no type 1`] = `"my_field = 'default_value'"`;
+
 exports[`Field > writes a field with a name, type, and value 1`] = `"my_int: int = 42"`;
 
 exports[`Field > writes a field with a tuple type 1`] = `


generators/pythonv2/codegen/src/ast/Type.ts
```diff
@@ -0,0 +1,211 @@
+import { assertNever } from "@fern-api/core-utils";
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+
+type InternalType = Int | Float | Bool | Str | Bytes | List | Set | Tuple | Dict | None | Optional | Union | Any;
```
noanflaherty, line 5: More types will be added here in future PRs. The most notable missing type here is `ClassReference`.

generators/pythonv2/codegen/src/ast/__test__/Class.test.ts
```diff
@@ -7,4 +7,14 @@ describe("class", () => {
...
     });
+
+    it("variables with annotation and initializer", async () => {
+        const clazz = python.class_({
+            name: "Car"
+        });
+        clazz.addVariable(
+            python.variable({ name: "color", type: python.annotation({ type: "str" }), initializer: "'red'" })
+        );
+        expect(clazz.toString()).toMatchSnapshot();
```
noanflaherty, line 18: Probably the most interesting test here in this PR

generators/pythonv2/codegen/src/ast/__test__/Type.test.ts
```diff
@@ -0,0 +1,88 @@
+import { Type } from "../Type";
+import { Writer } from "../core/Writer";
+
+describe("Type", () => {
```
noanflaherty, line 4: Cursor was a big help in generating a set of useful unit tests.

generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Class.test.ts.snap
```diff
@@ -1,3 +1,12 @@
...
 
-exports[`class > basic 1`] = `"class Car:"`;
+exports[`class > basic 1`] = `
+"class Car:
+"
+`;
+
+exports[`class > variables with annotation and initializer 1`] = `
+"class Car:
+    color: str = 'red'
```
noanflaherty, line 10: Shows that everything worked as expected for our desired usage.

generators/pythonv2/codegen/src/ast/Annotation.ts
```diff
@@ -0,0 +1,27 @@
...
+        super();
+        this.type = args.type;
+    }
+
+    public write(writer: Writer): void {
+        writer.write(": ");
+        if (typeof this.type === "string") {
+            writer.write(this.type);
+        } else {
+            this.type.write(writer);
```
dvargas92495, line 24: [q] How does the writer handle super long lines? like let's say the AstNode is a `Union` of like 30 types

generators/pythonv2/codegen/src/ast/Class.ts
```diff
@@ -11,12 +12,34 @@ export declare namespace Class {
...
 
+    private variables: Variable[] = [];
+
     constructor({ name }: Class.Args) {
         super();
         this.name = name;
     }
 
     public write(writer: Writer): void {
         writer.write(`class ${this.name}:`);
```
dvargas92495, line 23: [q] base classes coming in a future PR?

generators/pythonv2/codegen/src/ast/Type.ts
```diff
@@ -0,0 +1,211 @@
...
+    type: "list";
+    value: Type;
+}
+
+interface Set {
+    type: "set";
+    value: Type;
+}
+
+interface Tuple {
```
dvargas92495, line 45: [q] we may want to include a flag here for ellipsis, eg `Tuple[str, ...]`

generators/pythonv2/codegen/src/ast/Variable.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Annotation } from "./Annotation";
+
+export declare namespace Variable {
+    interface Args {
+        /* The name of the variable */
+        name: string;
+        /* The type annotation of the variable */
+        type?: Annotation;
+        /* The initializer for the variable */
+        initializer?: string;
```
dvargas92495, line 12: [nit] I'm surprised that initializer is a `string` and not an `AstNode`

generators/pythonv2/codegen/src/ast/__test__/Class.test.ts
```diff
@@ -7,4 +7,14 @@ describe("class", () => {
         });
         expect(clazz.toString()).toMatchSnapshot();
     });
+
+    it("variables with annotation and initializer", async () => {
+        const clazz = python.class_({
+            name: "Car"
+        });
+        clazz.addVariable(
+            python.variable({ name: "color", type: python.annotation({ type: "str" }), initializer: "'red'" })
```
dvargas92495, line 16: [oos] interesting question of whether we should be outputting `'red'` or `"red"`. Maybe there's a prettier/black postprocessing step we could do using the user's config if there is one, defaulting to our own?

dvargas92495 (approved):
Becoming a big fan of this `Writer` pattern

generators/pythonv2/codegen/src/ast/Annotation.ts
noanflaherty: The csharp package has a formatting step: https://github.com/fern-api/fern/blob/main/generators/csharp/codegen/src/ast/core/AstNode.ts#L85

Maybe we do something similar with `black`? I'd be keen on @dsinghvi's advice here when the time comes.

generators/pythonv2/codegen/src/ast/Class.ts
noanflaherty: Yeah OOS for this one.

generators/pythonv2/codegen/src/ast/Type.ts
noanflaherty: Gonna OOS this from here.

generators/pythonv2/codegen/src/ast/Variable.ts
noanflaherty: Hm I see a couple different patterns for initializers in csharp.

String: https://github.com/fern-api/fern/blob/main/generators/csharp/codegen/src/ast/Parameter.ts#L14
Codeblock: https://github.com/fern-api/fern/blob/main/generators/csharp/codegen/src/ast/Field.ts#L32

I'd be interested in @dsinghvi's preference here.

generators/pythonv2/codegen/src/ast/__test__/Class.test.ts
noanflaherty: Hm yeah interesting. My guess would be that we output the latter but then also have a postprocessing step, as commented above.

generators/pythonv2/codegen/src/ast/Annotation.ts
dsinghvi: I would recommend using this wasm library that uses ruff https://www.npmjs.com/package/@wasm-fmt/ruff_fmt

generators/pythonv2/codegen/src/ast/Variable.ts
```diff
@@ -0,0 +1,37 @@
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+import { Annotation } from "./Annotation";
+
+export declare namespace Variable {
+    interface Args {
+        /* The name of the variable */
+        name: string;
+        /* The type annotation of the variable */
+        type?: Annotation;
```
dsinghvi, line 10: thoughts on this just being a `Type.ts` -- forces someone to create a `Type` when instantiating a `Field`

generators/pythonv2/codegen/src/ast/Variable.ts
```diff
@@ -0,0 +1,37 @@
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+import { Annotation } from "./Annotation";
+
+export declare namespace Variable {
```
dsinghvi, line 5: from another comment (rename this to Field)

generators/pythonv2/codegen/src/ast/Type.ts
```diff
@@ -0,0 +1,195 @@
...
+                this.internalType.value.write(writer);
+                writer.write("]");
+                break;
+            case "set":
+                writer.write("Set[");
+                this.internalType.value.write(writer);
+                writer.write("]");
+                break;
+            case "tuple":
+                writer.write("Tuple[");
```
dsinghvi, line 154: @noanflaherty you will probably want to setup `ClassReference` and imports in this PR since we want to make sure that we are importing appropriately from the `typing` module 

`from typing import Tuple`
`from typing import Set`

generators/pythonv2/codegen/src/ast/Variable.ts
dsinghvi: And introduce a `CodeBlock AST Node` which can either be a string or can be a lambda that takes a writer. basically a glorified function that will allow you to write any code you want

generators/pythonv2/codegen/src/ast/__test__/Class.test.ts
dsinghvi: (commented on another thread to use ruff fmt)

generators/pythonv2/codegen/src/ast/Variable.ts
```diff
@@ -0,0 +1,37 @@
...
+        /* The type annotation of the variable */
+        type?: Annotation;
+        /* The initializer for the variable */
+        initializer?: string;
+    }
+}
+
+export class Variable extends AstNode {
+    public readonly name: string;
+    public readonly type: Annotation | undefined;
```
armandobelardo, line 18: Once/if this becomes `Field` then type should not be nullable

dvargas92495: annotations are optional in python though? eg this could be `color = 'red'`

generators/pythonv2/codegen/src/ast/Variable.ts
dvargas92495: When do you guys use `CodeBlock` vs stuff like [this](https://github.com/fern-api/fern/pull/4919/files#diff-377cd78484956f0278d9409f9c983abe8361f18f6f83d95ef7b557fc6a05e9ddR7)

Since this initializer is basically code, I would think we would want it all to be ast traversable

generators/pythonv2/codegen/src/ast/Variable.ts
armandobelardo: They're not really optional for fields, which is what this class is supposed to be (not variables)

generators/pythonv2/codegen/src/ast/ClassReference.ts
```diff
@@ -0,0 +1,28 @@
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+
+export class ClassReference extends AstNode {
+    constructor(private readonly name: string, private readonly modulePath: string[]) {
+        super();
+    }
```
noanflaherty, line 7: Make modulePath explicit

noanflaherty: Use args here.

generators/pythonv2/codegen/src/ast/__test__/Variable.test.ts
```diff
@@ -0,0 +1,75 @@
+import { python } from "../..";
+import { Writer } from "../core/Writer";
+
+describe("Field", () => {
```
dsinghvi, line 4: nit: rename to `Field.Test.ts`

dsinghvi (approved):
🚀 love it

PR #4927, opened 2024-10-15
feat(python): More Python AST Nodes

noanflaherty:
This PR introduces the following new ATS nodes for python. Most are directly influenced from csharp.

* `CodeBlock` – I also refactored existing initializers to use these instead of strings
* `Method` – Feels more Pythonic to call these "Functions," but I stuck to "Method" for consistency
* `Parameter` – Used to define arguments on a Method. At first my instinct was to reuse `Field` for this, but opted for consistency
* `MethodInvocation` – Used to invoke a method
* `MethodArgument` – Passed to MethodInvocation and a departure from csharp. Needed by python to support positional vs keyword arguments.

It also adds support for defining a class's parent class(es) and generic types.

## Summary by CodeRabbit

- **New Features**
	- Enhanced `.gitignore` to exclude additional workspace configuration files.
	- Added new functionalities for class handling, including inheritance and field management in the Python code generation module.
	- Introduced a structured way to define and generate methods, including support for static, instance, and class methods.

- **Bug Fixes**
	- Improved handling of parent class references and field management in class definitions.

- **Tests**
	- Expanded test coverage for class definitions, inheritance scenarios, and various field configurations.
	- Added comprehensive unit tests for method generation and representation.

- **Documentation**
	- Updated export structure to include new entities for better accessibility.

### DIFF
--- a/fern/pages/changelogs/cli/2024-10-15.mdx
+++ b/fern/pages/changelogs/cli/2024-10-15.mdx
@@ -1,0 +1,4 @@
+## 0.45.0-rc4
+**`(chore):`** SCIM has been added as a common initialism. 
+
+
--- a/fern/pages/changelogs/fastapi/2024-10-11.mdx
+++ b/fern/pages/changelogs/fastapi/2024-10-11.mdx
@@ -1,0 +1,4 @@
+## 1.5.0
+**`(fix):`** The snippet writer now correctly handles base64 strings.
+
+
--- a/fern/pages/changelogs/pydantic/2024-10-11.mdx
+++ b/fern/pages/changelogs/pydantic/2024-10-11.mdx
@@ -1,0 +1,4 @@
+## 1.4.7
+**`(fix):`** The snippet writer now correctly handles base64 strings.
+
+
--- a/fern/pages/changelogs/python-sdk/2024-10-11.mdx
+++ b/fern/pages/changelogs/python-sdk/2024-10-11.mdx
@@ -1,0 +1,4 @@
+## 4.2.8
+**`(fix):`** The snippet writer now correctly handles base64 strings.
+
+
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,3 +1,10 @@
+- changelogEntry: 
+    - summary: | 
+        SCIM has been added as a common initialism. 
+      type: chore
+  irVersion: 53
+  version: 0.45.0-rc4
+
 - changelogEntry:
     - summary: | 
         Numerous fixes to the Conjure API Importer such as reading in request bodies and query parameters. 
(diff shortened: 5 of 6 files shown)


generators/pythonv2/codegen/src/ast/Class.ts
```diff
@@ -6,21 +7,38 @@ export declare namespace Class {
     interface Args {
         /* The name of the Python class */
         name: string;
+        /* The parent classes that this class inherits from */
+        parentClassReferences?: ClassReference[];
```
dsinghvi, line 11: nit: would name this `extends`

generators/pythonv2/codegen/src/ast/Method.ts
```diff
@@ -0,0 +1,135 @@
...
+        /* The body of the method */
+        body?: CodeBlock;
+        /* The docstring for the method */
+        docstring?: string;
+        /* The type of the method, defaults to STATIC */
+        type?: MethodType;
+        /* The class this method belongs to, if any */
+        classReference?: ClassReference;
+        /* Any decorators to add to the method */
+        decorators?: string[];
```
dsinghvi, line 32: do decorators need to be imported often? I could see this being a `ClassReference`

generators/pythonv2/codegen/src/ast/MethodArgument.ts
```diff
@@ -0,0 +1,32 @@
+import { CodeBlock } from "./CodeBlock";
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+import { Type } from "./Type";
+
+export declare namespace MethodArgument {
```
dsinghvi, line 6: nit: we typically call this `Parameter`

generators/pythonv2/codegen/src/ast/MethodArgument.ts
```diff
@@ -0,0 +1,32 @@
...
+    interface Args {
+        /* If a kwarg, then the name of the parameter that this is a keyword argument for */
+        name?: string;
+        /* The value of the argument */
+        value: CodeBlock;
+    }
```
dsinghvi, line 12: whats the difference between `Parameter` and `MethodArgument` ?

dsinghvi: makes sense

dsinghvi (approved):

generators/pythonv2/codegen/src/ast/Method.ts
noanflaherty: I'm going to remove decorators from the scope of this PR and add them in separately and more hollistically.

I'm thinking about adding a new top-level `Decorator` ast node and adding `addDecorator` methods to both the `Class` and `Method` ast nodes.

PR #4936, opened 2024-10-17
feat(python): Introduce `PythonFile` AST Node

noanflaherty:
This introduces a new pattern for writing python files and managing absolute vs relative imports.

## Summary by CodeRabbit

- **New Features**
	- Introduced a `PythonFile` entity for improved management of Python files in the code generation process.
	- Added functionality to classes allowing optional body representation through a new `addBody` method.
	- Simplified the `Method` class by removing unnecessary properties, enhancing clarity and usability.

- **Bug Fixes**
	- Improved logic in the `write` method of classes to handle class bodies more effectively.

- **Tests**
	- Added comprehensive unit tests for the new features related to Python file and class body generation.

### DIFF
--- a/generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Class.test.ts.snap
+++ b/generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Class.test.ts.snap
@@ -6,12 +6,9 @@ exports[`class > basic 1`] = `
 `;
 
 exports[`class > class with a decorator 1`] = `
-"from dataclasses import dataclass
-    
-@dataclass
+"@dataclass
 class MyDataClass:
-    pass
-"
+    pass"
 `;
 
 exports[`class > fields with annotation and initializer 1`] = `
--- a/generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Decorator.test.ts.snap
+++ b/generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Decorator.test.ts.snap
@@ -1,16 +1,12 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`Decorator > toString > should generate a decorator with a reference 1`] = `
-"from decorators import decorator_with_reference
-    
-@decorator_with_reference
+"@decorator_with_reference
 "
 `;
 
 exports[`Decorator > toString > should generate a decorator with a reference and arguments 1`] = `
-"from decorators import complex_decorator
-    
-@complex_decorator('test', value=True)
+"@complex_decorator('test', value=True)
 "
 `;
 


generators/pythonv2/codegen/src/ast/Class.ts
```diff
@@ -15,8 +16,8 @@ export declare namespace Class {
 export class Class extends AstNode {
     public readonly name: string;
     public readonly extends_: Reference[];
-
     private fields: Field[] = [];
+    private body?: CodeBlock;
```
noanflaherty, line 20: Tangentially added support for defining the body of a class

generators/pythonv2/codegen/src/ast/Method.ts
```diff
@@ -26,10 +26,9 @@ export declare namespace Method {
         docstring?: string;
         /* The type of the method, defaults to STATIC */
         type?: MethodType;
-        /* The class this method belongs to, if any */
-        classReference?: Reference;
```
noanflaherty, line 30: Wasn't being used. Removing until there's reason for this.

generators/pythonv2/codegen/src/ast/PythonFile.ts
```diff
@@ -0,0 +1,103 @@
...
+import { Method } from "./Method";
+import { CodeBlock } from "./CodeBlock";
+import { Reference } from "./Reference";
+
+export declare namespace PythonFile {
+    interface Args {
+        /* The name of the Python module that this file belongs to*/
+        moduleName: string;
+        /* The path of the Python file relative to the module */
+        path: string[];
```
noanflaherty, line 13: These two are key for resolving relative imports

generators/pythonv2/codegen/src/ast/PythonFile.ts
```diff
@@ -0,0 +1,103 @@
...
+            }
+        });
+    }
+
+    /*******************************
+     * Helper Methods
+     *******************************/
+
+    private writeImports(writer: Writer): void {
+        const references = writer.getReferences();
```
noanflaherty, line 52: I found it more convenient for the writer to still be responsible for tracking references. I could see this living separately in a "context" object or similar, but keeping it simple by using the writer as the context for now. It felt heavy-handed to try to use the `PythonFile` node to keep track of all references.

generators/pythonv2/codegen/src/ast/PythonFile.ts
```diff
@@ -0,0 +1,103 @@
...
+            }
+        });
+
+        // Use the deduplicated references
+        const deduplicatedReferences = Array.from(uniqueReferences.values());
+
+        deduplicatedReferences.forEach((reference) => {
+            const refModulePath = reference.getModulePath();
+            if (refModulePath[0] === this.moduleName) {
+                // Relativize the import
```
noanflaherty, line 69: Cursor helped me with the algo for this, but tests show that it seems reasonable.

generators/pythonv2/codegen/src/ast/PythonFile.ts
```diff
@@ -0,0 +1,103 @@
...
+                uniqueReferences.set(fullyQualifiedPath, reference);
+            }
+        });
+
+        // Use the deduplicated references
+        const deduplicatedReferences = Array.from(uniqueReferences.values());
+
+        deduplicatedReferences.forEach((reference) => {
+            const refModulePath = reference.getModulePath();
+            if (refModulePath[0] === this.moduleName) {
```
noanflaherty, line 68: If the reference's path starts with module name, then we know it should be a relative import. Otherwise, we can infer it's an absolute import.

I welcome a sanity check on this though...

generators/pythonv2/codegen/src/ast/__test__/__snapshots__/PythonFile.test.ts.snap
```diff
@@ -0,0 +1,39 @@
...
+"from ..my_module.level_1 import OtherClass
+from ..my_module.level_1.level_2 import DeepClass
+
+class TestClass(OtherClass):
+    pass
+
+class DeeplyNestedTestClass(DeepClass):
+    pass
+"
+`;
```
noanflaherty, line 28: This is probably the most interesting use-case here. It shows the upwards traversal of relative imports to the module root.

generators/pythonv2/codegen/src/ast/core/Writer.ts
```diff
@@ -8,36 +8,17 @@ export declare namespace Writer {}
 
 export class Writer extends AbstractWriter {
     /* Import statements */
-    private references: Record<ModulePath, Reference[]> = {};
+    private references: Reference[] = [];
```
noanflaherty, line 11: The writer becomes way more simple as a result of all this. Notably it's still responsible for tracking references, as discussed above.

generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Class.test.ts.snap
```diff
@@ -2,35 +2,27 @@
 
 exports[`class > basic 1`] = `
 "class Car:
-    pass
-"
+    pass"
 `;
 
 exports[`class > fields with annotation and initializer 1`] = `
-"from typing import Dict
```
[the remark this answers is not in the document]
dsinghvi: @noanflaherty do you think you can flag this some how in the writer? (its important for fern to be able to toggle this behavior for snippets)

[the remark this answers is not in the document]
dsinghvi:
```ts
public toString(skipImports = false): string {
        if (!skipImports) {
            const imports = this.stringifyImports();
            if (imports.length > 0) {
                return `${imports}
    #nullable enable
    
    ${this.buffer}`;
            }
        }
        return this.buffer;
    }
```
    
Example from C#

dsinghvi (approved):

PR #4938, opened 2024-10-17
feat(python): Support Reference Aliases

noanflaherty:
Adds support for specifying the alias used for reference imports.

## Summary by CodeRabbit

- **New Features**
	- Enhanced import statement generation to support aliases for imported references.
	- Introduced a new optional alias property for references.

- **Tests**
	- Added test cases to verify the correct handling of imports with aliases in Python files.
	- Expanded test coverage for the `toString` method to include references with aliases.

### DIFF
--- a/generators/pythonv2/codegen/src/ast/__test__/__snapshots__/PythonFile.test.ts.snap
+++ b/generators/pythonv2/codegen/src/ast/__test__/__snapshots__/PythonFile.test.ts.snap
@@ -27,6 +27,22 @@ class DeeplyNestedTestClass(DeepClass):
 "
 `;
 
+exports[`PythonFile > Add a class with a relative import and alias 1`] = `
+"from ...test_module.test.sibling_dir import SiblingClass
+
+class TestClassWithRelativeAlias(SiblingClass as AliasedSibling):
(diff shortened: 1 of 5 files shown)


dsinghvi (approved):

generators/pythonv2/codegen/src/ast/__test__/__snapshots__/PythonFile.test.ts.snap
```diff
@@ -27,6 +27,22 @@ class DeeplyNestedTestClass(DeepClass):
 "
 `;
 
+exports[`PythonFile > Add a class with a relative import and alias 1`] = `
+"from ...test_module.test.sibling_dir import SiblingClass
+
+class TestClassWithRelativeAlias(SiblingClass as AliasedSibling):
```
dvargas92495, line 33: [blocking] this looks incorrect I don't think python supports inline `as` aliasing within the base class inheritance

noanflaherty: Ah true good catch. Will fix in a follow-up.

PR #4943, opened 2024-10-17
feat(python): Support Formatting via ruff

noanflaherty:
This PR introduces code formatting via ruff and updates tests accordingly.

### DIFF
--- a/generators/pythonv2/codegen/src/ast/core/AstNode.ts
+++ b/generators/pythonv2/codegen/src/ast/core/AstNode.ts
@@ -1,5 +1,6 @@
 import { AbstractAstNode } from "@fern-api/generator-commons";
 import { Writer } from "./Writer";
+import init, { format } from "@wasm-fmt/ruff_fmt";
 
 export abstract class AstNode extends AbstractAstNode {
     /**
@@ -9,5 +10,13 @@ export abstract class AstNode extends AbstractAstNode {
         const writer = new Writer();
         this.write(writer);
         return writer.toString();
+    }
+
+    /**
+     * Writes the node to a string and formats it.
+     */
+    public async toStringFormatted(): Promise<string> {
+        await init();
+        return format(this.toString());
     }
 }


generators/pythonv2/codegen/src/ast/PythonFile.ts
```diff
@@ -20,7 +20,7 @@ export class PythonFile extends AstNode {
     public readonly moduleName: string;
     public readonly path: string[];
     public readonly name: string;
-    private readonly statements: (Class | Method | CodeBlock)[] = [];
+    private readonly statements: AstNode[] = [];
```
noanflaherty, line 23: Unrelated change I'm sneaking in. You should be able to add any sort of node as a statement to a file imo.

generators/pythonv2/codegen/src/ast/__test__/Class.test.ts
```diff
@@ -1,11 +1,19 @@
 import { python } from "../..";
+import { Writer } from "../core/Writer";
 
 describe("class", () => {
+    let writer: Writer;
+
+    beforeEach(() => {
+        writer = new Writer();
+    });
```
noanflaherty, line 9: You'll find that I've updated all tests to call the writer's `toStringFormatted` method rather than adding a `toStringFormatted` method to the `AstNode`.

I describe why below.

generators/pythonv2/codegen/src/ast/core/Writer.ts
```diff
@@ -14,6 +15,11 @@ export class Writer extends AbstractWriter {
         return this.buffer;
     }
 
+    public async toStringFormatted(): Promise<string> {
+        await init();
+        return format(this.buffer);
+    }
```
noanflaherty, line 21: I intentionally chose to add this to the writer rather than the AstNode. This is because the writer has context needed for codegen that the AstNode doesn't have – primarily, references.

This was important for `PythonFile`, which relies on the fact that the writer has accumulated the references needed to generate imports for the file.

noanflaherty: Side Note: I wonder if there's a way to use a global to ensure that in a given process, we only ever call `await init();` once? I noticed tests were a bit slower and I suspect this is why.

generators/pythonv2/codegen/src/ast/PythonFile.ts
dsinghvi: nice

generators/pythonv2/codegen/src/ast/core/Writer.ts
dsinghvi:
```
export abstract class AstNode extends AbstractAstNode {
    /**
     * Writes the node to a string.
     */
    public toString(): string {
        const writer = new Writer();
        this.write(writer);
        return writer.toString();
    }

    public toStringFormatted(): string {
        return prettier.format(this.toString(), { parser: "typescript", tabWidth: 4, printWidth: 120 });
    }
}
```

dsinghvi: this is how i do it in typescript

dsinghvi (approved):
python guru 🚀

PR #4948, opened 2024-10-18
chore(python): Move References from `Writer` to `AstNode`

noanflaherty:
In trying to integrate fern's codegen with vellum, I've found it problematic for the `Writer` to be the thing that keeps track of references and instead, for it to be much more natural for an `AstNode` to keep track of its own references.

This came to a head with generating files and handling imports. `PythonFile` relied on `Writer`'s references to generate imports, but this depended on all of the file's statements being written first. This means only one of two bad options were possible:
1. Write imports then statements -> This led to empty imports becuase the writer didn't have any references yet because the statements hadn't been written yet; or
2. Write statements then imports –> This led to imports being underneath statements

With this fix, it both feels more natural and makes file writing possible.

### DIFF
--- a/generators/pythonv2/codegen/src/ast/Method.ts
+++ b/generators/pythonv2/codegen/src/ast/Method.ts
@@ -28,8 +28,6 @@ export declare namespace Method {
         docstring?: string;
         /* The type of the method if defined within the context of a class */
         type?: ClassMethodType;
-        /* The class this method belongs to, if any */
-        classReference?: Reference;
         /* The decorators that should be applied to this method */
         decorators?: Decorator[];
     }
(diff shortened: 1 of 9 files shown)


generators/pythonv2/codegen/src/ast/Method.ts
```diff
@@ -28,8 +28,6 @@ export declare namespace Method {
         docstring?: string;
         /* The type of the method if defined within the context of a class */
         type?: ClassMethodType;
-        /* The class this method belongs to, if any */
-        classReference?: Reference;
```
noanflaherty, line 32: Semi-unrelated change. Method's weren't using this yet.

generators/pythonv2/codegen/src/ast/__test__/__snapshots__/PythonFile.test.ts.snap
```diff
@@ -6,6 +6,15 @@ exports[`PythonFile > Add a Method 1`] = `
 "
 `;
 
+exports[`PythonFile > Add a class that inherits from a class imported from another file 1`] = `
+"from ..test_module.base import BaseClass
+
+
+class DerivedClass(BaseClass):
+    pass
```
noanflaherty, line 14: A test that proves this works now.

dsinghvi (approved):

PR #4949, opened 2024-10-18
feat(python): Support Adding Statements to Classes and Methods

noanflaherty:
With this, we can now add statements to classes and methods in the python ast.

### DIFF
--- a/generators/pythonv2/codegen/src/ast/Field.ts
+++ b/generators/pythonv2/codegen/src/ast/Field.ts
@@ -31,6 +31,19 @@ export class Field extends AstNode {
         this.name = name;
         this.type = type;
         this.initializer = initializer;
+
+        if (this.type) {
+            const typeReferences = this.type.getReferences();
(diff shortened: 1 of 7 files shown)


generators/pythonv2/codegen/src/ast/__test__/__snapshots__/Class.test.ts.snap
```diff
@@ -37,3 +37,11 @@ exports[`class > inherits from two parent classes 1`] = `
     pass
 "
 `;
+
+exports[`class > should generate a class with local classes 1`] = `
+"class OuterClass:
+    class InnerClass(ParentClass):
```
dvargas92495, line 43: [q] where does `ParentClass` come from? Are we missing an import statement, or is that oos for this PR?

dvargas92495 (approved):

dsinghvi (approved):

PR #4953, opened 2024-10-19
feat(python): Support Nested Attributes References

noanflaherty:
This allows you to reference nested attributes of an object in python.


generators/pythonv2/codegen/src/ast/Reference.ts
```diff
@@ -16,6 +16,8 @@ export declare namespace Reference {
         genericTypes?: Type[];
         /* The alias of the reference */
         alias?: string;
+        /* The path to the attribute of the reference */
+        attrPath?: string[];
```
noanflaherty, line 20: Open to feedback on naming for this.

dsinghvi: would probably name it attribute and type as `string[]` like you have it

dsinghvi (approved):

PR #4962, opened 2024-10-21
chore(python): Introduce Helper for Inheriting References

noanflaherty:
I found myself performing this pattern often enough, both in the fern repo and in the vellum repo, that I figured it was worth a quick helper.

### DIFF
--- a/generators/python-v2/codegen/src/ast/core/AstNode.ts
+++ b/generators/python-v2/codegen/src/ast/core/AstNode.ts
@@ -4,18 +4,18 @@ import init, { format } from "@wasm-fmt/ruff_fmt";
 import { Reference } from "../Reference";
 
 export abstract class AstNode extends AbstractAstNode {
-    private references: Reference[] = [];
+    protected references: Reference[] = [];
 
     public addReference(reference: Reference): void {
         this.references.push(reference);
     }
 
-    public getReferences(): Reference[] {
-        return this.references;
-    }
+    public inheritReferences(astNode: AstNode | undefined): void {
+        if (astNode === undefined) {
+            return;
+        }
 
-    public inheritReferences(astNode: AstNode): void {
-        astNode.getReferences().forEach((reference) => {
+        astNode.references.forEach((reference) => {
             this.addReference(reference);
         });
     }
--- a/generators/python-v2/codegen/src/ast/Field.ts
+++ b/generators/python-v2/codegen/src/ast/Field.ts
@@ -32,18 +32,8 @@ export class Field extends AstNode {
         this.type = type;
         this.initializer = initializer;
 
-        if (this.type) {
-            const typeReferences = this.type.getReferences();
-            typeReferences.forEach((reference) => {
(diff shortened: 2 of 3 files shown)


generators/python-v2/codegen/src/ast/Class.ts
```diff
@@ -64,10 +64,7 @@ export class Class extends AstNode {
 
     public addStatement(statement: AstNode): void {
         this.statements.push(statement);
-
-        statement.getReferences().forEach((reference) => {
-            this.addReference(reference);
-        });
+        this.inheritReferences(statement);
```
noanflaherty, line 67: I'm open to naming suggestions for this helper.

generators/python-v2/codegen/src/ast/core/AstNode.ts
```diff
@@ -14,6 +14,12 @@ export abstract class AstNode extends AbstractAstNode {
         return this.references;
     }
 
+    public inheritReferences(astNode: AstNode): void {
+        astNode.getReferences().forEach((reference) => {
+            this.addReference(reference);
+        });
+    }
```
dsinghvi, line 21: when would something have to inherit references? (asking to better understand the use case)

noanflaherty: "Inherit" might not be the right term since it's actually the reverse direction, but the idea is that if a "parent" AstNode has a "child" AstNode, and that child has a reference, then the parent should have the reference too, such that if you call `write` on the parent, it includes the imports needed by the child.

A real world example for us:

```python
chat_message = ChatMessage(role="USER", content=ChatMessageContent(type="FUNCTION_CALL", value=ChatMessageContentFunctionCall(...)))
```

now imagine our codegen looks like:
```
public ChatMessage extends AstNode {...}

public ChatMessageContent extends AstNode {...}

public ChatMessageContentFunctionCall extends AstNode {...}

chat_message.write(writer)
```

Calling `.write` on `ChatMessage` should include the imports of the children within. One thing to keep in mind is that as of [this PR](https://github.com/fern-api/fern/pull/4948), references are tracked on ast nodes rather than the writer, so having parent nodes inherit references from children nodes is now important.

dsinghvi: makes sense, i think we should model this as `protected` / `overriden` methods (ie your AST node should call super() and then add its own thing)

Is that something you'd be up to do?

dsinghvi: https://github.com/fern-api/fern/blob/bfc7aa0efe6800c7a6f1cdfa908778d68d954c98/generators/python-v2/codegen/src/ast/core/AstNode.ts#L7 TLDR lets make this `protected`

dsinghvi: and then all descendants can modify it how they like and when they like

dsinghvi (approved):

PR #4973, opened 2024-10-21
feat(python): Improve Interfaces for `MethodInvocation`

noanflaherty:
This generally makes the `MethodInvocation` AstNode more usable by doing three things:
1. Exports it from the `python` module
2. Let's you pass any `AstNode` as an arg instead of only `CodeBlock`s
3. Allows you to define a parent object that the method lives within.

### DIFF
--- a/generators/python-v2/codegen/src/ast/MethodInvocation.ts
+++ b/generators/python-v2/codegen/src/ast/MethodInvocation.ts
@@ -12,33 +12,33 @@ export declare namespace MethodInvocation {
         /* The arguments to pass to the method */
         arguments_: MethodArgument[];
         /* The parent object that the invoked method lives within, if any */
-        methodParent?: AstNode;
+        on?: AstNode;
     }
 }
 
 export class MethodInvocation extends AstNode {
     private methodName: string;
     private arguments: MethodArgument[];
-    private methodParent?: AstNode;
+    private on?: AstNode;
 
-    constructor({ method, arguments_, methodParent }: MethodInvocation.Args) {
+    constructor({ method, arguments_, on }: MethodInvocation.Args) {
         super();
 
         this.methodName = typeof method === "string" ? method : method.getName();
         this.arguments = arguments_;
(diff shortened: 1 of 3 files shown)


generators/python-v2/codegen/src/ast/__test__/__snapshots__/MethodInvocation.test.ts.snap
```diff
@@ -1,5 +1,10 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`MethodInvocation > should write a method invocation with a parent 1`] = `
+"parent_object.test_method("arg1", kwarg=42)
+"
```
noanflaherty, line 5: Shows the usage of the new `methodParent` property.

generators/python-v2/codegen/src/ast/MethodInvocation.ts
```diff
@@ -11,21 +11,37 @@ export declare namespace MethodInvocation {
         method: Method | string;
         /* The arguments to pass to the method */
         arguments_: MethodArgument[];
+        /* The parent object that the invoked method lives within, if any */
+        methodParent?: AstNode;
```
noanflaherty, line 15: See the test below for how this is used.

generators/python-v2/codegen/src/ast/MethodInvocation.ts
```diff
@@ -11,21 +11,37 @@ export declare namespace MethodInvocation {
...
         this.methodName = typeof method === "string" ? method : method.getName();
         this.arguments = arguments_;
+        this.methodParent = methodParent;
+
+        this.arguments.forEach((arg) => {
+            this.inheritReferences(arg);
+        });
+        if (this.methodParent) {
+            this.inheritReferences(this.methodParent);
+        }
```
noanflaherty, line 36: Improves reference management that probs should have been done previously.

PR #4977, opened 2024-10-21
feat(cli): Add IR for dynamic snippets

amckinney:
Re: https://github.com/fern-api/fern/pull/4842

This adds a new `dynamic` package to `ir-types-latest` to represent the `DynamicIntermediateRepresentation`. This IR is used as a more terse format tailored to generate dynamic snippets.

The `fern` CLI now includes a new `dynamic-ir` sub-command so it's easy to generate the dynamic IR for any given API definition, e.g.

```sh
$ fern dynamic-ir --api imdb ir.json
```

### DIFF
--- a/packages/cli/cli/tsconfig.json
+++ b/packages/cli/cli/tsconfig.json
@@ -11,6 +11,7 @@
     { "path": "../auth" },
     { "path": "../configuration" },
     { "path": "../docs-preview" },
+    { "path": "../dynamic-snippets" },
     { "path": "../generation/ir-generator" },
     { "path": "../generation/ir-migrations" },
     { "path": "../generation/local-generation/local-workspace-runner" },
--- a/packages/cli/ete-tests/package.json
+++ b/packages/cli/ete-tests/package.json
@@ -37,9 +37,7 @@
     "js-yaml": "^4.1.0",
     "node-fetch": "2.7.0",
     "strip-ansi": "^7.1.0",
-    "tmp-promise": "^3.0.3",
-    "url-join": "^5.0.0",
-    "qs": "6.11.2"
+    "tmp-promise": "^3.0.3"
   },
   "devDependencies": {
     "@types/jest": "^29.5.12",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4238,18 +4238,12 @@ importers:
       node-fetch:
         specifier: 2.7.0
         version: 2.7.0
-      qs:
-        specifier: 6.11.2
-        version: 6.11.2
       strip-ansi:
         specifier: ^7.1.0
         version: 7.1.0
       tmp-promise:
         specifier: ^3.0.3
         version: 3.0.3
-      url-join:
-        specifier: ^5.0.0
-        version: 5.0.0
     devDependencies:
       '@types/jest':
         specifier: ^29.5.12


fern/apis/dynamic-snippets/definition/endpoints.yml
```diff
@@ -0,0 +1,95 @@
...
+      token: commons.Name
+
+  EndpointHeaderAuth:
+    properties:
+      header: types.NamedParameter
+
+  EndpointOAuth:
+    properties:
+      clientId: commons.Name
+      clientSecret: commons.Name
```
dsinghvi, line 43: nit: would be nice to move these into an `auth.yml` for clarity

fern/apis/dynamic-snippets/definition/snippets.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  Values:
+    docs: |
+      Snippet values are represented as arbitrary key, value
+      pairs (i.e. JSON objects). The keys are expected to be
+      in the parameter's wire representation. For path parameters,
+      the name will match the parameter name.
+    type: map<string, unknown>
+
+  SnippetRequest:
```
dsinghvi, line 38: is it useful to encode a `SnippetResponse` (will make it easier to not break that API either)

amckinney: There's a lot we can add for responses to include in snippet generation (e.g. whether the endpoint responds with pagination, streaming, etc). 

The caller shouldn't need to specify any additional values in the `SnippetRequest` though - I imagine this would be an additional `response` property on the `Endpoint` type like so:

```yaml
types:
  Endpoint:
    properties:
      auth: optional<EndpointAuth>
      declaration: declaration.Declaration
      request: Request
      response: Response

  Request:
    docs: |
      Reperesents the request parameters required to call a specific endpoiont.
    union:
      body: BodyRequest
      inlined: InlinedRequest

  Response:
    docs: |
      Reperesents the response returned from the endpoint.
    union:
      unary: UnaryResponse
      fileDownload: FileDownloadResponse
      pagination: PaginationResponse
      streaming: StreamingResponse
```

With backwards compatibility in mind, I'm happy to start with just the `UnaryResponse` types as an empty object (it might someday have `json`, `text`, `bytes`, and so on). When this `UnaryResponse` with no properties is specified, the snippet will just be the traditional endpoint call like so:

```go
import (
  context "context"
  acme "github.com/buildwithfern/acme-go"
  client "github.com/buildwithfern/acme-go/client"
)

client := client.NewClient()
client.User.Create(
  context.TODO(),
  &acme.CreateUserRequest{
    Username: "john.doe",
    Status:   acme.StatusDeactivated,
  },
)
```

Then we can keep `response` typed as a required property, then roll out `FileDownload`, `PaginationResponse` and `StreamingResponse` later on in safe way.

--

Thoughts?

packages/ir-sdk/fern/apis/ir-types-latest/definition/dynamic/types.yml
```diff
@@ -0,0 +1,107 @@
...
+
+  NamedType:
+    docs: |
+      Represents the type of a parameter that can be used to generate a dynamic type.
+    union:
+      alias: AliasType
+      enum: EnumType
+      object: ObjectType
+      discriminatedUnion: DiscriminatedUnionType
+      undiscriminatedUnion: UndiscriminatedUnionType
```
dsinghvi, line 19: just a note for the future that we'll want to eventually add an `optional<boolean>` inline

packages/cli/dynamic-snippets/src/DynamicSnippetsConverter.ts
```diff
@@ -0,0 +1,404 @@
...
+    fernFilepath: FernFilepath;
+}
+
+export declare namespace DynamicSnippetsConverter {
+    interface Args {
+        ir: IntermediateRepresentation;
+    }
+}
+
+export class DynamicSnippetsConverter {
```
dsinghvi, line 47: 💯 so clean

dsinghvi (approved):

fern/apis/dynamic-snippets/definition/snippets.yml
amckinney: Discussed offline - the above looks good (and is included here), but we should still split out a separate `EndpointSnippetResponse` that contains a property for the snippet string. That way it's easy for us to support additional snippet variants later on.

PR #4980, opened 2024-10-22
chore(cli): ensure changelog generation does not duplicate entries

armandobelardo:
This PR clears out the changelog directory before regenerating it, attempting to ensure that the entries are unique
As a refresher, changelog generation works as follows:
- if the version is within our DB, then we use the date within the DB and write it to disk as is
- if the version is not within the DB, then it is assumed to be new, and todays date is used for the changelog entry

Fixes FER-3390

@coderabbitai summary


packages/seed/src/cli.ts
```diff
@@ -596,6 +596,12 @@ function addGenerateCommands(cli: Argv) {
                                     "Path to write the changelog to, if not provided, will write to cwd. Note this should be a directory, not a filename.",
                                 string: true,
                                 demandOption: false
+                            })
+                            .option("clean-directory", {
```
amckinney, line 600: Not sure what the _right_ name is here, but it's less important because this is scoped to `seed`.

amckinney (approved):

packages/seed/src/cli.ts
armandobelardo: Yeah totally, `overwrite` felt incorrect, etc
If we want to rename it down the line we can for sure since this isn't meant to be user facing

PR #4984, opened 2024-10-22
fix(cli): conjure importer reads from base-path

dsinghvi:
Previously, we didn't set the service base path but now we do.

### DIFF
--- a/packages/cli/api-importers/commons/src/index.ts
+++ b/packages/cli/api-importers/commons/src/index.ts
@@ -1,2 +1,7 @@
 export { APIDefinitionImporter } from "./APIDefinitionImporter";
-export { type FernDefinitionBuilder, FernDefinitionBuilderImpl, type FernDefinition, type HttpServiceInfo } from "./FernDefnitionBuilder";
+export {
+    type FernDefinitionBuilder,
+    FernDefinitionBuilderImpl,
+    type FernDefinition,
+    type HttpServiceInfo
+} from "./FernDefnitionBuilder";


packages/cli/api-importers/conjure/conjure-to-fern-tests/__test__/__snapshots__/trace.json
```diff
@@ -911,20 +883,44 @@
...
       "contents": {
-        "imports": {
-          "commons": "../commons/__package__.yml",
-          "playlist": "../PlaylistCrud/__package__.yml"
-        },
         "service": {
           "auth": false,
-          "base-path": "",
+          "base-path": "/v2/playlist",
```
dsinghvi, line 923: critical change

packages/cli/api-importers/commons/src/FernDefnitionBuilder.ts
```diff
@@ -125,16 +129,22 @@ export class FernDefinitionBuilderImpl implements FernDefinitionBuilder {
...
-        { displayName, docs }: { displayName?: string | undefined; docs?: string | undefined }
+        { auth, "base-path": basePath, "display-name": displayName, docs }: HttpServiceInfo
     ): void {
         const fernFile = this.getOrCreateFile(file);
         if (fernFile.service == null) {
             fernFile.service = {
-                auth: false,
-                "base-path": "",
+                auth: auth ?? false,
+                "base-path": basePath ?? "",
```
amckinney, line 138: This is a little redundant given the conditions below, but fine to keep as-is if you're worried about potential regressions.

dsinghvi: agreed was thinking the same, happy to just set it to defaults there

PR #4991, opened 2024-10-23
feat(go): Add dynamic snippet generator

amckinney:
This adds dynamic snippet generation in Go. The `DynamicSnippetsGenerator` receives an `EndpointSnippetRequest`, and responds with an `EndpointSnippetResponse` containing the generated snippet.

For example,

```go
import { DynamicSnippetsGenerator } from "@fern-api/go-dynamic-snippets";

// Initialize the dynamic IR and generator configuration.
const ir = { ... }
const config = { ... }

const generator = new DynamicSnippetsGenerator({ir, config});
const response = generator.generate({
  endpoint: {
    method: "POST",
    path: "/users"
  },
  requestBody: {
    username: "john.doe",
    status: "DEACTIVATED"
  },
});

console.log(response.snippet);

// Returns:
//
// package example
//
// import (
//     context "context"
//     acme "github.com/buildwithfern/acme-go"
//     client "github.com/buildwithfern/acme-go/client"
// )
// 
// func do() {
//     client := client.NewClient()
//     client.User.Create(
//         context.TODO(),
//         &acme.CreateUserRequest{
//             Username: "john.doe",
//             Status:   acme.StatusDeactivated,
//         },
//     )
// )
```


generators/go-v2/codegen/src/ast/Func.ts
```diff
@@ -0,0 +1,45 @@
...
+        return this.func.return_;
+    }
+
+    public get body(): CodeBlock | undefined {
+        return this.func.body;
+    }
+
+    public get docs(): string | undefined {
+        return this.func.docs;
+    }
```
dsinghvi, line 40: this is clean, feel like we should do this in other places as opposed to making attributes readonly

amckinney: Glad you noticed this! I was exploring options here when I asked myself, "What's the best way to restrict the scope of _this_ class, but delegate to a common structure?" Happy to use this pattern elsewhere as we go.

dsinghvi (approved):

PR #4996, opened 2024-10-24
fix(typescript): add jitter support and use fake timers in tests

dsinghvi:
Improvement: Add a variable jitter to the exponential backoff and retry.


carloslfu:
Cross-referencing this with vellum-ai/vellum-client-node#222 for people coming across the same bug.

PR #5000, opened 2024-10-25
feat(python): Class Instantiation & Improved Method Instantiation

noanflaherty:
This PR brings a new `ClassInstantiation` node and improves the existing `MethodInvocation` node.

It's pretty much all shared code between the two, so I included a `BaseInvocation` ast node.

### DIFF
--- a/generators/python-v2/codegen/src/ast/BaseInvocation.ts
+++ b/generators/python-v2/codegen/src/ast/BaseInvocation.ts
@@ -12,8 +12,6 @@ export declare namespace BaseInvocation {
         reference: Reference;
         /* The arguments to pass to the method */
         arguments_: MethodArgument[];
-        /* The parent object that the invoked method lives within, if any */
-        on?: AstNode;
     }
 }
 
@@ -21,7 +19,7 @@ export class BaseInvocation extends AstNode {
     protected reference: Reference;
     private arguments: MethodArgument[];
 
-    constructor({ reference, arguments_, on }: BaseInvocation.Args) {
+    constructor({ reference, arguments_ }: BaseInvocation.Args) {
         super();
 
         this.reference = reference;


generators/python-v2/codegen/src/ast/MethodInvocation.ts
```diff
@@ -4,54 +4,18 @@ import { AstNode } from "./core/AstNode";
...
-    interface Args {
-        /* The method to invoke */
-        method: Method | string;
-        /* The arguments to pass to the method */
-        arguments_: MethodArgument[];
-        /* The parent object that the invoked method lives within, if any */
-        on?: AstNode;
+    interface Args extends Omit<BaseInvocation.Args, "reference"> {
+        /* A reference to the method that you'd like to invoke */
+        methodReference: Reference;
```
amckinney, line 13: nit: Why remove `reference` and just add another `Reference` property? Thought son just specifying this instead?

```typescript
interface Args = BaseInvocation.Args
```

Alternatively,

```typescript
interface Args extends BaseInvocation.Args {}
```

amckinney (approved):

generators/python-v2/codegen/src/ast/MethodInvocation.ts
noanflaherty: At that point there truly is no difference between the two classes. This is kinda why I just went with the one `MethodInvocation` originally until the asks in [this thread](https://vellum-ai.slack.com/archives/C052UMCSFA8/p1729604635719589?thread_ts=1729551015.146029&cid=C052UMCSFA8). Happy to adjust once it's clear what we want. Do you prefer we:
A) Leave as is
B) Change to `reference` and keep the two classes essentially equivalent, but separate for future differentiation
C) Go back into consolidating into just one AstNode?

amckinney: Ah I see now. The issue is that `on` shouldn't be part of the `ClassInstantiaton` - that's only relevant to the `MethodInvocation` (since there's no instance to call the client constructor on, like there is with a method).

Two options:
1. Update `ClassInstantiation` to omit `on` (and not `reference).
2. Check out the `Func` AST node I wrote in [this PR](https://github.com/fern-api/fern/pull/4991/files#r1815440009) - it's very similar to what you're trying to accomplish, but tailored for Go:

```typescript
import { GoTypeReference } from "./GoTypeReference";
import { Comment } from "./Comment";
import { CodeBlock } from "./CodeBlock";
import { AstNode } from "./core/AstNode";
import { Writer } from "./core/Writer";
import { Method } from "./Method";
import { Parameter } from "./Parameter";
import { Type } from "./Type";

export declare namespace Func {
    type Args = Omit<Method.Args, "typeReference">;
}

export class Func extends AstNode {
    private func: Method;

    constructor({ name, parameters, return_, body, docs }: Func.Args) {
        super();
        this.func = new Method({ name, parameters, return_, body, docs });
    }

    public get name(): string {
        return this.func.name;
    }

    public get parameters(): Parameter[] {
        return this.func.parameters;
    }

    public get return_(): Type[] {
        return this.func.return_;
    }

    public get body(): CodeBlock | undefined {
        return this.func.body;
    }

    public get docs(): string | undefined {
        return this.func.docs;
    }

        writer.writeNode(this.func);
    }
}
```

noanflaherty: Ah, I failed to remove it from the Args interface, but my intention is to remove `on` entirely since `Reference` has `attributes`, which accomplishes the same thing already.

It's also valid in python to call a method like this `LevelA.LevelB.my_class_method()` and so classes should still have a way to perform `on` (again, imo through `Reference`'s `attributes`, as the test here shows).

noanflaherty: I've just pushed up a commit to remove `on` as intended. Hopefully this clears up how I see this being used.

amckinney (approved):

generators/python-v2/codegen/src/ast/MethodInvocation.ts
noanflaherty: Cool sounds like we'll go with option `A) Leave as is` from above, but know I'm happy to adjust to one of the other options as a fast follow if preferred!

PR #5005, opened 2024-10-25
fix(docs): section nodes should traverse and convert its children

abvthecity:
Fixes bug introduced in https://github.com/fern-api/fern/pull/4851 where section nodes were left out.

This also updates the way NodeId is generated so that 
- it doesn't get confused with the PageId.
- it is always unique

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,3 +1,11 @@
+
+- changelogEntry:
+    - summary: |
+        Fixes bug introduced in 0.45.0-rc20 where section children were dropped from the docs definition.
+      type: fix
+  irVersion: 53
+  version: 0.45.0-rc22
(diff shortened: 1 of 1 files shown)


packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
```diff
@@ -653,7 +654,7 @@ export class DocsDefinitionResolver {
             hidden: item.hidden,
             viewers: item.viewers,
             orphaned: item.orphaned,
-            children: [],
+            children: await Promise.all(item.contents.map((child) => this.toNavigationChild(id, child, slug))),
```
abvthecity, line 657: this is the main part that this PR fix.

dsinghvi (approved):

PR #5006, opened 2024-10-25
feat(python): Introduce `TypeInstantiation` AstNode

noanflaherty:
This introduces a new AstNode for instantiating primitive types.


generators/python-v2/codegen/src/ast/TypeInstantiation.ts
```diff
@@ -0,0 +1,179 @@
...
+    type: "none";
+}
+
+export class TypeInstantiation extends AstNode {
+    private constructor(private readonly internalType: InternalTypeInstantiation) {
+        super();
+    }
+
+    public static int(value: number): TypeInstantiation {
+        return new TypeInstantiation({ type: "int", value });
```
amckinney, line 68: Mind using `new this(...)` like we do elsewhere (e.g. [Go](https://github.com/fern-api/fern/blob/3839a27d51fb4a3c6c597b520999fb1cdbf75297/generators/go-v2/codegen/src/ast/TypeInstantiation.ts#L181))?

amckinney (approved):
Nice, this looks great!

generators/python-v2/codegen/src/ast/TypeInstantiation.ts
noanflaherty: Yeah no problem

PR #5007, opened 2024-10-25
feat(python): Support for UUIDs

noanflaherty:
This adds uuid support in `Type` and `TypeInstantiation`

### DIFF
--- a/generators/python-v2/codegen/src/ast/__test__/__snapshots__/TypeInstantiation.test.ts.snap
+++ b/generators/python-v2/codegen/src/ast/__test__/__snapshots__/TypeInstantiation.test.ts.snap
@@ -54,3 +54,8 @@ exports[`TypeInstantiation > tuple 1`] = `
 "(1, "two", True)
 "
 `;
+
+exports[`TypeInstantiation > uuid 1`] = `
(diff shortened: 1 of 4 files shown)


generators/python-v2/codegen/src/ast/__test__/__snapshots__/TypeInstantiation.test.ts.snap
```diff
@@ -54,3 +54,8 @@ exports[`TypeInstantiation > tuple 1`] = `
 "(1, "two", True)
 "
 `;
+
+exports[`TypeInstantiation > uuid 1`] = `
+"UUID("123e4567-e89b-12d3-a456-426614174000")
```
dsinghvi, line 59: would have expected an import to come through here, is UUID built in now?

noanflaherty: Imports are currently resolved when a `PythonFile` is written, but not included when a general AstNode is written. This is consistent with previous discussions and existing tests. I'm down to revisit this, but would ask that we do it as a separate effort/PR.

dsinghvi (approved):

PR #5016, opened 2024-10-27
chore(python): Make `TypeInstantiation` More Flexible

noanflaherty:
In integrating with the new `TypeInstantiation` AstNode, I found myself wishing that it was more flexible and allowed me to pass arbitrary other ast nodes to lists, sets, dicts, etc.


amckinney (approved):
Nice! This is often the direction AST nodes move in - we often start with a more strict interface, and extend out to arbitrary AST nodes as needed. Starting with this upfront works for me.

PR #5029, opened 2024-10-28
Start publishing python v2 codegen

dvargasfuertes:
@coderabbitai summary

### DIFF
--- a/.github/workflows/publish-generator-python-v2-codegen.yaml
+++ b/.github/workflows/publish-generator-python-v2-codegen.yaml
@@ -1,0 +1,80 @@
+# Generators are registered within seed.yml, once their tests are passing
+name: publish-and-register-python-v2-codegen
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "generators/pythonv2/codegen/versions.yml"
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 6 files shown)


.github/workflows/publish-generator-python-v2-codegen.yaml
```diff
@@ -0,0 +1,80 @@
+# Generators are registered within seed.yml, once their tests are passing
+name: publish-and-register-python-v2-codegen
```
dvargas92495, line 2: my general approach was to copy `csharp-model`

generators/python-v2/codegen/package.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "@fern-api/pythonv2-codegen",
+  "name": "@fern-api/python-v2-codegen",
```
dvargas92495, line 2: The directory has a dash, but the package name didn't. Not sure which you prefer

dvargas92495 (commented):
@dsinghvi could use additional guidance on:
- If there's anything I'm missing
- If there's anything I added that was unnecessary
- Anything else I need to resolve CI

dvargas92495:
closing and pushing directly to fern

PR #5031, opened 2024-10-28
feat(go): Add seed testing for dynamic snippets

amckinney:
## Overview

This adds support for generating dynamic snippet tests within `seed`. 

With this, we get a far better understanding of test coverage from every dynamic snippet generator. Not only are we confident the snippet we produce is valid Go code (i.e. it can be successfully formatted), we now verify that the snippet is fully compatible with the generated SDK against every generated example.

## Implementation

This ended up requiring a fair number of changes, the most important of which are outlined below:

1. The `AbstractFormatter` interface, which is optionally passed into the AST package (i.e. `@fern-api/go-codegen`).
    * The WASM-powered `gofmt` library is wrapped in a separate `@fern-api/go-formatter` package so that it can be omitted as needed (more on this below).
3. The `DynamicSnippetsTestSuite`, which defines a series of test cases for a particular API and generator configuration.
4. A few fixes to the Go SDK generator (all of which were caught with the help of the new dynamic snippet generator).
    * These will be released in version `0.28.1`.
5. The `DynamicSnippetsTestGenerator` produces valid Go files that can be compiled alongside the generated source files.
    * `seed` will generate each of these snippets as a separate package in the `dynamic-snippets` directory.
    * This is also compatible with `seed run` so we can easily test the dynamic snippets produced for arbitrary customer APIs.
6. The `seed` CLI is now built with a custom `build.cjs` file.
    * This excludes the `@fern-api/go-formatter` package due to what appears to be a WASM incompatibility with `tsup`.

### DIFF
--- a/seed/csharp-sdk/seed.yml
+++ b/seed/csharp-sdk/seed.yml
@@ -100,9 +100,17 @@ fixtures:
       outputFolder: read-only-memory
 allowedFailures:
   ## COMPILE FAILURES
-  - objects-with-imports
+  - alias-extends
+  - any-auth
+  - enum
   - examples:no-custom-config
   - examples:readme-config
+  - extends
+  - oauth-client-credentials
+  - oauth-client-credentials-default
+  - oauth-client-credentials-environment-variables
+  - oauth-client-credentials-nested-root
+  - objects-with-imports
   - mixed-file-directory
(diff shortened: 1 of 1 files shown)


dsinghvi (approved):
Super easy to understand, 🚀

PR #5035, opened 2024-10-29
fix(python): Fix Reference Inheritance for Callables & Types

noanflaherty:
We weren't correctly inheriting references for:
1. The callable used as a method invocation / class instantiation
2. Primitive types that can contain references (sets, lists, etc.)
3. Including imports for `Reference` objects

### DIFF
--- a/packages/cli/generation/ir-generator/src/__test__/test-definitions/mixed-case.json
+++ b/packages/cli/generation/ir-generator/src/__test__/test-definitions/mixed-case.json
@@ -2913,6 +2913,7 @@
                         }
                     },
                     "jsonExample": {
+                        "status": "ACTIVE",
                         "resource_type": "user",
                         "userName": "username",
                         "metadata_tags": [
@@ -3201,7 +3202,7 @@
                     "userSpecifiedExamples": [
                         {
                             "example": {
-                                "id": "625e8d23ceadb5d4efb47ccef83f3feebb7acd8051c753945c49f3c20f3dc769",
+                                "id": "0fed9fb2d42988bf41f50ad8d3024a6a6b993e9cc2b0e8abfeeacb915069fa0f",
                                 "name": null,
                                 "url": "/resource/rsc-xyz",
                                 "rootPathParameters": [],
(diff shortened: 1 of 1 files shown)


generators/python-v2/codegen/src/ast/Reference.ts
```diff
@@ -37,6 +37,7 @@ export class Reference extends AstNode {
         this.genericTypes = genericTypes ?? [];
         this.alias = alias;
         this.attribute = attribute ?? [];
+        this.references.push(this);
```
noanflaherty, line 40: This allows a `Reference` object to track itself as a reference. This simplifies lots of other logic such that we can safely look at `AstNode.references` to understand all of a nodes references, rather than add special conditionals in many places to check both `AstNode.references` and check to see whether the node is of type `Reference` and should explicitly be added.

generators/python-v2/codegen/src/ast/TypeInstantiation.ts
```diff
@@ -90,27 +91,40 @@ export class TypeInstantiation extends AstNode {
     }
 
     public static list(values: AstNode[]): TypeInstantiation {
-        return new this({ type: "list", values });
+        const list = new this({ type: "list", values });
+        values.forEach((value) => list.inheritReferences(value));
+        return list;
```
noanflaherty, line 96: We perform similar reference tracking in `python.Type` and failed to include it here in `python.TypeInstantiation`.

generators/python-v2/codegen/src/ast/__test__/TypeInstantiation.test.ts
```diff
@@ -67,12 +68,24 @@ describe("TypeInstantiation", () => {
         expect(await writer.toStringFormatted()).toMatchSnapshot();
     });
 
-    it("dict", async () => {
-        TypeInstantiation.dict([
-            { key: TypeInstantiation.str("one"), value: TypeInstantiation.int(1) },
-            { key: TypeInstantiation.str("two"), value: TypeInstantiation.bool(true) }
-        ]).write(writer);
-        expect(await writer.toStringFormatted()).toMatchSnapshot();
+    describe("dict", () => {
```
noanflaherty, line 71: Adds some tests to make sure that references within python primitives get picked up.

amckinney (approved):

PR #5038, opened 2024-10-29
fix(python): Fix Relative Import Paths

noanflaherty:
We were previously including the root module in the import path and going up an extra level. This is a bug that this PR fixes.

### DIFF
--- a/generators/python-v2/codegen/src/ast/PythonFile.ts
+++ b/generators/python-v2/codegen/src/ast/PythonFile.ts
@@ -102,7 +102,7 @@ export class PythonFile extends AstNode {
                 if (levelsUp > 0) {
                     relativePath += ".";
                 }
-                relativePath += refModulePath.slice(commonPrefixLength).join(".");
+                relativePath += refModulePath.slice(commonPrefixLength + 1).join(".");
 
                 // Write the relative import statement
                 writer.write(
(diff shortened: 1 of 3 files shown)


generators/python-v2/codegen/src/ast/__test__/PythonFile.test.ts
```diff
@@ -1,10 +1,7 @@
 import { python } from "../../../src";
 import { Writer } from "../core/Writer";
-import { Reference } from "../Reference";
-import { Class } from "../Class";
 import { Method } from "../Method";
 import { CodeBlock } from "../CodeBlock";
-import { Type } from "../Type";
```
noanflaherty, line 7: These were unused. It'd be cool if this package had lint rules to prevent this.

generators/python-v2/codegen/src/ast/PythonFile.ts
```diff
@@ -102,7 +102,7 @@ export class PythonFile extends AstNode {
                 if (levelsUp > 0) {
                     relativePath += ".";
                 }
-                relativePath += refModulePath.slice(commonPrefixLength).join(".");
+                relativePath += refModulePath.slice(commonPrefixLength + 1).join(".");
```
amckinney, line 105: Might be worth leaving a comment explaining why we add one. I assume it's because we assume it's under the same module root?

amckinney (approved):

noanflaherty:
Ah it turns out that this was an issue in my implementation, not on fern's side. Going to close this.

PR #5044, opened 2024-10-30
feat(php): Add content type to multipart uploads

ajgateno:


generators/php/sdk/src/endpoint/request/WrappedEndpointRequest.ts
```diff
@@ -214,19 +214,46 @@ export class WrappedEndpointRequest extends EndpointRequest {
...
+                        php.invokeMethod({
+                            method: "toMultipartFormDataPart",
+                            arguments_: [php.codeblock(`'${property.key.wireValue}'`)],
+                            on: php.codeblock(paramRef)
+                        })
+                    ],
+                    on: this.getRequestBodyArgument()
+                })
+            );
+        }
```
amckinney, line 256: I think we can split this out a bit - the only thing that is changing is the `arguments_` list, so how about this instead?

```typescript
const arguments_ [
  {
    name: "name",
    assignment: php.codeblock(`'${property.key.wireValue}'`)
  }
]
if (property.contentType != null) {
  arguments_.push(
    {
      name: "contentType",
      assignment: php.codeblock(`'${property.contentType}'`)
    }
  )
}
writer.writeNodeStatement(
  ...
)
```

I personally think it's fine to omit `multiline` in the `toMultipartFormDataPart` in both cases (since it's at most two parameters), but if you care to keep the formatting as-is, we could just base it on the length of arguments:

```typescript
{
  ...,
  multiline: arguments_.length > 1
}
```

ajgateno: done

amckinney (approved):
👏

generators/php/sdk/src/endpoint/request/WrappedEndpointRequest.ts
```diff
@@ -33,6 +33,7 @@ import {
 } from "./EndpointRequest";
 import { CodeBlock } from "@fern-api/generator-commons";
 import { assertNever } from "@fern-api/core-utils";
+import { FileUploadBodyProperty } from "@fern-fern/ir-sdk/serialization";
```
amckinney, line 36: I think we'll want to use `@fern-fern/ir-sdk/api` here, but I'm sure this works fine as-is because they share a common interface.

amckinney (commented):
Nicely done 👏

PR #5046, opened 2024-10-30
fix(cli): Log error when uploading proto API definitions fail to upload

fern-support:
@coderabbitai summary


amckinney (approved):
Nice 👍

PR #5057, opened 2024-10-31
feat(typescript): Implement type literal AST for TS

ajgateno:
This PR implements an AST node to represent type literals in the typescript codegen module.

### DIFF
--- a/generators/typescript/codegen/src/ast/TypeLiteral.ts
+++ b/generators/typescript/codegen/src/ast/TypeLiteral.ts
@@ -43,10 +43,6 @@ interface Tuple {
     values: TypeLiteral[];
 }
 
-interface IterableLiteral {
-    values: TypeLiteral[];
-}
-
 export class TypeLiteral extends AstNode {
     private constructor(public readonly internalType: InternalTypeLiteral) {
         super();
@@ -55,7 +51,7 @@ export class TypeLiteral extends AstNode {
     public write(writer: Writer): void {
         switch (this.internalType.type) {
             case "array": {
-                this.writeArray({ writer, array: this.internalType });
+                this.writeIterable({ writer, iterable: this.internalType });
                 break;
             }
             case "boolean": {
@@ -80,7 +76,7 @@ export class TypeLiteral extends AstNode {
                 break;
             }
             case "tuple": {
-                this.writeTuple({ writer, tuple: this.internalType });
+                this.writeIterable({ writer, iterable: this.internalType });
                 break;
             }
             default: {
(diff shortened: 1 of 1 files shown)


generators/typescript/codegen/src/ast/TypeLiteral.ts
```diff
@@ -0,0 +1,255 @@
...
+                    const parts = this.internalType.value.split("\n");
+                    const head = parts[0] + "\n";
+                    const tail = parts.slice(1).join("\n");
+                    writer.write(head.replaceAll("`", "\\`"));
+                    writer.writeNoIndent(tail.replaceAll("`", "\\`"));
+                    writer.write("`");
+                } else {
+                    writer.write(`"${this.internalType.value.replaceAll('"', '\\"')}"`);
+                }
+                break;
```
amckinney, line 98: Nice - for readability, what if we made a couple helper functions for this? I imagine something like:

```typescript
if (this.internalType.value.includes("\n") {
  this.writeStringWithBackticks(this.internalType.value);
  break;
}
this.writeStringWithQuotes(this.internalType.value);
break;
```

Then I imagine the helper might be as simple as the following:

```typescript
function writeStringWithBackticks({writer, value} : {writer: Writer; value: string; }) {
  writer.write("`");
  writer.writeNoIndent(this.internalType.value.replaceAll("`", "\\`"));
  writer.write("`");
}
```

Similarly, `writeStringWithQuotes` could use the same structure:

```typescript
function writeStringWithBackticks({writer, value} : {writer: Writer; value: string; }) {
  writer.write('"');
  writer.write(this.internalType.value.replaceAll('"', '\\"'));
  writer.write('"');
}
```

amckinney: Discussed offline - this actually handles multi-line strings better than I expected and we should do something similar elsewhere. Nicely done!

ajgateno: done

generators/typescript/codegen/src/ast/TypeLiteral.ts
```diff
@@ -0,0 +1,169 @@
...
+    type: "string";
+    value: string;
+}
+
+interface Tuple {
+    type: "tuple";
+    // TODO: In theory this should be a tuple type, not an array of types
+    valueTypes: Type[];
+    values: TypeLiteral[];
+}
```
amckinney, line 44: I wonder if the `value` and `valueType` should be a single interface so that it's impossible for the length to be inconsistent.

amckinney (approved):

PR #5058, opened 2024-10-31
feat(python): New `Operator` AstNode for Python

noanflaherty:
This PR introduces a new low level AstNode called `Operator`. For now, I'm just introducing a few operators that we at Vellum need immediately. The patterns are established to easily add more down the road (which we're likely to do). I didn't want to go too crazy adding operators until we've aligned on the pattern.

I'll note that this pattern is different than csharp, which currently seems to only have a single `Or` operator, modeled as its own AstNode. I assert this pattern is more DRY and extensible to support any future operator.

### DIFF
--- a/generators/python-v2/codegen/src/ast/Operator.ts
+++ b/generators/python-v2/codegen/src/ast/Operator.ts
@@ -1,8 +1,7 @@
 import { AstNode } from "./core/AstNode";
 import { Writer } from "./core/Writer";
 import { assertNever } from "@fern-api/core-utils";
-
-type OperatorType = "or" | "and" | "bitwiseLeftShift" | "bitwiseRightShift";
+import { OperatorType } from "./OperatorType";
 
 export declare namespace Operator {
     interface Args {
@@ -28,13 +27,13 @@ export class Operator extends AstNode {
 
     private getOperatorString(): string {
         switch (this.operator) {
-            case "or":
+            case OperatorType.Or:
                 return "or";
-            case "and":
+            case OperatorType.And:
                 return "and";
-            case "bitwiseLeftShift":
+            case OperatorType.LeftShift:
                 return "<<";
-            case "bitwiseRightShift":
+            case OperatorType.RightShift:
                 return ">>";
             default:
                 assertNever(this.operator);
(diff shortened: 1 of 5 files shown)


generators/python-v2/codegen/src/ast/Method.ts
```diff
@@ -30,7 +30,6 @@ export declare namespace Method {
         decorators?: Decorator[];
     }
 }
-[];
```
noanflaherty, line 33: Found some code that wasn't doing anything...

generators/python-v2/codegen/src/ast/Operator.ts
```diff
@@ -0,0 +1,49 @@
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+import { assertNever } from "@fern-api/core-utils";
+
+type OperatorType = "or" | "and" | "bitwiseLeftShift" | "bitwiseRightShift";
```
noanflaherty, line 5: We at Vellum actually only _really_ need `bitwiseRightShift` currently.

amckinney: Thoughts on defining an enum (like [Access](https://github.com/fern-api/fern/blob/9be2af2973a32c913af251e35c84cb7626231808/generators/php/codegen/src/ast/Access.ts#L3)) for this so it's a little easier to read at the call-site?

We could also remove `bitwise` since we're unlikely to need to use `leftShift` or `rightShift` for any other reason. It doesn't need to be an AstNode instance, but it could still be defined in its own `OperatorType.ts` file for discoverability:

```typescript
// OperatorType.ts
export type OperatorType = "or" | "and" | "leftShift" | "rightShift";

export const OperatorType = {
    Or: "or",
    And: "and",
    LeftShift: "leftShift",
    RightShift: "rightShift"
} as const;
```

From your test, we could instead write:
```typescript
const operator = python.operator({
    operator: OperatorType.LeftShift,
    lhs: python.TypeInstantiation.int(1),
    rhs: python.TypeInstantiation.int(2)
});
```

amckinney (approved):
Assuming you agree with the enum change, lgtm!

generators/python-v2/codegen/src/ast/Operator.ts
noanflaherty: Yep sounds good!

PR #5059, opened 2024-10-31
feat(go): Publish @fern-api/dynamic-ir-sdk

amckinney:
This publishes a new @fern-api/dynamic-ir-sdk package that includes the dynamic IR, as well as the `generator-exec` configuration used to run every generator.

With this, we can ship a single dependency that includes all the core dependencies required (excluding the language-specific snippet generator libraries)) to generate an offline SDK snippets module.

### DIFF
--- a/packages/ir-sdk/fern/apis/ir-types-latest/generators.yml
+++ b/packages/ir-sdk/fern/apis/ir-types-latest/generators.yml
@@ -20,6 +20,7 @@ groups:
         output:
           location: npm
           package-name: "@fern-api/dynamic-ir-sdk"
+          token: ${{NPM_TOKEN}}
         config:
           noOptionalProperties: true
           noSerdeLayer: true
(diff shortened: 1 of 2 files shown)


packages/ir-sdk/fern/apis/ir-types-latest/generators.yml
```diff
@@ -11,6 +11,18 @@ groups:
...
+      - dynamic
+    generators:
+      - name: fernapi/fern-typescript-node-sdk
+        version: 0.38.6
+        output:
+          location: npm
+          package-name: "@fern-api/dynamic-ir-sdk"
+        config:
+          includeUtilsOnUnionMembers: true
+          noOptionalProperties: true
```
dsinghvi, line 25: `noSerdeLayer: true` and no union utils

dsinghvi (approved):

PR #5062, opened 2024-11-01
feat(python): Support overriding default dev dependencies and mypy config in the python generator

dvargasfuertes:
We want to use mypy 1.11.1 in our sdk development. This allows us to specify it [here](https://github.com/vellum-ai/vellum-client-generator/pull/485/files#diff-1692c8a83676d09b3e8f5356903c7e554de5699701696de6943a96510a586931R86)

@coderabbitai summary


seed/python-sdk/alias-extends/pyproject.toml
dvargas92495, line 1: The changes in most of these files is mostly a side effect of dependencies now being alphabetized, which I think is likely preferable

seed/python-sdk/exhaustive/custom-mypy-config/.github/workflows/ci.yml
dvargas92495, line 1: The only file here that is intended to be affected by changes is `pyproject.toml`. It felt overkill for all of this other stuff to be generated too

I'm also not sure if I put this fixtures in the right spot. If I need to regen elsewhere, I'd love tips on how to run just my nested fixture so that I don't have to wait ~hour for tests to run
Screenshot 2024-11-07 at 8 30 20 AM

dvargas92495:
Confused why go-snippets would fail for python config specific tests?
Screenshot 2024-11-07 at 8 39 48 AM

PR #5066, opened 2024-11-01
feat(typescript): Add references to the new typescript AST library

ajgateno:
This PR adds references with imports to the typescript AST, which allows for implementing further nodes like function calls, class instantiation, etc.

### DIFF
--- a/generators/typescript/codegen/src/ast/core/Writer.ts
+++ b/generators/typescript/codegen/src/ast/core/Writer.ts
@@ -74,7 +74,11 @@ export class Writer extends AbstractWriter {
         const duplicates = [];
         for (const references of Object.values(this.imports)) {
             for (const ref of references) {
-                if (ref.importFrom?.type === "named" && ref.name === reference.name) {
+                if (
+                    ref.importFrom?.type === "named" &&
+                    ref.importFrom.moduleName !== reference.importFrom.moduleName &&
+                    ref.name === reference.name
+                ) {
                     duplicates.push(ref);
                 }
             }


generators/typescript/codegen/src/ast/core/Writer.ts
```diff
@@ -1,7 +1,81 @@
...
+                    throw new Error(
+                        `Cannot have multiple default imports for module ${reference.module.moduleName}: ` +
+                            `got ${reference.name} but already had ${existing.name}`
+                    );
+                }
+            }
+            let moduleImports = this.imports[reference.module.moduleName];
+            if (moduleImports == null) {
+                this.imports[reference.module.moduleName] = [];
+                moduleImports = this.imports[reference.module.moduleName];
```
dsinghvi, line 36: nit: 

```
this.imports[reference.module.moduleName] ??= [];
moduleImports = this.imports[reference.module.moduleName];
```

You can skip the `null` check with the fancy `??=` operator 😅

ajgateno: done

generators/typescript/codegen/src/ast/core/Writer.ts
```diff
@@ -1,7 +1,78 @@
...
+            this.imports[reference.module.moduleName] ??= [];
+            const moduleImports = this.imports[reference.module.moduleName];
+            if (moduleImports != null) {
+                const names = moduleImports.map((import_) => import_.name);
+                if (!names.includes(reference.name)) {
+                    moduleImports.push(reference);
+                }
+            }
+        }
+    }
```
amckinney, line 42: Thoughts on splitting this up a bit to reduce nesting? It's otherwise a little hard to follow at first glance given that there's a variety of conditions. What about something along the lines of the following? 

```typescript
public addImport(reference: Reference): void {
  if (reference.module == null) {
    return;
  }
  if (reference.module.defaultExport) {
    this.addDefaultExportOrThrow({ reference, moduleName: reference.module.name } );
    return;
  }
  const references = this.imports[reference.module.moduleName] ??= [];
  if (!references.includes(reference.name)) {
    references.push(reference);
  }
}
```

```typescript
private addReferenceOrThrow({ reference, moduleName }: { reference: Reference; moduleName: string; }): void {
  if (this.defaultImports[moduleName] != null) {
    throw new Error("Internal error; cannot have multiple default imports for module ...");
  }
  this.defaultImports[moduleName] = reference;
}
```

generators/typescript/codegen/src/ast/Reference.ts
```diff
@@ -0,0 +1,36 @@
+import { assertNever } from "@fern-api/core-utils";
+import { AstNode, Writer } from "./core";
+
+interface ModuleReference {
+    /* The name of the module the reference is imported from */
+    moduleName: string;
+    /* Whether the reference is the default export of the module */
+    defaultExport?: boolean;
+}
```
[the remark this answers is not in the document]
ajgateno: done

generators/typescript/codegen/src/ast/core/Writer.ts
ajgateno: done

ajgateno: Split it up into the different conditions

ajgateno: Ended up looking quite different than to this because I added stars

generators/typescript/codegen/src/ast/core/Writer.ts
```diff
@@ -1,7 +1,149 @@
...
+                    result += ` { ${stringifiedNonDefault} }`;
+                }
+                if (starImportAlias != null) {
+                    if (defaultImport != null || named.length > 0) {
+                        result += ", ";
+                    }
+                    result += ` * as ${starImportAlias}`;
+                }
+                result += ` from "${module}";\n`;
+            }
```
amckinney, line 145: We might be able to reduce nesting here too if we rearrange things a bit.

generators/typescript/codegen/src/ast/core/Writer.ts
```diff
@@ -1,7 +1,149 @@
...
+                this.starImportAliases[reference.importFrom.moduleName] = reference.importFrom.starImportAlias;
+                this.starImportAliasesInverse[reference.importFrom.starImportAlias] = reference.importFrom.moduleName;
+            } else if (existingAlias != null && existingAlias !== reference.importFrom.starImportAlias) {
+                throw new Error(
+                    "Cannot have more than one alias for non-named imports from a module: " +
+                        `got ${reference.importFrom.starImportAlias} but already have ${existingAlias}.`
+                );
+            }
+        }
+    }
```
amckinney, line 106: There's a fair amount of nesting in these functions - I'd try to reduce nesting as much as possible and split these out into separate functions so that it's a little easier to read and debug later on.

I think we can apply an approach similar to the one outlined in [this comment](https://github.com/fern-api/fern/pull/5066#discussion_r1826146433). For example, reversing the `if` condition at the top would reduce the entire function nesting:

```typescript
if (reference.importFrom?.type !== "star") {
  return;
}
...
```

In summary, I'd try to:
1. Reduce nesting and return early as much as possible. Although [this guide](https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting) is focused on Go, the same ideas apply here.
2. Add `*OrThrow` functions to handle error cases.
3. With the restructured control flow, simplify the conditions used in each `if` down to a minimum.

generators/typescript/codegen/src/ast/Reference.ts
```diff
@@ -0,0 +1,47 @@
+import { AstNode, Writer } from "./core";
+
+export declare namespace Reference {
```
amckinney, line 3: Nice, this AstNode looks great.

amckinney (approved):
The functionality looks great, so stamping as-is, but I think we can improve code readability a bit. I trust you to make edits as you see fit, and we can go from there!

generators/typescript/codegen/src/ast/core/Writer.ts
dsinghvi: What about something like this? We basically have a function that can turn a single module + multiple imports into statements. each statement is just an import

```ts
type Import = DefaultImport | NamedImport | StartImport; 

export function stringifyImports(): string[] {
  const importsByModule: Record<Module, Import[]>  = getImportsByModule();
  const statements = Object.entries(importsByModule).flatMap((module, imports) => {
     return  generateStatementsForImports(module, imports); 
  });
  return statements;
}

export function generateStatementsForImports(module, imports): string[] {
  const statements = []
  const nonStartImportStatement: string = "";
  for (const import of imports) {
    switch (import.type) {
       case "named": 
          nonStartImportStatement.push(named.value);
        case "startImport": 
           statements.push(`import * as ${import.alias} from ${module}`)
    }
  }
  return statements;
}
```

generators/typescript/codegen/src/ast/core/Writer.ts
```diff
@@ -1,7 +1,157 @@
...
+
+        const duplicates = [];
+        for (const references of Object.values(this.imports)) {
+            for (const ref of references) {
+                if (ref.importFrom?.type === "named" && ref.name === reference.name) {
+                    duplicates.push(ref);
+                }
+            }
+        }
+        // TODO: Need to be able to resolve conflicts instead of throwing
```
amckinney, line 82: Discussed offline - 

We'll need to build this so that it's resilient to import conflicts. For example, consider the following:

```typescript
export { File } from "./acme/foo/File";
export { File } from "./acme/bar/File";
```

We don't want to throw in these cases - the generator should handle the alias and assign it as needed.

```typescript
export { File } from "./acme/foo/File";
export { File as File_ } from "./acme/bar/File";
```

amckinney (approved):
Discussed offline - there's some things we'll want to revisit for the import management, but this is a great first step in the right direction. We can handle those changes in a follow-up. Well done!

PR #5068, opened 2024-11-01
feat(docs): broken link detector

abvthecity:
Closes FER-792

Checks if links referenced in the markdown are broken, in a fern check rule.

1. runs the DocsResolver inside of the workspace. This is to generate the `FernNavigation.RootNode`
2. traverses the RootNode to detect all visitable slugs (links)
3. traverses the markdown files to scrape for all `href` and `src` whether in html or jsx

----

For each `href`/`src`, check if
1. is it an external url? (if it has a protocol like `https:`, `data:`, `mailto:` it will be ignored)
2. if it's an external url, but the origin matches one of the docs instances (for example: `[link](https://buildwithfern.com/learn/api-reference)`), we will detect this and validate it as if it's an internal url
3. if it's an internal url, is it a file (ending in `.mdx` or `.md` or `.png` etc)? if so, check if the file exists on disk
4. if it's an internal url with no endings, and comes from `href`, validate that the slug matches a real markdown file
5. if the internal url matches a redirect rule, follow the redirect rule

----

Edge cases considered:
1. a markdown file may be attached to multiple slugs (i.e `/docs`, `/v1/docs`, `/v2/docs`). if so, relativized paths must match across all of these slugs

----

Future idea:
- we can probably do a `fetch()` on any `https://` links that are detected and do a quick and dirty external broken link checker inside the CLI itself.

----

tests were not written b/c vitest/CJS is borked

### DIFF
--- a/packages/cli/yaml/docs-validator/src/rules/playground-environments-exist/__test__/playground-environments-exist.disabled-test.ts
+++ b/packages/cli/yaml/docs-validator/src/rules/playground-environments-exist/__test__/playground-environments-exist.disabled-test.ts
@@ -1,3 +1,5 @@
+// This test is disabled because it fails due to an issue with fdr-sdk's package structure.
+// TODO: Re-enable when fdr-sdk is ESM compatible
 import { AbsoluteFilePath, RelativeFilePath, join } from "@fern-api/fs-utils";
 
 import { getViolationsForRule } from "../../../testing-utils/getViolationsForRule";


packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/collect-pathnames.ts
```diff
@@ -71,7 +90,12 @@ export function collectPathnamesToCheck(
 
         pathnamesToCheck.push({
             pathname,
-            position: link.position,
+            sourceFilepath: link.sourceFilepath,
+            position: link.position && {
+                ...link.position,
+                start: { ...link.position.start, line: link.position.start.line + frontmatterOffset },
+                end: { ...link.position.end, line: link.position.end.line + frontmatterOffset }
+            },
```
dsinghvi, line 98: nit: seems like we have the same logic copy pasta-ed in a couple places

```ts
sourceFilepath: link.sourceFilepath,
            position: link.position && {
                ...link.position,
                start: { ...link.position.start, line: link.position.start.line + frontmatterOffset },
                end: { ...link.position.end, line: link.position.end.line + frontmatterOffset }
            },
```

i could see abstracting that to a single helper

packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
```diff
@@ -160,18 +180,23 @@ export class DocsDefinitionResolver {
 
         // postprocess markdown files after uploading all images to replace the image paths in the markdown files with the fileIDs
         const basePath = this.getDocsBasePath();
+
+        // TODO: include more (canonical) slugs from the navigation tree
+        const markdownFilesToPathName: Map<AbsoluteFilePath, string> = new Map();
```
dsinghvi, line 185: nit: typically the rest of the codebase will use `Record` since it is more typesafe than a map

Defer to you what you think is better here (sometimes there are better reasons to use a map like preserving order of insertion)

packages/cli/cli/src/commands/validate/logViolations.ts
```diff
@@ -26,44 +27,79 @@ export function logViolations({
         logViolationsSummary({ context, stats, logWarnings });
     }
 
-    violations.forEach((violation) => {
-        if (violation.severity === "fatal" || violation.severity === "error") {
-            logViolation({ violation, context });
-        }
-    });
+    // Group all violations by nodePath
```
amckinney, line 30: nit: Redundant comment.

packages/cli/cli/src/commands/validate/validateDocsBrokenLinks.ts
```diff
@@ -0,0 +1,37 @@
...
+    errorOnBrokenLinks
+}: {
+    project: Project;
+    cliContext: CliContext;
+    errorOnBrokenLinks: boolean;
+}): Promise<void> {
+    const docsWorkspace = project.docsWorkspaces;
+
+    if (docsWorkspace == null) {
+        // Should this fail and throw?
```
amckinney, line 20: Probably, unless `validateDocsBrokenLink` is called on _every_ Project instance, regardless of whether or not it has a docs configuration.

A no-op like it is now is fine in that case.

packages/cli/cli/src/commands/validate/validateDocsWorkspaceAndLogIssues.ts
```diff
@@ -1,44 +1,53 @@
...
+    errorOnBrokenLinks?: boolean;
     logSummary?: boolean;
 }): Promise<{ hasErrors: boolean }> {
-    const violations = await validateDocsWorkspace(workspace, context, loadAPIWorkspace);
-    const { hasErrors } = logViolations({ violations, context, logWarnings, logSummary });
+    const violations = await validateDocsWorkspace(workspace, context, fernWorkspaces);
+    let { hasErrors } = logViolations({ violations, context, logWarnings, logSummary });
+
+    if (errorOnBrokenLinks) {
+        hasErrors = hasErrors || violations.some((violation) => violation.name === "valid-markdown-links");
```
amckinney, line 26: It might be worth updating the `valid-markdown-links` name to to `no-broken-links` here so that the flag matches the name upfront. But we can do this in a follow-up if you prefer.

packages/cli/docs-resolver/src/visitNavigationAst.ts
```diff
@@ -0,0 +1,137 @@
...
+        paginated: noop,
+        playground: noop,
+        flattened: noop,
+        path: noop,
+        page: noop,
+        contents: async (items: docsYml.RawSchemas.NavigationItem[] | undefined): Promise<void> => {
+            if (items == null) {
+                return;
+            }
+            items.map(async (item) => {
```
amckinney, line 101: Let's make sure to fix up the `eslint` errors before merging.

packages/cli/yaml/docs-validator/src/ValidationViolation.ts
```diff
@@ -2,6 +2,7 @@ import { NodePath } from "@fern-api/fern-definition-schema";
 import { RelativeFilePath } from "@fern-api/fs-utils";
 
 export interface ValidationViolation {
+    name?: string;
```
amckinney, line 5: Fine as-is for now - in the future I bet we'll make this required.

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * For a given pathname, there's two scenarios:
+ * 1. the pathname is a slug to a page in the docs (e.g. `/docs/my-page`)
+ * 2. the pathname is a path to a file in the docs (e.g. `/docs/my-file.md`)
+ *
+ * Crucially, the current file may referenced by two slugs, so if the target pathname contains a relativized path,
+ * we will need to check that if the relativized paths can be found on all of the slugs for the current file.
+ */
+export async function checkIfPathnameExists(
+    pathname: string,
```
amckinney, line 18: nit: Why not include the `pathname` arg alongside the rest of the in-lined argument struct? That way _every_ parameter is named rather than an anonymous parameter to begin.

I get what you're going for with the function name (i.e. the pathname is the thing I'm acting upon), but I personally like using the named struct for everything, especially for consistency with the rest of the codebase.

Feel free to take it or leave it.

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/collect-links.ts
```diff
@@ -0,0 +1,192 @@
...
+            return;
+        });
+    } while (contentQueue.length > 0);
+
+    return { links, sources };
+}
+
+// acorns and other errors are being thrown when we parse the markdown
+// so we'll just ignore these markdown files (or descriptions) for now.
+// TODO: fix this
```
amckinney, line 176: Should we do this here? If not, it might be worth adding a little more color to what needs to be fixed.

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/valid-markdown-link.ts
```diff
@@ -1,25 +1,198 @@
...
+
+                // if this happens, this probably means that the current file is omitted from the docs navigation
+                // most likely due to a slug collision. This should be handled in a different rule.
+                if (!slugs || slugs.length === 0) {
+                    return [];
+                }
+
+                // if this file cannot be indexed (noindex=true, or hidden=true), then we don't need to check for links
+                // since it will not hurt SEO.
+                // TODO: we should at least report this somehow, since it's a bad UX.
```
amckinney, line 91: Should we report this as a warning now that we've distinguished `fatal`, `error`, and `warning` more clearly?

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/valid-markdown-link.ts
```diff
@@ -1,25 +1,198 @@
...
+                                message
+                            };
                         });
+                    })
+                );
+
+                return [...violations, ...pathToCheckViolations.flat()];
+            },
+            apiSection: async ({ workspace: apiWorkspace, config }) => {
+                // TODO: this is duplicating work from above, but there's no clean way to associate this visitor with the docs resolver
```
amckinney, line 132: If we don't plan to solve it in the near term, I'd just drop the comment.

amckinney (approved):
Nice!! Mostly nits and a few other comments, but this generally looks great to me. It'd also be good to get @abvthecity's eyes on this one since he has some additional context.

packages/cli/cli/src/commands/validate/logViolations.ts
rpc333: I use AI to generate ONE thing 😭 (fixed)

packages/cli/yaml/docs-validator/src/ValidationViolation.ts
rpc333: Yeah I was thinking that'd be good to do either as a follow-up or just incrementally so that this PR doesn't get even more huge

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
rpc333: Nah I think you're right

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/collect-links.ts
rpc333: @abvthecity ?

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/valid-markdown-link.ts
rpc333: @abvthecity what do you think about making these ones `severity: "warning"` while keeping the others `severity: "error"` (though not `"fatal"`) in our new paradigm?

packages/cli/cli/src/commands/validate/validateDocsWorkspaceAndLogIssues.ts
rpc333: yeah was planning to do this as a follow-up

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/collect-links.ts
abvthecity: leveraging this logic will solve this issue: https://github.com/fern-api/fern-platform/blob/main/packages/fern-docs/mdx/src/sanitize/sanitize-mdx-expression.ts

rpc333: will come back and do a follow-up where we make `mdx` its own npm package and then use that down here

rpc333:
Just adding note for future reference that we're disabling the `playground-environments-exist` in `docs-validator` until issues in `fdr-sdk` are resolved. Discussion on that here:
https://buildwithfern.slack.com/archives/C05E0S0A9SN/p1736539366609429

PR #5071, opened 2024-11-02
feat(cli): Allow grpc and http endpoints in a single service

Swimburger:
@coderabbitai summary

@amckinney Let's review this next week.
I don't actually see the `encoding` property used anywhere, so I wonder if we can remove that.

### DIFF
--- a/packages/cli/ete-tests/src/tests/jsonschema/jsonschema.test.ts
+++ b/packages/cli/ete-tests/src/tests/jsonschema/jsonschema.test.ts
@@ -15,5 +15,5 @@ describe("jsonschema", () => {
             await readFile(join(pathOfDirectory, RelativeFilePath.of("schema.json")), "utf-8")
         );
         expect(jsonSchema).toMatchSnapshot();
-    }, 10_000);
+    }, 20_000);
 });


packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
```diff
@@ -371,19 +377,24 @@ export function getHeaderName({ headerKey, header }: { headerKey: string; header
...
-                proto: undefined
-            };
-        case "grpc":
-            return {
-                json: undefined,
-                proto: {}
-            };
-        default:
-            assertNever(transport);
+function convertEndpointSourceToEncoding(source: RawSchemas.SourceSchema | undefined): Encoding {
```
amckinney: Can we rename this `convertSourceToEncoding` and use it for both the service and endpoint? It looks like we have access to the `SourceSchema` in both instances, so we may as well keep it consistent.

packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
```diff
@@ -371,19 +377,24 @@ export function getHeaderName({ headerKey, header }: { headerKey: string; header
...
+        return {
+            json: {},
+            proto: undefined
+        };
+    }
+    if ("proto" in source) {
+        return {
+            json: undefined,
+            proto: {}
+        };
```
amckinney: We can use [isRawProtobufSourceSchema](https://github.com/fern-api/fern/blob/23bece6128e97891bd6ccb2954420732bdd8d0ba/packages/cli/fern-definition/schema/src/utils/isRawProtobufSourceSchema.ts#L3) for this - we could also roll out an `openapi` variant so that this reads a little more clearly.

packages/cli/generation/ir-generator/src/converters/services/convertTransport.ts
```diff
@@ -34,3 +34,59 @@ export async function convertTransport({
...
+    sourceResolver
+}: {
+    file: FernFileContext;
+    serviceTransport: Transport;
+    endpointDeclaration: RawSchemas.HttpEndpointSchema;
+    sourceResolver: SourceResolver;
+}): Promise<Transport | undefined> {
+    const isGrpcService = serviceTransport.type === "grpc";
+    const isHttpService = serviceTransport.type === "http";
+    const isGrpcEndpoint = endpointDeclaration.source !== undefined && "proto" in endpointDeclaration.source;
```
amckinney, line 51: Let's use [isRawProtobufSourceSchema](https://github.com/fern-api/fern/blob/23bece6128e97891bd6ccb2954420732bdd8d0ba/packages/cli/fern-definition/schema/src/utils/isRawProtobufSourceSchema.ts#L3) here too.

packages/cli/generation/ir-generator/src/converters/services/convertTransport.ts
```diff
@@ -34,3 +34,59 @@ export async function convertTransport({
...
+    const isHttpEndpoint = !isGrpcEndpoint;
+    if (isHttpService && isHttpEndpoint) {
+        // if the service is http, the endpoint should inherit the service transport
+        return undefined;
+    }
+    if (isHttpService && isGrpcEndpoint) {
+        throw new Error("Cannot have a grpc endpoint on an http service");
+    }
```
amckinney, line 59: I imagine we'll eventually want to support this, but it seems fine to keep this as-is for now. This should be impossible for now based on how we determine the service-level transport.

packages/cli/generation/ir-generator/src/converters/services/convertTransport.ts
```diff
@@ -34,3 +34,59 @@ export async function convertTransport({
...
+                source: resolvedSource,
+                serviceNameOverride
+            });
+            if (protobufService == null) {
+                throw new Error(`Failed to resolve service name from ${resolvedSource.relativeFilePath}.`);
+            }
+            return Transport.grpc({
+                service: protobufService
+            });
+        }
```
amckinney, line 88: Can we reuse the `getTransportForService` logic here? It looks like this is doing the same thing. Maybe we can refactor the signature a bit so a common helper could be used by both the transport and endpoint functions?

packages/cli/generation/ir-generator/src/converters/services/convertTransport.ts
```diff
@@ -34,3 +34,59 @@ export async function convertTransport({
...
+    } else {
+        throw new Error("This should never happen");
```
amckinney, line 90: If we must throw in this case, let's give a more informative error. Something like
```suggestion
        throw new Error(`Internal error; failed to determine endpoint "${endpoint.name}" transport`);
```

packages/cli/generation/ir-generator/src/converters/services/convertTransport.ts
```diff
@@ -34,3 +34,59 @@ export async function convertTransport({
         service: protobufService
     });
 }
+
+export async function getTransportForEndpoint({
```
amckinney, line 38: I left a few comments based on the current structure, but I wonder if we could simplify this to:
1. If the service's transport is not gRPC, return undefined.
2. If the service's source and endpoint's source don't match, set the endpoint's transport (i.e. it's an override).

That way, it's technically possible for a single Fern service to use multiple gRPC services, or a different combination of transports (e.g. gRPC and HTTP).

Thoughts?

amckinney (commented):
Looks good so far! A few notes on how to adjust the control flow in determining the endpoint's transport, but otherwise on the right track.

packages/cli/generation/ir-generator/src/converters/services/convertTransport.ts
Swimburger: My thinking here too!

packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
Swimburger: This has changed with the commit I didn't push.
It's now converting it based on the calculated transport and whether there are any HTTP endpoints mixed in.

packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
Swimburger: Encoding is only in service at the moment. I can add it to the endpoint, too, but I don't know if we even need this property. It's not used anywhere AFAIK.

packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
```diff
@@ -380,10 +386,25 @@ function convertTransportToEncoding(transport: Transport): Encoding {
             };
         case "grpc":
             return {
-                json: undefined,
+                json: serviceHasHttpEndpoints(service) ? {} : undefined,
```
amckinney, line 389: Discussed offline - we can nix this for now. Eventually we'll want to support per-endpoint level encoding, but for now we'll keep the contract as:

1. By default, all endpoints inherit the transport/encoding specified on the service.
2. If the endpoint defines its own transport/encoding, it will override the service-level value.

amckinney (approved):
Nicely done!

amckinney (commented):
You can update and release the version by editing [this](https://github.com/fern-api/fern/blob/c8f28b15e78f50434e83ba2d5e23426b67fba4d4/packages/ir-sdk/fern/apis/ir-types-latest/VERSION#L1) to a new minor. Let's do that in this PR, too!

Swimburger:
@amckinney I updated the version. I also updated the snapshots of the tests, should this be part of this PR?

amckinney (approved):
🎉

Issue #3453, opened 2024-04-24
How to set user-agent string in generated SDKs


Arkessler:
### What can be improved

I'm trying to figure out if it's currently possible to set a user-agent string in generated SDKs. 

I found this Issue about it: https://github.com/fern-api/fern/issues/2264. Hard to tell from the history on that Issue whether support was added, and if it was I can't find "user-agent" anywhere in the docs.

If this is supported, could somebody advise on how to set it?

mstade:
@dannysheridan you closed #2264 that this ticket references – any chance you can provide any guidance on this? Thank you! 🙏

dannysheridan:
Howdy @Arkessler and @mstade! Sending a `user-agent` string (e.g., `User-Agent: ${package}/${version}`) is not currently supported.

Today Fern includes the following request headers for the same purpose:
```ts
"X-Fern-Language": "${language}",
"X-Fern-SDK-Name": "${package}",
"X-Fern-SDK-Version": "${version}",
```

A user of Fern has added a UserAgent themselves. If helpful, here's a link the PR: https://github.com/AssemblyAI/assemblyai-java-sdk/pull/105/files

dannysheridan:
@amckinney do you plan to support this feature? If not, let's close this issue as `not planned`.

amckinney:
This feature is being rolled out now that its part of Fern's IR ([PR](https://github.com/fern-api/fern/pull/4232)). For example, this is available in [TypeScript](https://github.com/fern-api/fern/pull/4234) and [C#](https://github.com/fern-api/fern/pull/4380). 

Fern SDKs will send the `User-Agent` header in the `<package>/<version>` format. If you want to send a custom `User-Agent` header, you can set an arbitrary HTTP header using the specific language's mechanism. 

For example, consider the following Go use case:

```go
header := make(http.Header)
header.Add("User-Agent", "your/custom/user-agent")
client := acmeclient.NewClient(
  option.WithHTTPHeader(header), // These headers are sent on every request.
)
```

PR #5085, opened 2024-11-03
fix(typescript): alias imported types in (de)serialization schemas

mstade:
This fixes an issue where a type imported in a (de)serialization schema has the same name as the exported schema, causing build errors.

Without this fix, we'd have schemas generated looking like so:

```ts
/**
 */

import * as serializers from "../../../index";
import * as Engine from "../../../../api/index";
import * as core from "../../../../core";
import { SecurityFinding } from "../../ocsf/resources/v110/resources/securityfinding/resources/classes/types/SecurityFinding";

export const SecurityFinding: core.serialization.ObjectSchema<serializers.SecurityFinding.Raw, Engine.SecurityFinding> = SecurityFinding;

export declare namespace SecurityFinding {
    type Raw = SecurityFinding.Raw;
}
```

The imported `SecurityFinding` type shares the same name as the serialization schema, which then causes build errors that are hard and awkward to fix in post processing of the generated source code.

After this fix, the generated code looks like so:

```ts
/**
 */

import * as serializers from "../../../index";
import * as Engine from "../../../../api/index";
import * as core from "../../../../core";
import { SecurityFinding as SecurityFindingType } from "../../ocsf/resources/v110/resources/securityfinding/resources/classes/types/SecurityFinding";

export const SecurityFinding: core.serialization.ObjectSchema<serializers.SecurityFinding.Raw, Engine.SecurityFinding> = SecurityFindingType;

export declare namespace SecurityFinding {
    type Raw = SecurityFindingType.Raw;
}
```

The added `Type` suffix ensures the names do not conflict, solving any build issues that would otherwise arise.

@coderabbitai summary


generators/typescript/sdk/generator/src/contexts/type-schema/TypeSchemaContextImpl.ts
```diff
@@ -146,7 +146,7 @@ export class TypeSchemaContextImpl implements TypeSchemaContext {
                       useDynamicImport: false,
                       namespaceImport: "serializers"
                   }
-                : { type: "direct" },
+                : { type: "direct", alias: `${typeName.name.originalName}Type` },
```
mstade, line 149: The alias here is identical to the alias on line 178. If they aren't the same type is imported twice, with different aliases. (Or indeed, if no alias is specified with the conflicting name.)

generators/typescript/sdk/generator/src/contexts/type-schema/TypeSchemaContextImpl.ts
```diff
@@ -175,7 +175,7 @@ export class TypeSchemaContextImpl implements TypeSchemaContext {
                         return { type: "fromRoot", useDynamicImport: false, namespaceImport: "serializers" };
                     } else if (isGeneratingSchema) {
                         // Return default import strategy or another strategy based on your logic
-                        return { type: "direct" };
+                        return { type: "direct", alias: `${typeName.name.originalName}Type` };
```
mstade, line 178: See previous comment about line 149, and the aliases being identical. It may be meaningful to extract this into a utility function to make this less prone to breakage. On the other hand, importing the same type twice with different aliases isn't going to break anything, even if it is redundant.

mstade (commented):
## Author Notes

@dsinghvi we've been stuck on v 0.12.7 of the TS SDK for some time due to this issue. I believe the issue was introduced with #3206, but the changes there seem sound and does improve performance. The changes I'm proposing here does not seem to break anything else in _our_ project, but I have to admit all the indirection in this project made it extremely hard to even pinpoint where the issue is introduced, let alone trying to fix it.

This _seems_ like a safe change to make, but your guidance here is much appreciated.

mstade:
The failed [preview-docs / run (pull_request)](https://github.com/fern-api/fern/actions/runs/11655191008/job/32449579614?pr=5085#step:4:47) check  does not seem to be related to these changes, unless I'm missing something, here's the error message:

```
Failed to read markdown file "../snippets/peace-lily.mdx" referenced in /home/runner/work/fern/fern/fern/pages/fern-docs/content/reusable-snippets.mdx
```

mstade:
The failed [seed / typescript-sdk (pull_request)](https://github.com/fern-api/fern/actions/runs/11655191004/job/32449587230?pr=5085) check however does seem related to these changes. Looks like the test snapshots need updating. I've tried running `pnpm test:update` but alas, it fails with errors like:

```
Missing required dependency; please install 'protoc-gen-openapi' to continue (e.g. 'brew install go && go install github…
```

The error message is cut, but not when trying to run tests in which case it says to run:

```sh
brew install go && go install github.com/google/gnostic/cmd/protoc-gen-openapi@latest
```

I did try this, but to no avail. I'm not quite sure how to update these snapshots now.

mstade:
@dsinghvi I wonder if maybe you have some insights into how I may update the snapshots?

mstade:
> @dsinghvi do you think there's any value in this change?
> 
> It fixes a very real issue in production but of course if this is the wrong approach or a won't fix I'd be happy to receive such feedback so we can adjust accordingly.

@amckinney apologies if you're not the right person to ping, I just saw your comment on #3453 and figured since I've not been able to get a response from @dsinghvi maybe you could help? Is this PR worth pursuing you think, or should we resolve this issue some other way?

mstade:
Obviously not a priority so I'll just close this and use a fork instead. Thank you.

Swimburger:
This has been fixed in [#12899](https://github.com/fern-api/fern/pull/12899), shipping in TypeScript SDK generator v3.51.5.

Root cause: The serialization schema generator used direct imports (import { Foo } from "...") for cross-package type references. When two packages defined types with the same exported name (e.g., Foo in both folder-b and folder-c), the generated schema file would contain duplicate identifiers.

Type files were unaffected because they already use namespace imports (import * as SeedCrossPackageTypeNames from "..."), but serialization schemas did not.

Fix: The schema generator now tracks which exported names have been directly imported per schema file. When a collision is detected (same name from a different package, or matching the schema's own name), it switches to a namespace import (serializers.folderC.Foo) instead of a direct import. Non-colliding types continue to use direct imports — no changes to existing generated output for any other fixture.

PR #5086, opened 2024-11-03
feat(typescript): add `outputProjectFiles` option when generating locally

### LINKED ISSUE #3257: fernapi/fern-typescript-browser-sdk does not generate project files
## Describe the Bug

Output from running the TypeScript generator with output `local-file-system` does not include `package.json` and other project files. While this can be useful in so far as allowing greater project customization than the `npm` output target, it makes it much harder to keep up with things like managing the dependencies the SDK requires.

### CLI Version

`0.19.18`

### Generator Version

`fernapi/fern-typescript-browser-sdk@0.12.7`

Full `generators.yml`:

```yaml
default-group: local
groups:
  local:
    generators:
      - name: fernapi/fern-typescript-browser-sdk
        version: 0.12.7
        output:
          location: local-file-system
          path: ../generated/sdk-typescript
```

### API Definition

The minimal OpenAPI specification or Fern definition required to reproduce the bug.

```yaml
types:
  Pet:
    name: string
    age: double
```

### Actual SDK

Only source files are included.

### Expected SDK

Expected project files like `package.json` as well as the generated source code.

## Additional Context

If this behaviour is intentional, it should probably be documented here: https://docs.buildwithfern.com/generate-sd-ks/generators/type-script-node-js-sdk#local

mstade:
This adds a new option `outputProjectFiles` which, when generating an SDK locally, will output the entire project as opposed to just the source files. This option takes precedence over the `outputSourceFiles` option.

Fixes #3257.

@coderabbitai summary


generators/typescript/utils/abstract-generator-cli/src/AbstractGeneratorCli.ts
```diff
@@ -123,7 +123,12 @@ export abstract class AbstractGeneratorCli<CustomConfig> {
                     });
                 },
                 downloadFiles: async () => {
-                    if (this.outputSourceFiles(customConfig)) {
+                    if (this.outputProjectFiles(customConfig)) {
+                        await typescriptProject.copyProjectAsZipTo({
+                            logger,
+                            destinationZip
+                        });
```
mstade, line 130: This is the key to outputting the project files locally. Because this option outputs the entire project, the `outputSourceFiles` option is essentially implied since the `src` directory is included in the project. That's why this should take priority.

mstade (commented):
## Author Notes

@dsinghvi this should fix #3257.

PR #5087, opened 2024-11-04
feat(python): fastapi generator with pydantic v2

dsinghvi:


generators/python-v2/base/src/asIs/core.ts
```diff
@@ -0,0 +1,6 @@
+import { python } from "@fern-api/python-ast";
+
+export const serialize_datetime = python.reference({
```
amckinney, line 3: I'd consider moving this into a separate directory - the `asIs` directory is usually used for all the files copied directly into the SDK (e.g. [C#](https://github.com/fern-api/fern/tree/main/generators/csharp/codegen/src/asIs), [PHP](https://github.com/fern-api/fern/tree/main/generators/php/codegen/src/asIs)), whereas this is used as a library for model generation (e.g. ObjectGenerator).

generators/python-v2/base/src/project/WriteablePythonFile.ts
```diff
@@ -0,0 +1,23 @@
...
+    interface Args {
+        filename: string;
+        /* Directory of the filepath */
+        directory: RelativeFilePath;
+
+        contents: python.PythonFile;
+    }
+}
+
+export class WriteablePythonFile extends File {
```
amckinney, line 15: Nice, naming makes sense to me here.

generators/python-v2/base/src/context/AbstractPythonGeneratorContext.ts
```diff
@@ -0,0 +1,57 @@
...
+        return false;
+    }
+
+    public getPascalCaseSafeName(name: Name): string {
+        return name.pascalCase.safeName;
+    }
+
+    public getSnakeCaseSafeName(name: Name): string {
+        return name.snakeCase.safeName;
+    }
```
amckinney, line 54: Ideally we have functions that better represent the constructs of the language so that the generator author never has to remember what casing convention to use (e.g. `getPropertyName`, `getClassName`, etc).

With that said, it's still nice to have these helpers so that we can just delegate from those other helpers later.

generators/python-v2/base/src/context/AbstractPythonGeneratorContext.ts
```diff
@@ -0,0 +1,57 @@
...
+            const typeDeclaration = this.getTypeDeclarationOrThrow(typeReference.typeId);
+            if (typeDeclaration.shape.type === "alias") {
+                return this.isTypeReferenceOptional(typeDeclaration.shape.aliasOf);
+            }
+        }
+        return false;
+    }
+
+    public getPascalCaseSafeName(name: Name): string {
+        return name.pascalCase.safeName;
```
amckinney, line 49: Should we be using `unsafeName` here since they shouldn't ever conflict with Python keywords (e.g. we can use `Str` instead of `Str_`)? Does this preserve behavior with the current generator?

amckinney (approved):
Structure and implementation looks great to me. Only minor comments to consider.

generators/python-v2/base/src/context/AbstractPythonGeneratorContext.ts
dsinghvi: Agreed, ill aim to refactor those as i get deeper into the generator

generators/python-v2/base/src/context/AbstractPythonGeneratorContext.ts
dsinghvi: Yup copied the existing generator for now, and was hoping to tackle as i went deeper to see what could be improved (but also not create compile breaks)

dsinghvi:
@amckinney adressed some comments, will address more as I continue to build out this generator.

generators/python-v2/base/src/asIs/core.ts
dsinghvi: Makes sense (will make this change in a FLUP)

PR #5088, opened 2024-11-04
feat(python): Support Formatting Configuration

noanflaherty:
Ruff fmt supports passing in a config object to change the way it formats code. We'd like to be able to modify the max line length, and so exposing this config makes sense.

This PR also cleans up some imports to reduce the number of circular dependencies, as identified via `madge`.


amckinney (approved):
Nice lgtm!

PR #5093, opened 2024-11-04
fix(cli): Set `instance` when migrating Mintlify > Fern

mattblank11:
Related to [this task](https://linear.app/buildwithfern/issue/FER-3471/set-instance-when-migrating-mintlify-fern)

Here's what the `docs.yml` file looks like for Bland AI with the instances value set:
```
instances:
  - url: https://bland-ai.docs.buildwithfern.com
```

@coderabbitai summary


packages/cli/docs-importers/mintlify/src/MintlifyImporter.ts
```diff
@@ -95,6 +95,13 @@ export class MintlifyImporter extends DocsImporter<MintlifyImporter.Args> {
             }
         }
         this.context.logger.debug("Imported OpenAPI specs");
+
+        // We're currently setting one instance so that clients can generate docs out of the box,
+        // but they'll have the option to add additional instances and custom domains later.
+        builder.setInstance({
+            companyName: mint.name
+        });
+        this.context.logger.debug("Set instance");
```
RohinBhargava, line 104: do we want to output the name here, as well?

mattblank11: Good idea, updating now!

dsinghvi (approved):

packages/cli/docs-importers/mintlify/src/MintlifyImporter.ts
mattblank11: @RohinBhargava are you thinking that we should output the name from the mint file or the name we auto-generate?

RohinBhargava: good question! I think you have the most context, but I would assume `mint.name`, given the locality of the debug statement to the invocation

dsinghvi: @mattblank11 could see this being:  `Added instance ${url} to docs.yml`

mattblank11: Appreciate the input!

I agree with @dsinghvi on returning the url, if anything. IMO it's weird to return a string from the `setInstance` function but I don't love the alternative of creating a separate `getInstanceUrl` function. Feel free to push back on this!

PR #5099, opened 2024-11-04
feat(go): Improve dynamic snippets error reporting

amckinney:
This adds the `ErrorReporter` into the dynamic snippet generator's core flow. With this, we collect all errors that occur during snippet generation, but continue to compose the snippet as best we can (leveraging `go.TypeInstantiation.nop` as needed).

In the future, we plan to supplement this validation with the help of JSON schema, but this at least builds out the hand-crafted solution that will be used by the regions of the snippet that won't be covered by JSON schema (e.g. query parameters, headers, etc).

### DIFF
--- a/generators/go-v2/dynamic-snippets/src/__test__/__snapshots__/examples.test.ts.snap
+++ b/generators/go-v2/dynamic-snippets/src/__test__/__snapshots__/examples.test.ts.snap
@@ -1,5 +1,18 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`examples (errors) > invalid request body 1`] = `
+[
+  {
+    "message": "Expected string but got number",
+    "path": [
+      "requestBody",
+      "title",
+    ],
+    "severity": "CRITICAL",
+  },
+]
+`;
+
 exports[`examples > 'GET /metadata (allow-multiple)' 1`] = `
 "package example
 
--- a/generators/go-v2/dynamic-snippets/src/__test__/examples.test.ts
+++ b/generators/go-v2/dynamic-snippets/src/__test__/examples.test.ts
@@ -113,3 +113,43 @@ describe("examples", () => {
         expect(response.snippet).toMatchSnapshot();
     });
 });
+
+describe("examples (errors)", () => {
+    it("invalid request body", async () => {
+        const generator = buildDynamicSnippetsGenerator({
+            irFilepath: join(DYNAMIC_IR_TEST_DEFINITIONS_DIRECTORY, RelativeFilePath.of("examples.json")),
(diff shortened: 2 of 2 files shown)


generators/go-v2/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
```diff
@@ -72,65 +80,95 @@ export class DynamicSnippetsGeneratorContext {
...
+                severity: Severity.Critical,
+                message: this.newParameterNotRecognizedError(discriminantFieldName).message
+            });
+            return undefined;
         }
         if (typeof discriminantValue !== "string") {
-            throw new Error(`Expected discriminant value to be a string but got: ${JSON.stringify(discriminantValue)}`);
+            this.errors.add({
+                severity: Severity.Critical,
+                message: `Expected discriminant value to be a string but got: ${JSON.stringify(discriminantValue)}`
```
dsinghvi, line 160: `${JSON.stringify(discriminantValue)}` this might be a little tough for anyone to render in a pretty way -- maybe we just add the type of `discriminantValue` ?

dsinghvi (approved):
Approving to unblock, couple things crossed my mind: 

1. If we add `JSON.stringify()` in the error messages, its going to be hard for a consumer to render in a pretty way. 
2. It would be nice to decouple validation from generator specific code (i.e. python/typescript should not have to worry about dealing with validation)
3. Reimplementing validation will never be as fully-featured as a third party lib that'll take things like `min`, `max`, `regex` into account as well (so still pushing json schema here) 

Still great to have this since we are going from a world with no validation to real validation. Curious for your thoughts on the above!

generators/go-v2/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
amckinney: Good call, happy to adjust the message here.

amckinney:
> ... couple things crossed my mind:

1. Good call - i'll see what I can do to remove all instances of `JSON.stringify` in the error messages before merging.
2. Another good call - we should be able to do this by pushing more logic into the to-be-named `AbstractDynamicSnippetsGeneratorContext` that will be shared between all dynamic snippet generators. I'll keep this in mind once we start rolling out the next dynamic snippets generator.
3. 100% agree, this won't replace the JSON schema validation we have planned.
    * I imagine the final flow will be receive request -> validate against JSON schema -> generate snippet and collect warnings along the way.
    * Many (if not all) of the "critical" errors we define here can be factored out and only warnings that can't be easily captured by JSON schema will remain from this abstraction.

PR #5102, opened 2024-11-04
feat(python): support wrapped aliases in pydantic v2

dsinghvi:


dsinghvi:
Keeping this open since we may merge it after @matthew-giglio does some testing

PR #5104, opened 2024-11-05
feat(csharp): Support HTTP endpoints in GRPC service

Swimburger:
@coderabbitai summary

### DIFF
--- a/generators/csharp/sdk/src/SdkGeneratorContext.ts
+++ b/generators/csharp/sdk/src/SdkGeneratorContext.ts
@@ -432,7 +432,7 @@ export class SdkGeneratorContext extends AbstractCsharpGeneratorContext<SdkCusto
         return `${endpoint.name.pascalCase.safeName}Async`;
     }
 
-    public endpointUsesGrpcTransport(service: HttpService, endpoint: HttpEndpoint) {
+    public endpointUsesGrpcTransport(service: HttpService, endpoint: HttpEndpoint): boolean {
         return service.transport?.type === "grpc" && endpoint.transport?.type !== "http";
     }
 


generators/csharp/sdk/src/endpoint/EndpointGenerator.ts
```diff
@@ -32,7 +32,9 @@ export class EndpointGenerator extends AbstractEndpointGenerator {
         rawClient: RawClient;
         grpcClientInfo: GrpcClientInfo | undefined;
     }): csharp.Method {
-        if (grpcClientInfo != null) {
+        // If the service is a grpc service, grpcClientInfo will not be null or undefined,
+        // so any endpoint will be generated as a grpc endpoint, unless the transport is overriden by setting type to http
```
amckinney, line 36: nit: I think we can remove this comment - the control flow speaks for itself in my opinion. Thoughts?

generators/csharp/sdk/src/wrapped-request/WrappedRequestGenerator.ts
```diff
@@ -46,6 +46,9 @@ export class WrappedRequestGenerator extends FileGenerator<CSharpFile, SdkCustom
             type: csharp.Class.ClassType.Record
         });
 
+        const service = this.context.getHttpServiceOrThrow(this.serviceId);
+        // If the service transport is grpc and the endpoint transport is not http, we need to generate a proto request
+        const isProtoRequest = service.transport?.type === "grpc" && this.endpoint.transport?.type !== "http";
```
amckinney, line 51: Thoughts on building out a helper for this on the `context`? That way we can drop the comment and reuse it elsewhere later if need be. Plus, we can someday leverage the encoding field in this function more naturally.

```typescript
const isProtoRequest = this.context.endpointHasProtoRequest({ service, endpoint: this.endpoint });
```

amckinney (approved):
Nice, this looks great - don't forget to fill out the [versions.yml](https://github.com/fern-api/fern/blob/b98d1e80a2016e72026d741450a703290fdb6451/generators/csharp/sdk/versions.yml#L9) file before merging!

PR #5105, opened 2024-11-05
feat(python): Multi-line Strings

noanflaherty:
This PR explores adding support for multi-line strings in python.


generators/python-v2/ast/src/TypeInstantiation.ts
```diff
@@ -200,6 +204,20 @@ export class TypeInstantiation extends AstNode {
...
         }
     }
+
+    private writeStringWithTripleQuotes({ writer, value }: { writer: Writer; value: string }): void {
+        writer.write('"""');
+        const parts = value.split("\n");
+        const head = parts[0] + "\n";
+        const tail = parts.slice(1).join("\n");
+        writer.write(this.escapeString(head));
+        writer.writeNoIndent(this.escapeString(tail));
```
noanflaherty, line 214: You'll notice that this `writeNoIndent` is used, but as the test below shows, it doesn't work as I'd expect.

generators/python-v2/ast/src/TypeInstantiation.ts
```diff
@@ -200,6 +204,20 @@ export class TypeInstantiation extends AstNode {
...
+        writer.write('"""');
+    }
+
+    private escapeString(value: string): string {
+        return value.replaceAll('"', '\\"');
```
noanflaherty, line 219: Tbh I'm not sure if this is the right escape strategy here.

generators/python-v2/ast/src/__test__/Class.test.ts
```diff
@@ -103,4 +103,20 @@ describe("class", () => {
...
+
+    it("class with de-indented multi-line string", async () => {
+        const clazz = python.class_({
+            name: "MyClass"
+        });
+        clazz.add(
+            python.field({
+                name: "multiline_string",
+                type: python.Type.str(),
+                initializer: python.TypeInstantiation.str("Hello\nWorld")
```
noanflaherty, line 115: This test doesn't yet produce the snapshot I'd expect.

generators/python-v2/ast/src/__test__/TypeInstantiation.test.ts
```diff
@@ -31,9 +31,26 @@ describe("TypeInstantiation", () => {
...
+            expect(await writer.toStringFormatted()).toMatchSnapshot();
+        });
+
+        it("should render a string containing newline characters", async () => {
+            TypeInstantiation.str("\n\n####\n\n").write(writer);
+            expect(await writer.toStringFormatted()).toMatchSnapshot();
+        });
+
+        it("should render a string containing newline characters and quotes", async () => {
+            TypeInstantiation.str('She said "Hi"\nHe said "bye"\nShe said "okay then"').write(writer);
```
noanflaherty, line 51: This test doesn't yet produce the snapshot I'd expect.

generators/python-v2/ast/src/__test__/__snapshots__/Class.test.ts.snap
```diff
@@ -13,6 +13,13 @@ class MyDataClass:
 "
 `;
 
+exports[`class > class with de-indented multi-line string 1`] = `
+"class MyClass:
+    multiline_string: str = """Hello
+    World"""
+"
```
noanflaherty, line 20: I would expect for `writer.writeNoIndent` to not indent this line. There should be no whitespace before `World` here

generators/python-v2/ast/src/__test__/__snapshots__/TypeInstantiation.test.ts.snap
```diff
@@ -50,11 +50,32 @@ exports[`TypeInstantiation > set 1`] = `
...
+####
+
+"""
+"
+`;
+
+exports[`TypeInstantiation > str > should render a string containing newline characters and quotes 1`] = `
+""""She said \\"Hi\\"
+He said \\"bye\\"
+She said \\"okay then\\" """
```
noanflaherty, line 70: I'm not sure how the trailing space after `then\\"` is being generated 🤔

generators/python-v2/ast/src/TypeInstantiation.ts
dvargas92495: I wouldn't expect to need this during multiline strings

generators/python-v2/ast/src/TypeInstantiation.ts
```diff
@@ -143,7 +143,11 @@ export class TypeInstantiation extends AstNode {
                 }
                 break;
             case "str":
-                writer.write(`"${this.internalType.value}"`);
+                if (this.internalType.value.includes("\n")) {
```
dvargas92495, line 146: Idk if it's necessarily true that a newline in the string means it should be triple quoted. the user may want to have `"A\\nstring"` and so the decision to use single line or triple quote might need to come from the consumer. And at least for our use case, I think we could have templating node templates and jinja block templates always proactively generate triple quoted strings in our opinionated default path

dvargas92495: Just saw the thread @amckinney was advising otherwise. I also checked python's `ast` library and it seems like it treats it as the same AST node as well:
Screenshot 2024-11-17 at 9 24 18 AM

so ultimately could go either way here, but will just note that I think this could cause drift in some edge cases between what the user specifies and what we generate

generators/commons/src/ast/AbstractWriter.ts
```diff
@@ -38,7 +38,10 @@ export class AbstractWriter {
      * @param text
      */
     public writeNoIndent(text: string): void {
-        this.writeInternal(text);
+        const currIndentLevel = this.indentLevel;
+        this.indentLevel = 0;
+        this.write(text);
+        this.indentLevel = currIndentLevel;
```
noanflaherty, line 44: This didn't work as expected. I claim this was the desired behavior all along.

generators/python-v2/ast/src/Field.ts
```diff
@@ -48,7 +48,8 @@ export class Field extends AstNode {
         }
 
         if (this.initializer !== undefined) {
-            writer.write(` = ${this.initializer}`);
+            writer.write(" = ");
+            this.initializer.write(writer);
```
noanflaherty, line 52: Fixes a bug. We should have been doing this the whole time.

amckinney (approved):

PR #5106, opened 2024-11-05
feat(cli): Turn Mintlify migrator into a CLI command

mattblank11:
Related to [this card](https://linear.app/buildwithfern/issue/FER-3472/turn-mintlify-migrator-into-a-cli)

This PR includes the functionality to run Mintlify migrations through the CLI. Some highlights in the PR:
- Running fern init --mintlify <path-to-mint.json> will create the Fern docs in the directory that the command is run through
- There's a function called runMintlifyMigration in @fern-api/mintlify-importer that actually runs the migration functionality
- There's a function called initializeWithMintlify in /cli/init that ensures the mint.json file is valid, then calls runMintlifyMigration
- There are unit tests for initializeWithMintlify to ensure that runMintlifyMigration only gets called if the mint.json file is valid
- There's an ETE test for running the migration command, though I mentioned that's currently broken. I'm going to fix that this evening

@coderabbitai summary


packages/commons/fs-utils/src/isUrl.ts
mattblank11, line 1: Let me know if you want me to move this back - I could argue that it belongs here or in a separate shared utils folder

packages/cli/ete-tests/src/tests/init/init.ts
```diff
@@ -4,8 +4,9 @@ import { runFernCli } from "../../utils/runFernCli";
 
 export async function init({
     directory,
-    openApiArg
-}: { directory?: AbsoluteFilePath; openApiArg?: string } = {}): Promise<AbsoluteFilePath> {
+    openApiArg,
+    mintJsonArg
+}: { directory?: AbsoluteFilePath; openApiArg?: string; mintJsonArg?: string } = {}): Promise<AbsoluteFilePath> {
```
dsinghvi, line 9: nit: would probably make this accept a discriminated union of args and it can either be openapi or mintlify (cause i dont think `--openapi` and `--mintlify` together would work)

packages/commons/fs-utils/src/isUrl.ts
dsinghvi: this sounds great

dsinghvi:
Up to add a changelog entry to `cli/versions/versions.yml`

packages/cli/ete-tests/src/tests/init/init.ts
mattblank11: @dsinghvi Took a slightly different path vs. a discriminated union but achieved a similar outcome. Let me know what you think!

packages/cli/docs-importers/mintlify/src/runMintlifyMigration.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    await writeFile(
+        join(
+            AbsoluteFilePath.of(outputPath),
+            RelativeFilePath.of(FERN_DIRECTORY),
+            RelativeFilePath.of(PROJECT_CONFIG_FILENAME)
+        ),
+        JSON.stringify(
+            {
+                version: "*",
```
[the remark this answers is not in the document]
mattblank11: Good catch! I believe I did this properly - let me know if not

packages/cli/cli/versions.yml
```diff
@@ -1,3 +1,10 @@
+- changelogEntry:
+    - summary: |
+        The Fern CLI now supports initializing with a Mintlify project.
+      type: fix
+  irVersion: 53
+  version: 0.45.0-rc33
```
dsinghvi, line 6: would just pull from main to confirm there is no merge conflict

dsinghvi (approved):
We should fix `init.test.ts.snap` and make sure its easy to see the generated fern directory in a string format, otherwise good to merge! 🚀

PR #5115, opened 2024-11-05
feat(python): Support wrapped aliases for Pydantic v2

matthew-giglio:
- Wrapped aliases are not supported for v2
- We need to support both the wrapped aliases themselves along with validators for them
- The v2 wrapped aliases leverage `RootModel` and `model_validator` on their validators
- `RootModel` does not support `extra` as a keyword nor does it support `frozen` behavior well`, so we omit those from the model behavior here
- Tested by generating Pydantic v1 and v2 wrapped aliases with validators
- We don't support this for "both" (have tests demonstrating) since there's no great in between model that works well between both versions

### DIFF
--- a/generators/python/src/fern_python/external_dependencies/pydantic.py
+++ b/generators/python/src/fern_python/external_dependencies/pydantic.py
@@ -43,10 +43,6 @@ class Pydantic:
         return _export("RootModel")
 
     @staticmethod
-    def ModelValidator() -> AST.ClassReference:
-        return _export("model_validator")
-
-    @staticmethod
     def PydanticImport() -> AST.ReferenceImport:
         return AST.ReferenceImport(
             module=AST.Module.external(


generators/python/src/fern_python/external_dependencies/pydantic.py
```diff
@@ -16,12 +16,7 @@ class PydanticVersionCompatibility(str, enum.Enum):
 
 def _export(*export: str) -> AST.ClassReference:
     return AST.ClassReference(
-        import_=AST.ReferenceImport(
-            module=AST.Module.external(
-                dependency=PYDANTIC_DEPENDENCY,
-                module_path=("pydantic",),
-            ),
-        ),
+        import_=Pydantic.PydanticImport(),
```
matthew-giglio, line 19: Added this for easier reference in other codepaths

generators/python/src/fern_python/generators/pydantic_model/fern_aware_pydantic_model.py
```diff
@@ -110,6 +115,15 @@ def to_reference(self) -> LocalClassReference:
...
 
+    def _model_validator(self, pre: bool = False) -> AST.FunctionInvocation:
+        return AST.FunctionInvocation(
+            function_definition=AST.Reference(
+                qualified_name_excluding_import=("model_validator",),
+                import_=Pydantic.PydanticImport(),
+            ),
+            kwargs=[("mode", AST.Expression(expression='"before"' if pre else '"after"'))],
+        )
+
```
matthew-giglio, line 126: I don't like having this as part of the class, but I also don't know where you store utility files in your repo. Feel free to move!

PR #5122, opened 2024-11-07
chore(csharp): Update vulnerable System.Text.Json dependency

Swimburger:
@coderabbitai summary


dsinghvi (approved):

amckinney (approved):
lgtm, although is there any way we can pin to ^8.0 so that we pick up these patches automatically?

Swimburger:
> lgtm, although is there any way we can pin to ^8.0 so that we pick up these patches automatically?

In .NET, you can use `8.*.*` but nobody ever uses this feature.
I looked at various SDKs, and none of them have done so. I will dig deeper and make a separate PR if it is a good practice.

PR #5130, opened 2024-11-08
feat(csharp): Let users customize MSBuild properties using props file

Swimburger:
@coderabbitai summary

### DIFF
--- a/generators/csharp/sdk/src/SdkCustomConfig.ts
+++ b/generators/csharp/sdk/src/SdkCustomConfig.ts
@@ -14,7 +14,7 @@ export const SdkCustomConfigSchema = z.strictObject({
         .record(z.string())
         .optional()
         .describe(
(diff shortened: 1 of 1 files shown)


generators/csharp/codegen/src/project/CsharpProject.ts
```diff
@@ -198,6 +199,12 @@ export class CsharpProject extends AbstractProject<AbstractCsharpGeneratorContex
             join(absolutePathToProjectDirectory, RelativeFilePath.of(`${this.name}.csproj`)),
             templateCsProjContents
         );
+
+        await writeFile(
+            join(absolutePathToProjectDirectory, RelativeFilePath.of(`${this.name}.Custom.props`)),
```
amckinney, line 204: Should the file be named `Custom.Props` or `Custom.props`? The description below says `Custom.Props`, so as long as we're consistent.

amckinney (approved):

PR #5132, opened 2024-11-08
feat(python): Allow for Statements Above Imports

noanflaherty:
There are times where you want to specify certain statements above all imports in a python file. This PR adds support for this concept, and also removes the `name` field from python files since it's not actually used anywhere.

### DIFF
--- a/generators/python-v2/ast/src/PythonFile.ts
+++ b/generators/python-v2/ast/src/PythonFile.ts
@@ -9,8 +9,6 @@ export declare namespace PythonFile {
         moduleName: string;
         /* The path of the Python file relative to the module */
         path: ModulePath;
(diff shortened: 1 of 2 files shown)


generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
```diff
@@ -84,6 +84,17 @@ var_b: List[str]
 "
 `;
 
+exports[`PythonFile > Include statements above imports 1`] = `
+"# flake8: noqa: F401, F403
+
+from itertools import chain
```
noanflaherty, line 90: This test shows the end goal of this PR.

PR #5155, opened 2024-11-11
fix(cli): handle matching request + response examples correctly

dsinghvi:
Previously we were not matching requests and responses according to the example name in OpenAPI. This is now fixed.

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,5 +1,13 @@
 - changelogEntry:
     - summary: |
+        The OpenAPI importer now supports correlating request and response examples by name. When an example name is shared 
+        between a request body and response, they will be paired together in the generated Fern definition.
+      type: fix
+  irVersion: 53
+  version: 0.45.0-rc39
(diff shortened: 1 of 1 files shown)


packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/ExampleEndpointFactory.ts
```diff
@@ -412,9 +411,8 @@ function consolidateRequestResponseExamples(
 
         let paired = false;
         for (let idx = 0; idx < responseExamples.length; idx++) {
-            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             const [responseId, responseExample] = responseExamples[idx]!;
-            if (responseId != null || visitedResponseIdx.has(idx)) {
```
dsinghvi, line 417: Changing `!=` to `==` is the key piece of logic

RohinBhargava (approved):

PR #5156, opened 2024-11-11
feat(cli): Add docs to generated object properties when running `fern jsonschema`

pattobrien:
This PR adds the following changes:
- feat(cli): add `description` value to object properties in generated Json Schema definitions (`fern jsonschema`)
- docs: add CLI changelog entry
- tests: added new `property-docs` test definitions
- tests: updated `ir-to-jsonschema` snapshots to reflect new `description` values

### DIFF
--- a/fern/pages/changelogs/cli/2024-11-18.mdx
+++ b/fern/pages/changelogs/cli/2024-11-18.mdx
@@ -1,4 +1,4 @@
-## 0.45.0-rc45
+## 0.45.0-rc46
 **`(fix):`** The IR handles converting example unions that are aliases.
 
 


packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/alias/type__Type.json
```diff
@@ -2,10 +2,12 @@
   "type": "object",
   "properties": {
     "id": {
-      "$ref": "#/definitions/TypeId"
+      "$ref": "#/definitions/TypeId",
+      "title": "id"
```
dsinghvi, line 6: @pattobrien thoughts on making the `title` the actual name of the type so `TypeId` ?

packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/examples/type_types_ExtendedMovie.json
```diff
@@ -12,22 +13,29 @@
...
     "from": {
-      "type": "string"
+      "type": "string",
+      "title": "from"
     },
     "rating": {
-      "type": "number"
+      "type": "number",
+      "title": "rating",
+      "description": "The rating scale is one to five stars"
```
dsinghvi, line 30: this looks good to me

packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/examples/type_types_ExtendedMovie.json
```diff
@@ -12,22 +13,29 @@
...
     },
     "from": {
-      "type": "string"
+      "type": "string",
+      "title": "from"
     },
     "rating": {
-      "type": "number"
+      "type": "number",
+      "title": "rating",
```
dsinghvi, line 29: is there value in having a title here?

pattobrien: @dsinghvi I looked into how other popular schemas / schema tools handle titles (e.g. VSCode settings.json Intellisense), and I now agree that there's little to no value in having the property key in the title.

Regarding using the type name instead: since JSONSchema suggests that `title` and `description` should document **purpose**, I think there's also little value from doing that, since the name of a type might not provide any useful info in a doc comment that the schema validation doesn't already provide (I can provide more detail on this if you'd like). But its possible for us to put something like `type: FooSchema` at the top of the doc comment, but I'm indifferent about it since I don't see other schema tools adding that info in.

packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/alias/type__Type.json
pattobrien: Resolving, based on other comment thread.

packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/examples/type_types_ExtendedMovie.json
pattobrien: Removed `title` generation.

amckinney:
Closing this as-is for now, we'll restore this (or something similar) when it's relevant again.

PR #5165, opened 2024-11-12
feat(csharp): Set Content-Type header for HTTP requests when specified in the API spec/definition

fern-support:
When a content type is specified in the Fern definition, the default content type will be overwritten before sending the HTTP request. (the default being JSON or octet stream depending on the request type).

Notes:

-  Fix: Setting the content type on the HTTP request throws an exception in .NET, you need to set it on the headers of the content of the request.
- I'm using `MediaTypeHeaderValue.Parse` which also accepts "...; charset=utf8" in case that's specified in the spec/definition.

### DIFF
--- a/generators/go-v2/dynamic-snippets/src/__test__/__snapshots__/ir.test.ts.snap
+++ b/generators/go-v2/dynamic-snippets/src/__test__/__snapshots__/ir.test.ts.snap
@@ -2122,6 +2122,146 @@ package example
 
 import (
 	context "context"
+	acme "github.com/acme/acme-go"
+	client "github.com/acme/acme-go/client"
+	option "github.com/acme/acme-go/option"
+	types "github.com/acme/acme-go/types"
+	uuid "github.com/google/uuid"
+)
+
+func do() {
+	client := client.NewClient(
+		option.WithToken(
+			"<token>",
+		),
+	)
+	client.Endpoints.ContentType.PostJSONPatchContentType(
+		context.TODO(),
+		&types.ObjectWithOptionalField{
+			String: acme.String(
+				"string",
+			),
+			Integer: acme.Int(
+				1,
+			),
+			Long: acme.Int64(
+				1000000,
+			),
+			Double: acme.Float64(
+				1.1,
+			),
+			Bool: acme.Bool(
+				true,
+			),
+			Datetime: acme.Time(
+				acme.MustParseDateTime(
+					"2024-01-15T09:30:00Z",
+				),
+			),
+			Date: acme.Time(
+				acme.MustParseDateTime(
+					"2023-01-15",
+				),
+			),
+			UUID: acme.UUID(
+				uuid.MustParse(
+					"d5e9c84f-c2b2-4bf4-b4b0-7ffd7a9ffc32",
+				),
+			),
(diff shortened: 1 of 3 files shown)


generators/csharp/sdk/src/endpoint/http/RawClient.ts
```diff
@@ -85,6 +85,12 @@ export class RawClient {
                 assignment: csharp.codeblock(headerBagReference)
             });
         }
+        if (endpoint.requestBody && "contentType" in endpoint.requestBody && endpoint.requestBody.contentType) {
```
amckinney, line 88: We normally try to be more explicit when we resolve specific properties from discriminated unions. That way if we ever change the shape of the type in a later IR version, we know exactly where we need to change the code and we aren't ever surprised if we break Content-Type detection somewhere.

What are your thoughts on something like this instead?

```typescript
function getContentTypeFromRequestBody(requestBody: HttpRequestBody): string | undefined {
  switch (requestBody.type) {
    case "inlinedRequestBody":
      return requestBody.contentType;
    case "reference":
      return requestBody.contentType;
    case "fileUpload":
      return undefined;
    case "bytes":
      return requestBody.contentType;
    default:
      assertNever(requestBody);
  }
}
```

generators/csharp/sdk/src/test-generation/mock-server/MockServerTestGenerator.ts
```diff
@@ -91,6 +91,14 @@ export class MockServerTestGenerator extends FileGenerator<CSharpFile, SdkCustom
                         writer.write(`.WithHeader("${header.name.wireValue}", "${maybeHeaderValue}")`);
                     }
                 }
+                if (
+                    this.endpoint.requestBody &&
+                    "contentType" in this.endpoint.requestBody &&
+                    this.endpoint.requestBody.contentType
+                ) {
```
amckinney, line 98: Same thing with the recommendation above.

generators/csharp/sdk/src/endpoint/utils/getContentTypeFromRequestBody.ts
```diff
@@ -0,0 +1,14 @@
...
+    return endpoint.requestBody._visit<string | undefined>({
+        inlinedRequestBody: (body) => body.contentType,
+        reference: (body) => body.contentType,
+        fileUpload: (_) => undefined,
+        bytes: (body) => body.contentType,
+        _other: (body) => {
+            throw new Error(`Unexpected request body type: ${body.type}`);
+        }
+    });
+};
```
amckinney, line 14: nit: Why not use a function here? 
```suggestion
export function getContentTypeFromRequestBody(endpoint: HttpEndpoint): string | undefined {
    if (endpoint.requestBody == null) {
      return undefined;
    }
    return endpoint.requestBody._visit<string | undefined>({
        inlinedRequestBody: (body) => body.contentType,
        reference: (body) => body.contentType,
        fileUpload: (_) => undefined,
        bytes: (body) => body.contentType,
        _other: (body) => {
            throw new Error(`Unexpected request body type: ${body.type}`);
        }
    });
};
```

amckinney (approved):
Assuming the `const` is updated to a function, lgtm!

PR #5166, opened 2024-11-12
fix(cli): support non-standard http 498 & `x-fern-examples` validation.

eyw520:
@coderabbitai summary

### DIFF
--- a/packages/cli/ete-tests/src/tests/write-definition/__snapshots__/writeDefinition.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/write-definition/__snapshots__/writeDefinition.test.ts.snap
@@ -1603,6 +1603,11 @@ name: api
       data: optional<string>
     source:
       openapi: openapi/cohere.yaml
+  InvalidTokenErrorBody:
+    properties:
+      data: optional<string>
+    source:
+      openapi: openapi/cohere.yaml
   NotFoundErrorBody:
     properties:
       data: optional<string>
@@ -1678,6 +1683,14 @@ errors:
       - value: {}
     status-code: 500
     type: unknown
+  InvalidTokenError:
+    docs: >
+      This error is returned when a request or response contains a deny-listed
+      token.
+    examples:
+      - value: {}
+    status-code: 498
+    type: unknown
   NotFoundError:
     docs: >
       This error is returned when a resource is not found. This could be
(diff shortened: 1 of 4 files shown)


packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/operation/convertStreamingOperation.ts
```diff
@@ -222,7 +222,6 @@ export function isStreamingExample(
         unknown: (unknownExample) => {
             const maybeFernExample = RawSchemas.serialization.ExampleEndpointCallSchema.parse(unknownExample);
             if (!maybeFernExample.ok) {
-                context.logger.error("Failed to parse example", maybeFernExample.errors.toString());
```
amckinney, line 225: Should this log actually be removed? Was this resulting in double logs with your updated change below?

packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/convertToHttpError.ts
```diff
@@ -35,6 +35,7 @@ export const ERROR_NAMES_BY_STATUS_CODE: Record<number, string> = {
     449: "RetryWithError", // non-standard
     450: "BlockedByWindowsParentalControlsError", // non-standard
     451: "UnavailableForLegalReasonsError", // non-standard
+    498: "InvalidTokenError", // non-standard
```
amckinney, line 38: Not required here, but I imagine we might eventually need to support all the unofficial status codes listed [here](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) (since it looks like we've been adding them one by one).

amckinney (approved):
If the removed log message is actually intentional and we aren't losing any coverage, lgtm!

packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/operation/convertStreamingOperation.ts
dsinghvi: Yup intentional -- should be removed and replaced with the log line elsewhere for a better UX

PR #5168, opened 2024-11-13
feat(cli): Add OpenAPI tree-shaking setting

amckinney:
This adds the `only-include-referenced-schemas` setting to implement OpenAPI tree shaking for endpoints, webhooks, and websocket channels.

Users can opt-in with the following configuration:

```yaml
# generators.yml
api:
  specs:
    - openapi: ../openapi.yml
      settings:
        only-include-referenced-schemas: true
```

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildFernDefinition.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildFernDefinition.ts
@@ -75,7 +75,6 @@ export function buildFernDefinition(context: OpenApiIrConverterContext): FernDef
 
     const convertedServices = buildServices(context);
     const sdkGroups = convertedServices.sdkGroups;
-    let schemaIdsToExclude = convertedServices.schemaIdsToExclude;
 
     context.setInState(State.Webhook);
     buildWebhooks(context);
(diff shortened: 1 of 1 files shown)


packages/cli/api-importers/openapi/openapi-ir-to-fern/src/OpenApiIrConverterContext.ts
```diff
@@ -49,22 +62,54 @@ export class OpenApiIrConverterContext {
...
+     * in the generated definition (e.g. webhook tree-shaking).
+     */
+    private inWebhook = false;
+
+    /**
+     * Whether the current schema being processed is part of a websocket channel.
+     * This is used to determine whether certain properties should be included
+     * in the generated definition (e.g. channel tree-shaking).
+     */
+    private inChannel = false;
```
amckinney, line 111: I kept things in their current state for consistency, but we can also roll out an enum-based approach if we want to reduce some of the copying between the `isIn{Channel,Endpoint,Webhook}` helper functions.

I imagine we'd want something like:

```typescript
enum State {
  Endpoint,
  Channel,
  Webhook,
  Request,
}
```

```typescript
private inState: State | undefined;
```

```typescript
/**
 * Sets that we're currently processing an endpoint
 */
public setInEndpoint(): void {
    this.setInState(State.Endpoint);
}

private setInState(state: State): void {
    this.inState = state;
}
```

dsinghvi (approved):

packages/cli/api-importers/openapi/openapi-ir-to-fern/src/OpenApiIrConverterContext.ts
amckinney: Decided to go with this approach after all!

PR #5172, opened 2024-11-13
fix(cli): support piping query parameter validation into fern definition generation & fix file typo.

eyw520:
@coderabbitai summary

### DIFF
--- a/packages/cli/ete-tests/src/tests/update-api-v2/__snapshots__/update-api.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/update-api-v2/__snapshots__/update-api.test.ts.snap
@@ -1148,27 +1148,37 @@ exports[`fern api update unioned > fern api update unioned 1`] = `
           "type": {
             "type": "string",
             "description": "A URI reference that identifies the problem type",
-            "example": "https://example.com/probs/out-of-credit"
+            "examples": [
+              "https://example.com/probs/out-of-credit"
+            ]
           },
           "title": {
             "type": "string",
             "description": "A short, human-readable summary of the problem type",
-            "example": "You do not have enough credit."
+            "examples": [
+              "You do not have enough credit."
+            ]
           },
           "detail": {
             "type": "string",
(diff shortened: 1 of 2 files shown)


packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildQueryParameter.ts
```diff
@@ -67,6 +67,13 @@ export function buildQueryParameter({
         queryParameterSchema.availability = convertAvailability(queryParameter.availability);
     }
 
+    if (typeof typeReference.value === "object" && "type" in typeReference.value) {
+        const detailedSchema = typeReference.value as RawSchemas.TypeReferenceDetailedSchema;
+        if (detailedSchema.validation !== undefined) {
+            queryParameterSchema.validation = detailedSchema.validation;
+        }
+    }
```
amckinney, line 75: nit: We can add a type guard function for this. That way we get the type narrowing on line 71 for free. Something along the lines of the following:

```typescript
function isRawTypeReferenceDetailedSchema(rawTypeReference: RawSchemas.TypeReference): rawTypeReference is RawSchemas.TypeReferenceDetailedSchema {
      return (rawTypeReference as RawSchemas.TypeReferenceDetailedSchema).type != null;
}
```

This is similar to the other `isRaw` helpers you'll see elsewhere (like [this](https://github.com/fern-api/fern/blob/ed573cc105a27a648e3ad5dfac2cc18a4de7e3ce/packages/cli/fern-definition/schema/src/utils/visitRawTypeDeclaration.ts#L42)). You can keep the function un-exported in this file for now if that's simpler.

amckinney (approved):
You'll want to update the test definitions with the following:

```sh
$ pnpm test:ete:update
```

packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildQueryParameter.ts
eyw520: Implemented the suggested change.

PR #5173, opened 2024-11-14
feat(go): Improve file layout for types

amckinney:
## Overview

This makes changes in both the Fern CLI and the Go generator to improve the file layout of the Go SDK. 

## Implementation

This updates the IR's `ServiceTypeReferenceInfo` so that it now captures _transitive_ types referenced from each service. Additionally, this simplifies and improves the Go SDK's file layout resolution according to the latest changes in the IR.

## Effect

In combination, types can more accurately be placed according to the service they're used by whenever possible. The SDK still limits how many top-level files it can create for GitHub repository aesthetics - there is at most one file for every package we generate. 

For large APIs, this reduces the size of the generated `types.go` file by ~50% and preserves backwards compatibility.


dsinghvi (approved):
🚀

PR #5174, opened 2024-11-14
feat(csharp): Add support for idempotency headers

Swimburger:
@coderabbitai summary

### DIFF
--- a/generators/csharp/sdk/src/options/IdempotentRequestOptionsGenerator.ts
+++ b/generators/csharp/sdk/src/options/IdempotentRequestOptionsGenerator.ts
@@ -41,7 +41,10 @@ export class IdempotentRequestOptionsGenerator extends FileGenerator<
                     writer.indent();
                     for (const header of this.context.getIdempotencyHeaders()) {
                         const type = this.context.csharpTypeMapper.convert({ reference: header.valueType });
-                        const isString = type.internalType.type === "string";
+                        const isString =
+                            type.internalType.type === "string" ||
+                            (type.internalType.type === "optional" &&
+                                type.internalType.value.internalType.type === "string");
                         const toString = isString ? "" : ".ToString()";
                         // In header values, we only accept simple types, so we can assume that none are nullable (apart from string),
                         // unless the type is optional
--- a/generators/csharp/sdk/src/options/BaseOptionsGenerator.ts
+++ b/generators/csharp/sdk/src/options/BaseOptionsGenerator.ts
@@ -136,7 +136,8 @@ export class BaseOptionsGenerator {
                 name: header.name.name.pascalCase.safeName,
                 get: true,
                 init: true,
-                type: this.context.csharpTypeMapper.convert({ reference: header.valueType })
+                type: this.context.csharpTypeMapper.convert({ reference: header.valueType }),
+                summary: header.docs
             })
         );
     }


generators/csharp/sdk/src/options/IdempotentRequestOptionsGenerator.ts
```diff
@@ -0,0 +1,69 @@
...
+                parameters: [],
+                return_: csharp.Type.reference(this.context.getHeadersClassReference()),
+                interfaceReference: this.context.getIdempotentRequestOptionsInterfaceClassReference(),
+                type: csharp.MethodType.INSTANCE,
+                body: csharp.codeblock((writer) => {
+                    writer.writeLine("return new Headers(new Dictionary<string, string>");
+                    writer.writeLine("{");
+                    writer.indent();
+                    for (const header of this.context.getIdempotencyHeaders()) {
+                        // TODO: find a way to check if a header is nullable, and add nullable check
```
Swimburger, line 43: Is there an existing method to check whether something will be nullable?
- string
- object
- classes
- int?
- double?
but not
- int
- double
- etc.

I can't apply `?.` to everything as it won't compile for non-nullable types.

dsinghvi: can you check if the type of the header is a container + optional ?

dsinghvi: Don't see an existing method for it, but i could imagine it taking in a type reference and returning true/false

seed/csharp-sdk/idempotency-headers/src/SeedIdempotencyHeaders/Core/IIdempotentRequestOptions.cs
```diff
@@ -0,0 +1,8 @@
+namespace SeedIdempotencyHeaders.Core;
+
+internal interface IIdempotentRequestOptions : IRequestOptions
+{
+    public string IdempotencyKey { get; init; }
+    public int IdempotencyExpiration { get; init; }
+    internal Headers GetIdempotencyHeaders();
+}
```
dsinghvi, line 8: what's the benefit of having the interface here as opposed to `IdempotentRequstOptions` just inheriting from `IRequestOptions` ?

seed/csharp-sdk/oauth-client-credentials/src/SeedOauthClientCredentials/Core/Public/RequestOptions.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace SeedOauthClientCredentials;
 
-public partial class RequestOptions
+public partial class RequestOptions : IRequestOptions
```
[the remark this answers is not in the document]
Swimburger: Yes, and for future variations of request options. I'm using interfaces for the lack of multiple inheritance in C#.

seed/csharp-sdk/idempotency-headers/src/SeedIdempotencyHeaders/Core/IIdempotentRequestOptions.cs
Swimburger: Another options variation may be introduced in the future that may inherit from `IRequestOptions` or from `IIdempotentRequestOptions`

dsinghvi (approved):

PR #5180, opened 2024-11-15
feat(cli): Add RetryConfig to IR

amckinney:
This adds a new `RetryConfig` to the IR so that we can customize and centralize what HTTP status codes each generator should act upon. 

Eventually, we imagine this could hold other information like default retry attempts, and minimum/maximum delay, which are also SDK-specific for the time being.


dsinghvi (approved):

amckinney:
Closing this for now since we'll want to come back to this and add more properties (e.g. `minRetryDelay`, `maxRetryDelay`, etc).

PR #5181, opened 2024-11-15
fix(python): add `test_generator_cli.py`; upgrade to use latest generator-cli version.

eyw520:
@coderabbitai summary

### DIFF
--- a/fern/pages/changelogs/go-sdk/2024-11-15.mdx
+++ b/fern/pages/changelogs/go-sdk/2024-11-15.mdx
@@ -1,0 +1,4 @@
+## 0.31.1
+**`(internal):`** Adds additional tests to confirm the behavior of the `core.Retrier`.
+No functional, user-facing changes are included.
+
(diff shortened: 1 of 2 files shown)


generators/python/sdk/versions.yml
```diff
@@ -1,4 +1,11 @@
 # For unreleased changes, use unreleased.yml
+- version: 4.3.6
+  irVersion: 53
+  changelogEntry:
+    - type: fix
+      summary: |
+        Add `test_generator_cli.py`; upgrade to use latest cli version and compatibility.
```
dsinghvi, line 7: This is a user facing changelog -- thoughts on just talking about how readme generation is now fixed?

dsinghvi (approved):

generators/python/sdk/versions.yml
eyw520: That makes sense to me. Fixing.

PR #5184, opened 2024-11-15
fix(cli): recursive type checking exceeding max call stack.

eyw520:
@coderabbitai summary


packages/cli/generation/ir-generator/src/examples/validateObjectExample.ts
```diff
@@ -19,7 +19,8 @@ export function validateObjectExample({
     exampleResolver,
     workspace,
     example,
-    breadcrumbs
+    breadcrumbs,
+    depth = 0
```
dsinghvi, line 23: thoughts on not defaulting `depth` so that each caller is forced to pass it in and make sure it can't randomly get reset to 0?

packages/cli/generation/ir-generator/src/examples/validateTypeExample.ts
```diff
@@ -70,18 +75,22 @@ export function validateTypeExample({
...
                 exampleResolver,
                 workspace,
-                breadcrumbs
+                breadcrumbs,
+                depth
             });
         },
         undiscriminatedUnion: (rawUnion) => {
+            // console.log("validateUndiscriminatedUnionExample");
+            // console.log("validateUndiscriminatedUnionExample with Union", JSON.stringify(rawUnion, null, 2));
```
dsinghvi, line 84: nit: lets remove these before merging!

dsinghvi (approved):

packages/cli/generation/ir-generator/src/examples/validateObjectExample.ts
eyw520: Good catch, I agree with enforcing this.

packages/cli/generation/ir-generator/src/examples/validateTypeExample.ts
eyw520: Oops, should've caught these. Thanks for catching this.

PR #5187, opened 2024-11-15
feat(csharp): add pagination helper methods to C# SDK generator

Swimburger:
This feature works, but the code is still a little rough and needs some cleaning.

- The new `HttpPagerEndpointGenerator` generates pager methods that use the main endpoint methods for users to paginate over items in a paginated endpoint.
- `AsIs/Page.Template.cs` holds a list of items in a single page
- `AsIs/Pager.Template.cs` has a public abstract `Pager<Item>` class for a user to interact with, and two internal implementations `OffsetPager` and CursorPager`. The bulk of the logic to paginate is stored within these implementations, and the remaining parts are passed into the pagers as delegates.

⚠️ Assumption: For cursor-based pagination, the cursor is assumed to be a string. This is probably correct in 95% of cases, but that's just a guess. Does the cursor type need to be configurable?

⚠️ Big change: The generator thus far assumes that each endpoint generates one method on a client. This PR changes that so that an endpoint generator can generate multiple methods. For generated SDKs, this isn't too big of a change, but this also affects generated snippets and references. There's some code duplication to quickly get snippets and references to work as expected with this pager variation method for the endpoint
1. Should we stick to one method per endpoint to reduce the amount of change in this generator and downstream dependencies? Or
2. Should we push forward on an endpoint with multiple methods and refactor the generator at a larger scale for this to be supported seamlessly?

### DIFF
--- a/generators/csharp/codegen/src/AsIs.ts
+++ b/generators/csharp/codegen/src/AsIs.ts
@@ -39,14 +39,14 @@ export const AsIsFiles = {
         RawClientTests: "test/RawClientTests.Template.cs",
         StringEnumSerializerTests: "test/StringEnumSerializerTests.Template.cs",
         Pagination: [
-            "test/Pagination/GuidCursorTestCase.Template.cs",
-            "test/Pagination/HasNextPageOffsetTestCase.Template.cs",
-            "test/Pagination/IntOffsetTestCase.Template.cs",
-            "test/Pagination/LongOffsetTestCase.Template.cs",
-            "test/Pagination/NoRequestCursorTestCase.Template.cs",
-            "test/Pagination/NoRequestOffsetTestCase.Template.cs",
-            "test/Pagination/StepOffsetTestCase.Template.cs",
-            "test/Pagination/StringCursorTestCase.Template.cs"
+            "test/Pagination/GuidCursorTest.Template.cs",
+            "test/Pagination/HasNextPageOffsetTest.Template.cs",
+            "test/Pagination/IntOffsetTest.Template.cs",
+            "test/Pagination/LongOffsetTest.Template.cs",
+            "test/Pagination/NoRequestCursorTest.Template.cs",
+            "test/Pagination/NoRequestOffsetTest.Template.cs",
+            "test/Pagination/StepOffsetTest.Template.cs",
+            "test/Pagination/StringCursorTest.Template.cs"
         ]
     }
 };
--- a/generators/csharp/codegen/src/asIs/test/Pagination/GuidCursorTest.Template.cs
+++ b/generators/csharp/codegen/src/asIs/test/Pagination/GuidCursorTest.Template.cs
@@ -4,19 +4,20 @@ using <%= namespace%>.Core;
 namespace <%= namespace%>.Test.Core.Pagination;
 
 [TestFixture(Category = "Pagination")]
-public class NoRequestCursorTestCase
+public class GuidCursorTest
 {
     [Test]
-    public async Task CursorPagerShouldWorkWithStringCursor()
+    public async Task CursorPagerShouldWorkWithGuidCursors()
     {
         var pager = CreatePager();
         await AssertPager(pager);
     }
 
-    private const string? Cursor1 = null;
-    private const string Cursor2 = "cursor2";
-    private const string Cursor3 = "cursor3";
-    private string? _cursorCopy;
+    private static readonly Guid? Cursor1 = null;
+    private static readonly Guid Cursor2 = new("00000000-0000-0000-0000-000000000001");
+    private static readonly Guid Cursor3 = new("00000000-0000-0000-0000-000000000001");
+    private Guid? _cursorCopy;
+
     private Pager<object> CreatePager()
     {
         var responses = new List<Response>
@@ -27,7 +28,7 @@ public class NoRequestCursorTestCase
                 {
                     Items = ["item1", "item2"]
                 },
-                Cursor = new ()
+                Cursor = new()
                 {
                     Next = Cursor2
                 }
@@ -38,7 +39,7 @@ public class NoRequestCursorTestCase
                 {
                     Items = ["item1"]
                 },
-                Cursor = new ()
+                Cursor = new()
                 {
                     Next = Cursor3
                 }
@@ -49,7 +50,7 @@ public class NoRequestCursorTestCase
                 {
                     Items = []
                 },
-                Cursor = new ()
+                Cursor = new()
                 {
                     Next = null
                 }
@@ -57,13 +58,16 @@ public class NoRequestCursorTestCase
         }.GetEnumerator();
         _cursorCopy = Cursor1;
         Pager<object> pager = new CursorPager<
-            Request?,
+            Request,
             object?,
             Response,
-            string,
+            Guid?,
             object
         >(
-            null,
+            new()
+            {
+                Cursor = Cursor1
+            },
             null,
             (_, _, _) =>
             {
@@ -109,7 +113,7 @@ public class NoRequestCursorTestCase
 
     private class Request
     {
-        public required string? Cursor { get; set; }
+        public required Guid? Cursor { get; set; }
     }
 
     private class Response
@@ -125,6 +129,6 @@ public class NoRequestCursorTestCase
 
     private class Cursor
     {
-        public required string? Next { get; set; }
+        public required Guid? Next { get; set; }
     }
 }
--- a/generators/csharp/codegen/src/asIs/test/Pagination/HasNextPageOffsetTest.Template.cs
+++ b/generators/csharp/codegen/src/asIs/test/Pagination/HasNextPageOffsetTest.Template.cs
@@ -4,7 +4,7 @@ using <%= namespace%>.Core;
 namespace <%= namespace%>.Test.Core.Pagination;
 
 [TestFixture(Category = "Pagination")]
-public class HasNextPageOffsetTestCase
+public class HasNextPageOffsetTest
 {
     [Test]
     public async Task OffsetPagerShouldWorkWithHasNextPage()
--- a/generators/csharp/codegen/src/asIs/test/Pagination/IntOffsetTest.Template.cs
+++ b/generators/csharp/codegen/src/asIs/test/Pagination/IntOffsetTest.Template.cs
@@ -4,7 +4,7 @@ using <%= namespace%>.Core;
 namespace <%= namespace%>.Test.Core.Pagination;
 
 [TestFixture(Category = "Pagination")]
-public class IntOffsetTestCase
+public class IntOffsetTest
 {
     [Test]
     public async Task OffsetPagerShouldWorkWithIntPage()
--- a/generators/csharp/codegen/src/asIs/test/Pagination/LongOffsetTest.Template.cs
+++ b/generators/csharp/codegen/src/asIs/test/Pagination/LongOffsetTest.Template.cs
@@ -1,10 +1,10 @@
 using NUnit.Framework;
-using SeedPagination.Core;
+using <%= namespace%>.Core;
 
-namespace SeedPagination.Test.Core.Pagination;
+namespace <%= namespace%>.Test.Core.Pagination;
 
 [TestFixture(Category = "Pagination")]
-public class LongOffsetTestCase
+public class LongOffsetTest
 {
     [Test]
     public async Task OffsetPagerShouldWorkWithLongPage()
@@ -17,12 +17,43 @@ public class LongOffsetTestCase
     {
         var responses = new List<Response>
         {
-            new() { Data = new() { Items = ["item1", "item2"] } },
-            new() { Data = new() { Items = ["item1"] } },
-            new() { Data = new() { Items = [] } },
+            new()
+            {
+                Data = new()
+                {
+                    Items = ["item1", "item2"]
+                }
+            },
+            new()
+            {
+                Data = new()
+                {
+                    Items = ["item1"]
+                }
+            },
+            new()
+            {
+                Data = new()
+                {
+                    Items = []
+                }
+            }
         }.GetEnumerator();
-        Pager<object> pager = new OffsetPager<Request, object?, Response, long, object?, object>(
-            new() { Pagination = new() { Page = 1 } },
+        Pager<object> pager = new OffsetPager<
+            Request,
+            object?,
+            Response,
+            long,
+            object?,
+            object
+        >(
+            new()
+            {
+                Pagination = new()
+                {
+                    Page = 1
+                }
+            },
             null,
             (_, _, _) =>
             {
--- a/generators/csharp/codegen/src/asIs/test/Pagination/NoRequestCursorTest.Template.cs
+++ b/generators/csharp/codegen/src/asIs/test/Pagination/NoRequestCursorTest.Template.cs
@@ -1,10 +1,10 @@
 using NUnit.Framework;
-using SeedPagination.Core;
+using <%= namespace%>.Core;
 
-namespace SeedPagination.Test.Core.Pagination;
+namespace <%= namespace%>.Test.Core.Pagination;
 
 [TestFixture(Category = "Pagination")]
-public class NoRequestCursorTestCase
+public class NoRequestCursorTest
 {
     [Test]
     public async Task CursorPagerShouldWorkWithStringCursor()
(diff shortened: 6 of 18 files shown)


Swimburger:
The `mockResponse` is generated incorrectly in `ListWithExtendedResultsAndOptionalDataTest` and `ListWithExtendedResultsTest`.
The mockresponse JSON should have the `data` property. The list within `data` is optional so that's fine, but there should be a `data: {}` property since that isn't optional in Fern definition.

Swimburger:
Improvements:

- Support for different types in cursor pagination
- Support for int and long in offset pagination
- Initialize request if null before passing request to `setCursor` and `setOffset`
- Initialize nested properties towards cursor and offset property.

For initializing towards cursor and offset property, I assume the types don't have required fields and initialize an empty instance using `?? = new()`.

I'm worried if one of these types does have a required field, at which point the compiler will complain.

generators/csharp/sdk/src/endpoint/AbstractEndpointGenerator.ts
```diff
@@ -39,7 +53,60 @@ export abstract class AbstractEndpointGenerator {
...
+    }
+
+    protected getPaginationItemType(endpoint: HttpEndpoint) {
+        this.assertHasPagination(endpoint);
+        const listItemType = this.context.csharpTypeMapper.convert({
+            reference: endpoint.pagination._visit({
+                offset: (pagination) => pagination.results.property.valueType,
+                cursor: (pagination) => pagination.results.property.valueType,
+                _other: (pagination) => {
+                    throw new Error(`Unsupported pagination type: ${pagination.type}`);
```
amckinney, line 94: Prefer `switch` w/ `assertNever` if possible.

generators/csharp/sdk/src/endpoint/EndpointGenerator.ts
```diff
@@ -31,22 +31,32 @@ export class EndpointGenerator extends AbstractEndpointGenerator {
...
+    }
+
+    private isGrpcEndpoint(
+        grpcClientInfo: GrpcClientInfo | undefined,
+        endpoint: HttpEndpoint
+    ): grpcClientInfo is GrpcClientInfo {
+        // If the service is a grpc service, grpcClientInfo will not be null or undefined,
+        // so any endpoint will be generated as a grpc endpoint, unless the transport is overriden by setting type to http
+        return grpcClientInfo != null && endpoint.transport?.type !== "http";
     }
```
amckinney, line 61: Nice improvement here 👍

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                        writer.writeLine(
+                            `response => ${this.nullableDotGet("response", pagination.results)}?.ToList()`
+                        );
+                        writer.dedent();
+                        writer.writeTextStatement(")");
+                        writer.writeTextStatement("return pager");
+                    },
+                    _other: (pagination) => {
+                        throw new Error(`Unsupported pagination type: ${pagination.type}`);
+                    }
```
amckinney, line 441: nit: Prefer `switch` w/ `assertNever` here.

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                        if (pagination.hasNextPage) {
+                            writer.writeLine(`response => ${this.nullableDotGet("response", pagination.hasNextPage)}`);
+                        } else {
+                            writer.writeLine("null");
+                        }
+                        writer.dedent();
+                        writer.writeTextStatement(")");
+                        writer.writeTextStatement("return pager");
+                    },
+                    cursor: (pagination) => {
```
amckinney, line 404: nit: Could we put all of this in a separate function (e.g. `generateCursorPagerMethod`).

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                    throw new Error("Request parameter is required for pagination");
+                }
+                const unpagedEndpointMethodName = this.context.getEndpointMethodName(endpoint);
+                const unpagedEndpointResponseType = getEndpointReturnType({ context: this.context, endpoint });
+                if (!unpagedEndpointResponseType) {
+                    throw new Error("No return type found for endpoint");
+                }
+
+                endpoint.pagination?._visit({
+                    offset: (pagination) => {
```
amckinney, line 353: nit: Could we put all of this in a separate function (e.g. `generateOffsetPagerMethod`).

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                        writer.writeLine(`${optionsParamName},`);
+                        writer.writeLine(`${unpagedEndpointMethodName},`);
+                        writer.writeLine(`request => ${this.nullableDotGet("request", pagination.page)} ?? 0,`);
+
+                        writer.writeLine("(request, offset) => {");
+                        writer.indent();
+                        this.initializeNestedObjects(writer, "request", pagination.page);
+                        writer.writeTextStatement(`${this.dotGet("request", pagination.page)} = offset`);
+                        writer.dedent();
+                        writer.writeLine("},");
```
amckinney, line 385: Is it possible to use `csharp.instantiateClass` for this? We're otherwise repeating the logic required to write every individual parameter.

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+            return_,
+            body: csharp.codeblock((writer) => {
+                const requestParam = endpointSignatureInfo.requestParameter;
+                if (!requestParam) {
+                    throw new Error("Request parameter is required for pagination");
+                }
+                const unpagedEndpointMethodName = this.context.getEndpointMethodName(endpoint);
+                const unpagedEndpointResponseType = getEndpointReturnType({ context: this.context, endpoint });
+                if (!unpagedEndpointResponseType) {
+                    throw new Error("No return type found for endpoint");
```
amckinney, line 349: If we have to throw an error, let's be more specific. Something along the lines of:
```suggestion
                    throw new Error("Internal error; a response type is required for pagination endpoints");
```

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                        writer.writeLine(`${requestParam.name},`);
+                        writer.writeLine(`${optionsParamName},`);
+                        writer.writeLine(`${unpagedEndpointMethodName},`);
+
+                        writer.writeLine("(request, cursor) => {");
+                        writer.indent();
+                        this.initializeNestedObjects(writer, "request", pagination.page);
+                        writer.writeTextStatement(`${this.dotGet("request", pagination.page)} = cursor`);
+                        writer.dedent();
+                        writer.writeLine("},");
```
amckinney, line 429: Same thing as above w/ `csharp.instantiateClass`.

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+        const service = this.context.ir.services[serviceId];
+        if (service == null) {
+            throw new Error(`Unexpected no service with id ${serviceId}`);
+        }
+        const serviceFilePath = service.name.fernFilepath;
+        const requestBodyType = endpoint.requestBody?.type;
+        // TODO: implement these
+        if (requestBodyType === "fileUpload" || requestBodyType === "bytes") {
+            return undefined;
+        }
```
amckinney, line 603: Is this even relevant here?

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+        clientVariableName: string;
+        serviceId: ServiceId;
+        parseDatetimes: boolean;
+        additionalEndParameters?: csharp.CodeBlock[];
+        getResult?: boolean;
+    }): csharp.MethodInvocation | undefined {
+        const service = this.context.ir.services[serviceId];
+        if (service == null) {
+            throw new Error(`Unexpected no service with id ${serviceId}`);
+        }
```
amckinney, line 597: Can we use or add a `this.context.getServiceOrThrow`?

seed/csharp-sdk/pagination/src/SeedPagination/Core/Pager.cs
```diff
@@ -0,0 +1,211 @@
...
+            if (hasStep)
+            {
+                longOffset += items?.Count ?? 1;
+            }
+            else
+            {
+                longOffset++;
+            }
+
+            // ensure there's a request object to set the offset on
```
amckinney, line 129: nit: I'd remove most of the comments in this file, especially given that it will be generated as a core utility.

seed/csharp-sdk/pagination/src/SeedPagination/Users/UsersClient.cs
```diff
@@ -29,7 +29,390 @@ internal UsersClient(RawClient client)
...
+            (request, cursor) =>
+            {
+                request.Pagination ??= new();
+                request.Pagination.Cursor = cursor;
+            },
+            response => response?.Page?.Next?.StartingAfter,
+            response => response?.Data?.ToList()
+        );
+        return pager;
+    }
```
amckinney, line 88: The generated code looks great 👏

generators/csharp/sdk/src/test-generation/mock-server/MockServerTestGenerator.ts
```diff
@@ -135,29 +129,44 @@ export class MockServerTestGenerator extends FileGenerator<CSharpFile, SdkCustom
...
+                            writer.addReference(this.context.getFluentAssetionsJsonClassReference());
+                            writer.writeNode(this.context.getJTokenClassReference());
+                            writer.write(".Parse(mockResponse).Should().BeEquivalentTo(");
+                            writer.writeNode(this.context.getJTokenClassReference());
+                            writer.write(".Parse(");
+                            writer.writeNode(this.context.getJsonUtilsClassReference());
+                            writer.writeTextStatement(".Serialize(response)))");
+                        } else if (responseBodyType === "text") {
+                            writer.writeTextStatement("Assert.That(response, Is.EqualTo(mockResponse))");
+                        }
```
amckinney, line 165: Let's use `switch` w/ `assertNever` - that way we get compile breaks if we ever add more response body types.

When we have the switch, it might be worth rolling out helper functions for these cases (but I'll leave it to you to decide).

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+            : this.generateHttpEndpointSnippet({
+                  example,
+                  endpoint,
+                  clientVariableName,
+                  serviceId,
+                  additionalEndParameters,
+                  getResult,
+                  parseDatetimes
+              });
+    }
```
amckinney, line 520: Nice 👏

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +309,347 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+        this.assertHasPagination(endpoint);
+        const endpointSignatureInfo = this.getEndpointSignatureInfo({ serviceId, endpoint });
+        const parameters = [...endpointSignatureInfo.baseParameters];
+        const optionsParamName = this.getRequestOptionsParamNameForEndpoint({ endpoint });
+        const requestOptionsParam = this.getRequestOptionsParameter({ endpoint });
+        const requestOptionsType = requestOptionsParam.type;
+        parameters.push(requestOptionsParam);
+        const itemType = this.getPaginationItemType(endpoint);
+        const return_ = this.getPagerReturnType(endpoint);
+        const snippet = this.getHttpPagerMethodSnippet({ endpoint });
```
amckinney, line 333: There's a lot of parameters required here - I wonder if it's be worth rolling out a `PaginatedEndpointGenerator` class (or something similar) that could define all these properties as member variables, so that the following implementation could be reduced / a little more concise and readable.

Not totally necessary to do that here, but let me know your thoughts.

generators/csharp/sdk/src/endpoint/AbstractEndpointGenerator.ts
```diff
@@ -7,6 +7,8 @@ import { getEndpointRequest } from "./utils/getEndpointRequest";
 import { getEndpointReturnType } from "./utils/getEndpointReturnType";
 import { EndpointSignatureInfo } from "./EndpointSignatureInfo";
 
+type PagingEndpoint = HttpEndpoint & { pagination: NonNullable<HttpEndpoint["pagination"]> };
```
amckinney, line 10: Interesting, we might want to use this more elsewhere!

amckinney (commented):
Overall looking great - mostly comments on implementation, and only minor details for the generated code (e.g. comments, etc).

Swimburger:
I added a shallow clone using the record `with {}` feature. Future versions of C# may include a DeepClone feature for records.

.vscode/settings.json
```diff
@@ -4,16 +4,24 @@
...
+  "eslint.workingDirectories": [
+    "./packages",
+    "./generators"
+  ],
   "prettier.prettierPath": "node_modules/prettier/index.js",
   "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.enablePromptUseWorkspaceTsdk": true,
-  "editor.formatOnSave": true,
-  "editor.codeActionsOnSave": ["source.fixAll", "source.sortMembers"],
+  "editor.formatOnSave": false,
```
amckinney, line 17: Was this intentional? It might be better to put this in your user settings that aren't committed to the repository.

generators/csharp/codegen/src/asIs/Pager.Template.cs
```diff
@@ -0,0 +1,207 @@
...
+            switch (offset)
+            {
+                case int:
+                    _setOffset(_request, (TOffset)(object)(int)longOffset);
+                    break;
+                case long:
+                    _setOffset(_request, (TOffset)(object)longOffset);
+                    break;
+                default:
+                    throw new InvalidOperationException("Offset must be int or long");
```
amckinney, line 138: Fine to leave this in, but ideally we cover this during SDK generation so that it's effectively unreachable.

generators/csharp/sdk/src/endpoint/AbstractEndpointGenerator.ts
```diff
@@ -39,7 +54,63 @@ export abstract class AbstractEndpointGenerator {
...
+            unboxOptionals: true
+        });
+
+        if (listItemType.internalType.type !== "list") {
+            throw new Error(
+                `Pagination result type for endpoint ${endpoint.name.originalName} must be a list, but is ${listItemType.internalType.type}.`
+            );
+        }
+        const itemType = listItemType.internalType.value;
+        return itemType;
```
amckinney, line 110: nit: We can return directly here:
```suggestion
        return listItemType.internalType.value;
```

generators/csharp/sdk/src/endpoint/AbstractEndpointGenerator.ts
```diff
@@ -70,6 +141,20 @@ export abstract class AbstractEndpointGenerator {
...
+            return false;
+        }
+        return endpoint.pagination !== undefined;
+    }
+
+    protected assertHasPagination(endpoint: HttpEndpoint): asserts endpoint is PagingEndpoint {
+        if (this.hasPagination(endpoint)) {
+            return;
+        }
+        throw new Error(`Endpoint ${endpoint.name.originalName} is not a pager endpoint`);
```
amckinney, line 155: nit:
```suggestion
        throw new Error(`Endpoint ${endpoint.name.originalName} is not a paginated endpoint`);
```

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +312,401 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                        writer.writeTextStatement(`${this.dotGet("request", pagination.page)} = cursor`);
+                        writer.dedent();
+                        writer.writeLine("}");
+                    }),
+                    csharp.codeblock(`response => ${this.nullableDotGet("response", pagination.next)}`),
+                    csharp.codeblock(`response => ${this.nullableDotGet("response", pagination.results)}?.ToList()`)
+                ]
+            })
+        );
+        writer.writeTextStatement("return pager");
```
amckinney, line 510: nit: Should we just return the pager directly rather than using `var`?

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -312,4 +312,401 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                    csharp.codeblock(`response => ${this.nullableDotGet("response", pagination.results)}?.ToList()`),
+                    csharp.codeblock(
+                        pagination.hasNextPage
+                            ? `response => ${this.nullableDotGet("response", pagination.hasNextPage)}`
+                            : "null"
+                    )
+                ]
+            })
+        );
+        writer.writeTextStatement("return pager");
```
amckinney, line 456: nit: Same thing with returning directly.

amckinney (approved):
Nice!! 🎉

PR #5191, opened 2024-11-17
Add Block str support for python templated strings

dvargasfuertes:
@coderabbitai summary

### DIFF
--- a/generators/python-v2/ast/src/TypeInstantiation.ts
+++ b/generators/python-v2/ast/src/TypeInstantiation.ts
@@ -4,7 +4,7 @@ import { Writer } from "./core/Writer";
 import { Type } from "./Type";
 import { Reference } from "./Reference";
 
-type InternalTypeInstantiation = Int | Float | Bool | Str | Bytes | List | Set | Tuple | Dict | None | Uuid;
+type InternalTypeInstantiation = Int | Float | Bool | Str | BlockStr | Bytes | List | Set | Tuple | Dict | None | Uuid;
 
 interface Int {
     type: "int";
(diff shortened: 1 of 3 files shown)


generators/python-v2/ast/src/TypeInstantiation.ts
```diff
@@ -86,6 +91,10 @@ export class TypeInstantiation extends AstNode {
         return new this({ type: "str", value });
     }
 
+    public static blockStr(value: string): TypeInstantiation {
+        return new this({ type: "block_str", value });
+    }
+
```
amckinney, line 97: Let's consolidate this functionality in the `str` helper above. Check out PHP's `multiline` option in [this method](https://github.com/fern-api/fern/blob/1908ee6716e2779e1af4a71865673397fb26f55b/generators/php/codegen/src/ast/Type.ts#L305), which prevents the caller from specifying `multiline` unless they actually care about it. That way there's a single `str` AST node, and it knows how to render itself based on the options provided.

I imagine we'd have something like:

```typescript
public static str(value: string, { block }: { block?: boolean } = {}): Type {
```

generators/python-v2/ast/src/TypeInstantiation.ts
```diff
@@ -145,6 +154,11 @@ export class TypeInstantiation extends AstNode {
             case "str":
                 writer.write(`"${this.internalType.value}"`);
                 break;
+            case "block_str": {
+                const suffix = this.internalType.value.endsWith("\n") ? "\\" : "\n";
+                writer.write(`"""\\\n${this.internalType.value}${suffix}"""`);
```
amckinney, line 159: This behavior would be handled in the `str` case in `if (this.internalType.block) { ... }`.

amckinney (approved):
Stamping to unblock, assuming that we move `block_str` into `str` as an option.

dvargas92495:
Closing in favor of another previously merged pr by our team

PR #5222, opened 2024-11-19
fix(docs): update SDK documentation IA with redirects

dannysheridan:
Reverts fern-api/fern#5220


dannysheridan:
@chdeskur I think your original PR had formatting changes. It'd be nice to remove that in future PRs (unless the whole PR is just a formatting change one-time)

PR #5231, opened 2024-11-20
feat(cli): Add support for in-lined path parameters

amckinney:
This adds support for including `path-parameters` in the `request` section of the endpoint definition. This includes:

1. A new `fern check` rule which ensures that only one `path-parameters` block is present for any endpoint.
2. The `inline-path-parameters` OpenAPI importer setting.

With this, both `path-parameters` formats are supported so that users can configure an endpoint like so:

```yaml
service:
  auth: false
  base-path: /base/{baseParameter}
  path-parameters:
    fakeBaseParameter: string
  endpoints:
    legacy:
      path: /path/{parameter}/to/endpoint
      path-parameters:
        parameter: string
      method: POST

    inlined:
      path: /path/{parameter}/to/endpoint
      request:
        name: InlinedRequest
        path-parameters:
          parameter: string
      method: POST
```

Note that a new `path-parameters` test definition was added, and some generators might produce an in-lined request type alongside positional path parameters. This case should be handled within each generator with a generator-specific `inlinePathParameters` configuration option.

### DIFF
--- a/fern/pages/changelogs/cli/2024-11-21.mdx
+++ b/fern/pages/changelogs/cli/2024-11-21.mdx
@@ -1,6 +1,4 @@
-
-## 0.45.0-rc51
-**`(feat):`** Add support for the `inline-path-parameters` setting in the OpenAPI importer.
-
 ## 0.45.0-rc50
 **`(fix):`** Increase max recursive depth allowed for example validation.
+
+


packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildEndpoint.ts
```diff
@@ -606,3 +623,15 @@ function getRequest({
...
+function endpointRequestSupportsInlinedPathParameters({
+    context,
+    request
+}: {
+    context: OpenApiIrConverterContext;
+    request: Request | undefined;
+}): boolean {
+    // octet-stream requests do not support named request wrappers,
+    // so we can't inline path parameters for them.
+    return context.inlinePathParameters && request?.type !== "octetStream";
```
dsinghvi, line 636: nit: if we make this a switch it would be great so that folks are forced to update if we add a new type

packages/cli/fern-definition/validator/src/rules/valid-path-parameters-configuration/valid-path-parameters-configuration.ts
```diff
@@ -0,0 +1,26 @@
...
+                httpEndpoint: ({ endpoint }) => {
+                    if (
+                        endpoint["path-parameters"] != null &&
+                        typeof endpoint.request !== "string" &&
+                        endpoint?.request?.["path-parameters"] != null
+                    ) {
+                        return [
+                            {
+                                severity: "error",
+                                message: "Endpoint has path-parameters defined in both endpoint and request."
```
dsinghvi, line 17: nit: would update this error message to `path-parameters cannot be defined in both endpoint and request`

packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
```diff
@@ -181,6 +181,8 @@ types:
     properties:
       wrapperName: commons.Name
       bodyKey: commons.Name
+      includePathParameters: boolean
+      onlyPathParameters: boolean
```
dsinghvi, line 185: chatted offline -- this needs to become optional

dsinghvi (approved):

packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildEndpoint.ts
amckinney: Good call, updated!

PR #5233, opened 2024-11-20
fix(php): add `list` to PHP reserved keywords; prevent escaping PHP method names.

eyw520:
@coderabbitai summary


generators/php/sdk/src/SdkGeneratorContext.ts
```diff
@@ -74,6 +74,12 @@ export class SdkGeneratorContext extends AbstractPhpGeneratorContext<SdkCustomCo
     }
 
     public getEndpointMethodName(endpoint: HttpEndpoint): string {
+        // TODO: Propogate reserved keywords through IR via CasingsGenerator.
+        const reservedKeywords = ["use", "clone", "list"];
```
amckinney, line 78: nit: Can we declare this in [constants.ts](https://github.com/fern-api/fern/blob/main/generators/php/sdk/src/constants.ts) as `RESERVED_METHOD_NAMES`?

amckinney (approved):

PR #5236, opened 2024-11-20
fix(python): Include content-type header in endpoint generation

eyw520:
## Description
Include content-type header in endpoint generation as part of fix for FER-3584

## Changes Made
- Linting fixes to endpoint_function_generator.py
- Include content_type value when available in generating request headers.

## Testing
- Basis Theory generated Python SDK includes custom patch header.

### DIFF
--- a/seed/python-sdk/grpc-proto-exhaustive/src/seed/dataservice/client.py
+++ b/seed/python-sdk/grpc-proto-exhaustive/src/seed/dataservice/client.py
@@ -75,6 +75,9 @@ class DataserviceClient:
                 ),
                 "namespace": namespace,
             },
+            headers={
+                "content-type": "application/json",
+            },
             request_options=request_options,
             omit=OMIT,
         )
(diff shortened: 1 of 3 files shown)


generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
```diff
@@ -1165,6 +1173,12 @@ def _get_headers_for_endpoint(
         if endpoint.idempotent:
             ir_headers += idempotency_headers
 
+        if endpoint.request_body is not None:
+            unioned_value = endpoint.request_body.get_as_union()
+            if isinstance(unioned_value, ir_types.InlinedRequestBody):
```
dsinghvi, line 1178: can we check the discriminant here instead? `unioned_value.type === "inlined..."`

dsinghvi (approved):

generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
eyw520: resolved.

PR #5244, opened 2024-11-21
feat(go): Add inlinePathParameters configuration

amckinney:
This adds the new `inlinePathParameters` configuration option so that users can generate their in-lined request types to contain the path parameters specified by the endpoint (which are generated as positional parameters by default).

With this, users are now presented with path parameter endpoint signatures that look like the following:

```go
func (c *Client) GetUser(
	ctx context.Context,
	request *acme.GetUserRequest,
	opts ...option.RequestOption,
) error {
        ...
}
```

Note that this also updates a couple other minor details:

1. Updates the Dynamic IR to contain the `includePathParameters` and `onlyPathParameters` fields in the `InlinedRequest`.
2. Updates the DynamicSnippetsGenerator to fix the dynamic snippets produced when this configuration is toggled.
    * Note that a follow-up is required to use the new configuration introduced in the PR here (denoted by a TODO).
3. Updates the `AbstractDynamicSnippetGeneratorContext` to reference the same IR version.
    * This is a temporary fix - in a follow-up, we'll remove the constraint that the two IR versions are consistent between these packages.

### DIFF
--- a/fern/pages/changelogs/go-sdk/2024-11-21.mdx
+++ b/fern/pages/changelogs/go-sdk/2024-11-21.mdx
@@ -1,3 +1,9 @@
 ## 0.33.0
 **`(feat):`** Add support for the `inlinePathParameters` configuration option, which generates path parameters in the generated request type (if any) instead of as separate positional parameters.
+```yaml # generators.yml
+- name: fern-api/fern-go-sdk
+  version: 0.33.0
+  config:
+    inlinePathParameters: true
+```
 


generators/go/sdk/versions.yml
```diff
@@ -1,3 +1,11 @@
+- version: 0.33.0
+  changelogEntry:
+    - type: feat
+      summary: >-
+        Add support for the `inlinePathParameters` configuration option, which generates
+        path parameters in the generated request type (if any) instead of as separate
+        positional parameters.
```
dsinghvi, line 7: nit: can you add an example of the `generators.yml` including that configuration

dsinghvi:
```yml generators.yml 
groups: 
  - go-sdk: 
       name: fernapi/fern-go-sdk
       version: 0.33.0
       config: 
          inlinePathParameters: true
```

dsinghvi (approved):

generators/go/sdk/versions.yml
amckinney: Good call, will add it to the earlier changelog entry I had for `inlineFileProperties`, too.

PR #5247, opened 2024-11-21
fix(typescript): add endpoint name to TimeoutError message

eyw520:
## Description
Improve generated `GeneratedTimeoutSdkError` to include endpoint name in message body for improved debugging.

## Changes Made
- Modify `GeneratedTimeoutSdkError` error class to take a message parameter.
- Endpoint error case generation passes the endpoint name into the thrown error.

## Testing
- [x] Regenerated TS SDKs -- compiled and verified behavior.

### DIFF
--- a/generators/typescript/sdk/generator/package.json
+++ b/generators/typescript/sdk/generator/package.json
@@ -55,6 +55,7 @@
     "@types/uuid": "^9.0.8",
     "ts-morph": "^15.1.0",
     "ts-poet": "^6.7.0",
+    "url-join": "^5.0.0",
     "uuid": "^9.0.1"
   },
   "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2526,6 +2526,9 @@ importers:
       ts-poet:
         specifier: ^6.7.0
         version: 6.7.0
+      url-join:
+        specifier: ^5.0.0
+        version: 5.0.0
       uuid:
         specifier: ^9.0.1
         version: 9.0.1


generators/typescript/sdk/client-class-generator/src/endpoints/default/endpoint-response/GeneratedThrowingEndpointResponse.ts
```diff
@@ -772,7 +772,12 @@ export class GeneratedThrowingEndpointResponse implements GeneratedEndpointRespo
                         ),
                         [
                             ts.factory.createThrowStatement(
-                                context.timeoutSdkError.getGeneratedTimeoutSdkError().build(context)
+                                context.timeoutSdkError
+                                    .getGeneratedTimeoutSdkError()
+                                    .build(
+                                        context,
+                                        `Timeout exception generated in endpoint ${this.endpoint.name.camelCase.unsafeName}.`
```
amckinney, line 779: We might actually want to include the full endpoint name representation (e.g. `POST /users/{userId}`). You can use a combination of the endpoint method and [this function](https://github.com/fern-api/fern/blob/cf894f08ab4d3f0dad59f4512a3c822ac558530e/generators/typescript/sdk/generator/src/SdkGenerator.ts#L1074) to determine that. 

Otherwise, it'll be too difficult to know what endpoint actually timed out (e.g. `list` could be `user.list` or `organization.list`, etc).

amckinney (commented):
Implementation looks good, will want to see the `seed` testdata to make sure everything is good to go though.

generators/typescript/sdk/client-class-generator/src/endpoints/default/endpoint-response/GeneratedThrowingEndpointResponse.ts
dsinghvi: From slack +1, ideal message would be `Timeout of 60s exceeded in POST /users/v1` but the timeout is configurable

eyw520: Great, implemented!

dsinghvi (approved):
🔥

PR #5248, opened 2024-11-21
feat(cli): Add `inline` property to type references from properties in Fern definition with OpenAPI importer

Swimburger:
## Description
In the Fern definition, you can express your desire to inline a type from a property on a type or request:

```
service:
  endpoints:
    endpointName:
      ...
      request:
        ...
        body:
          ...
          properties:
            bar:
              type: InlineType

types:
  RootType:
    properties:
      bar:
        type: InlineType
  InlineType:
    inline: true
    properties:
      ...
```

The IR generated from this Fern definition will store the desire to inline a type on the type declaration.

## Changes Made

- Update Fern definition to allow `inline` true on declarations
- Update OpenAPI parser and OpenAPI IR to pass `inline` to the generated Fern definition
- Update Fern to IR generator to pass on `inline` to types

## Testing

Added more test definitions  + evaluate existing test definition snapshots.


fern/apis/fern-definition/definition/types.yml
```diff
@@ -1,223 +1,224 @@
...
       - BaseTypeReferenceSchema
       - commons.WithName
       - commons.WithAudiences
-    properties: 
+    properties:
       type: string
-  
-  BaseTypeDeclarationSchema: 
-    extends: 
+      inline: optional<boolean>
```
dsinghvi, line 83: i assume this is the only thing that changed? nit: would be nice to remove the formatting changes for this PR

packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
```diff
@@ -727,3 +731,49 @@ function filterServiceTypeReferenceInfoForAudiences(
         typesReferencedOnlyByService: filteredTypesReferencedOnlyByService
     };
 }
+
+type NamedObjectProperty = ObjectProperty & { valueType: TypeReference.Named };
+function markInlineTypes(types: Record<string, TypeDeclaration>) {
```
dsinghvi, line 736: nit: would name this `markInlineTypeDeclarations`

packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
```diff
@@ -727,3 +731,49 @@ function filterServiceTypeReferenceInfoForAudiences(
...
+    // find types that have properties containing inline types
+    const namedProps = Object.values(types).flatMap((type) => {
+        if (type.shape.type !== "object") {
+            return [];
+        }
```
dsinghvi, line 741: what happens if an inline type is "extended" or if a discriminated union references the inline type as a variant?

dsinghvi: I think you need to iterate over every single TypeReference and store the reference. Maybe we make `parseTypeReference` stateful? 

https://github.com/fern-api/fern/blob/cf894f08ab4d3f0dad59f4512a3c822ac558530e/packages/cli/generation/ir-generator/src/converters/type-declarations/convertObjectTypeDeclaration.ts#L36C24-L36C67

packages/cli/api-importers/openapi/openapi-ir/fern/definition/parseIr.yml
```diff
@@ -283,6 +283,7 @@ types:
         type: string
         docs: |
           A unique name for the property.
+      inline: optional<boolean>
```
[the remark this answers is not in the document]
dsinghvi: In other words, in openapi-ir-to-fern when you call `buildTypeDeclaration()` and then that calls `buildTypeReference` and that calls `buildTypeDeclaration` again, you know that the second declaration call is for an inlined type.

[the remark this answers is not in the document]
dsinghvi: So if we just tracked depth of call, then i think we should be able to control inline based on that

fern/apis/fern-definition/definition/types.yml
```diff
@@ -1,223 +1,231 @@
...
       encoding: optional<encoding.EncodingSchema>
       source: optional<source.SourceSchema>
-  
-  ### Aliases #### 
 
-  AliasSchema: 
+  ### Aliases ####
+
+  AliasSchema:
     extends: BaseTypeDeclarationSchema
```
amckinney, line 97: Should `WithInline` just be part of `BaseTypeDeclarationSchema`? It looks like we omitted it from `AliasSchema` which is reasonable, but it's theoretically possible that we generate an alias with a bad name too, and we could just as easily create an in-lined alias type.

That way any TypeDeclaration moving forward will _always_ be able to be in-lined. Curious to hear your thoughts though - I realize this case is less important overall.

packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildTypeDeclaration.ts
```diff
@@ -610,3 +652,7 @@ function convertPropertyTypeReferenceToTypeDefinition(
         };
     }
 }
+
+function getInline(declarationDepth: number): boolean | undefined {
+    return declarationDepth > 0 ? true : undefined;
+}
```
amckinney, line 658: It might be worth leaving a function-level comment on this one - I could see us forgetting the rationale around how we're detecting whether or not a type is in-lined. It looks like this was inspired by [this comment](https://github.com/fern-api/fern/pull/5248#discussion_r1852938762)?

amckinney (approved):
Looks good to me - only a couple comments worth considering before merging!

PR #5254, opened 2024-11-22
fix(python): Fix Reference Tracking for `Class` and `Method` AST Nodes

noanflaherty:
## Description
Python Classes and Methods weren't correctly tracking their internal references. This PR fixes these issues.

## Changes Made
- Update constructor of `Class` AST Node
- Update constructor of `Method` AST Node

## Testing
- [ X] Unit tests added/updated

### DIFF
--- a/generators/python-v2/ast/src/Method.ts
+++ b/generators/python-v2/ast/src/Method.ts
@@ -50,6 +50,20 @@ export class Method extends AstNode {
         this.type = type;
         this.decorators = decorators ?? [];
         this.static_ = static_ ?? false;
+
+        this.parameters.forEach((parameter) => {
+            this.inheritReferences(parameter);
+        });
+
+        this.inheritReferences(this.return);
+
+        this.decorators.forEach((decorator) => {
+            this.inheritReferences(decorator);
+        });
+
+        this.statements.forEach((statements) => {
+            this.inheritReferences(statements);
+        });
     }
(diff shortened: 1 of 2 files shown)


generators/python-v2/ast/src/Class.ts
```diff
@@ -35,6 +35,10 @@ export class Class extends AstNode {
         this.extends_.forEach((parentClassReference) => {
             this.inheritReferences(parentClassReference);
         });
+
+        this.decorators.forEach((decorator) => {
+            this.inheritReferences(decorator);
```
dvargas92495, line 40: [q] the decorator could be a function with arguments. Does this catch those parameters?

generators/python-v2/ast/src/Method.ts
```diff
@@ -50,6 +50,20 @@ export class Method extends AstNode {
         this.type = type;
         this.decorators = decorators ?? [];
         this.static_ = static_ ?? false;
+
+        this.parameters.forEach((parameter) => {
+            this.inheritReferences(parameter);
```
dvargas92495, line 55: [q] do parameters automatically import their type annotations if necessary?

dvargas92495 (commented):
👍

amckinney (approved):
Assuming @dvargas92495's comments are addressed, this lgtm

PR #5255, opened 2024-11-22
fix(cli): docs generation now preserves original model schema names.

eyw520:
## Description
Fix issue where docs generation fails when presented non-alphanumeric object schemas in non-discriminated unions.

## Changes Made

- Add `--preserve-schemas` flag to enable docs fern-definition generation
- `fern generate --docs` now preserves schemas when generating IR for docs.

## Testing


dsinghvi (approved):
Would be great to add a fixture for this in `openapi-ir-to-fern-tests`!

PR #5273, opened 2024-11-25
feat(docs): enable yaml-language-server in `generators.yml` across Fern's Fern Definitions

dannysheridan:
quite a meta title.


fern/apis/fhir/generators.yml
```diff
@@ -1,4 +1,5 @@
-# default-group: sdks
```
chdeskur, line 1: any reason to remove this

chdeskur (approved):

PR #5275, opened 2024-11-25
fix(cli): validated example generation

eyw520:
## Description
Use parameter validation information, when available, to generate validated primitives.

## Changes Made
- Update `generatePrimitiveExample` and add helper functions.
- Update validation test snapshots.

## Testing


packages/cli/generation/ir-generator/src/examples/generator/generatePrimitiveExample.ts
```diff
@@ -16,8 +23,12 @@ export function generatePrimitiveExample({
 }: generatePrimitiveExample.Args): ExampleGenerationSuccess<ExamplePrimitive> {
     switch (primitiveType.v1) {
         case "STRING": {
-            const jsonExample = fieldName ?? Examples.STRING;
-            return { type: "success", example: ExamplePrimitive.string({ original: jsonExample }), jsonExample };
+            let validation = undefined;
+            if (primitiveType.v2?.type === "string") {
+                const stringType = primitiveType.v2 as PrimitiveTypeV2.String;
+                validation = stringType.validation;
+            }
```
amckinney, line 30: Can we roll out a few helper functions for this? Something along the lines of `maybeDoubleValidation`, `maybeStringValidation`, and so on? That way we can get rid of the dangling `let` and just call it in-line (since `undefined` is a valid value. 

I'd also make it so that the new function below takes in named arguments so the call-site looks like the following:

```typescript
return generatePrimitiveStringExample({
    fieldName,
    validation: maybeDoubleValidation(primitiveType.v2)
})
```

I'll leave a comment below to demonstrate how that's done.

packages/cli/generation/ir-generator/src/examples/generator/generatePrimitiveExample.ts
```diff
@@ -74,3 +95,64 @@ export function generatePrimitiveExample({
             assertNever(primitiveType.v1);
     }
 }
+
+function generatePrimitiveStringExample(
+    fieldName: string | undefined,
+    validation: StringValidationRules | undefined
+): ExampleGenerationSuccess<ExamplePrimitive> {
```
amckinney, line 102:
```suggestion
function generatePrimitiveStringExample({
    fieldName,
    validation,
}: {
    fieldName: string | undefined,
    validation: StringValidationRules | undefined
}}
    fieldName: string | undefined,
    validation: StringValidationRules | undefined
): ExampleGenerationSuccess<ExamplePrimitive> {
```

packages/cli/generation/ir-generator/src/examples/generator/generatePrimitiveExample.ts
```diff
@@ -74,3 +95,64 @@ export function generatePrimitiveExample({
...
+    if (validation) {
+        const minLength = validation.minLength;
+        const maxLength = validation.maxLength;
+        if (maxLength) {
+            const maxLengthExample = "a".repeat(maxLength);
```
amckinney, line 107: Is this desirable? Showing examples with all `aaaaa` at the specific length isn't going to look great (even if they conform to the correct length).

I know it's tough to have a word generator that produces valid results for all the possible validation values, but I wonder if there's something more creative we can do here like:

1. Support some static string values up to a certain length (e.g. up to 32 characters like, "a", "be", "the", "food", and so on).
2. For anything longer, come up with some default message we use, then append continuous ellipses? 

Then  we could generate an example like this for strings that have a minimum of 80 characters:

```typescript
{
    message: "The content should be at least 80 characters: ................................."
}
```

 cc @dsinghvi

dsinghvi: Open to ideas here -- i think ideal is AI generating solutions, but the current solution/your solution are both better than the status quo

dsinghvi: I think 1 + 2 are ideal though, like you suggested

packages/cli/generation/ir-generator/src/examples/generator/generatePrimitiveExample.ts
eyw520: implemented

packages/cli/generation/ir-generator/src/examples/generator/generatePrimitiveExample.ts
```diff
@@ -74,3 +89,110 @@ export function generatePrimitiveExample({
...
+                jsonExample: maxLengthExample
+            };
+        } else if (minLength) {
+            const minLengthExample = getStringExampleOfLength(minLength);
+            return {
+                type: "success",
+                example: ExamplePrimitive.string({ original: minLengthExample }),
+                jsonExample: minLengthExample
+            };
+        }
```
amckinney, line 156: Let's swap the order here - we should prefer the minimum whenever possible. And in the case that we only have a maxmium and no minimum, we should prefer a length that is strictly smaller than the maximum. Otherwise users will see examples with a ton of content (e.g. 255 characters).

eyw520: Done!

PR #5278, opened 2024-11-25
fix(cli): Adds check for overwriting files in generate output

ajkpersonal:
## Description
This aims to fix the high vulnerability from the pen test that could lead to users accidentally overwriting important files using the generate command.

## Changes Made

- We add a check and user confirmation if the output directory for a generation has existing files that will be overwritten
- We provide an option to save approved output paths so that they won't be prompted again for the same path.
- Don't apply checks or prompts in CI environments.
- Add a --force flag that skips any prompts.

## Testing

- [x] Unit tests
- [x] Integ tests for CI env

## Visual
Here's quick loom showing what the prompting looks like to the user: https://www.loom.com/share/5839cfd9bd5f4e80a34338ab8b775302?sid=54ba7010-b2b4-4056-be94-0ad355e6ce21

### DIFF
--- a/fern/pages/changelogs/cli/2024-12-09.mdx
+++ b/fern/pages/changelogs/cli/2024-12-09.mdx
@@ -1,4 +1,4 @@
-## 0.45.2-rc0
+## 0.45.4-rc0
 **`(fix):`** The CLI prompts the user to confirm output directory overwrites on fern generate.
 
 


packages/cli/cli/src/utils/isCI.ts
```diff
@@ -0,0 +1,10 @@
+export function isCI(): boolean {
+    return Boolean(
+        process.env.CI || // Generic CI
```
dsinghvi, line 3: > Prefer using nullish coalescing operator (`??`) instead of a logical or (`||`)

small eslint warning that we should fix!

packages/cli/cli/src/persistence/getPathToOutputDirectoriesFile.ts
```diff
@@ -0,0 +1,13 @@
+import { AbsoluteFilePath, join, RelativeFilePath } from "@fern-api/fs-utils";
+import { homedir } from "os";
+
+const APPROVED_OUTPUT_DIRECTORIES_FILENAME = "approved-output-directories";
```
dsinghvi, line 4: nit: lets rename this to `allowed-directories`

packages/cli/cli/src/persistence/getPathToOutputDirectoriesFile.ts
```diff
@@ -0,0 +1,13 @@
...
+const LOCAL_STORAGE_FOLDER = process.env.LOCAL_STORAGE_FOLDER ?? ".fern";
```
dsinghvi, line 5: nit: there is a `consts` file that we can use for some of this 

https://github.com/fern-api/fern/blob/main/packages/cli/cli/src/constants.ts

packages/cli/cli/src/persistence/getPathToOutputDirectoriesFile.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function getPathToOutputDirectoriesFile(): AbsoluteFilePath {
```
dsinghvi, line 7: nit: `getPathToAllowedDirectoriesFile`

packages/cli/cli/src/commands/generate/generateAPIWorkspaces.ts
```diff
@@ -52,6 +54,23 @@ export async function generateAPIWorkspaces({
         token = currentToken;
     }
 
+    if (!isCI()) {
```
dsinghvi, line 57: nit: thoughts on adding a `--force` option so that users never get prompted? 

(not blocking merge at all)

packages/cli/cli/src/commands/generate/generateAPIWorkspaces.ts
```diff
@@ -52,6 +54,23 @@ export async function generateAPIWorkspaces({
...
+        for (const workspace of project.apiWorkspaces) {
+            for (const generator of workspace.generatorsConfiguration?.groups.flatMap((group) => group.generators) ??
+                []) {
+                if (generator.absolutePathToLocalOutput) {
+                    const { shouldProceed } = await checkOutputDirectory(
+                        generator.absolutePathToLocalOutput,
+                        cliContext
+                    );
+                    if (!shouldProceed) {
+                        cliContext.failAndThrow("Generation cancelled - output directory not empty");
```
dsinghvi, line 67: should be enough to just say `Generation cancelled` since the user opts for that

packages/cli/cli/src/persistence/storeOutputDirectories.ts
```diff
@@ -0,0 +1,8 @@
+import { mkdir, writeFile } from "fs/promises";
+import path from "path";
+import { getPathToOutputDirectoriesFile } from "./getPathToOutputDirectoriesFile";
+
+export async function storeOutputDirectories(outputDirectories: string[]): Promise<void> {
```
dsinghvi, line 5: How do you handle relative/absolute filepaths here? i.e. if a directory has been approved and its a relative path then should we normalize to absolute here?

dsinghvi: we have some type safe wrappers called `AbsoluteFilepath` and `RelativeFilepath` that may be helpful and also account for windows/mac/etc

packages/cli/cli/package.json
```diff
@@ -114,5 +114,8 @@
     "vitest": "^2.1.4",
     "yaml": "^2.4.5",
     "yargs": "^17.4.1"
+  },
+  "dependencies": {
+    "@inquirer/prompts": "^7.1.0"
```
dsinghvi, line 119: lets put all dependencies in `devDependencies` (this is for `tsup` which we use as a bundling solution in the CLI)

packages/cli/cli/src/commands/generate/checkOutputDirectory.ts
```diff
@@ -0,0 +1,82 @@
...
+            shouldProceed: false,
+            didSaveDirectory: false
+        };
+    }
+
+    // Ask if they want to save this as a valid output path
+    const shouldSave = await cliContext.confirmPrompt(
+        "Would you like to save this as a valid output directory for future generations?",
+        true
+    );
```
dsinghvi, line 72: nit: should we skip this step and assume yes -- one less step for the user

packages/cli/cli/src/cli-context/CliContext.ts
```diff
@@ -306,6 +307,19 @@ export class CliContext {
...
+     * @param message The message to display to the user
+     * @param defaultValue Optional default value (defaults to false)
+     * @returns Promise<boolean> representing the user's choice
+     */
+    public async confirmPrompt(message: string, defaultValue = false): Promise<boolean> {
+        return await confirm({
+            message,
+            default: defaultValue
+        });
+    }
```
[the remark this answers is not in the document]
ajkpersonal: Sure, added!

packages/cli/cli/src/persistence/getPathToOutputDirectoriesFile.ts
ajkpersonal: I prefer output directories to allowed directories - particularly if we autosave all output directories so that they are allowed (based on later comment to remove the 2nd prompt)

packages/cli/cli/src/commands/generate/generateAPIWorkspaces.ts
ajkpersonal: I think this would be great to have, will add tomo

packages/cli/cli/src/persistence/storeOutputDirectories.ts
ajkpersonal: We normalize to absolute when we check the filepath, and only pass in absolute filepaths here, I'll make that clear by updating the type of outputDirectories to AbsoluteFilepath

packages/cli/cli/src/commands/generate/checkOutputDirectory.ts
ajkpersonal: Sure, makes sense - mostly just needed for testing

dsinghvi (approved):

PR #5279, opened 2024-11-26
feat(csharp): Update main dependencies + misc changes

Swimburger:
## Description
Generated C# SDKs had transitive dependencies on vulnerable packages.

## Changes Made
* Add two dependencies that previously were transitive dependencies to ensure the generated SDKs use the patched versions without vulnerabilities.
  * `System.Net.Http` >= `4.3.4`
  * `System.Text.RegularExpressions` >= `4.3.1`
* Update other dependencies to the latest version:
  * `Portable.System.DateTimeOnly` = `8.0.2` (on net462 & netstandard2.0)
  * `PolySharp` = `1.15.0`
  * `OneOf` = `3.0.271`
  * `OneOf.Extended` = `3.0.271`
* Always include `PolySharp` and `PolySharpIncludeRuntimeSupportedAttributes`
* Remove `<NuGetAudit>false</NuGetAudit>` to ensure vulnerable packages are reported on build
* Use `TestContext.Out.WriteLine` instead of `TestContext.WriteLine` as advised by NUnit analyzer warnings.
* Call `Server.Dispose();` in `OneTimeTearDown` marked method as forced by NUnit analyzer error.
* rimraf C# seed folders to remove orphaned folders

## Testing

### DIFF
--- a/generators/csharp/codegen/src/project/CsharpProject.ts
+++ b/generators/csharp/codegen/src/project/CsharpProject.ts
@@ -504,10 +504,6 @@ class CsProj {
 
 ${projectGroup.join("\n")}
 
-    <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
-        <Reference Include="System.Net.Http" />
-    </ItemGroup>
-
     <ItemGroup Condition="'$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'netstandard2.0'">
         <PackageReference Include="Portable.System.DateTimeOnly" Version="8.0.2" />
     </ItemGroup>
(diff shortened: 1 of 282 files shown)


generators/csharp/codegen/src/project/CsharpProject.ts
```diff
@@ -504,23 +504,12 @@ class CsProj {
...
     <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
         <Reference Include="System.Net.Http" />
     </ItemGroup>
 
-    <ItemGroup Condition="'$(TargetFramework)' == 'net7.0' Or '$(TargetFramework)' == 'net6.0' Or '$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'netstandard2.0'">
-        <PackageReference Include="PolySharp" Version="1.14.1">
-            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-            <PrivateAssets>all</PrivateAssets>
-        </PackageReference>
+    <ItemGroup Condition="'$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'netstandard2.0'">
```
amckinney, line 511: Nice, this is way simpler - we don't lose any functionality here?

amckinney (approved):

generators/csharp/codegen/src/project/CsharpProject.ts
Swimburger: We're always building for all TFMs, so we'll always restore the package anyway.
Technically, the highest TFM doesn't require it so it may improve build speed, but insignificantly.

PR #5281, opened 2024-11-26
feat(cli): [WIP] Add support for OpenAPI input w/out fs

amckinney:
This starts to split out the abstractions required to parse the OpenAPI file directly as input rather than interacting with the user's local filesystem. The primary goal of this exploration is to make it possible to generate Fern's IR directly in the browser.

**Challenges**:
1. Far reaching usage of node and/or FS libraries (e.g. `path`, `fs`, `stream`, `os`, `util`, etc).
    * The transitive closure of the IR generator _and_ the dynamic snippets generator cannot reference any of these libraries.
        * This includes `@fern-api/go-ast`, `@fern-api/generator-commons`, etc, which already include heavy use.
    * If we don't incorporate any tree-shaking, _all_ packages referenced must omit these Node-specific libraries.
2. Update the `@fern-api/go-dynamic-snippets` to consume the latest dynamic IR.
   * The dynamic IR purposefully enables `noSerdeLayer` to reduce the size of the library for browser use.
   * This causes an incompatibility with the generated `fern-fern/ir-sdk` and `fern-api/dynamic-ir-sdk` types due to the custom [_type](https://github.com/fern-api/fern/blob/2d64e0ab81abc241333b5af824d44958d59babc0/packages/ir-sdk/fern/apis/ir-types-latest/definition/dynamic/types.yml#L23) discriminator in the dynamic IR.
   * In order for these to be consistent and have the monorepo in a compile-able state we have two options:
       1. Maintain a mapper between the types.
       2. Update the dynamic IR so that the results between `noSerdeLayer: {false, true}` are always consistent


amckinney:
This was implemented across a variety of other PRs.

PR #5285, opened 2024-11-26
chore(cli): Add @fern-api/path-utils

amckinney:
This adds the new `@fern-api/path-utils` package, which replaces a subset of `@fern-api/fs-utils` package. With this, packages that need to be browser-compatible can opt into a slimmed down version of the original package that contains no node-specific dependencies.

Note that we continue to export the `AbsoluteFilePath` and `RelativeFilePath` types from `@fern-api/fs-utils` so that old callers can still import the types from a single location. New packages can opt into `@fern-api/path-utils` as needed.

### DIFF
--- a/packages/commons/path-utils/src/AbsoluteFilePath.ts
+++ b/packages/commons/path-utils/src/AbsoluteFilePath.ts
@@ -9,7 +9,7 @@ export type AbsoluteFilePath = string & {
 
 export const AbsoluteFilePath = {
     of: (value: string): AbsoluteFilePath => {
-        if (!path.isAbsolutePath(value)) {
+        if (!path.isAbsolute(value)) {
             throw new Error("Filepath is not absolute: " + value);
         }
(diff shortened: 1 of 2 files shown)


packages/commons/fs-utils/src/AbsoluteFilePath.ts
```diff
@@ -1,15 +1,3 @@
...
-    of: (value: string): AbsoluteFilePath => {
-        if (!path.isAbsolute(value)) {
-            throw new Error("Filepath is not absolute: " + value);
-        }
-        return convertToOsPath(value) as AbsoluteFilePath;
-    }
-};
+// For convenience, we re-export the AbsoluteFilePath type for any caller
+// that requires fs-utils.
+export { AbsoluteFilePath } from "@fern-api/path-utils";
```
dsinghvi, line 3: Genius 🔥

dsinghvi (approved):

PR #5288, opened 2024-11-26
chore(cli): Add browser compatible base generator

amckinney:
This splits the `@fern-api/generator-commons` package into two new packages:

1. `@fern-api/base-generator`
2. `@fern-api/browser-compatible-base-generator`

Similar to the approach used in `@fern-api/path-utils`, the `@fern-api/base-generator` exports the types defined in `@fern-api/browser-compatible-base-generator` so that any callers only need to import a single dependency.

Any package that needs to be browser compatible will use `@fern-api/browser-compatible-base-generator` directly.

### DIFF
--- a/generators/browser-compatible-base/package.json
+++ b/generators/browser-compatible-base/package.json
@@ -18,8 +18,8 @@
   "scripts": {
     "clean": "rm -rf ./lib && tsc --build --clean",
     "compile": "tsc --build",
-    "test": "vitest --run",
-    "test:update": "vitest --run -u",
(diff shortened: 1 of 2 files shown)


generators/base/src/browserCompatibleExports.ts
```diff
@@ -0,0 +1,6 @@
+// Export everything from @fern-api/browser-compatible-base-generator
+// so that generators can simply consume @fern-api/base-generator.
```
dsinghvi, line 2: nice 🔥

dsinghvi (approved):

PR #5292, opened 2024-11-27
feat(cli): add new openapi-fdr command for direct openapi - fdr parsing

RohinBhargava:
## Description
Adds an optional flag, `--v2` for command `fdr` that writes out an fdr read shape directly from an OpenApi spec.

This is a VERY nascent implementation and will likely evolve to support multiple OAS in the future.

## Changes Made
- Adds `--v2` flag on command
- Adds `generateOpenApiToFdrApiDefinitionForWorkspaces` function
- Installs `@fern-api/docs-parsers` package to CLI

## Testing
- [ x ] Manual testing completed


packages/cli/cli/src/commands/generate-openapi-fdr/generateOpenApiToFdrApiDefinitionForWorkspaces.ts
```diff
@@ -0,0 +1,71 @@
...
+                await cliContext.runTaskForWorkspace(workspace, async (context) => {
+                    if (workspace instanceof LazyFernWorkspace) {
+                        context.logger.info("Skipping, API is specified as a Fern Definition.");
+                        return;
+                    } else if (!(workspace instanceof OSSWorkspace)) {
+                        return;
+                    }
+                    const openAPISpecs = await getAllOpenAPISpecs({ context, specs: workspace.specs });
+
+                    const openApi = openAPISpecs[0];
```
dsinghvi, line 36: can your parser handle taking more than one ? this is going to be important in most places

packages/cli/cli/src/cli.ts
```diff
@@ -618,6 +620,34 @@ function addFdrCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
     );
 }
 
+function addOpenAPIFdrCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
+    cli.command(
+        "openapi-fdr <path-to-output>",
```
dsinghvi, line 625: nit: lets call this `fdr v2`

dsinghvi: actually thoughts on just passing in a `--v2` command to `fern fdr fdr.json` and if that flag is passed in then it delegates to your library?

packages/cli/cli/src/commands/generate-openapi-fdr/generateOpenApiToFdrApiDefinitionForWorkspaces.ts
dsinghvi: if not, we should log an error if > 0 and track that work

RohinBhargava: The underlying referencing is what would need to support multiple inputs -- this boils down to a few functions in the parser OR exposing a document merger. I think we should discuss this.

The intention is to support multiple files -- I need to understand how the files can be merged together a bit before implementation. I'll leave a TODO, an error, and track the work!

dsinghvi (approved):

PR #5295, opened 2024-11-28
chore(cli): Add @fern-api/source-resolver

amckinney:
## Overview

This adds the new `@fern-api/source-resolver` and `@fern-api/cli-source-resolver` packages which are primarily used in the `@fern-api/ir-generator`.

## Rationale

The primary `SourceResolver` implementation requires access to the filesystem, so we abstract this interface away in a separate package, and require it as input to `generateIntermediateRepresentation`.

The implementation defined in `@fern-api/cli-source-resolver` is explicitly provided in every node-compatible call-site (e.g. the CLI), whereas browser-compatible callers can provide the `NopSourceResolver`, and can still receive valid IR output.

With this, the `@fern-api/ir-generator` no longer depends on `fs/promises`.

### DIFF
--- a/packages/cli/cli-source-resolver/tsconfig.json
+++ b/packages/cli/cli-source-resolver/tsconfig.json
@@ -19,6 +19,9 @@
       "path": "../fern-definition/schema"
     },
     {
+      "path": "../source-resolver"
+    },
+    {
       "path": "../task-context"
     },
     {


packages/cli/generation/ir-generator/src/generateIntermediateRepresentation.ts
```diff
@@ -54,31 +54,33 @@ import { generateEndpointExample } from "./examples/generator/generateSuccessEnd
...
-    includeOptionalRequestPropertyExamples,
     audiences,
     readme,
     packageName,
     version,
-    context
+    context,
+    fdrApiDefinitionId,
+    includeOptionalRequestPropertyExamples,
+    sourceResolver = new NopSourceResolver()
```
amckinney, line 69: @dsinghvi this will be fixed before merge - if everything else look good I'll make this a required property so that every caller needs to explicitly pass it in themselves. It's otherwise too easy for us to miss it.

amckinney: Changed to a required property in [this commit](https://github.com/fern-api/fern/pull/5295/commits/62c507e319827b52a1d5c37bbed57a54968e7452).

dsinghvi (approved):

amckinney:
@dsinghvi sounds great to me! I'll handle this in a follow up commit so that the other PRs branched off this one don't need to be rebased.

amckinney:
@dsinghvi follow-up PR here: https://github.com/fern-api/fern/pull/5312

PR #5311, opened 2024-11-30
feat(python): Comments, Star Imports, and Python File Improvements

noanflaherty:
## Description
This PR includes everything needed to generate more custom code at the top of a python file, including comments abd star import (e.g. `from .my_module import *`).

## Changes Made
- Introduced a new `Comment` AstNode, following patterns established by Go.
- Added new `topOfFileComments` arg to the existing `PythonFile` AstNode to support writing those comments at the top of a python file.
- Added new `statements` arg to `PythonFile` AstNode to support initializing a python file with known statements rather than forcing that all statements be added after-the-fact.
- Introduce a new `StarImport` AstNode that subclasses the existing `Reference` AstNode
- Added new `starImports` arg to `PythonFile` to support passing star imports upon initialization.

## Testing

### DIFF
--- a/generators/python-v2/ast/src/PythonFile.ts
+++ b/generators/python-v2/ast/src/PythonFile.ts
@@ -15,8 +15,8 @@ export declare namespace PythonFile {
         isInitFile?: boolean;
         /* Any comments that should be at the top of the file */
         comments?: Comment[];
-        /* Any star imports that should be included */
-        starImports?: StarImport[];
+        /* Any explicit imports that should be included */
+        imports?: StarImport[];
     }
 }
 
@@ -26,7 +26,7 @@ export class PythonFile extends AstNode {
     private readonly statements: AstNode[] = [];
     private readonly comments: Comment[];
 
-    constructor({ path, statements, isInitFile = false, comments, starImports }: PythonFile.Args) {
+    constructor({ path, statements, isInitFile = false, comments, imports }: PythonFile.Args) {
         super();
         this.path = path;
         this.isInitFile = isInitFile;
@@ -35,7 +35,7 @@ export class PythonFile extends AstNode {
 
         this.comments = comments ?? [];
 
-        starImports?.forEach((starImport) => this.addReference(starImport));
+        imports?.forEach((import_) => this.addReference(import_));
     }
 
     public addStatement(statement: AstNode): void {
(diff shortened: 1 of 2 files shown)


generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -1,26 +1,41 @@
...
 export declare namespace PythonFile {
     interface Args {
         /* The path of the Python file relative to the module */
         path: ModulePath;
+        /* The list of statements in the Python file. More can be added following initialization. */
+        statements?: AstNode[];
         /* Whether or not this represents the root of a Python module */
         isInitFile?: boolean;
+        /* Any comments that should be at the top of the file */
+        topOfFileComments?: Comment[];
```
amckinney, line 17: nit: I think `comments` is sufficient here since comments associated with a file are always written first.

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -1,26 +1,41 @@
...
+        /* Any star imports that should be included */
+        starImports?: StarImport[];
```
amckinney, line 19: nit: Thoughts on just naming the field `imports` even if it's typed as `StarImport`?

generators/python-v2/ast/src/StarImport.ts
```diff
@@ -0,0 +1,18 @@
...
+export class StarImport extends Reference {
+    public readonly docs: string | undefined;
+
+    public constructor({ modulePath }: StarImport.Args) {
+        super({ name: "*", modulePath });
+    }
+
+    public write(writer: Writer): void {
+        throw new Error("Not intended to be written outside the context of a PythonFile.");
+    }
```
amckinney, line 17: nit: Hmm.. not sure whether we should explicitly throw or just no-op here. If you'd rather throw for now to prevent buggy implementations, do you mind updating the message a bit so it's more clear this is coming from StarImport?

```suggestion
    public write(writer: Writer): void {
        throw new Error("Internal error; StarImport node cannot be written outside of PythonFile.");
    }
```

amckinney (approved):

generators/python-v2/ast/src/PythonFile.ts
noanflaherty: As a consumer of this package, this alternative naming might get me spooked and make me think that I’m supposed to explicitly include all imports, whereas most importing is handled for me via References.

I’m down for an alternative name though. Previously you had suggested “sideEffectImports” - wdyt about that?

generators/python-v2/ast/src/StarImport.ts
noanflaherty: I’m down to no-op

generators/python-v2/ast/src/PythonFile.ts
noanflaherty: I'll roll with `imports` and if we think of something better later, we can change it.

PR #5317, opened 2024-12-02
feat(cli): Add @fern-api/browser-compatible-fern-workspace

amckinney:
This adds the `@fern-api/browser-compatible-fern-workspace` package, which implements a new in-memory OpenAPI document workspace. With this, users can use the `@fern-api/ir-generator` in a browser compatible manner (i.e. with no filesystem access).

All of the common logic used in the `OSSWorkspace` are moved to the brower-compatible package so that the same configuration is shared between both implementations.

Note that this also refactors the `workspace` packages and consolidates a lot of previously duplicated logic and types (e.g. `Source`, `Spec`, etc). With this, we have the following `workspace` directory layout:

```sh
./packages/cli/workspace/
├── browser-compatible-fern-workspace
├── commons
├── lazy-fern-workspace
└── loader
```


dsinghvi (approved):
<discussed offline>

PR #5323, opened 2024-12-03
fix(typescript): support object type declarations with extraProperties

dsinghvi:
## Description
This PR updates the TypeScript generator to leverage `.extraProperties`. Types are now generated with `[key: string]: any` and corresponding zurg schemas have a `.passthrough()` method that is similar to [zod](https://github.com/colinhacks/zod?tab=readme-ov-file#passthrough). 

## Changes Made
- Add `passthrough` to zurg and corresponding tests. 
- Update ObjectGenerator to create a property with `[key: string]: any`

## Testing


amckinney (approved):
Nice, `passthrough` looks great 👏

PR #5332, opened 2024-12-04
fix(java): publishing to maven central requires more publishing metadata

dsinghvi:
## Description
This PR adds additional logic to the java generator to properly publish to maven central with a secret key, passphrase, and key id. If these are not specified in generators.yml then we publish to OSSRH instead which doesn't require these. 

## Changes Made
- Updated build.gradle generation to add a variety of custom blocks

## Testing


seed/java-sdk/exhaustive/signed_publish/build.gradle
```diff
@@ -3,6 +3,7 @@ plugins {
     id 'maven-publish'
```
dsinghvi, line 3: Changes to this file are the critical piece

amckinney (approved):

PR #5339, opened 2024-12-05
chore(cli): Add @fern-api/docs-formatter

amckinney:
This adds the new `@fern-api/docs-formatter` package which abstracts away the IR generator's usage of `prettier`. With this, the `@fern-api/ir-generator` removes another browser-incompatible dependency.


amckinney:
Closing this in favor of https://github.com/fern-api/fern/pull/5340 - the changes required to thread in an abstract DocsFormatter isn't worth it, and ends up complicating a variety of code paths. It's far simpler to just use a well-scoped conditional import.

PR #5347, opened 2024-12-05
fix(cli): Move swagger2openapi to workspace loader

amckinney:
Re: https://github.com/fern-api/fern/pull/5344

This moves the `swagger2openapi` dependency into the workspace loader so that the `@fern-api/openapi-parser` only ever acts upon OpenAPI v3. 

If we ever decide to implement our own OpenAPI v2 -> v3 converter, then it may make sense to move this conversion back into the `@fern-api/openapi-parser` layer.


packages/cli/api-importers/openapi/openapi-ir-parser/src/parse.ts
```diff
@@ -78,25 +77,14 @@ export async function parse({
...
-                    ir = merge(ir, openapiIr);
-                }
+                const openapiIr = generateIrFromV3({
+                    taskContext: context,
+                    openApi: document.value,
+                    options: getParseOptions({ specSettings: document.settings, overrides: options }),
+                    source,
+                    namespace: document.namespace
+                });
+                ir = merge(ir, openapiIr);
```
dsinghvi, line 87: nice!!

dsinghvi (approved):

PR #5350, opened 2024-12-05
feat(typescript): Support inline types in TypeScript generator

fern-support:
Add support for generating inline everything


generators/typescript/model/type-reference-converters/src/AbstractTypeReferenceToTypeNodeConverter.ts
```diff
@@ -41,6 +71,60 @@ export abstract class AbstractTypeReferenceToTypeNodeConverter extends AbstractT
...
+        propertyName,
+        genericIn
+    }: ConvertTypeReferenceParams.InlinePropertyTypeParams): ts.TypeNode {
+        let name = ts.factory.createQualifiedName(
+            ts.factory.createIdentifier(parentTypeName),
+            ts.factory.createIdentifier(propertyName)
+        );
+        switch (genericIn) {
+            case "list":
+                name = ts.factory.createQualifiedName(name, "Item");
```
amckinney, line 111: Let's refer to that same exported constants here - this might mean we place those constants elsewhere so it's easier to import from a shared location.

generators/typescript/sdk/endpoint-error-union-generator/src/GeneratedEndpointErrorUnionImpl.ts
```diff
@@ -73,7 +76,9 @@ export class GeneratedEndpointErrorUnionImpl implements GeneratedEndpointErrorUn
             includeOtherInUnionTypes: true,
             includeSerdeLayer,
             noOptionalProperties,
-            retainOriginalCasing
+            retainOriginalCasing,
+            inlineInlineTypes,
+            inline: false
```
amckinney, line 81: Is this intentional? If so, a lightweight comment justifying why we always use `false` here might be useful.

generators/typescript/sdk/generator/src/SdkGenerator.ts
```diff
@@ -540,6 +545,17 @@ export class SdkGenerator {
               });
     }
 
+    private getTypesToGenerate(): Record<TypeId, TypeDeclaration> {
+        if (this.config.inlineInlineTypes) {
+            return Object.fromEntries(
+                Object.entries(this.intermediateRepresentation.types).filter(
+                    ([_, typeDeclaration]) => typeDeclaration.inline !== true
```
amckinney, line 552: nit: 
```suggestion
                    ([_, typeDeclaration]) => !typeDeclaration.inline
```

generators/typescript/sdk/generator/src/contexts/type-schema/TypeSchemaContextImpl.ts
```diff
@@ -154,8 +164,12 @@ export class TypeSchemaContextImpl implements TypeSchemaContext {
         });
     }
 
+    private generateForInlineUnion(typeName: DeclaredTypeName): ts.TypeNode {
+        throw new Error("Inline unions are not supported in schemas.");
```
amckinney, line 168: These errors would only ever occur from a generator author issue, right? It will be impossible for some customer to hit this on their own? If so, let's preface this with `Internal error` like so:

```suggestion
        throw new Error("Internal error; inline unions are not supported in schemas.");
```

This might be relevant for other error messages, too.

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
```diff
@@ -198,6 +225,138 @@ export class GeneratedRequestWrapperImpl implements GeneratedRequestWrapper {
...
+        return {
+            kind: StructureKind.Module,
+            name: inlinedRequestBody.name.pascalCase.safeName,
+            isExported: true,
+            hasDeclareKeyword: false,
+            declarationKind: ModuleDeclarationKind.Namespace,
+            statements: Array.from(inlineProperties.entries()).flatMap(
+                ([objectProperty, typeDeclaration]: [ObjectProperty, TypeDeclaration]) => {
+                    const typeName = objectProperty.name.name.pascalCase.safeName;
+                    const listOrSetStatementGenerator = () => {
```
amckinney, line 259: Similar to above w.r.t. closures. But I think all of this will be resolved based on your suggestion to split this out into a shared component.

packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertAdditionalProperties.ts
```diff
@@ -104,20 +105,68 @@ export function convertAdditionalProperties({
...
+        case "reference":
+            break;
+        case "unknown":
+            break;
+        default:
+            assertNever(schema);
+    }
+
+    return schema;
+}
```
amckinney, line 168: nit: You can group these cases together for slightly better readability. We also don't need the final `return` if we use `assertNever` like you already are.

```suggestion
function addInline(schema: SchemaWithExample): SchemaWithExample {
    switch (schema.type) {
        case "array":
        case "enum":
        case "map":
        case "object":
        case "oneof":
            schema.inline = true;
            break;
        case "nullable":
        case "optional":
            schema.inline = true;
            schema.value = addInline(schema.value);
            break;
        case "literal":
        case "primitive":
        case "reference":
        case "unknown":
            break;
        default:
            assertNever(schema);
    }
}
```

amckinney (commented):
Nice! Mainly just nits, but this is looking good!

.eslintrc.js
```diff
@@ -67,12 +67,7 @@ module.exports = {
                 ignoreRestSiblings: true
             }
         ],
-        "@typescript-eslint/no-namespace": [
-            "error",
-            {
-                allowDeclarations: true
-            }
-        ],
+        "@typescript-eslint/no-namespace": "off",
```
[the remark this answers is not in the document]
Swimburger: I did mean to do this, it allows you to declare const enums in a namespace:
```ts
export namespace ConvertTypeReferenceParams {
    ...
    export const GenericIn = {
        List: "list",
        Map: "map",
        Set: "set"
    } as const;
    export type GenericIn = typeof GenericIn[keyof typeof GenericIn];
}
```

packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertAdditionalProperties.ts
Swimburger: I think eslint didn't let me do this, but I'll try again

amckinney (approved):
Nice, the refactors look great!

Issue #5343, opened 2024-12-05
Validate Swagger/OpenAPI response when running `fern init --openapi {url}`


thechori:
### Feature Description

Add a simple validation check to prevent users from accidentally running the `fern init` command against the Swagger HTML endpoint and inform them that they need to point it at an endpoint that returns pure JSON or Y(A)ML.

I was guilty of personally making this mistake on two occasions. Would be a nice bump in DX.

```bash
# returns JSON - good ✅
$ fern init --openapi http://localhost:3000/api-json # returns JSON
```

```
# returns HTML - bad ❌
$ fern init --openapi http://localhost:3000/api 
```

### Proposed Generated Code (Optional)

Simple approach:
```ts
const response = await axios.get(url);
const contentType = response.headers['content-type'];

if (contentType.includes('text/html')) {
  throw Error("Please point to a URL that returns JSON or YML")
} 

// continue...
```

More robust:
```ts
const response = await axios.get(url);
const contentType = response.headers['content-type'];

const validContentTypes = ['application/json', 'application/x-yaml', 'text/yaml'];

if (!validContentTypes.some(type => contentType.includes(type))) {
  throw new Error("Please point to a URL that returns one of the supported content types: ", validContentTypes.join(", "));
}

// continue...
```

thechori:
@dannysheridan

PR #5352, opened 2024-12-06
fix(cli): add logic to validate and properly handle url response when running `fern init --openapi {url}`

thechori:
## Description
Fixes bug that caused the `fern init --openapi {url}` to generate invalid yml code against NestJS/Swagger's yml endpoint ([see original issue here](https://github.com/fern-api/fern/issues/5342))

Before, the `loadOpenAPIFromUrl` function assumed that the API endpoint url provided was always returning JSON which led to issues downstream when trying to convert this code to YML. This led to issues such as a leading "|" char in the output yml which resulted from running a yml transformation on yml.

## Changes Made
- Check content-type of API response to ensure it needs manipulation (yml-based endpoints are good by definition, json endpoints need to be parsed, html throws error)

## Testing

## Notes
This code also satisfies the needs of [this feature request](https://github.com/fern-api/fern/issues/5343)

:v:

### DIFF
--- a/packages/cli/init/src/utils/loadOpenApiFromUrl.ts
+++ b/packages/cli/init/src/utils/loadOpenApiFromUrl.ts
@@ -55,5 +55,5 @@ async function fetchOpenAPIFromUrl({ url, logger }: { url: string; logger: Logge
     logger.warn(
         `Unrecognized Content-Type "${contentType}" from endpoint ${url}. Please ensure you're pointing to a URL that returns JSON or YAML and not HTML (e.g. Swagger UI webpage)`
     );
-    return response.data;
+    return dump(response.data);
 }


thechori:
@dannysheridan here we go!! 🔥

packages/cli/init/src/utils/loadOpenApiFromUrl.ts
```diff
@@ -24,8 +24,23 @@ export interface FailedLoadOpenAPI {
...
+
+        if (contentType.includes("json")) {
+            yamlData = dump(response.data);
+        } else if (contentType.includes("yaml")) {
+            yamlData = response.data;
+        } else {
+            throw new Error(
+                "Unsupported Content-Type from endpoint. Please ensure you're pointing to a URL that returns JSON or Y(A)ML and not HTML (e.g., Swagger UI webpage)"
+            );
+        }
```
amckinney, line 42: Nice! Mind putting this into a helper function so that it cleans up the control flow a bit? Plus, we might also want to be more permissive and start with a debug log instead of an error, just in case the Content-Type is empty for YAML content. In combination, I'm thinking of something like this:

```typescript
export async function loadOpenAPIFromUrl({ url, logger }: { url: string; logger: Logger }): Promise<LoadOpenAPIResult> {
    try {
        const yamlData = await fetchOpenAPIFromUrl({ url, logger });
        const tmpDir = await tmp.dir();
        const filePath = join(tmpDir.path, "openapi.yml");
        logger.debug("tmpDir", tmpDir.path);
        logger.debug("filePath", filePath);
        await writeFile(filePath, yamlData);
        return {
            status: LoadOpenAPIStatus.Success,
            filePath
        };
    } catch (error) {
        logger.debug(`Encountered an error while loading OpenAPI spec: ${JSON.stringify(error)}`);
        const errorMessage = `Failed to load OpenAPI spec from ${url}`;
        return {
            status: LoadOpenAPIStatus.Failure,
            errorMessage
        };
    }
}

async function fetchOpenAPIFromUrl({ url, logger }: { url: string; logger: Logger }): Promise<string> {
    const response = await axios.get(url);
    const contentType = response.headers["content-type"] ?? "";
    if (contentType.includes("json")) {
        return dump(response.data);
    }
    if (contentType.includes("yaml")) {
        return response.data;
    }
    logger.debug(`Unrecognized Content-Type "${contentType}" from endpoint. Please ensure you're pointing to a URL that returns JSON or YAML and not HTML (e.g. Swagger UI webpage)`);
    return response.data;
}
```

thechori: Sounds good! I'm on it now 👍

thechori: @amckinney I took your advice and implemented the suggestions! 

I also converted the `logger.debug` to a `logger.warn` so that users could see this warning message on their first run without having to try again with `--log-level debug`. Figured this could make for slightly better DX. If you'd prefer I swap this out for the original debug message, just let me know! ✌️

packages/cli/init/src/utils/loadOpenApiFromUrl.ts
```diff
@@ -44,3 +42,18 @@ export async function loadOpenAPIFromUrl({ url, logger }: { url: string; logger:
...
+    if (contentType.includes("json")) {
+        return dump(response.data);
+    }
+    if (contentType.includes("yaml")) {
+        return response.data;
+    }
+    logger.warn(
+        `Unrecognized Content-Type "${contentType}" from endpoint ${url}. Please ensure you're pointing to a URL that returns JSON or YAML and not HTML (e.g. Swagger UI webpage)`
+    );
+    return response.data;
```
amckinney, line 58: Accidentally missed this from my last review, but do you mind preserving the original behavior in the default case? We can assume we've received JSON and dump it to YAML like so:

```suggestion
    return dump(response.data);
```

amckinney (approved):

dsinghvi (approved):

PR #5363, opened 2024-12-08
feat(cli): @fern-api/ir-generator is sync

amckinney:
This updates the `@fern-api/ir-generator` package completely synchronous. This involved the following changes:

1. Split out the `fern check` rules focused on `generators.yml` into `@fern-api/generators-validator`.
2. Remove `prettier` from the IR generator altogether, which was previously used to format documentation on the fly.
3. Update the Fern definition visitor abstractions to by synchronous.
    * None of the methods actually required them to be `async`.

With this, the `@fern-api/snippets-core` library is entirely synchronous, too.

### DIFF
--- a/seed/csharp-model/alias-extends/src/SeedAliasExtends.sln
+++ b/seed/csharp-model/alias-extends/src/SeedAliasExtends.sln
@@ -3,9 +3,9 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.0.31903.59
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SeedAliasExtends", "SeedAliasExtends\SeedAliasExtends.csproj", "{C0E8F871-A099-4DBF-A5A5-BAFE2FD31D93}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SeedAliasExtends", "SeedAliasExtends\SeedAliasExtends.csproj", "{A375DF48-6767-4394-8139-0AC6831A51DB}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SeedAliasExtends.Test", "SeedAliasExtends.Test\SeedAliasExtends.Test.csproj", "{D8363324-1CC0-47B0-9510-F11A51372297}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SeedAliasExtends.Test", "SeedAliasExtends.Test\SeedAliasExtends.Test.csproj", "{B2BF2007-0311-43F3-9608-4D9CDFF2F88B}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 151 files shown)


packages/cli/generation/ir-generator/src/formatDocs.ts
```diff
@@ -1,13 +1,16 @@
...
-    const prettier = await import("prettier");
-    const formattedDocs = prettier.format(docs, {
-        parser: "markdown"
-    });
+    // TODO: Find a way to support this in the browser without an async conditional import.
+    //
+    // const prettier = import("prettier");
+    // const formattedDocs = prettier.format(docs, {
+    //     parser: "markdown"
+    // });
```
dsinghvi, line 12: can you have the caller pass in a function to `generateIntermediateRepresentation` here ?

dsinghvi: then we can only pass it in for sdk generation but leave it out for snippets

amckinney: I took a look at that earlier, but it still requires calling an `async` function (even when it's abstracted away). I experimented with moving the docs formatting up a layer - we can handle this in the workspace before its passed onto the IR generator so that the IR generator can remain totally sync.

I explored doing this in the OSSWorkspace and LazyFernWorkspace [here](https://github.com/fern-api/fern/pull/5363/commits/b318ba849dae417a1a1c18b3574c247d116aca29), but if this looks good I think we could  move it to the `AbstractAPIWorkspace` as a protected method so that every `async` variant has access to the same implementation in a single place.

Let me know what you think!

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/suger.json
```diff
@@ -2372,12 +2372,12 @@ in the google.rpc.Status.details field, or localized by the client.",
...
               "displayUnitDescription": "optional<string>",
               "id": {
-                "docs": "The usage metering dimension key, in format of "{plan_key}_{usage_dimension_key}"",
+                "docs": "The usage metering dimension key, in format of "{plan*key}*{usage_dimension_key}"",
                 "type": "optional<string>",
               },
               "metricKind": "optional<string>",
               "name": {
-                "docs": "Name: The resource name of the metric descriptor, in format of "{service_name}/{plan_key}_{usage_dimension_key}"",
+                "docs": "Name: The resource name of the metric descriptor, in format of "{service*name}/{plan_key}*{usage_dimension_key}"",
```
amckinney, line 2380: Running this on more test definitions illustrates that we might not actually want to run `prettier` in general. 

It seems like the main thing we want to do is clean up leading/trailing whitespace, but even that might best be shipped in a separate formatter tool that users can independently control. Otherwise we might end up with surprising behavior like this.

amckinney: I removed `prettier` altogether in [this commit](https://github.com/fern-api/fern/pull/5363/commits/7f04efccd2998cf1a438bcb482c12b98b8463132) to explore the results a bit more.

packages/cli/ete-tests/src/tests/validate/__snapshots__/validate.test.ts.snap
```diff
@@ -13,7 +13,8 @@ exports[`validate > no-api 1`] = `
 exports[`validate > no-generator 1`] = `"Misconfigured fern directory: please see the docs at https://buildwithfern.com/learn/api-definition/introduction/what-is-the-fern-folder"`;
 
 exports[`validate > simple 1`] = `
-"[api]: Found 2 errors and 0 warnings. Run fern check --warnings to print out the warnings.
+"[api]: ✓ All checks passed
```
amckinney, line 16: We'll need to update this - ideally the prefix is `[generators]` now that we have a separate `@fern-api/generators-validator`, but we'll need to make more changes to customize the prefix [here](https://github.com/fern-api/fern/blob/d93860839ab4c4995cb2cf5857ebb3d94d387c69/packages/cli/cli/src/cli-context/CliContext.ts#L208).

amckinney: Fixed in [this commit](https://github.com/fern-api/fern/pull/5363/commits/33d07c895d690956556472bcd45fb391720841d8).

dsinghvi (approved):

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/suger.json
amckinney: Discussed offline - we'll remove `prettier` altogether for these reasons.

packages/cli/generation/ir-generator/src/formatDocs.ts
amckinney: Resolved in [this comment](https://github.com/fern-api/fern/pull/5363#discussion_r1876188319).

PR #5365, opened 2024-12-08
fix(python): WIP – Fix Imports w/ Same Name

noanflaherty:
## Description
This PR showcases via a test some erroneous behavior when:
1. There are two imported references with the same name; and
2. There is a class/function/variable that's given the same name as that of an import

### DIFF
--- a/generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
+++ b/generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
@@ -112,6 +112,17 @@ my_variable: ImportedClass.nested.attribute
 "
 `;
 
+exports[`PythonFile > Write duplicative import names 1`] = `
+"from .cars import Car
+from .vehicles import Car
+
+
+class Car:
+    car = Car()
+    vehicle = Car()
+"
+`;
+
 exports[`PythonFile > Write star imports 1`] = `
 "# flake8: noqa: F401, F403
 
--- a/generators/python-v2/ast/src/__test__/PythonFile.test.ts
+++ b/generators/python-v2/ast/src/__test__/PythonFile.test.ts
@@ -316,4 +316,44 @@ describe("PythonFile", () => {
         expect(await writer.toStringFormatted()).toMatchSnapshot();
         expect(file.getReferences()).toHaveLength(2);
     });
+
+    it("Write duplicative import names", async () => {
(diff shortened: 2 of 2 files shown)


generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
```diff
@@ -112,6 +112,17 @@ my_variable: ImportedClass.nested.attribute
 "
 `;
 
+exports[`PythonFile > Write duplicative import names 1`] = `
+"from .cars import Car
+from .vehicles import Car
```
noanflaherty, line 117: I have ideas for how to fix this – it seems the other language ast modules handle this case by prepending `_` to an alias name. 

So, this would become 
```
from .cars import Car
from .vehicles import Car as _Car
```

Is that right? If so, I can explore solutions that produce this end result

generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
```diff
@@ -112,6 +112,17 @@ my_variable: ImportedClass.nested.attribute
...
+
+
+class Car:
+    car = Car()
+    vehicle = Car()
```
noanflaherty, line 122: This seems like the harder case. Here, we create a class called `Car`, which has a naming conflict with the imports above. Do other language ast modules handle this case? Is it generally left up to the consumer to avoid? One could argue the desired generated code should look like:

```python
from .cars import Car as _Car
from .vehicles import Car as __Car

class Car:
    car = _Car()
    vehicle = __Car()
```

dsinghvi: Yup the AST should allow a consumer to pass in any overrides to the alias (i.e. Fern has important info that allows us to alias in a nicer way)

amckinney: > Should the AST be smart enough to see that lower down in the file, we have a class named Car, therefore it should automatically produce an alias for all imports of other objects named Car

This one - the AST should know how to guarantee that it can write itself to a file, and do so in a deterministic way.

noanflaherty:
Closing now that we've aligned on the desired end result. I'll open a fresh PR once there's some of this implemented.

PR #5369, opened 2024-12-09
feat(openapi): Add OpenAPI endpoint filter

amckinney:
This adds a new `filter` setting in the `generators.yml`, which allows users to filter the OpenAPI document down to a subset of endpoints.

For example,

```yaml
# generators.yml
api:
  specs:
    - openapi: ../openapi.yml
      settings:
        filter:
          endpoints:
            - POST /companies
        only-include-referenced-schemas: true
```

For now we only allow _including_ `endpoints`, but we could easily roll out the `not` or `excludeEndpoints` variant later as needed.


dsinghvi (approved):
Does the filter work for specific methods as well `POST /companies` ?

amckinney:
> Does the filter work for specific methods as well `POST /companies` ?

No, but it should - I might want to refactor this as an `endpoint` filter so that it assumes the form like this:

```yaml
# generators.yml
api:
  specs:
    - openapi: ../openapi.yml
      settings:
        filter:
          endpoints:
            - POST /companies
        only-include-referenced-schemas: true
```

PR #5372, opened 2024-12-09
chore(docs): linked to most recent versions in publish-your-sdk.mdx

cooperbosch:
## Description
Added text to the publish-your-sdk.mdx that links to fern-api README where most up to date version table is accessible

## Testing
Will look at site preview, unsure if that information will always be stored there


cooperbosch:
This change doesn't appear properly in preview, looking into it now

PR #5377, opened 2024-12-10
fix(python): Handle Duplicative Import Names

noanflaherty:
## Description
Sibling PR: https://github.com/fern-api/fern/pull/5365

The goal of this PR is to handle auto-aliasing of imports when there are naming conflicts.

## Changes Made
- Track ref name overrides on the `Writer`
- Build up the map of overrides as part of `PythonFile`'s `write` method

## Testing

### DIFF
--- a/generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
+++ b/generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
@@ -114,12 +114,14 @@ my_variable: ImportedClass.nested.attribute
 
 exports[`PythonFile > Write duplicative import names 1`] = `
 "from .cars import Car as CarsCar
+from .transportation.vehicles import Car as VehiclesCar
 from .automobiles import Car as AutomobilesCar
 from .vehicles.automobiles import Car as VehiclesAutomobilesCar
 
 
 class Car:
     car = CarsCar()
+    car = VehiclesCar()
     automobile = AutomobilesCar()
     vehicle = VehiclesAutomobilesCar()
 "
--- a/generators/python-v2/ast/src/__test__/PythonFile.test.ts
+++ b/generators/python-v2/ast/src/__test__/PythonFile.test.ts
@@ -339,6 +339,19 @@ describe("PythonFile", () => {
 
         local_class.add(
             python.field({
+                name: "car",
+                initializer: python.instantiateClass({
+                    classReference: python.reference({
+                        modulePath: ["root", "transportation", "vehicles"],
+                        name: "Car"
+                    }),
+                    arguments_: []
+                })
+            })
+        );
+
+        local_class.add(
+            python.field({
                 name: "automobile",
                 initializer: python.instantiateClass({
                     classReference: python.reference({
@@ -367,6 +380,6 @@ describe("PythonFile", () => {
 
         file.write(writer);
         expect(await writer.toStringFormatted()).toMatchSnapshot();
-        expect(file.getReferences()).toHaveLength(3);
+        expect(file.getReferences()).toHaveLength(4);
     });
 });


generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -44,47 +48,99 @@ export class PythonFile extends AstNode {
     }
 
     public write(writer: Writer): void {
+        const uniqueReferences = this.deduplicateReferences();
+
+        this.updateWriterRefNameOverrides({ writer, uniqueReferences });
```
noanflaherty, line 53: This method is where the magic happens. Within it, we add state to `Writer`

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -44,47 +48,99 @@ export class PythonFile extends AstNode {
...
+        this.writeImports({ writer, uniqueReferences });
         this.statements.forEach((statement, idx) => {
             statement.write(writer);
             writer.newLine();
             if (idx < this.statements.length - 1) {
                 writer.newLine();
             }
         });
+
+        writer.unsetRefNameOverrides();
```
noanflaherty, line 65: Wipe the state we had set on the writer just to be safe.

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -44,47 +48,99 @@ export class PythonFile extends AstNode {
...
+            } else if (statement instanceof Field) {
+                reservedNames.add(statement.name);
+            }
+        });
+
+        return reservedNames;
     }
 
-    private writeImports(writer: Writer): void {
+    private deduplicateReferences() {
```
noanflaherty, line 133: I hope to get rid of the need for this method altogether later. Right now, the data structure we use allows for duplicative refs, but there's no reason it needs to. We can deduplicate upon adding a new ref.

generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
```diff
@@ -112,6 +112,17 @@ my_variable: ImportedClass.nested.attribute
 "
 `;
 
+exports[`PythonFile > Write duplicative import names 1`] = `
+"from .cars import Car as Car_1
+from .vehicles import Car as Car_2
```
noanflaherty, line 117: I chose a `_n` suffix, but am open to suggestions here.

generators/python-v2/ast/src/Reference.ts
```diff
@@ -43,7 +43,8 @@ export class Reference extends AstNode {
     }
 
     public write(writer: Writer): void {
-        writer.write(this.alias ?? this.name);
+        const nameOverride = writer.getRefNameOverride(this);
+        writer.write(nameOverride.name);
```
noanflaherty, line 47: Notably a `Reference` object now gets its name from the `Writer` during its own `write` method.

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -44,47 +48,99 @@ export class PythonFile extends AstNode {
     }
 
     public write(writer: Writer): void {
+        const uniqueReferences = this.deduplicateReferences();
```
noanflaherty, line 51: I hope to get rid of the need for this method altogether later. Right now, the data structure we use for ref tracking allows for duplicative refs, but there's no reason it needs to. We can deduplicate upon adding a new ref.

In any case, this is just a refactor to take existing logic that was there and move it into a helper so its return value can be used in two places directly below.

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -99,6 +155,35 @@ export class PythonFile extends AstNode {
...
+            comment.write(writer);
+        });
+
+        if (this.comments.length > 0) {
+            writer.newLine();
+        }
+    }
+
+    private getImportName({ writer, reference }: { writer: Writer; reference: Reference }): string {
+        const nameOverride = writer.getRefNameOverride(reference);
```
noanflaherty, line 172: Here is one of two places where we use the state that was added to the `Writer`.

generators/python-v2/ast/src/core/Writer.ts
```diff
@@ -1,9 +1,32 @@
 import { AbstractWriter } from "@fern-api/base-generator";
 import { Config } from "@wasm-fmt/ruff_fmt";
+import { Reference } from "../Reference";
 
 export declare namespace Writer {}
 
 export class Writer extends AbstractWriter {
+    private completeRefPathsToNameOverrides: Record<string, { name: string; isAlias: boolean }> = {};
```
noanflaherty, line 8: This shows the state that's now tracked on `Writer`s and helps for interacting with it.

generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
noanflaherty: @dvargas92495 lmk if you have a preferred naming convention here.

dvargas92495: usually do something module dependent. eg

```python
from .cars import Car as CarsCar
from .vehicles import Car as VehiclesCar
```

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -44,47 +48,99 @@ export class PythonFile extends AstNode {
...
+    }: {
+        writer: Writer;
+        uniqueReferences: Map<string, { modulePath: ModulePath; references: Reference[]; referenceNames: Set<string> }>;
+    }): void {
+        const reservedNames = this.getReservedNames();
+        const references: Reference[] = Array.from(uniqueReferences.values()).flatMap(({ references }) => references);
+
+        // Build up a map of refs to their name overrides, keeping track of howmany times we've seen a name as we go.
+        const completeRefPathsToNameOverrides: Record<string, { name: string; isAlias: boolean }> = {};
+        const nameUsageCounts: Record<string, number> = {};
```
amckinney, line 84: Do we actually need to track the counts? We only care if it's used at all right? Could we just use `reservedNames` instead?

generators/python-v2/ast/src/Reference.ts
```diff
@@ -70,4 +71,8 @@ export class Reference extends AstNode {
     public getFullyQualifiedModulePath(): string {
         return this.modulePath.join(".");
     }
+
+    public getCompletePath(): string {
+        return `${this.getFullyQualifiedModulePath()}.${this.name}`;
+    }
```
amckinney, line 77: nit: Thoughts `getFullyQualifiedPath`?

generators/python-v2/ast/src/core/Writer.ts
amckinney: Let's define an `interface` for the value and name it. Something like:

```typescript
interface ImportedName {
  name: string,
  isAlias?: boolean;
}

...

private fullyQualifiedPathsToImportedNames = Record<string, ImportedName> = {};
```

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -44,47 +48,99 @@ export class PythonFile extends AstNode {
...
+                reservedNames.add(statement.name);
+            } else if (statement instanceof Method) {
+                reservedNames.add(statement.name);
+            } else if (statement instanceof Field) {
+                reservedNames.add(statement.name);
+            }
+        });
+
+        return reservedNames;
     }
```
amckinney, line 131: I'd update the name here - when I think of reserved names I think of keywords like `def`, `__init__`, etc. Maybe just `getUsedNames`?

noanflaherty: Yep good call 👍

generators/python-v2/ast/src/Reference.ts
noanflaherty: This helper had existed, but yep happy to change as part of this PR.

generators/python-v2/ast/src/PythonFile.ts
noanflaherty: We do for the current naming convention scheme, but once I adapt to match Vargas's proposal, we won't need to.

generators/python-v2/ast/src/core/Writer.ts
noanflaherty: Will do

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -44,47 +49,98 @@ export class PythonFile extends AstNode {
...
 
-    private writeComments(writer: Writer): void {
-        this.comments.forEach((comment) => {
-            comment.write(writer);
+    private updateWriterRefNameOverrides({
+        writer,
+        uniqueReferences
+    }: {
+        writer: Writer;
+        uniqueReferences: Map<string, { modulePath: ModulePath; references: Reference[]; referenceNames: Set<string> }>;
```
amckinney, line 78: nit: Should we define an interface for this map value?

generators/python-v2/ast/src/Reference.ts
```diff
@@ -67,7 +68,11 @@ export class Reference extends AstNode {
...
     }
 
-    public getFullyQualifiedModulePath(): string {
+    public getFullyQualifiedPath(): string {
         return this.modulePath.join(".");
     }
+
+    public getCompletePath(): string {
+        return `${this.getFullyQualifiedPath()}.${this.name}`;
+    }
```
amckinney, line 77: nit: I would use `getFullyQualifiedModulePath` and `getFullyQualifiedPath` (since omitting the `Module` component implies its for the reference), but feel free to ignore if you disagree. Not a big deal either way, and it's easy to change later (if ever).

generators/python-v2/ast/src/__test__/__snapshots__/PythonFile.test.ts.snap
```diff
@@ -112,6 +112,19 @@ my_variable: ImportedClass.nested.attribute
...
+exports[`PythonFile > Write duplicative import names 1`] = `
+"from .cars import Car as CarsCar
+from .automobiles import Car as AutomobilesCar
+from .vehicles.automobiles import Car as VehiclesAutomobilesCar
+
+
+class Car:
+    car = CarsCar()
+    automobile = AutomobilesCar()
+    vehicle = VehiclesAutomobilesCar()
```
amckinney, line 124: 👏

generators/python-v2/ast/src/core/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+            }
+            // Preserve words that are all uppercase and longer than one character
+            if (word.length > 1 && word === word.toUpperCase() && !/^\d+$/.test(word)) {
+                return word;
+            }
+            // Capitalize first letter, lowercase rest
+            return word.charAt(0).toUpperCase() + word.slice(1).toLowerCase();
+        })
+        .join("");
+}
```
amckinney, line 48: There's a lot of custom regex in here - could we have used some combination of [lodash](https://github.com/lodash/lodash) functions for the same effect? This might be as simple as just calling `upperFirst(camelCase(name))` to generate PascalCase like [this](https://github.com/fern-api/fern/blob/f5529b192d991f3c3a6ec2f0458e409ad5a968b3/packages/cli/generation/ir-generator/src/casings/CasingsGenerator.ts#L38C34-L38C60) (plus potentially some other minor cleanups).

Would you mind looking into this a bit to see if we can remove this altogether?

generators/python-v2/ast/src/PythonFile.ts
```diff
@@ -44,47 +49,98 @@ export class PythonFile extends AstNode {
...
+            let isAlias = !!reference.alias;
+
+            while (usedNames.has(nameOverride)) {
+                isAlias = true;
+
+                const module = reference.modulePath[modulePathIdx];
+                if (modulePathIdx < 0 || !module) {
+                    nameOverride = `_${nameOverride}`;
+                } else {
+                    nameOverride = `${createPythonClassName(module)}${nameOverride}`;
```
amckinney, line 101: Note that if we don't use a simple `_` strategy, the generated names can potentially get _long_ unless you maintain a collision map and assign type names more carefully.

For example, imagine the following module name (which we actually sometimes see in practice):

```python
from .resources.v1.api.vehicles.automobiles import Car as ResourcesV1ApiVehiclesAutomobilesCar
class Car:
    vehicle = ResourcesV1ApiVehiclesAutomobilesCar()
```

For what it's worth, we do this in the original Go generator but only prepend names as needed (until we find a unique name). Given that we traverse the statements in a deterministic order, the import alias assignment is always guaranteed to be unique. You can check out the method docs [here](https://github.com/fern-api/fern/blob/f5529b192d991f3c3a6ec2f0458e409ad5a968b3/generators/go/internal/gospec/scope.go#L60), and [this](https://github.com/fern-api/fern/blob/f5529b192d991f3c3a6ec2f0458e409ad5a968b3/generators/go/internal/gospec/scope_test.go#L17) test case.

amckinney (approved):
Stamping to unblock, but there's a few things I think we can improve here. I trust you to consider them before merging, but I'll leave you to making a judgement call for your use case.

generators/python-v2/ast/src/PythonFile.ts
noanflaherty: Perhaps I'm misunderstanding, but I assert the current implementation _does_ only prepend names as needed until it comes up with one that's unique. It goes from right to left, adding one module to the name at a time, until it finds that it's come up with a unique name.

noanflaherty: Updating the test to make this clear makes sense. Will do!

PR #5385, opened 2024-12-11
fix(java): Don't generate pagination with nonempty path

ajgateno:
Skips generating pagination logic where the parameter path is nonempty, since we don't support object nesting in such requests in Java for the time being.


dsinghvi (approved):
You'll want to update `java/sdk/versions.yml` for the release

ajgateno:
This fixed https://github.com/fern-api/fern/pull/5379

PR #5388, opened 2024-12-11
feat(cli): support bytes in the fern definition

dsinghvi:
## Description
This PR supports responses that are bytes. The motivation here is to not 1) force users to specify `file` when they actually return bytes and 2) allow the generators to generate more idiomatic code like returning an `ArrayBuffer`. 

## Changes Made
- Introduced a new union variant for `HttpResponseBody` as part of IR V54
- Added a reverse migration (i casted to `any` bc union visiting is annoying and we know the types will align in a specific case)

## Testing

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,5 +1,18 @@
 - changelogEntry:
     - summary: |
+        The Fern Definition now supports `bytes` as a response type.
+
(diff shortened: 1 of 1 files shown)


chdeskur (approved):

chdeskur:
docs look good to me

amckinney (approved):

PR #5400, opened 2024-12-12
chore(internal): Replace yarn with pnpm

Swimburger:
## Description
We use pnpm at Fern, but still have some straggling usage of yarn.
This PR removes usage of Yarn.

I also removed @yarnpkg/esbuild-plugin-pnp as I can't find any reference to us using it.
Maybe it's required for a reason I don't know about.

## Changes Made

- Replace `yarn` with `pnpm`
- Remove @yarnpkg/esbuild-plugin-pnp dependency

## Testing

  - [x] Ran ts-sdk and ts-express seed


dsinghvi (approved):

amckinney (dismissed):
🚀

dsinghvi (changes requested):
chatted offline (about potentially scoping changes to our use of typescript) but not the typescript generator

PR #5402, opened 2024-12-12
feat(typescript): support `bytes` responses 

dsinghvi:
## Description
This PR supports `bytes` responses in the TypeScript SDK. A `bytes` response is returned as an `ArrayBuffer`. 

## Changes Made
- Upgrade to IR `v54`
- Handle `HttpResponseBody.Bytes`
- Introduce a new endpoint to the `bytes` Fern Definition for bytes download (affects seed)
- Release a new CLI version that will run appropriate IR migrations for v55 -> v54

## Testing


seed/ts-sdk/bytes/src/api/resources/service/client/Client.ts
```diff
@@ -78,4 +78,53 @@ export class Service {
                 });
         }
     }
+
+    /**
+     * @param {Service.RequestOptions} requestOptions - Request-specific configuration.
+     */
+    public async download(requestOptions?: Service.RequestOptions): Promise<ArrayBuffer> {
```
dsinghvi, line 85: Critical change

amckinney (dismissed):
Looks great 👏

seed/ts-sdk/bytes/src/api/resources/service/client/Client.ts
remorses: In that case Response is the best option, it can be streamed, turned into a buffer, blob, etc

PR #5407, opened 2024-12-13
PHP discriminated unions proposal

ajgateno:
Our PHP SDK generator currently does not support discriminated unions. This PR is a proposal detailing how they should look and feel. The following description outlines a simple example with both the request and response use cases.

## API

Consider the following `Shape` union, which is defined by the following API specification:

```yaml
types:
  Shape:
    union:
      circle: Circle
      square: Square

  Circle: 
    properties:
      radius: double

  Square: 
    properties:
      length: double
```

## UX

With the generated `Shape` class, users are presented a couple static constructors to easily build an instance of the `Shape` and send it in a request:

```php
$this->client = new AcmeClient();

$shape = Shape::circle(
  new Circle([
    'radius' => 1.0
  ])
)

$response = $this->client->shapes->create(
  request: new CreateShapeRequest([
    [
       'shape' => $shape
    ]
  ])
)
```

Then on the response end, you can either switch on the generated discriminant property (i.e. `type`), or use the generated type guard to act upon the `Shape` variant:

```php
$response = $this->client->shapes->get(
  request: new GetShapeRequest([
    'id' => "..."
  ])
);

$shape = $response->shape;

// Use a switch statement to handle the discriminant.
switch ($shape->type) {
    case 'circle':
        $circle = $shape->asCircle();
        echo "Circle with radius: " . $circle->radius;
        break;

    case 'square':
        $square = $shape->asSquare();
        echo "Square with side length: " . $square->sideLength;
        break;

    default:
        echo "Unknown shape type: " . $shape->type;
        break;
}

// Alternatively, use the type directly.
if ($shape->type === 'circle') {
    $circle = $shape->asCircle();
    echo "Circle with radius: " . $circle->radius;
} elseif ($shape->type === 'square') {
    $square = $shape->asSquare();
    echo "Square with side length: " . $square->sideLength;
} else {
    echo "Unknown shape type: " . $shape->type;
}
```

## Class design

The generated PHP class is shown below:

```php
class Shape extends DiscriminatedUnion
{
    /**
     * @var 'circle'|'square'|'_unknown' $type 
     */
    #[JsonProperty('type')]
    #[Discriminant(['circle' => Circle::class, 'square' => Square::class])]
    public string $type;

    /**
     * @var Circle|Square|mixed
     */
    #[JsonSkip]
    public mixed $value;

    /**
     * @param ?array{
     *   type?: 'circle'|'square'|'_unknown',
     *   value?: Circle|Square|mixed,
     * } $options
     */
    private function __construct(
        private readonly ?array $options = null,
    ) {
        $this->type = $this->options['type'] ?? '_unknown';
        $this->value = $this->options['value'] ?? null;
    }

    public static function circle(
        Circle $circle
    ): Shape {
        return new Shape([
            'type' => 'circle',
            'value' => $circle
        ]);
    }

    public static function square(
        Square $square
    ): Shape {
        return new Shape([
            'type' => 'square',
            'value' => $square
        ]);
    }

    public static function _unknown(
        mixed $_unknown
    ): Shape {
        return new Shape([
            'value' => $_unknown
        ]);
    }

    public function asCircle(): Circle
    {
        if ($this->type != 'circle') {
            throw new \Exception(
                "Expected type to be 'circle'; got '$this->type.'"
            );
        }

        if (!($this->value instanceof Circle)) {
            throw new \Exception(
                "Expected value to be instance of Circle."
            );
        }

        return $this->value;
    }

    public function asSquare(): Square
    {
        if ($this->type != 'square') {
            throw new \Exception(
                "Expected type to be 'square'; got '$this->type.'"
            );
        }

        if (!($this->value instanceof Square)) {
            throw new \Exception(
                "Expected value to be instance of Square."
            );
        }

        return $this->value;
    }
}
```


seed/php-sdk/unions/src/Types/Types/UnionWithLiteral.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    #[JsonProperty('type')]
+    #[Discriminant(['fern' => 'fern'])]
+    public string $type;
+
+    /**
+     * @var 'fern'|mixed
+     */
+    #[JsonSkip]
+    public mixed $value;
```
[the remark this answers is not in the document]
amckinney: Does this literal even need to support `mixed`? Thoughts on always just requiring it to be `fern`? It would make interacting with the literal string a lot easier from the user's perspective.

I can see that you're using `mixed` to handle the unknown values from the server (which is great), but I wonder if we should be more strict with literal values to start.

I realize you could apply the same thinking here though - if a server ever starts supporting a new value, we ideally don't want to throw. However, failing hard on this literal value is valuable elsewhere (e.g. undiscriminated unions).

seed/php-sdk/unions/src/Types/Types/UnionWithLiteral.php
```diff
@@ -0,0 +1,86 @@
...
+            throw new \Exception(
+                "Expected type to be 'fern'; got '$this->type.'"
+            );
+        }
+
+        if ($this->value !== 'fern') {
+            throw new \Exception(
+                "Expected value to be \"fern\"."
+            );
+        }
```
amckinney, line 82: Depending on the phpstan literal support and whether or not we make this _always_ be the literal value, we might not need this check.

With that said, we should definitely validate against literal values in the [de]serialization layer. For example, if this is a response type and the server sends the following, we should fail to deserialize the type:

```php
{
  "fern": "something"
}
```

All that to say - the more we can move this validation logic to the serialization layer the better. That way when users have a constructed instance of the type, it's guaranteed to be valid.

seed/php-sdk/unions/src/Union/Types/Shape.php
```diff
@@ -0,0 +1,99 @@
...
+use Seed\Union\Types\Circle;
+use Seed\Union\Types\Square;
+
+class Shape extends DiscriminatedUnion
+{
+    /**
+     * @var 'circle'|'square'|'_unknown' $type 
+     */
+    #[JsonProperty('type')]
+    #[Discriminant(['circle' => Circle::class, 'square' => Square::class,])]
```
amckinney, line 18: nit: Trailing comma here:
```suggestion
    #[Discriminant(['circle' => Circle::class, 'square' => Square::class])]
```

seed/php-sdk/unions/src/Union/Types/Shape.php
```diff
@@ -0,0 +1,99 @@
+<?php
```
amckinney, line 1: This file is so sweet, really shows the whole UX well 👏

seed/php-sdk/unions/tests/Seed/Core/Json/DiscriminatedUnionTest.php
```diff
@@ -0,0 +1,127 @@
...
+        );
+
+        $shape = Shape::fromJson($expected_json);
+
+        $this->assertTrue($shape instanceof Shape, "Deserialized object must be of type Shape");
+        $this->assertEquals($shape->type, 'circle', "Type property must be circle");
+        $this->assertNotNull($shape->value, "Shape value must not be null");
+        $this->assertTrue($shape->value instanceof Circle);
+        $this->assertEquals($shape->value->radius, 1.0, "Circle radius must match definition");
+        $this->assertJsonStringEqualsJsonString($shape->toJson(), $expected_json);
```
amckinney, line 34: Love these round trip tests 👏

seed/php-sdk/unions/tests/Seed/Core/Json/DiscriminatedUnionTest.php
```diff
@@ -0,0 +1,127 @@
...
+        $this->assertJsonStringEqualsJsonString($primitiveUnion->toJson(), $expected_json);
+    }
+
+    public function testRoundTripSerdeWithLiteral(): void
+    {
+        $expected_json = json_encode(
+            [
+                'base' => 'base',
+                'type' => 'fern',
+                'fern' => 'fern',
```
amckinney, line 113: Depending on the result of the conversation above, would be great to include a test that throws when the literal doesn't match.

amckinney (commented):
This looks _great_ - mostly just thinking out loud on my comments, but I think this is shaping up really well.

seed/php-sdk/unions/src/Types/Types/UnionWithLiteral.php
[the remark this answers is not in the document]
ajgateno: I'd imagine this is the role of `asFern()`, right?

seed/php-sdk/unions/src/Types/Types/UnionWithLiteral.php
```diff
@@ -0,0 +1,86 @@
...
+     *   type?: 'fern'|'_unknown',
+     *   value?: 'fern'|mixed,
+     * } $options
+     */
+    public function __construct(
+        private readonly ?array $options = null,
+    ) {
+        if (is_null($this->options)) {
+            throw new \Exception("Missing all required properties");
+        }
```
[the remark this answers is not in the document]
ajgateno: Done!

seed/php-sdk/unions/src/Union/Types/Shape.php
ajgateno: done

seed/php-sdk/unions/src/Core/Json/JsonSerializableType.php
```diff
@@ -121,15 +165,60 @@ public static function jsonDeserialize(array $data): static
...
+                $arrayType = null;
+                $unionType = null;
+                $typeName = null;
+
+                // Handle discriminated union
+                if ($reflectionClass->isSubclassOf(DiscriminatedUnion::class)) {
+                    $discriminantAttr = $property->getAttributes(Discriminant::class)[0] ?? null;
+
+                    if ($discriminantAttr && count($discriminantAttr->getArguments()) === 1) {
+                        $propertyName = 'value';
```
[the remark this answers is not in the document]
ajgateno: done

seed/php-sdk/unions/src/Types/Types/UnionWithLiteral.php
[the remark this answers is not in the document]
ajgateno: Actually I think I see what you mean here... I'm going to take another look at how much validation we can comfortably replace with serialization-step logic and come back to whether we can drop this mixed type.

PR #5422, opened 2024-12-16
feat(typescript): Basic Dynamic Snippet Generation for Typescript-v2

Schnides123:
added rough draft code and tests for functions, parameters, comments, and function invocations for TS-v2.

## Description

Adds dynamic snippet generation to typescript-v2 generator package

## Changes Made

introduces basic endpoint and dynamic snippet generation logic, with basic test coverage

## Testing

- added basic imdb snapshot test cases
- verified that imdb snapshot generates correct output and errors, minus import handling


generators/typescript-v2/dynamic-snippets/src/EndpointSnippetGenerator.ts
```diff
@@ -1,6 +1,11 @@
 import { FernIr } from "@fern-api/dynamic-ir-sdk";
 import { ts } from "@fern-api/typescript-ast";
 
+const SNIPPET_PACKAGE_NAME = "example";
+const SNIPPET_IMPORT_PATH = "fern";
+const SNIPPET_FUNC_NAME = "do";
+const CLIENT_VAR_NAME = "client";
```
amckinney, line 7: These values might not be relevant to TypeScript snippets - they're largely Go-specific. The goal we're after is something like the following:

```typescript
import { CohereClientV2 } from 'cohere-ai';

const cohere = new CohereClientV2({
  token: '<YOUR_API_KEY>',
});
await cohere.chat({
  model: 'command-r-plus',
  messages: [
    {
      role: 'user',
      content: 'hello world!',
    },
  ],
});
```

generators/typescript-v2/ast/src/ast/__test__/Function.test.ts
```diff
@@ -0,0 +1,63 @@
...
+                arguments_: [ts.TypeLiteral.string("https://api.example.com")]
+            });
+            expect(actual.toStringFormatted()).toMatchSnapshot();
+        });
+    });
+
+    describe("function", () => {
+        it("Should generate a function", () => {
+            const actual = ts.func({
+                name: "WithBaseURL",
```
amckinney, line 17: nit: For all these tests, let's use `withBaseURL` so that it's more idiomatic TypeScript. Exported functions are often `camelCase`'d in TypeScript.

generators/browser-compatible-base/src/ast/AbstractWriter.ts
```diff
@@ -150,6 +150,24 @@ export class AbstractWriter {
...
+        this.write(prefix);
+        if (nodes.length > 0) {
+            writeFunction(nodes[0]);
+            for (let i = 1; i < nodes.length; i++) {
+                this.write(delimiter);
+                writeFunction(nodes[i]);
+            }
+        }
+        this.write(suffix);
+    }
```
amckinney, line 169: Nice, I feel like we've needed something like this for a while. Couple things we can do to improve the readability at the call-site though:

1. Let's use an in-lined request with named parameters.
2. I'd probably drop the `prefix` and `suffix` arguments to start. I feel like it's marginally more readable just to include those at the call-site so that users can clearly see where the beginning and end of a `( ... )` is.
3. Should we write the arguments across newlines and indent/dedent the inner values by default? We could also add a `newline` option to control that behavior if we prefer.

In combination, I imagine something like this:

```typescript
public delimit({
  nodes,
  writeFunc,
  delimiter,
  newline,
}: {
  nodes: AbstractAstNode[],
  writeFunc: (node: AbstractAstNode) => void,
  delimiter: string,
  newline?: boolean
}): void {
  if (nodes.length === 0) {
    return;
  }
  ...
}
```

With that said, I imagine we'll eventually need something like Go's writeArguments to handle no-op arguments well. It might be simpler to repurpose that method as a separate TypeScript utility instead (which uses ts.TypeLiteral instead of go.TypeInstantiation).

I'll leave it to you though - either way works!

generators/typescript-v2/ast/src/ast/Func.ts
```diff
@@ -0,0 +1,70 @@
...
+        }
+        writer.write(" => ");
+        writer.writeLine("{");
+        writer.indent();
+        this.body?.write(writer);
+        writer.dedent();
+        writer.writeNewLineIfLastLineNot();
+        writer.writeLine("}");
+    }
+}
```
amckinney, line 70: Nice, this looks great!

generators/typescript-v2/dynamic-snippets/src/EndpointSnippetGenerator.ts
Schnides123: Gotcha, so no need to hold onto them if not required for generating snippets

generators/browser-compatible-base/src/ast/AbstractWriter.ts
Schnides123: For the prefix suffix, I was mostly thinking about leaving those in for cases like a union type, where you could use this for `< Type1 | Type2 | Type3 >` with a delimiter of "|" and diamond brackets for the prefix/suffix. I can see it being more ergonomic to simplify it for the common case though

generators/typescript-v2/ast/src/ast/__test__/__snapshots__/Function.test.ts.snap
Swimburger, line 1: The snapshots look great

generators/typescript-v2/ast/src/ast/__test__/__snapshots__/Method.test.ts.snap
Swimburger, line 1: Snapshot looks great

Swimburger (approved):

PR #5427, opened 2024-12-17
feat(typescript): Support retainOriginalName/noSerdeLayer in inline path parameters

Swimburger:
## Description
Support retainOriginalName/noSerdeLayer in inline path parameters

## Changes Made
- positional path parameters respect retainOriginalName, and ignore noSerdeLayer
- root path parameters respect retainOriginalName, and ignore noSerdeLayer
- inline path parameters respect retainOriginalName and noSerdeLayer

## Testing
- [x] Seed tests

### DIFF
--- a/packages/cli/generation/ir-generator-tests/src/dynamic-snippets/__test__/test-definitions/path-parameters.json
+++ b/packages/cli/generation/ir-generator-tests/src/dynamic-snippets/__test__/test-definitions/path-parameters.json
@@ -328,7 +328,7 @@
             },
             "location": {
                 "method": "GET",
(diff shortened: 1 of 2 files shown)


generators/typescript/utils/commons/src/codegen-utils/getParameterNameForPathParameter.ts
```diff
@@ -0,0 +1,86 @@
+import { Name, PathParameter } from "@fern-fern/ir-sdk/api";
```
amckinney, line 1: Nice, localizing all this logic in a single place looks great 👏

amckinney (approved):
👏

PR #5434, opened 2024-12-18
fix(typescript): Fix inline path body request bug

Swimburger:
## Description
Fix a bug where client would send request wrapper instead of the body of the request wrapper, when the request has inline path parameters and a body property.

## Testing

- [x] Seed testing

### DIFF
--- a/packages/cli/generation/ir-generator-tests/src/dynamic-snippets/__test__/test-definitions/path-parameters.json
+++ b/packages/cli/generation/ir-generator-tests/src/dynamic-snippets/__test__/test-definitions/path-parameters.json
@@ -1074,6 +1074,354 @@
                 "type": "json"
             }
         },
+        "endpoint_user.createUser": {
+            "auth": null,
+            "declaration": {
+                "name": {
+                    "originalName": "createUser",
+                    "camelCase": {
+                        "unsafeName": "createUser",
+                        "safeName": "createUser"
+                    },
+                    "snakeCase": {
+                        "unsafeName": "create_user",
+                        "safeName": "create_user"
+                    },
+                    "screamingSnakeCase": {
+                        "unsafeName": "CREATE_USER",
+                        "safeName": "CREATE_USER"
+                    },
+                    "pascalCase": {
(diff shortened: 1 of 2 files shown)


generators/typescript/sdk/client-class-generator/src/request-parameter/RequestBodyParameter.ts
```diff
@@ -42,6 +42,10 @@ export class RequestBodyParameter extends AbstractRequestParameter {
         throw new Error("Cannot reference query parameter because request is not wrapped");
     }
 
+    public getReferenceToPathParameter(): ts.Expression {
+        throw new Error("Cannot reference path parameter because request is not wrapped");
+    }
```
amckinney, line 47: Does this need to be conditional? If inlinePathParameters is false, `getReferenceToPathParameter` should just return the name of the positional path parameter argument, right? Or is that not actually relevant here?

seed/ts-sdk/path-parameters/inline-path-parameters-no-serde/src/api/resources/user/client/Client.ts
```diff
@@ -90,6 +91,138 @@ export class User {
...
+     *             name: "name",
+     *             tags: ["tags", "tags"]
+     *         }
+     *     })
+     */
+    public async updateUser(
+        request: SeedPathParameters.UpdateUserRequest,
+        requestOptions?: User.RequestOptions
+    ): Promise<SeedPathParameters.User> {
+        const { user_id: userId, body: _body } = request;
```
amckinney, line 175: Such a simple solution, nicely done 👍

amckinney (approved):

generators/typescript/sdk/client-class-generator/src/request-parameter/RequestBodyParameter.ts
Swimburger: This class is already in the context of a body request parameter, so if the body request parameter is not a wrapper, there's no path parameters on the body request.

PR #5435, opened 2024-12-18
feat(typescript): Added AST classes for typescript-v2 dynamic snippet generation

Schnides123:
## Description
Adds AST classes and tests for types, type literals, functions, and methods.

## Changes Made
Implemented basic AST classes for typescript-v2 snapshot generation

## Testing

- Added unit test coverage for new classes
- Ran tests and checked snapshot files for accuracy


generators/browser-compatible-base/src/ast/AbstractWriter.ts
```diff
@@ -150,6 +150,24 @@ export class AbstractWriter {
         this.indentLevel--;
     }
 
+    public delimit<T extends AbstractAstNode>(
+        nodes: T[],
+        delimiter: string,
+        writeFunction: (node: T) => void,
+        prefix: string = "",
```
amckinney: Let's make sure to fix these `eslint` errors.

generators/browser-compatible-base/src/ast/AbstractWriter.ts
```diff
@@ -150,6 +150,24 @@ export class AbstractWriter {
...
+        suffix: string = ""
+    ): void {
+        this.write(prefix);
+        if (nodes.length > 0) {
+            writeFunction(nodes[0]!);
```
amckinney: We might need to refactor things a bit to prevent any need for the `!` operator.

What if we did a length check then iterated one by one? We can also start by always using newlines to separate arguments (otherwise lines could get huge).

```typescript
if (nodes.length === 0) {
  this.write(prefix);
  this.write(suffix);
  return;
}
this.writeLine(prefix);
for (const node of nodes) {
  writeFunction(node);
  this.writeLine(delimiter);
}
this.writeLine(suffix);
```

generators/browser-compatible-base/src/ast/AbstractWriter.ts
amckinney: Let's use the in-lined request params I mentioned in an [earlier comment.](https://github.com/fern-api/fern/pull/5422#discussion_r1887542602). Otherwise the call-sites are difficult to follow.

generators/typescript-v2/ast/src/ast/Func.ts
```diff
@@ -0,0 +1,81 @@
...
+                if (returnType != null) {
+                    returnType.write(writer);
+                }
+            } else {
+                writer.write("(");
+                writer.delimit(this._return_, ", ", (type) => type.write(writer), "(", ")");
+                writer.write(")");
+            }
+        }
+        writer.write(" => ");
```
amckinney, line 73: It looks like you're modeling this as an anonymous function (i.e. `() => ...`). I don't think we need this yet, so we might just want to leave this whole AST node out for now, especially becasue I think we'll reserve this AST node for defining top-level functions:

```typescript
function doSomething(...) {
  ...
}
```

In fact, I'm not sure we ever use anonymous functions within the generated SDK.

generators/typescript-v2/ast/src/ast/FunctionInvocation.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    constructor({ func, arguments_ }: FunctionInvocation.Args) {
+        super();
+        this.func = func;
+        this.arguments_ = arguments_;
+    }
+
+    public write(writer: Writer): void {
+        writer.writeNode(this.func);
+        writer.delimit(this.arguments_, ", ", (argument) => argument.write(writer), "(", ")");
```
amckinney, line 25: Thoughts on using an approach like `writeArguments` like I mentioned [here](https://github.com/fern-api/fern/pull/5422#discussion_r1887542602)? Otherwise we would need to refactor this to handle the `nop` nodes again separately.

generators/typescript-v2/ast/src/ast/Type.ts
```diff
@@ -2,12 +2,99 @@ import { assertNever } from "@fern-api/core-utils";
...
+    | Map_
+    | Tuple
+    | Enum
+    | Any
+    | Unknown
+    | Void
+    | Undefined
+    | Null
+    | Never
+    | Nop;
```
Swimburger, line 21: What's `Nop`? Is it supposed to be `NoOp` or `Nope`?

generators/typescript-v2/ast/src/ast/Method.ts
```diff
@@ -0,0 +1,51 @@
...
+        reference: Reference;
+        name: string;
+        parameters: Parameter[];
+        return_: Type[];
+        body?: CodeBlock;
+        docs?: string;
+    };
+}
+
+export class Method extends Func {
```
amckinney, line 21: Similar to above, I don't think we need this node to start. If there are still TODOs, it's probably best to just leave it out for now. Plus, it looks like the snapshots are actually generating top-level functions instead of methods?

```typescript
function withBaseURL(baseUrl: string): string {
    console.log("foo");
    return baseUrl + "foo";
}
```

generators/typescript-v2/ast/src/ast/MethodInvocation.ts
```diff
@@ -0,0 +1,32 @@
...
+        this.on = on;
+        this.method = method;
+        this.arguments_ = arguments_;
+    }
+
+    public write(writer: Writer): void {
+        this.on.write(writer);
+        writer.write(".");
+        writer.write(this.method);
+        writer.delimit(this.arguments_, ", ", (argument) => argument.write(writer), "(", ")");
```
amckinney, line 30: Same note about `writeArugments`.

Swimburger (commented):
LGTM, but wait for Alex since he has more context.
Make sure to fix the eslint errors.

generators/typescript-v2/ast/src/ast/Type.ts
```diff
@@ -18,8 +105,58 @@ export class Type extends AstNode {
             case "string":
                 writer.write("string");
                 break;
+            case "number":
+                writer.write("number");
+                break;
+            case "bigint":
+                writer.write("bigint");
```
amckinney, line 112: Let's add a TODO here since this will actually require using some configuration. We only support `bigInt` when we've actually configured the generator to do so ([link](https://github.com/fern-api/fern/blob/0c3d46efacba93049a6485cce44394955c0eac5d/generators/typescript/sdk/cli/src/SdkGeneratorCli.ts#L64)).

generators/typescript-v2/ast/src/ast/Type.ts
```diff
@@ -2,12 +2,99 @@ import { assertNever } from "@fern-api/core-utils";
...
     type: "string";
 }
 
+interface Number_ {
+    type: "number";
+}
+
+interface BigInt_ {
+    type: "bigint";
+}
```
Swimburger, line 33: Is BigInt something we should care about in TS?

generators/typescript-v2/ast/src/typescript.ts
```diff
@@ -12,6 +12,27 @@ export function reference(arg: AST.Reference.Args): AST.Reference {
...
+export function invokeFunction(args: AST.FunctionInvocation.Args): AST.FunctionInvocation {
+    return new AST.FunctionInvocation(args);
+}
+
+export function invokeMethod(args: AST.MethodInvocation.Args): AST.MethodInvocation {
+    return new AST.MethodInvocation(args);
+}
+
+export function parameter(args: AST.Parameter.Args): AST.Parameter {
+    const foo = AST.Parameter;
```
amckinney, line 32: What's this used for?

generators/typescript-v2/ast/src/typescript.ts
```diff
@@ -12,6 +12,27 @@ export function reference(arg: AST.Reference.Args): AST.Reference {
...
+    return new AST.Method(args);
+}
+
+export function invokeFunction(args: AST.FunctionInvocation.Args): AST.FunctionInvocation {
+    return new AST.FunctionInvocation(args);
+}
+
+export function invokeMethod(args: AST.MethodInvocation.Args): AST.MethodInvocation {
+    return new AST.MethodInvocation(args);
+}
```
amckinney, line 29: Based on the other comments, I _think_ we only need these to start.

generators/typescript-v2/ast/src/ast/Type.ts
Schnides123: It's no-op. In the go-v2 generator, we use the same abbreviation so I'm keeping it as nop for now.

generators/typescript-v2/ast/src/ast/Type.ts
amckinney: We unfortunately will need to for some customers.

generators/typescript-v2/ast/src/ast/Type.ts
amckinney: This is the node I used in [the other AST](https://github.com/fern-api/fern/blob/0c3d46efacba93049a6485cce44394955c0eac5d/generators/go-v2/ast/src/ast/TypeInstantiation.ts#L22) - it's a word I've used in the past for `no-op` since it's always weird to see `Noop` or `NoOp`. We can change this later though.

generators/browser-compatible-base/src/ast/AbstractWriter.ts
Schnides123: Yep, chugging away at them now...

generators/browser-compatible-base/src/ast/AbstractWriter.ts
Schnides123: True, I think the benefit of parameterizing the delimiter is that we could plumb this through to a config eventually for how a given customer wants to format it

generators/browser-compatible-base/src/ast/AbstractWriter.ts
Schnides123: I like dropping the prefix/suffix, but I think we could probably drop the proposed newline arg. The user can supply writeLine for the write function, which would have the desired effect.

generators/typescript-v2/ast/src/ast/Func.ts
Schnides123: okay, I'll update it to the other syntax

generators/typescript-v2/ast/src/ast/Type.ts
Schnides123: can do

generators/typescript-v2/ast/src/typescript.ts
Schnides123: I think that was a debug breakpoint at one point

generators/typescript-v2/ast/src/typescript.ts
Schnides123: up to you, I'm happy to toss them in now if you're up for reviewing them

generators/typescript-v2/ast/src/ast/Method.ts
Schnides123: Yeah, we can probably leave a lot of it stubbed.

generators/typescript-v2/ast/src/ast/Func.ts
```diff
@@ -0,0 +1,89 @@
...
+    interface Args {
+        name: string;
+        parameters: Parameter[];
+        return_: Type[];
+        body: CodeBlock | undefined;
+        docs: string | undefined;
+    }
+}
+
+export class Func extends AstNode {
```
amckinney, line 18: nit: Mind renaming this to `Function`? We try to keep the AST nodes named based on the language. For Go we refer to them as `Func`, but for TypeScript we can say `Function`.

generators/typescript-v2/ast/src/ast/Func.ts
```diff
@@ -0,0 +1,89 @@
...
+        writer.writeNode(new Comment({ docs: this.docs }));
+        writer.write("function ");
+        writer.write(`${this.name}`);
+        writer.write("(");
+        writer.delimit({
+            nodes: this._parameters,
+            delimiter: ",\n",
+            writeFunction: (parameter) => parameter.writeWithType(writer)
+        });
+        writer.write("\n)");
```
amckinney, line 64: We should use `writer.writeLine(")")`.

generators/typescript-v2/ast/src/ast/Func.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public write(writer: Writer): void {
+        writer.writeNode(new Comment({ docs: this.docs }));
+        writer.write("function ");
+        writer.write(`${this.name}`);
+        writer.write("(");
+        writer.delimit({
+            nodes: this._parameters,
+            delimiter: ",\n",
```
amckinney, line 61: This feels off - including the `\n` within the delimiter feels a little bug prone, especially because we have the `writer.writeLine` method already.

With that said, it might be simpler just to write the implementation more explicitly in each call-site, even if you end up with the same shape of code across different AST nodes.

We can leave this as-is for now and come back to it though.

generators/typescript-v2/ast/src/ast/Func.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    public get docs(): string | undefined {
+        return this._docs;
+    }
+
+    public write(writer: Writer): void {
+        writer.writeNode(new Comment({ docs: this.docs }));
+        writer.write("function ");
+        writer.write(`${this.name}`);
+        writer.write("(");
```
amckinney, line 58: Should this be `writer.writeLine("(")`?

amckinney (commented):
Alright one more pass of comments but after these are resolved think we're good to go.

generators/typescript-v2/ast/src/ast/Func.ts
Schnides123: Sure, can do

generators/typescript-v2/ast/src/ast/Func.ts
Schnides123: makes sense

generators/typescript-v2/ast/src/ast/Func.ts
Schnides123: yeah, true. I'm going to mull it over for a bit

generators/typescript-v2/ast/src/ast/Func.ts
Schnides123: Yeah, probably

generators/typescript-v2/ast/src/ast/Func.ts
Swimburger: This could be an option inside the delimit function telling it each item has to be on a new line.

generators/typescript-v2/ast/src/ast/TypeLiteral.ts
```diff
@@ -43,6 +48,10 @@ interface Tuple {
     values: TypeLiteral[];
 }
 
+interface Nop {
```
Swimburger, line 51: What's `Nop` here? Should it be `NoOp`?

generators/typescript-v2/ast/src/ast/Function.ts
```diff
@@ -0,0 +1,89 @@
...
+        docs: string | undefined;
+    }
+}
+
+export class Function extends AstNode {
+    protected _name: string;
+    protected _parameters: Parameter[];
+    protected _return_: Type[];
+    protected _body: CodeBlock | undefined;
+    protected _docs: string | undefined;
```
amckinney, line 23: nit: We can drop the `_` prefix on these parameters. We use the `_` suffix for keywords.

generators/typescript-v2/ast/src/ast/core/AstNode.ts
```diff
@@ -6,13 +6,19 @@ export abstract class AstNode extends AbstractAstNode {
     /**
      * Writes the node to a string.
      */
-    public toString(): string {
+    public async toString(): Promise<string> {
+        const writer = new Writer();
+        this.write(writer);
+        return writer.toString();
+    }
```
Swimburger, line 13: Is there a reason why there would be an async version?
If there's a good reason, I would make a `toStringAsync` rather than a `toStringSync`.
Since all object in JS already have `.toString()` which is sync, I'd keep the sync version as `toString` and if async is necessary, `toStringAsync`.

generators/typescript-v2/ast/src/ast/__test__/Function.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        });
+    });
+
+    describe("function with a body", () => {
+        it("Should generate a function with a body", () => {
+            const actual = ts.function_({
+                name: "withBaseURL",
+                parameters: [ts.parameter({ name: "baseUrl", type: ts.Types.string() })],
+                return_: [ts.Types.string()],
+                body: ts.codeblock("console.log('foo');\nreturn baseUrl + 'foo';"),
```
amckinney, line 49: We should ideally use the structured `ts.codeblock` that I mentioned in another comment earlier.

generators/typescript-v2/ast/src/ast/Function.ts
```diff
@@ -0,0 +1,89 @@
...
+import { Parameter } from "./Parameter";
+import { Type } from "./Type";
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+
+export declare namespace Function {
+    interface Args {
+        name: string;
+        parameters: Parameter[];
+        return_: Type[];
```
Swimburger, line 12: Return types in TS should be a single value, not an array IMO.
If we need multiple values it should be a single Type, and the type in question would be an array or something similar.

amckinney (approved):
Sweet looks good! We'll probably need to come back to some of these details (e.g. the `newline` config), but this at least gets things in a good place to iterate on.

Swimburger (approved):

generators/typescript-v2/ast/src/ast/Func.ts
amckinney: @Swimburger Agreed, I actually had the same thought [here](https://github.com/fern-api/fern/pulls#discussion_r1887542602).

generators/typescript-v2/ast/src/ast/Function.ts
amckinney: Yea, this is probably a relic of the Go AST, which often requires multiple return args (for `error`).

generators/typescript-v2/ast/src/ast/core/AstNode.ts
amckinney: Yea this is on me - we should go back and refactor to prefer the sync method as the `toString` option.

generators/typescript-v2/ast/src/ast/TypeLiteral.ts
amckinney: https://github.com/fern-api/fern/pull/5435#discussion_r1890671038

Swimburger (approved):

PR #5438, opened 2024-12-18
chore(java): Test java inline path parameters

ajgateno:
This PR adds a test to verify that the Java SDK generator is handling inline path parameters correctly


seed/java-sdk/java-inline-path-parameters/java-inline-path-parameters/src/main/java/com/seed/api/resources/imdb/ImdbClient.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public Movie getMovie(String movieId) {
+        return getMovie(movieId, GetMovieRequest.builder().build());
+    }
+
+    public Movie getMovie(String movieId, GetMovieRequest request) {
+        return getMovie(movieId, request, null);
+    }
+
+    public Movie getMovie(String movieId, GetMovieRequest request, RequestOptions requestOptions) {
```
ajgateno, line 89: @amckinney Looks like this was already implemented. If we want, we can make this test more general and run it on other generators to see which ones we're missing to support this feature.

amckinney: In this case, we actually want `movieId` to be included in `GetMovieRequest`. That's what in-lined in the request is referring to.

ajgateno: Gotcha, will get on this then!

PR #5449, opened 2024-12-19
fix(python): Pagination correctly uses specified defaults to support custom schemes.

eyw520:
## Description
Pagination correctly uses default page values when specified in the Fern definition.

## Changes Made
- Added a integer default extractor util to extract offset parameter default.
- Correctly use extracted default in default `page_number` init.

## Testing


generators/python/src/fern_python/generators/sdk/client_generator/request_properties.py
```diff
@@ -3,3 +3,36 @@
...
+                boolean=lambda _: None,
+                long_=lambda _: None,
+                big_integer=lambda _: None,
+                uint=lambda _: None,
+                uint_64=lambda _: None,
+                date=lambda _: None,
+                date_time=lambda _: None,
+                uuid_=lambda _: None,
+                base_64=lambda _: None,
+                float_=lambda _: None,
```
amckinney, line 27: We might need to handle default values typed on other numeric types (e.g. [long](https://github.com/fern-api/fern/blob/6d3eb48a4c090afe7a7627d960027a302c783ca2/packages/ir-sdk/fern/apis/ir-types-latest/definition/types.yml#L258) and [double](https://github.com/fern-api/fern/blob/6d3eb48a4c090afe7a7627d960027a302c783ca2/packages/ir-sdk/fern/apis/ir-types-latest/definition/types.yml#L267) both support default values today). 

For example, if the user defines the parameter in the OpenAPI spec as a `number` instead of an `integer` (re: [docs](https://swagger.io/docs/specification/v3_0/data-models/data-types/#numbers)).

amckinney (approved):
Nice! One comment to consider, but otherwise looks good!

generators/python/src/fern_python/generators/sdk/client_generator/request_properties.py
eyw520: Great point. What would be the ideal behavior for long / doubles? Collapsing these values into ints when defaulting the pagination offset?

PR #5454, opened 2024-12-19
feat(cli): docs dev v2 with new OpenApi Parser

RohinBhargava:
## Description
Enables the new openapi parser in docs dev command.

## Changes Made
- Adds a converter from FDR Latest -> Fern Nav structure
- Adds an experimental config for use with preview and publish
- Adds a skeleton for publish command (to be done in next PR)
- Refactors parsing logic into a reusable function
- Bumps FDR version (API change), and parser version

## Testing


packages/cli/cli/src/cli.ts
```diff
@@ -992,6 +992,10 @@ function addDocsPreviewCommand(cli: Argv<GlobalCliOptions>, cliContext: CliConte
                     string: true,
                     hidden: true,
                     description: "Path of the local docs bundle to use"
+                })
+                .option("v2", {
+                    boolean: true,
+                    description: "Use openapi parser v2"
```
dsinghvi, line 998: Can we add this to the `api: ` block in `docs.yml` instead? That way someone can't docs dev v2 but then generate on v1 and encounter issues

Additionally if someone has 5 OpenAPIs, this gives you some upgrade flexibility

packages/cli/docs-resolver/src/ApiReferenceNodeConverterLatest.ts
```diff
@@ -0,0 +1,1000 @@
+import { docsYml } from "@fern-api/configuration-loader";
```
RohinBhargava, line 1: @dsinghvi This is the main event

packages/cli/cli/src/commands/write-docs-definition/writeDocsDefinitionForProject.ts
```diff
@@ -22,6 +24,17 @@ export async function writeDocsDefinitionForProject({
...
+        const ossWorkspaces = (
+            await Promise.all(
+                project.apiWorkspaces.map(async (workspace) => {
+                    if (workspace instanceof OSSWorkspace) {
+                        return workspace as OSSWorkspace;
+                    }
+                    return null;
+                })
+            )
+        ).filter(isNonNullish);
```
dsinghvi, line 36: nit: seeing this logic in a couple places -- thoughts on making it a helper and reusing? that way if a new workspace type is introduced we'll only need to handle in one place

packages/cli/docs-preview/src/previewDocs.ts
```diff
@@ -142,3 +168,28 @@ class ReferencedAPICollector {
...
+    constructor(private readonly context: TaskContext) {}
+
+    public addReferencedAPI({ api }: { api: FdrAPI.api.latest.ApiDefinition }): APIDefinitionID {
+        try {
+            this.apis[api.id] = api;
+            return api.id;
+        } catch (e) {
+            // Print Error
+            const err = e as Error;
+            this.context.logger.error(`Failed to read referenced API: ${err?.message} ${err?.stack}`);
```
dsinghvi, line 184: what does this error message look like? if its going to print out something thats hard for a user to diagnose it might be better to only print this on `--debug`

dsinghvi: Or something more human readable `Failed to read OpenAPI spec at /some/path/... Please file an issue`

dsinghvi (approved):

PR #5461, opened 2024-12-20
feat(typescript): use specified defaults for pagination offset parameters during SDK generation

eyw520:
## Description
Offset parameter uses specified default value in specs definition.

## Changes Made
- Extract default query parameter value in pagination offset.
- Use default value as fallback for optional offset.

## Testing


Swimburger:
Looks good to me, but can we add or update a test definition to demonstrate the output of this change?

dsinghvi (approved):
feel free to add to the existing `pagination` test definition

PR #5476, opened 2024-12-24
fix(typescript): Remove node-fetch in TypeScript SDK

### LINKED ISSUE #5468: Remove `node-fetch` and other dependencies from typescript sdk
### Feature Description

Currently even the `fernapi/fern-typescript-browser-sdk` generator import node-fetch and `form-data`, this code will fail to build in tools like Vite, because the bundler tries to import Node.js specific libraries and it fails.

Please remove `node-fetch` which is no longer needed, since in Node.js 18 fetch and FormData are already globals.

Here is the full list of packages you can remove:
- qs: you can inline a simple function instead of qs.stringify, should be very simple
- node-fetch: `fetch` is already available everywhere
- form-data and formdata-node: `FormData` is already available everywhere, use that instead
- form-data-encoder: you can do this instead
  ```ts
  new Response(formData, { headers: {'content-type': 'multipart/form' }}).text()
   ```
- js-base64: `atob` and `btoa` are available everywhere

remorses:
## Description
Removes node-fetch, Node.js 18 already comes with fetch global, Node 18 is the oldest version currently supported.

This change makes it possible to bundle the generated sdk for browser with bundlers like Vite and Rollup without marking node-fetch external

Fix #5468 

## Changes Made
Remove node-fetch

## Testing


dsinghvi:
@remorses this change will require some discussion on our end because there may be breaking changes related to file download and upload endpoints (where the current generated SDKs leverage `node-fetch`). 

cc @Swimburger

remorses:
The global fetch should already be used by default if available, this means that features like upload already work with it

generators/typescript/utils/core-utilities/fetcher/src/form-data-utils/__private__test__/multipart.upload.test.ts
```diff
@@ -48,7 +47,6 @@ describe("Multipart Form Data Tests", () => {
 
         await fdw.appendFile("file", new Blob([new Uint8Array(y)]), "asda");
 
-        const fetch = await getFetchFn();
```
remorses, line 51: This is the main change

generators/typescript/utils/core-utilities/fetcher/src/form-data-utils/FormDataWrapper.ts
```diff
@@ -19,7 +19,7 @@ export interface CrossPlatformFormData {
 
     appendFile(key: string, value: unknown, fileName?: string): Promise<void>;
 
-    getRequest(): MaybePromise<FormDataRequest<unknown>>;
```
Swimburger, line 22: We usually use `unknown` instead of `any`. Any reason why this was changed here?

Swimburger (changes requested):

generators/typescript/utils/core-utilities/fetcher/src/form-data-utils/FormDataWrapper.ts
remorses: yes, now fetch type is more strict and no longer accepts a RequestInit<unknown> so many type checks in tests for example fail

That unknown has no reason to exist because there is no way to override it with a generic

benmccann:
> @remorses this change will require some discussion on our end because there may be breaking changes related to file download and upload endpoints (where the current generated SDKs leverage node-fetch).

@dsinghvi @Swimburger @dannysheridan have y'all gotten a chance to have that discussion yet?

I see `getFetchFn.ts` has:
```
export async function getFetchFn(): Promise<any> {
    // In Node.js 18+ environments, use native fetch
    if (RUNTIME.type === "node" && RUNTIME.parsedVersion != null && RUNTIME.parsedVersion >= 18) {
        return fetch;
    }

    ...
}
```

Perhaps it could be part of the generator to specify which environments you want to run on? For folks who only are going to run on LTS versions of Node, it would be really nice to skip the unnecessary dependencies. In the next major, the version of that flag could be flipped so that `node-fetch` is being skipped by default in favor of the built-in `fetch` and then the user could opt-in to the older `node-fetch` when they require it

benmccann:
Ah, it looks like there was a follow-up PR to move this forward, but I hadn't seen it mentioned here or in the issue: https://github.com/fern-api/fern/pull/5869

Swimburger:
> > @remorses this change will require some discussion on our end because there may be breaking changes related to file download and upload endpoints (where the current generated SDKs leverage node-fetch).
> 
> @dsinghvi @Swimburger @dannysheridan have y'all gotten a chance to have that discussion yet?
> 
> I see `getFetchFn.ts` has:
> 
> ```
> export async function getFetchFn(): Promise<any> {
>     // In Node.js 18+ environments, use native fetch
>     if (RUNTIME.type === "node" && RUNTIME.parsedVersion != null && RUNTIME.parsedVersion >= 18) {
>         return fetch;
>     }
> 
>     ...
> }
> ```
> 
> Perhaps it could be part of the generator to specify which environments you want to run on? For folks who only are going to run on LTS versions of Node, it would be really nice to skip the unnecessary dependencies. In the next major, the version of that flag could be flipped so that `node-fetch` is being skipped by default in favor of the built-in `fetch` and then the user could opt-in to the older `node-fetch` when they require it

That's the plan, we'll provide feature flags so these dependencies will be gone in the future, but existing customers can maintain backwards compatibility.

Swimburger:
Great news, we have added options to remove all dependencies.
These options are now enabled by default in the [2.0.0 TypeScript SDK generator](https://github.com/fern-api/fern/pull/7763).

PR #5481, opened 2024-12-26
fix(cli): Correctly support `readOnly` query parameters in `PATCH` endpoint generation.

eyw520:
## Description
This fix correctly allows PATCH endpoints to omit readOnly query parameters.

## Changes Made
- Update if condition where PATCH endpoints were omitted from respecting readOnly params.

## Testing


dsinghvi (approved):
https://github.com/fern-api/fern/blob/85b22b41f80099a163e81fa1609a6b0918acc3ab/packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/ExampleEndpointFactory.ts#L59 

Here's another place we check readonly (it may be worth consolidating these checks into a single shared helper function)

PR #5483, opened 2024-12-26
fix(cli): load both `generators.yml` or `generators.yaml`

dsinghvi:
## Description
This PR fixes the Fern CLI to load both `generators.yml` or `generators.yaml`. 

## Changes Made
- Update the Fern CLI to load both `generators.yml` or `generators.yaml`
- Adds tests to confirm

## Testing


amckinney (approved):
Looks great! 

I wonder if we'll eventually need to apply the same treatment to other files (e.g. __package__.yml), or if we already have all our bases covered. Regardless, out of scope from this change.

PR #5488, opened 2024-12-26
fix(cli): Boolean default values are now propagated from the Fern Definition through to docs generation.

eyw520:
## Description
This PR correctly plumbs default values for booleans through to the IR -> FDR -> docs. 

## Changes Made
- Use default information when converting Boolean values in `visitRawTypeReference`.

## Testing

### DIFF
--- a/fern/pages/changelogs/cli/2024-12-27.mdx
+++ b/fern/pages/changelogs/cli/2024-12-27.mdx
@@ -1,0 +1,4 @@
+## 0.46.14
+**`(fix):`** Boolean default values are now propagated from the Fern Definition through to docs generation.
+
+


dsinghvi (approved):

packages/cli/fern-definition/schema/src/utils/visitRawTypeReference.ts
```diff
@@ -104,7 +104,9 @@ export function visitRawTypeReference<R>({
         case RawPrimitiveType.boolean:
             return visitor.primitive({
                 v1: PrimitiveTypeV1.Boolean,
-                v2: undefined
+                v2: PrimitiveTypeV2.boolean({
+                    default: _default != null ? (_default as boolean) : undefined
+                })
```
dsinghvi, line 109: @eyw520 do we need to add these v2 constructors to all the `undefined` values above?

eyw520: I've added it for all the ones that currently support them, but there seems to be gaps in the primitiveTypeV2 definitions. For instance, we support defaults for `long` values, but not `float` values.

PR #5492, opened 2024-12-27
fix(go): handle missing delimiter case during stream parse method.

eyw520:
## Description
Fixes an issue where stream delimiter length was included while reading stream data, leading to a `bufio.ErrAdvanceTooFar` error.

## Changes Made
- modify `parse` method.

## Testing


generators/go/internal/generator/sdk/core/stream.go
```diff
@@ -162,11 +162,15 @@ func (s *scannerStreamReader) parse(bytes []byte) (int, []byte, error) {
...
+	if delimIndex < 0 {
+		return len(bytes), data, nil
 	}
-	n := start + len(data) + len(s.options.delimiter)
-	return n, data, nil
+	endIdx := delimIndex + len(s.options.delimiter)
+	parsedData := data[:endIdx]
+	n := start + endIdx
+
+	return n, parsedData, nil
```
amckinney, line 173: nit: Thoughts on updating these variable names a bit so it's more consistent? Now that we're using the `*Index` suffix, we can do the same for `start` -> `startInde` (which would require a couple changes above):

```go
func (s *scannerStreamReader) parse(bytes []byte) (int, []byte, error) {
	var startIndex int
	if s.options != nil && s.options.prefix != "" {
		if i := strings.Index(string(bytes), s.options.prefix); i >= 0 {
			startIndex = i + len(s.options.prefix)
		}
	}
	data := bytes[startIndex:]
	delimIndex := strings.Index(string(data), s.options.delimiter)
	if delimIndex < 0 {
		return len(bytes), data, nil
	}
	endIndex := delimIndex + len(s.options.delimiter)
	parsedData := data[:endIndex]
	n := startIndex + endIndex
	return n, parsedData, nil
}
```

generators/go/internal/generator/sdk/core/stream.go
```diff
@@ -162,11 +162,15 @@ func (s *scannerStreamReader) parse(bytes []byte) (int, []byte, error) {
...
-	if i := strings.Index(string(data), s.options.delimiter); i >= 0 {
-		data = data[:i+len(s.options.delimiter)]
+	delimIndex := strings.Index(string(data), s.options.delimiter)
+	if delimIndex < 0 {
+		return len(bytes), data, nil
 	}
-	n := start + len(data) + len(s.options.delimiter)
-	return n, data, nil
+	endIdx := delimIndex + len(s.options.delimiter)
+	parsedData := data[:endIdx]
```
amckinney, line 170: Do we need to add any additional handling for the fix we made in TypeScript [here](https://github.com/fern-api/fern/pull/5471), or is this working well already?

amckinney (approved):
Looks great! Just a nit on variable naming.

generators/go/internal/generator/sdk/core/stream.go
```diff
@@ -162,11 +162,15 @@ func (s *scannerStreamReader) parse(bytes []byte) (int, []byte, error) {
 		}
 	}
 	data := bytes[start:]
-	if i := strings.Index(string(data), s.options.delimiter); i >= 0 {
-		data = data[:i+len(s.options.delimiter)]
+	delimIndex := strings.Index(string(data), s.options.delimiter)
+	if delimIndex < 0 {
+		return len(bytes), data, nil
```
amckinney, line 167: Also, I think `n` should be set to `start + len(data)` here - `len(bytes)` won't cut it since we need to consume the prefix, too.

generators/go/internal/generator/sdk/core/stream.go
eyw520: Done! Good note.

generators/go/internal/generator/sdk/core/stream.go
eyw520: It is working well -- the logic there ends when no additional delimiter is found.

PR #5494, opened 2024-12-27
fix(cli): resolve bug where user specified examples would be omitted in favor of autogenerated examples.

eyw520:
## Description
Resolve bug in FDR generation where user specified examples were omitted when autogenerated examples were null.

## Changes Made
- Fix logic in `convertService` where null autogenerated examples prevented user specified examples from being parsed.
- Amend `convertHttpEndpointExample`.

## Testing


packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
```diff
@@ -679,7 +678,10 @@ function convertHttpEndpointExample({
             return undefined;
         })
         .filter(isNonNullish);
-    const example = userSpecifiedExample?.example ?? autogeneratedExample.example;
+    const example = userSpecifiedExample?.example ?? autogeneratedExample?.example;
+    if (example === undefined) {
```
dsinghvi, line 682: nit: example == null

packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
```diff
@@ -679,7 +678,10 @@ function convertHttpEndpointExample({
...
+        throw new Error("Example is unexpectedly undefined.");
```
dsinghvi, line 683: we should not be throwing in this case and blocking docs generation -- can we return undefined and handle that upstream or rearchitect the code so that we don't have to throw?

packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
eyw520: fixed.

packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
eyw520: Checks out. Fixed.

packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
```diff
@@ -70,25 +70,27 @@ function convertService(
...
-                        autogeneratedExample,
-                        userSpecifiedExample,
-                        irEndpoint,
-                        ir
-                    })
-                )
+                ...irEndpoint.userSpecifiedExamples
+                    .map((userSpecifiedExample) =>
+                        convertHttpEndpointExample({
+                            userSpecifiedExample,
```
dsinghvi, line 78: can you update this function to take an autogeneratedExample if present -- basically sometimes users can create partial examples and they need to be combined with the autogenerated one

dsinghvi (approved):

packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
eyw520: Done

PR #5499, opened 2024-12-28
fix(cli): Handle OpenAPI type arrays better

Swimburger:
## Description
Improve parsing of OpenAPI schemas with an array in the `type` property.
        * If the array contains `"null"`, it is interpreted as nullable, and removed from the array.
        * If there is only a single item in the array (after removing "null"), it previously defaulted to `unknown`, but now the specified type is used.

## Testing
- [x] Snapshot testing


amckinney (approved):
Discussed offline - it looks like this usage of "null" is now more widely supported, but using an explicit 'nullable' setting is preferred.

https://github.com/OAI/OpenAPI-Specification/issues/3148

Issue #5501, opened 2024-12-29
Multiple showcases


zachkirsch:
### Documentation Page URL

buildwithfern.com/learn

### Issue Description

When you search for showcase there are three different pages

### Steps to Reproduce (Optional)

![IMG_0744](https://github.com/user-attachments/assets/a02f7ff8-890e-4cbb-aecc-547d1b946ab7)

### Suggested Improvement

I think these should be combined or the pages renamed to be more descriptive

PR #5506, opened 2024-12-30
chore(docs): update showcase titles

chdeskur:
Resolving https://github.com/fern-api/fern/issues/5501


dsinghvi (approved):

dsinghvi:
note: we should double check any breaking URLs

PR #5519, opened 2025-01-02
feat(internal): Add turborepo

amckinney:
This migrates this repository to use [turborepo](https://turbo.build/repo/docs). The developer experience remains the exact same - users can simply run `pnpm compile`, `pnpm test`, `pnpm test:update`, and so on. These commands delegate to `turbo` so that results are cached and only the relevants tests are rerun when the user edits particular files.

In particular, this migrates the following scripts, which are included in every `package.json` defined in the workspace. Plus, our CI workflows do not need to be updated at all - we can simply delegate to `pnpm` just like we can locally (re: Turbo's [GitHub Actions guide](https://turbo.build/repo/docs/guides/ci-vendors/github-actions)). Given that the [@fern-api/ete-tests](https://github.com/fern-api/fern/blob/8164037bb6c63bab4cbea1f7dae3aa006a13a52d/packages/cli/ete-tests/package.json#L2) have their own requirements to run (i.e. compiling the CLI from source [here](https://github.com/fern-api/fern/blob/8164037bb6c63bab4cbea1f7dae3aa006a13a52d/package.json#L18)), those scripts remain unchanged.

In my own testing, performance improvements are substantial. We can continue to build upon this and explore other improvements as we go (e.g. [remote caching](https://turbo.build/repo/docs/core-concepts/remote-caching)).

Note that this also updates the `lint-staged.config.js` file so that we only run `prettier` on the files that have changed. This improves the performance of `git commit` substantially.


dsinghvi (approved):
Do we need to update the github workflows to use a turbo token?

amckinney:
> Do we need to update the github workflows to use a turbo token?

To enable remote caching, yes. We can either do that here or in a follow-up. It looks like it's as easy as [this](https://github.com/fern-api/fern-platform/blob/1d25cfcc7d2f62e9c34c0d1d30a47be1f2cf818b/.github/workflows/ci.yml#L9). 

Let me know if/when you add the token (unless it's already configured in this repository, too)!

amckinney:
@dsinghvi Remote caching is enabled now with those environment variables, but we might need to add some additional configuration in Vercel (or perhaps the secret doesn't exist yet). For example, rom [this job](https://github.com/fern-api/fern/actions/runs/12590672443/job/35092598079?pr=5519) you'll notice:

>  WARNING  Insufficient permissions to write to remote cache. Please verify that your role has write access for Remote Cache Artifact at https://vercel.com/docs/accounts/team-members-and-roles/access-roles/team-level-roles?resource=Remote+Cache+Artifact

PR #5532, opened 2025-01-06
fix(cli): add unions to settings.

eyw520:
## Description
Adds the unions config as a valid option to generators.yml

## Changes Made
- Update generators.yml.

## Testing

### DIFF
--- a/generators-yml.schema.json
+++ b/generators-yml.schema.json
@@ -1366,6 +1366,16 @@
             }
           ]
         },
+        "prefer-undiscriminated-unions-with-literals": {
+          "oneOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
         "object-query-parameters": {
           "oneOf": [
             {


dsinghvi (commented):
@eyw520 you'll need to make sure we are reading from this field if populated https://github.com/fern-api/fern/blob/a71e81e62e98f94a69e5a2b9d68487686540f4f6/packages/cli/configuration-loader/src/generators-yml/convertGeneratorsConfiguration.ts#L341

fern/apis/generators-yml/definition/generators.yml
```diff
@@ -222,6 +222,9 @@ types:
       title-as-schema-name: optional<boolean>
       optional-additional-properties: optional<boolean>
       coerce-enums-to-literals: optional<boolean>
+      unions:
```
dsinghvi, line 225: nit: thoughts on naming this `prefer-undiscriminated-unions-with-literals` ? its not the cleanest but more useful than `unions`

eyw520: Agreed

dsinghvi (approved):

PR #5542, opened 2025-01-06
chore(internal): Specify our Google Tag Manager ID in docs.yml

rpc333:
## Description

Now that we support specifying GA/GTM IDs in `docs.yml`, use that for our docs.

## Changes Made

Add `/gtm/container-id` key under `analytics`

## Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "eslint-plugin-react-hooks": "^5.1.0",
     "eslint-plugin-tailwindcss": "^3.17.5",
     "execa": "^5.1.1",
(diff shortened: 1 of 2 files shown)


dsinghvi (approved):

dsinghvi:
@rpc333 looks like the `lint` step is failing cause fern is unintentionally pinned to a version in the [package.json](https://github.com/fern-api/fern/blob/7f7e1ff40d96113fd8a06b0a905a4e543aed7427/package.json#L81). One option is to bump it to the latest version you just released or to have it be `latest`

PR #5551, opened 2025-01-08
feat(typescript): Add DynamicTypeLiteralMapper

amckinney:
This adds the `DynamicTypeLiteralMapper`, which maps dynamic IR types into instances of TypeScript types. With this, we now generate all the method arguments used in the TypeScript client constructor, including:

1. BaseURL
2. Environments (including multi-URL support)
3. All authentication schemes (including arbitrary header types)

This also adds a couple new test cases to verify the behavior of the above use cases, but additional tests will follow in `seed` in the future.

### DIFF
--- a/generators/typescript-v2/dynamic-snippets/src/__test__/multi-url-environment.test.ts
+++ b/generators/typescript-v2/dynamic-snippets/src/__test__/multi-url-environment.test.ts
@@ -126,6 +126,7 @@ describe("multi-url-environment", () => {
                 type: "bearer",
                 token: "<YOUR_API_KEY>"
             },
+            baseURL: undefined,
             environment: {
                 ec2: "https://custom.ec2.aws.com"
             },


generators/browser-compatible-base/src/dynamic-snippets/AbstractDynamicSnippetsGeneratorContext.ts
```diff
@@ -245,6 +245,54 @@ export abstract class AbstractDynamicSnippetsGeneratorContext {
...
+                    this.errors.add({
+                        severity: Severity.Critical,
+                        message: `The provided environments are invalid; got: [${Object.keys(multiEnvironmentUrlValues).join(", ")}], expected: [${keys(firstEnvironment.urls).join(", ")}]`
+                    });
+                    return false;
+                }
+                return true;
+            }
+        }
+    }
```
ajgateno, line 294: Should we handle receiving an unknown environment type?

ajgateno (approved):
Looks good!

generators/browser-compatible-base/src/dynamic-snippets/AbstractDynamicSnippetsGeneratorContext.ts
amckinney: That's actually already prevented since the `switch` is exhaustive - as soon as we add more types, we'll get a compile break (as desired).

PR #5554, opened 2025-01-08
fix(typescript): Fix ESM output

Swimburger:
## Description
The ESM generated by `tsc` isn't compatible with vanilla Node.js in some cases.
Node.js's ESM loader doesn't support directory loading and requires you to add the file extension.

## Changes Made
- After SDK code is generated, ts-morph updates the TS code to add `index.js` to directory imports and adds `.js` to imports where necessary.
- When SDK code is build using `yarn build` the ESM output is patched to use the `mjs` and `d.mts` extension. This is required by Node.js or it will assume the syntax is the same as what is specified in `package.json => type` which is `CommonJS`.
- 

## Testing

### DIFF
--- a/seed/ts-sdk/alias-extends/scripts/rename-to-esm-files.js
+++ b/seed/ts-sdk/alias-extends/scripts/rename-to-esm-files.js
@@ -35,9 +35,13 @@ async function findFiles(rootPath) {
 }
 
 async function updateFiles(files) {
+    const updatedFiles = [];
     for (const file of files) {
-        await updateFileContents(file);
+        const updated = await updateFileContents(file);
+        updatedFiles.push(updated);
     }
+
+    console.log(`Updated imports in ${updatedFiles.length} files.`);
 }
 
 async function updateFileContents(file) {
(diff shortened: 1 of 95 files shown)


generators/typescript/utils/commons/src/scripts/ScriptsManager.ts
```diff
@@ -0,0 +1,15 @@
...
+import path from "path";
+
+import { AbsoluteFilePath } from "@fern-api/fs-utils";
+
+import { renameToEsmFilesFile } from "./RenameToEsmFilesFile";
+
+export class ScriptsManager {
+    public async copyScripts({ pathToRoot }: { pathToRoot: AbsoluteFilePath }): Promise<void> {
+        // make sure the scripts directory exists
+        const scriptsDir = path.join(pathToRoot, "scripts");
```
amckinney, line 11: Should we use the new `SCRIPTS_DIRECTORY_NAME` constant here? If it isn't natural to import from that package, it's ok to leave this as-is.

dsinghvi (approved):

amckinney (approved):

PR #5559, opened 2025-01-08
feat(cli): `fern check` now runs simple rules on openapi specs & overrides.

eyw520:
## Description
Fern check now validates spec workspaces and throws errors when distinct overrides resolve to the same SDK method & group name permutation.

## Changes Made
- Define new `oss-validator` package and Rule schema, with initial rule `no-duplicate-overrides`.
- Refactor `OpenAPILoader` to parse out helper utils.
- Define test helpers and initial simple fixtures.

## Testing


packages/cli/workspace/oss-validator/src/ValidationViolation.ts
```diff
@@ -0,0 +1,9 @@
+import { NodePath } from "@fern-api/fern-definition-schema";
+import { RelativeFilePath } from "@fern-api/fs-utils";
+
+export interface ValidationViolation {
+    severity: "error" | "warning";
+    relativeFilepath: RelativeFilePath;
+    nodePath: NodePath;
+    message: string;
+}
```
dsinghvi, line 9: worth centralizing in a follow up

packages/cli/workspace/oss-validator/src/rules/no-duplicate-overrides/__test__/no-duplicate-overrides.test.ts
```diff
@@ -0,0 +1,30 @@
...
+            ),
+            cliVersion: "0.1.3-rc0"
+        });
+
+        const expectedViolations: ValidationViolation[] = [
+            {
+                severity: "error",
+                relativeFilepath: RelativeFilePath.of("openapi/openapi.yml"),
+                nodePath: ["paths", "/a/b", "get"],
+                message: "Duplicate SDK method found: group 'a' with method 'b'"
```
dsinghvi, line 24: It might be a bit cleaner to do `Encountered duplicate x-fern-sdk-method-name users.create`

dsinghvi: `Type {name of type} already exists`
`SDK method users.get already exists (x-fern-sdk-group-name: users, x-fern-sdk-method-name: get)`

dsinghvi (approved):
🚀 Excited to get this in! 

Two things we chatted about offline: 
1) Adding `snapshots` for the test case
2) Updating error message

PR #5569, opened 2025-01-09
chore(internal): Add new "fatal" level for violations, freeing up "error" for new uses

rpc333:
## Description

Instead of having 2 levels of violation severity (`"error" | "warning"`), have 3 (`"fatal" | "error" | "warning"`) with existing "error"-level violations being bumped up to "fatal".

This frees up the "error" level to be used for things that we want to always print to console etc but not necessarily exit with non-zero codes on. First use case for this will be broken links, but there will undoubtedly be others. 

## Changes Made

- Add new "fatal" severity level
- Upgrade all existing "error" references to "fatal" 
- Update severity handling logic in a few places

## Testing


amckinney (approved):
Nice 👏

PR #5579, opened 2025-01-10
feat(cli): IR uses all request properties from OAuth getToken to support custom schemas.

eyw520:
## Description
This PR updates the IR to include additional customProperties (beyond `scopes`, `clientId`, `clientSecret`) to better support custom OAuth getToken endpoints.

## Changes Made
- Update the IR `auth.yml` schema to support customProperties.
- Ingest all request body properties from the getToken endpoint into IR.

## Testing


packages/cli/generation/ir-generator/src/converters/convertOAuthTokenEndpoint.ts
```diff
@@ -22,6 +22,19 @@ export function convertOAuthTokenEndpoint({
...
+        typeof resolvedEndpoint.endpoint.request === "object" &&
+        resolvedEndpoint.endpoint.request.body != null &&
+        typeof resolvedEndpoint.endpoint.request.body === "object" &&
+        "properties" in resolvedEndpoint.endpoint.request.body
+            ? (resolvedEndpoint.endpoint.request.body.properties ?? {})
+            : {};
+
+    const customPropertyNames = Object.keys(requestBodyProperties).filter(
+        (propertyName) => propertyName !== "client_id" && propertyName !== "client_secret" && propertyName !== "scopes"
+    );
```
amckinney, line 36: I think we'll need to do some additional logic here - it's possible that the user's property might not actually match this convention (e.g. `clientId`, `clientSecret`, `clientSECRET`, etc).

We'll need to come up with a way to actually recognize these parameters in particular, ideally from the configured `clientId`, `clientSecret`, and `scopes` settings.

eyw520: Fixed -- we now use the actual Oauth configuration, which specifies potential aliases for `clientId`, `clientSecret`, etc.

packages/cli/generation/ir-generator/src/converters/convertOAuthTokenEndpoint.ts
```diff
@@ -22,6 +22,23 @@ export function convertOAuthTokenEndpoint({
...
+            ? (resolvedEndpoint.endpoint.request.body.properties ?? {})
+            : {};
+
+    const customPropertyNames = Object.keys(requestBodyProperties).filter(
+        (propertyName) =>
+            !tokenEndpoint.requestProperties.client_id.includes(propertyName) &&
+            !tokenEndpoint.requestProperties.client_secret.includes(propertyName) &&
+            (tokenEndpoint.requestProperties.scopes == null ||
+                !tokenEndpoint.requestProperties.scopes.includes(propertyName))
+    );
```
amckinney, line 40: nit: It might be worth adding a helper function for this so that we can just call it within the `return` statement below.

Something like:

```typescript
return {
  ...
  requestProperties: {
    ...
    customProperties: resolveCustomRequestProperties(...)
  }
}
```

amckinney (approved):
Looks good overall! It'd be great if we had a test that demonstrated how the IR changes with this addition. 

Maybe we could modify one of the `test-definitions` to include an endpoint with custom properties so that we can see that they actually get plugged in the way we expect?

Stamping to unblock, assuming the testing goes well.

packages/cli/generation/ir-generator/src/converters/convertOAuthTokenEndpoint.ts
eyw520: DOne

packages/ir-sdk/fern/apis/ir-types-latest/changelog/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [v54.0.1] - 2024-01-10
```
dsinghvi, line 9: nit: i would make this a minor 54.1.0

PR #5580, opened 2025-01-10
chore(cli): Add changelog entry for broken link checks

rpc333:
Forgot to add changelog entry in #5068 before merge!

### DIFF
--- a/fern/pages/changelogs/cli/2025-01-10.mdx
+++ b/fern/pages/changelogs/cli/2025-01-10.mdx
@@ -1,5 +1,7 @@
 ## 0.47.2
-**`(feat):`** The CLI now supports checking for broken links in your docs. You will see warnings in `fern check` if your docs link to any page that can't be resolved, and
-you can also run `fern docs broken-links` to only check for broken links.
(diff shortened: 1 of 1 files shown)


packages/cli/cli/versions.yml
```diff
@@ -1,3 +1,11 @@
+- changelogEntry:
+  - summary: |
+      The CLI now supports checking for broken links in your docs. You will see warnings in `fern check` if your docs link to any page that can't be resolved, and
+      you can also run `fern docs broken-links` to only check for broken links.
+    type: feat
```
dsinghvi, line 5: @rpc333 no pressure -- but would be great to document all the different CLI commands you showed or link to a page in the documentation for this

dsinghvi (approved):

packages/cli/cli/versions.yml
rpc333: @dsinghvi I added some extra info to changelog + updated the documentation, but the latter in particular might want some extra attention as a follow-up based on conversation @abvthecity @amckinney and I were having in Slack

PR #5587, opened 2025-01-12
fix(typescript): Write valid version.ts file

Swimburger:
## Description
The output of the `version.ts` file was `[object Object]`.
This change makes it `export const SDK_VERSION = "0.0.1";`.

## Changes Made

- Rename `.toString` to `.toStringSync` as all objects have a sync `.toString` function by default.
- Rename `.toString` to `.toStringAsync`.
- Add `.toString` to writer

## Testing

- [x] Seed


dsinghvi (approved):
cc @amckinney on the toStringSync vs toStringAsync naming

PR #5598, opened 2025-01-13
fix(typescript): Use serializer for multipart form data when noSerde: false

Swimburger:
## Description
- Use Serde to convert types to mapped value, and JSON stringify if not already a string.
- Use Serde for named types and array/set of named types in query strings, except for aliases 
  to primitive types.
  If the named type is an alias of the primitive type, use the primitive type's stringify.

This fixes multipart file uploads with JSON parts when serde is enabled.

## Testing
- [x] Seed

### DIFF
--- a/generators/typescript/model/type-reference-converters/package.json
+++ b/generators/typescript/model/type-reference-converters/package.json
@@ -29,7 +29,6 @@
     "@fern-fern/ir-sdk": "53.23.0",
     "@fern-api/core-utils": "workspace:*",
     "@fern-typescript/commons": "workspace:*",
-    "@fern-typescript/resolvers": "workspace:*",
     "@fern-typescript/contexts": "workspace:*",
     "ts-morph": "^15.1.0"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2296,9 +2296,6 @@ importers:
       '@fern-typescript/contexts':
         specifier: workspace:*
         version: link:../../utils/contexts
-      '@fern-typescript/resolvers':
-        specifier: workspace:*
-        version: link:../../utils/resolvers
       ts-morph:
         specifier: ^15.1.0
         version: 15.1.0


generators/typescript-v2/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
```diff
@@ -237,10 +237,6 @@ async function main() {
     const client = new AcmeClient();
     await client.service.post({
         file: new Blob(["Hello, world!"]),
-        fileList: [
-            new Blob(["First"]),
-            new Blob(["Second"]),
-        ],
```
amckinney, line 243: I'm surprised by this one since this should just use the `@fern-api/typescript-ast` package. If the change isn't obvious, I can take a closer look in a follow-up.

generators/typescript/express/generator/src/ExpressGenerator.ts
```diff
@@ -564,7 +564,9 @@ export class ExpressGenerator {
             includeSerdeLayer: this.config.includeSerdeLayer,
             retainOriginalCasing: this.config.retainOriginalCasing,
             useBigInt: this.config.useBigInt,
-            enableInlineTypes: false
+            enableInlineTypes: false,
+            allowExtraFields: this.config.allowExtraFields,
+            omitUndefined: false
```
amckinney, line 569: Is `this.config.omitUndefined` not available here?

generators/typescript/model/type-reference-converters/src/TypeReferenceToStringExpressionConverter.ts
```diff
@@ -202,6 +255,12 @@ export class TypeReferenceToStringExpressionConverter extends AbstractTypeRefere
         );
     }
 
+    /**
+     * Example:
+     * ```ts
+     * typeof REFERENCE === "string" ? REFERENCE : JSON.stringify(REFERENCE)
+     * ```
+     */
```
amckinney, line 263: Good call on adding these 👏

packages/cli/generation/ir-generator-tests/src/dynamic-snippets/__test__/test-definitions/file-upload.json
```diff
@@ -514,7 +514,7 @@
                         {
                             "type": "fileArray",
                             "name": {
-                                "originalName": "fileList",
+                                "originalName": "file_list",
```
amckinney, line 517: Ahh this explains the dynamic snippets failure - would you mind updating the `DynamicSnippetsGenerator.test.ts` case so that the provided parameter matches? Updating the wire name changes the behavior.

amckinney (approved):
lgtm!

generators/typescript/express/generator/src/ExpressGenerator.ts
Swimburger: No, it's not part of the express config. We can add it in the future if we ever need it.
I defaulted to the default value that is in the SDK config.

PR #5601, opened 2025-01-14
feat(cli): Add support for nullable<T>

amckinney:
This adds support for a new Fern definition keyword, `nullable`. With this, users can explicitly specify whether or not a type is allowed to be serialized as `null`, which is valid for both required and `optional` properties.

With this, users can specify APIs like the following:

```yaml
types:
  User:
    properties:
      name: string
      email: nullable<string>
      createdAt: optional<string>
      updatedAt: optional<string>

service:
  auth: false
  base-path: /
  endpoints:
    createUser:
      method: POST
      path: /users
      request:
        name: CreateUserRequest
        body:
          properties:
            name: string
            email: nullable<string>
      response: User
```

A few things to note:
1. The `nullable` type is migrated to `optional` whenever the target generator doesn't support the latest IR.
2. A follow-up will introduce an opt-in OpenAPI setting that converts `nullable` OpenAPI properties to `nullable`.
    * By default, these properties are converted to `optional`.
3. In another follow-up the dynamic snippet generators will be updated to allow for explicit `null` values.

### DIFF
--- a/fern/pages/changelogs/cli/2025-01-14.mdx
+++ b/fern/pages/changelogs/cli/2025-01-14.mdx
@@ -1,5 +1,5 @@
 ## 0.48.0
-**`(feat):`** Adds support for nullable<T> types in the Fern definition, such as the following:
+**`(feat):`** Adds support for nullable types in the Fern definition, such as the following:
 
 ```yaml
 types:
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,6 +1,6 @@
 - changelogEntry:
   - summary: |
-      Adds support for nullable<T> types in the Fern definition, such as the following:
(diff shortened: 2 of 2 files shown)


packages/cli/generation/ir-generator/src/examples/validateTypeReferenceExample.ts
```diff
@@ -488,6 +503,15 @@ function areResolvedTypesEquivalent({ expected, actual }: { expected: ResolvedTy
                             ? actual.container.itemType
                             : actual
                 });
+            case "nullable":
+                // special case: if expected is a nullable but actual is not, that's okay
```
dsinghvi, line 507: nice

dsinghvi (approved):

PR #5603, opened 2025-01-14
feat(docs): document LLMs.txt and LLMs-full.txt

dannysheridan:
Add this feature to Fern's documentation: https://github.com/fern-api/fern-platform/pull/1828


chdeskur:
@dannysheridan - looks like the image path needs to be resolved so a preview can generate

PR #5611, opened 2025-01-14
feat(openapi): Add respect-nullable-schemas config

amckinney:
This adds a new OpenAPI converter setting to preserve `nullable` properties in the generated SDKs. By default, `nullable` properties are mapped to `optional` (which nearly gives us the same effect), but this allows for a more explicit setting that distinguishes whether or not an explicit `null` is valid. Plus, in practice it's possible that a nullable property is actually `required`.

For backwards compatibility, this setting is opt-in and can be enabled like so:

```yaml
api:
  specs:
    - openapi: ./path/to/openapi.yml
      settings:
        respect-nullable-schemas: true
```


dsinghvi (approved):
snapshots look great to me

PR #5612, opened 2025-01-14
feat(typescript): Add support explicit null values

amckinney:
Now that the Fern definition supports `nullable` properties (re: #5601), this upgrades the TypeScript generator to IRv55 to support this feature. For example, consider the following Fern definition:

```yaml
types:
  User:
    properties:
      name: string
      tags: nullable<list<string>>
      metadata: optional<nullable<Metadata>>

  Metadata:
    properties:
      createdAt: datetime
      updatedAt: datetime
      avatar: nullable<string>
      activated: optional<nullable<boolean>>
```

The generated TypeScript types look like the following:

```typescript
export interface User {
    name: string;
    tags: string[] | null;
    metadata?: Metadata | null;
}

export interface Metadata {
    createdAt: Date;
    updatedAt: Date;
    avatar: string | null;
    activated?: boolean | null;
}
```

Note that this also temporarily disables the dynamic snippet tests so that we can more confidently test the TypeScript SDK generator until example generation is fixed.

### DIFF
--- a/fern/pages/changelogs/cli/2025-01-15.mdx
+++ b/fern/pages/changelogs/cli/2025-01-15.mdx
@@ -1,4 +1,7 @@
 ## 0.50.0
 **`(internal):`** The CLI is capable of migrating the latest TypeScript generator to IRv55.
 
+## 0.49.1
+**`(fix):`** The OpenAPI v2 parser now supports `x-fern-global-headers` and fixes an issue with generating webhook content.
+
 
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -6,6 +6,13 @@
   version: 0.50.0
 
 - changelogEntry:
+  - summary: |
+      The OpenAPI v2 parser now supports `x-fern-global-headers` and fixes an issue with generating webhook content.
+    type: fix
+  irVersion: 55
+  version: 0.49.1
+
+- changelogEntry:
     - summary: |
         The OpenAPI importer now supports respecting nullable properties in schemas. When enabled, nullable properties will
         be preserved in the gnerated SDK. By default (without this setting), nullable properties are treated as `optional`.
--- a/packages/cli/docs-resolver/package.json
+++ b/packages/cli/docs-resolver/package.json
@@ -31,7 +31,7 @@
     "@fern-api/configuration-loader": "workspace:*",
     "@fern-api/core-utils": "workspace:*",
     "@fern-api/docs-markdown-utils": "workspace:*",
-    "@fern-api/docs-parsers": "^0.0.26",
+    "@fern-api/docs-parsers": "^0.0.37",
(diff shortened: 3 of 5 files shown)


generators/typescript/sdk/generator/src/TemplateGenerator.ts
```diff
@@ -638,6 +642,16 @@ export class TemplateGenerator {
                       })
                     : undefined;
             },
+            nullable: (optionalType) =>
+                this.getTemplateFromTypeReference({
```
Swimburger, line 646: `optionalType` => `nullableType`

generators/typescript/sdk/client-class-generator/src/endpoints/default/endpoint-response/GeneratedThrowingEndpointResponse.ts
```diff
@@ -405,6 +408,7 @@ export class GeneratedThrowingEndpointResponse implements GeneratedEndpointRespo
                     literal: () => undefined,
                     list: () => undefined,
                     set: () => undefined,
+                    nullable: (optType) => this.getDefaultPaginationValue({ type: optType }),
```
Swimburger, line 411: `optType` => `nullableType`

seed/ts-sdk/nullable/src/api/resources/nullable/client/Client.ts
```diff
@@ -0,0 +1,206 @@
...
+     */
+    public async getUsers(
+        request: SeedNullable.GetUsersRequest = {},
+        requestOptions?: Nullable.RequestOptions,
+    ): Promise<SeedNullable.User[]> {
+        const { usernames, avatar, activated, tags } = request;
+        const _queryParams: Record<string, string | string[] | object | object[] | null> = {};
+        if (usernames != null) {
+            if (Array.isArray(usernames)) {
+                _queryParams["usernames"] = usernames.map((item) => item);
```
Swimburger, line 53: unrelated, but are we creating a copy of an array here? I think there are simpler ways to do this.

Swimburger (approved):

generators/typescript/sdk/client-class-generator/src/endpoints/default/endpoint-response/GeneratedThrowingEndpointResponse.ts
amckinney: Thanks for catching these 👏

seed/ts-sdk/nullable/src/api/resources/nullable/client/Client.ts
amckinney: Yea it looks like it - I have a feeling this is because we always map over the array values and (optionally) perform a `string` mapper function (e.g. `item.toString()`), so it's simpler to _always_ use the `.map` expression.

We can take a closer look here though.

PR #5613, opened 2025-01-14
feat(cli): specify openrpc in docs.yml

dsinghvi:
## Description
This PR allows you to specify an `openrpc` document to render as part of an API section in `docs.yml`. 

## Changes Made
- Update `docs.yml` schema to include openrpc document
-`fern check` rule which validates that the openrpc document exists
- DocsResolver now supports reading `openrpc` and translating it into fdr latest

**Note**: I discovered that you can't have `apisV2` and `apis` both at the same time so local preview does not entirely work. That's okay for now since this is an initial version. 

## Testing
- [] Unit tests added/updated

### DIFF
--- a/docs-yml.schema.json
+++ b/docs-yml.schema.json
@@ -1437,6 +1437,16 @@
             }
           ]
         },
+        "openrpc": {
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
         "audiences": {
           "oneOf": [
             {


packages/cli/configuration/src/docs-yml/schemas/sdk/api/resources/docs/types/ApiReferenceConfiguration.ts
abvthecity, line 1: Nit: just use availability tag?

abvthecity (approved):

amckinney (approved):

packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
```diff
@@ -593,6 +594,35 @@ export class DocsDefinitionResolver {
         item: docsYml.DocsNavigationItem.ApiSection,
         parentSlug: FernNavigation.V1.SlugGenerator
     ): Promise<FernNavigation.V1.ApiReferenceNode> {
+        if (item.openrpc != null) {
+            const absoluteFilepathToOpenrpc = resolve(
+                this.docsWorkspace.absoluteFilePath,
+                RelativeFilePath.of(item.openrpc)
+            );
+            if (!(await doesPathExist(absoluteFilepathToOpenrpc))) {
+                throw new Error(`OpenRPC file does not exist at path: ${absoluteFilepathToOpenrpc}`);
```
abvthecity, line 603: should we use the logger to throw these errors?

packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
```diff
@@ -593,6 +594,35 @@ export class DocsDefinitionResolver {
...
+            }
+            const api = await generateFdrFromOpenrpc(absoluteFilepathToOpenrpc, this.taskContext);
```
abvthecity, line 605: it's probably ok that we're doing this inline but i think it would be better for performance if we could push this step upstream.

packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
dsinghvi: so i have a fern check rule for this, so it should really never get here cause we only run when fern check succeeds

dsinghvi: The CLI also catches errors and logs with error level

packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
dsinghvi: I see -- is the case we would be optimizing for if the same openrpc file is used multiple times? or are there other situations?

PR #5622, opened 2025-01-15
fix(typescript): Handle null query parameters

amckinney:
This fixes an issue where nullable query parameters were not null-safe in their method invocations. We now appropriately guard against `null` values like so:

```typescript
const _queryParams: Record< ... >;
if (value !== undefined) {
    _queryParams["value"] = value?.toString() ?? null;
}
```

Note that this also updates the IRv55 -> IRv54 migration so that we prevent double optional type references that were created from types specified as `optional<nullable<T>>`.

### DIFF
--- a/generators/typescript/sdk/CHANGELOG.md
+++ b/generators/typescript/sdk/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ```typescript
   const _queryParams: Record< ... >;
   if (value !== undefined) {
-      _queryParams["value"] = value?.toString ?? null;
+      _queryParams["value"] = value?.toString() ?? null;
   }
   ```
 


seed/ts-sdk/nullable/src/api/resources/nullable/client/Client.ts
```diff
@@ -76,6 +77,10 @@ export class Nullable {
             }
         }
 
+        if (extra !== undefined) {
+            _queryParams["extra"] = extra?.toString ?? null;
+        }
```
Swimburger, line 82: I thought it was a typo in the description, but I think `.toString` should be invoked.

amckinney: Oof, surprised this works at all then - taking a closer look.

Swimburger (approved):

PR #5623, opened 2025-01-16
fix(typescript): Fix null value parsing

amckinney:
This fixes a couple issues with supporting `null` values in the generated serialization layer. We now roll out a separate `optionalNullable` option to simplify the zod-like generated schemas. This also fixes an issue that prevented `null` values from being serialized in record/map types.

### DIFF
--- a/seed/ts-express/alias-extends/core/schemas/builders/schema-utils/getSchemaUtils.ts
+++ b/seed/ts-express/alias-extends/core/schemas/builders/schema-utils/getSchemaUtils.ts
@@ -120,7 +120,7 @@ export function optionalNullable<Raw, Parsed>(
                     value: undefined,
                 };
             }
-            if (raw == null) {
+            if (raw === null) {
                 return {
                     ok: true,
                     value: null,
@@ -135,7 +135,7 @@ export function optionalNullable<Raw, Parsed>(
                     value: undefined,
                 };
             }
-            if (parsed == null) {
+            if (parsed === null) {
                 return {
                     ok: true,
                     value: null,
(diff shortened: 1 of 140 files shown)


generators/typescript/utils/core-utilities/zurg/src/builders/schema-utils/getSchemaUtils.ts
```diff
@@ -107,6 +109,49 @@ export function optional<Raw, Parsed>(
...
+): Schema<Raw | null | undefined, Parsed | null | undefined> {
+    const baseSchema: BaseSchema<Raw | null | undefined, Parsed | null | undefined> = {
+        parse: (raw, opts) => {
+            if (raw === undefined) {
+                return {
+                    ok: true,
+                    value: undefined
+                };
+            }
+            if (raw == null) {
```
Swimburger, line 123: nit: use `raw === null`

generators/typescript/utils/core-utilities/zurg/src/builders/schema-utils/getSchemaUtils.ts
```diff
@@ -107,6 +109,49 @@ export function optional<Raw, Parsed>(
...
+            return schema.parse(raw, opts);
+        },
+        json: (parsed, opts) => {
+            if (parsed === undefined) {
+                return {
+                    ok: true,
+                    value: undefined
+                };
+            }
+            if (parsed == null) {
```
Swimburger, line 138: nit: use `raw === null`

Swimburger (approved):

PR #5626, opened 2025-01-16
feat(typescript): Enhance support for bigint

Swimburger:
## Description
When `useBigInt` SDK configuration is set to `true`, a customized JSON serializer & deserializer is used that will preserve the precision of `bigint`'s, as opposed to the native `JSON.stringify` and `JSON.parse` function which converts `bigint`'s to `number`'s losing precision.

  When combining `useBigInt` with our serialization layer (`no-serde: false` (default)), both the request and response properties that are marked as `long` and `bigint` in OpenAPI/Fern spec, will consistently be `bigint`'s.
  However, when disabling the serialization layer (`no-serde: true`), they will be typed as `number | bigint`.

  Here's an overview of what to expect from the generated types when combining `useBigInt` and `noSerde` with the following Fern definition:

  **Fern definition**
  ```yml
  types:
    ObjectWithOptionalField:
      properties:
        longProp: long
        bigIntProp: bigint
  ```

  **TypeScript output**
  ```typescript
  // useBigInt: true
  // noSerde: false
  interface ObjectWithLongAndBigInt {
    longProp: bigint;
    bigIntProp: bigint;
  }

  // useBigInt: true
  // noSerde: true
  interface ObjectWithLongAndBigInt {
    longProp: bigint | number;
    bigIntProp: bigint | number;
  }

  // useBigInt: false
  // noSerde: false
  interface ObjectWithLongAndBigInt {
    longProp: number;
    bigIntProp: string;
  }

  // useBigInt: false
  // noSerde: true
  interface ObjectWithLongAndBigInt {
    longProp: number;
    bigIntProp: string;
  }
  ```

Here's an explanation of what happens under the hood:
* If `useBigInt: true` + `noSerde: false`:
  * convert object to JSON:
    1. user-friendly JS object is mapped to raw (bigint => bigint)
    2. raw is converted to JSON using `JSON.stringify` with replacer function that converts bigint to string, suffixed with `n`.
    3. JSON string is replaced using regex to remove the quotes around the bigint and remove the `n` suffix.
    The result is valid JSON with untruncated bigint's that are not strings.

  * convert JSON to object:
    1. Add quotes and `n` suffix to ints that are big using regex
    2. Use `JSON.parse` with reviver function that converts bigint strings to bigint's.
       **Important note**: props that can be bigints will be deserialized as `number` if they aren't too big.
    3. Raw object is mapped to user-friendly JS object (number | bigint => bigint).
    The result is users always receive a bigint, regardless of the size of the integer.

* If `useBigInt: true` + `noSerde: true`:
  * convert object to JSON:
    1. object is converted to JSON using `JSON.stringify` with replacer function that converts bigint to string, suffixed with `n`.
    2. JSON string is replaced using regex to remove the quotes around the bigint and remove the `n` suffix.
    The result is valid JSON with untruncated bigint's that are not strings.

  * convert JSON to object:
    1. Add quotes and `n` suffix to ints that are big using regex
    2. Use `JSON.parse` with reviver function that converts bigint strings to bigint's.
       **Important note**: props that can be bigints will be deserialized as `number` if they aren't too big.
    The result is users receive a number or bigint, depending on the size of the integer.

## Testing
- [x] Seed test

TODO: fix dynamic snippets. I tried to update the code for dynamic snippets, but it wouldn't have effect on the output.

### DIFF
--- a/generators/typescript-v2/ast/src/ast/TypeLiteral.ts
+++ b/generators/typescript-v2/ast/src/ast/TypeLiteral.ts
@@ -130,8 +130,7 @@ export class TypeLiteral extends AstNode {
                     return;
                 }
                 writer.write(`"${this.internalType.value.toString()}"`);
-                    return;
-                }
+                return;
                 break;
             }
             case "datetime": {


generators/typescript-v2/ast/src/ast/TypeLiteral.ts
```diff
@@ -123,15 +125,17 @@ export class TypeLiteral extends AstNode {
                 break;
             }
             case "bigint": {
-                if (writer.customConfig?.noSerdeLayer) {
-                    writer.writeNode(TypeLiteral.string(this.internalType.value.toString()));
+                if (useBigInt === true) {
+                    writer.write(`BigInt(${this.internalType.value.toString()})`);
+                    return;
+                } else {
+                    writer.write(`"${this.internalType.value.toString()}"`);
```
amckinney, line 132: nit:
```suggestion
                if (useBigInt) {
                    writer.write(`BigInt(${this.internalType.value.toString()})`);
                    return;
                }
                writer.write(`"${this.internalType.value.toString()}"`);
```

generators/typescript/sdk/CHANGELOG.md
```diff
@@ -5,6 +5,55 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.48.0] - 2025-01-16
+
+- Feat: When `useBigInt` SDK configuration is set to `true`, a customized JSON serializer & deserializer is used that will preserve the precision of `bigint`'s, as opposed to the native `JSON.stringify` and `JSON.parse` function which converts `bigint`'s to `number`'s losing precision.
```
amckinney, line 10: Thanks for writing this up, looks great 👏

amckinney (approved):

PR #5637, opened 2025-01-16
fix(java): Fix union type name conflict resolution

ajgateno:
This PR makes it so that a union called Visitor gets a Visitor called Visitor_ and adds resolution to undiscriminated unions to make sure their variants don't clash with the main class name.


generators/java/model/versions.yml
```diff
@@ -1,3 +1,10 @@
+- changelogEntry:
+    - summary: |
+        Fix union inline type name conflit resolution.
```
Swimburger, line 3: `conflit` typo :)

amckinney (approved):

Issue #5638, opened 2025-01-16
link checker reports good links as broken


cwcowell:
### CLI Version (from `fern.config.json`)

0.50.5

### Minimal API Specification

n/a

### Actual UI Behavior

When we run `fern docs dev` or `fern docs broken-links` it reports many links as broken when they in fact work fine.

For example, this `href` is reported as broken, even though it works fine when we click this rendered card.

```
  <Card
    title="Release notes"
    icon="fa-solid fa-newspaper"
    href="release-notes"
  >
    Check out the latest AI Hub features, improvements, and fixes.
  </Card>
```

Could it be because the `href` value here is the page’s slug and not its absolute or relative path in the filesystem?

Here’s the frontmatter of the page that the card links to.

```
---
title: Release notes
description: Important release updates for Instabase AI Hub.
slug: release-notes
---
```

The linked-to page lives at `fern/docs/pages/releases/releasenotes.mdx` in the repo.

### Expected UI Behavior

this link (and other similar links) should not be flagged as broken

dsinghvi:
@cwcowell would you be able to give the latest CLI a try https://github.com/fern-api/fern/pull/5642 and see if you run into issues. I tested on your spec but didn't have access to the underlying OpenAPI spec, so was only able to partially test.

cwcowell:
@dsinghvi Yes, all the links reported as broken now are truly broken. Thank you for fixing!

PR #5642, opened 2025-01-17
fix(cli): reduce false positives from broken link checker

dsinghvi:
## Description
This PR reduces the false positives from the broken link checker. Addresses https://github.com/fern-api/fern/issues/5638

## Changes Made
- whenever checking in `pageSlugs` we should make sure that we removeLeadingSlash. There should really be a class that controls access to remove the burden from the consumer. 
- 

## Testing

### DIFF
--- a/packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
+++ b/packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
@@ -42,7 +42,7 @@ export async function checkIfPathnameExists({
 }): Promise<true | string[]> {
     const slugs = absoluteFilepath != null ? (absoluteFilePathsToSlugs.get(absoluteFilepath) ?? []) : [];
 
-    // base case: empty pathname is valid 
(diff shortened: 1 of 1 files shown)


packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
```diff
@@ -82,13 +81,11 @@ export async function checkIfPathnameExists({
...
 
     // if that fails, we need to check if the path exists against all of the slugs for the current file
-
     const brokenSlugs: string[] = [];
     for (const slug of slugs) {
         const url = new URL(`/${slug}`, wrapWithHttps(baseUrl.domain));
         const targetSlug = withRedirects(new URL(pathname, url).pathname, baseUrl, redirects);
-
-        if (!pageSlugs.has(targetSlug)) {
+        if (!pageSlugs.has(removeLeadingSlash(targetSlug))) {
```
dsinghvi, line 88: critical change

eyw520 (approved):

abvthecity (approved):

PR #5644, opened 2025-01-17
fix(go): Support special literal values in undiscriminated unions

amckinney:
This fixes an issue where special literal values that represent invalid Go identifiers (e.g. "$foo") weren't generated correctly in undiscriminated unions. This updates the generated code so that we map the identifier into something that is valid, and also export the properties so that the undiscriminated union is easier to use.

Along this way, this also fixes an issue where custom names were not respected with the basic auth scheme.


dsinghvi (approved):
blind +1

PR #5652, opened 2025-01-19
chore(cli): validate links with trailing slashes

chdeskur:


dsinghvi (approved):
youll need to bump the version as well since i just released 0.50.9

PR #5672, opened 2025-01-20
feat(typescript): Add support for explicit null in snippets

amckinney:
This updates the dynamic snippet generators to support the new `nullable` container type introduced in #5601. With this, the dynamic snippets API will render `null` values when the property is `nullable` like so:

```typescript

async function main() {
    const client = new AcmeClient({
        environment: "https://api.example.com",
    });
    await client.users.create({
        username: "john.doe",
        tags: [
            "admin",
        ],
        metadata: {
            avatar: null,
            activated: null,
        },
        avatar: null,
    });
}
main();
```


Swimburger (approved):
TS part LGTM

PR #5688, opened 2025-01-21
fix(typescript): Support basic auth environment variables

amckinney:
This fixes the TypeScript generator to support custom environment variables for basic auth. Now, if auth is required by the SDK, we throw an error if either of the settings are not configured.

### DIFF
--- a/seed/ts-express/ts-inline-types/api/types/UndiscriminatedUnion1.ts
+++ b/seed/ts-express/ts-inline-types/api/types/UndiscriminatedUnion1.ts
@@ -19,9 +19,6 @@ export type UndiscriminatedUnion1 =
     | SeedObject.UndiscriminatedUnion1DiscriminatedUnion1
     /**
      * lorem ipsum */
-    | SeedObject.UndiscriminatedUnion1DiscriminatedUnion1
-    /**
-     * lorem ipsum */
     | SeedObject.UndiscriminatedUnion1InlineEnum1
     /**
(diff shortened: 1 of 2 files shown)


Swimburger (approved):

seed/ts-sdk/basic-auth-environment-variables/src/api/resources/basicAuth/client/Client.ts
```diff
@@ -198,9 +198,23 @@ export class BasicAuth {
...
+        if (username == null) {
+            throw new errors.SeedBasicAuthEnvironmentVariablesError({
+                message: "Please specify USERNAME when instantiating the client.",
+            });
+        }
+
+        const accessToken = (await core.Supplier.get(this._options.accessToken)) ?? process?.env["PASSWORD"];
+        if (accessToken == null) {
+            throw new errors.SeedBasicAuthEnvironmentVariablesError({
+                message: "Please specify PASSWORD when instantiating the client.",
```
dsinghvi, line 211: nit: `Please specify a password by either passing it in to the constructor or initializing a PASSWORD environment variable`

Something like this ^ would be a better error message i think

amckinney: Sweet, I just pattern matched on the original message we used for bearer. We can update across the board.

PR #5695, opened 2025-01-21
fix(java): Prevent accept header from being removed in wrapped requests

ajgateno:
Previously the order in which headers were being appended in wrapped requests meant we were clearing the "Accept" header. This PR fixes that.


dsinghvi (approved):
nit: you probably dont have to release the model generator!

PR #5705, opened 2025-01-22
fix(typescript): Fix snippets and examples with BigInt

Swimburger:
## Testing

- [x] Seed
- [x] Snapshot


dsinghvi (approved):
diff looks good to me!

PR #5708, opened 2025-01-22
fix(java): Fix usage of cursor pagination causing endpoint return type to be void

ajgateno:
This PR addresses a use case where a complex return type will make cursor paginated endpoints return void. PR description update pending more investigation.


generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/AbstractEndpointWriter.java
```diff
@@ -890,82 +918,111 @@ public com.fern.ir.model.types.TypeReference _visitUnknown(Object o) {
...
+                        com.fern.ir.model.types.TypeReference numberType = pageType.getContainer()
+                                .map(containerType -> containerType.visit(new ContainerTypeToUnderlyingType()))
+                                .orElse(pageType);
+                        TypeName numberTypeName =
+                                clientGeneratorContext.getPoetTypeNameMapper().convertToTypeName(true, numberType);
+
+                        String one = INTEGER_ONE;
+                        if (numberTypeName.equals(TypeName.FLOAT) || numberTypeName.equals(TypeName.DOUBLE)) {
+                            one = DECIMAL_ONE;
+                        }
```
amckinney, line 1002: Nice, this is great 👏

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/PaginationPathUtils.java
```diff
@@ -0,0 +1,586 @@
...
+import com.squareup.javapoet.CodeBlock;
+import com.squareup.javapoet.ParameterizedTypeName;
+import com.squareup.javapoet.TypeName;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class PaginationPathUtils {
```
amckinney, line 31: Do we imagine this abstraction will stick around after the IR change to include better property path information? Whether it's this type or the other `Enriched*` helper class's, let's make sure to include a note for ourselves so it's clear what we can drop later on.

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/PaginationPathUtils.java
```diff
@@ -0,0 +1,586 @@
...
+    public static Optional<EnrichedCursorPathItem> enriched(Name name, TypeReference pathPropertyType) {
+        return pathPropertyType.visit(new TypeReferenceEnricher(name));
+    }
+
+    public static Optional<TypeName> unboxedTypeName(
+            AbstractGeneratorContext<?, ?> generatorContext, TypeReference reference) {
+        return reference.visit(new UnboxedTypeNameGetter(generatorContext, reference));
+    }
+
+    public static List<EnrichedCursorPathSetter> getPathSetters(
```
amckinney, line 94: This function is quite long, so it could probably benefit from a variety of helper functions scoped to individual control flow statements. 

It's fine to leave this as-is for now, but potentially something for us to revisit when we include those IR improvements to simplify this whole process.

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/PaginationPathUtils.java
```diff
@@ -0,0 +1,586 @@
...
+                            .getSafeName()
+                            .equals(desiredPropertyName.getCamelCase().getSafeName()))
+                    .map(InlinedRequestBodyProperty::getValueType)
+                    .findAny();
+        }
+
+        @Override
+        public Optional<TypeReference> visitReference(HttpRequestBodyReference httpRequestBodyReference) {
+            // TODO(ajgateno): Should not happen for this use-case since we should only allow named types for
+            //  pagination.
```
amckinney, line 402: I'd drop the TODO here since there's nothing for us to actually change.

seed/java-sdk/seed.yml
```diff
@@ -125,3 +127,4 @@ allowedFailures:
   - unions
   - streaming-parameter
   - literal
+  - java-pagination-deep-cursor-path
```
amckinney, line 130: Do you think we still need this test definition alongside the testdata already included in the pagination testdata? 

I think we have a lot of nested cursor pagination examples there, but feel free to leave this as-is if I'm wrong. If not here, wventually we'll want to consolidate the endpoints here in the `pagination` test definition so everyone can benefit.

amckinney (approved):
Nice! Left a few comments but this otherwise looks good to me as soon as the `java-sdk` snapshots are resolved (re: our discussion offline).

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/PaginationPathUtils.java
ajgateno: Thanks, meant this as a `// NOTE` instead.

seed/java-sdk/seed.yml
ajgateno: I'd rather leave it as is until the java one deals with these cases

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/PaginationPathUtils.java
ajgateno: I'll make a linear ticket and put it on backlog

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/PaginationPathUtils.java
ajgateno: These utils might stick around, but the enriched types should probably go

ajgateno: I'll add comments on those

amckinney (approved):
The new `ZeroValueUtils.java` looks great 👏

PR #5732, opened 2025-01-23
fix(python): Fix Indentation of Python Class Docs

noanflaherty:
## Description
Docstrings weren't being properly indented in Python Classes. 

## Changes Made
- Fixes the bug
- Adds test coverage

### DIFF
--- a/generators/python-v2/ast/src/__test__/__snapshots__/Class.test.ts.snap
+++ b/generators/python-v2/ast/src/__test__/__snapshots__/Class.test.ts.snap
@@ -1,5 +1,23 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`class > Renders docs correctly 1`] = `
+"class MyClass:
+    """This is a class"""
+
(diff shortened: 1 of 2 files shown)


generators/python-v2/ast/src/__test__/__snapshots__/Class.test.ts.snap
```diff
@@ -1,5 +1,23 @@
...
+
+    pass
+"
+`;
+
+exports[`class > Renders docs with multi-line strings correctly 1`] = `
+"class MyClass:
+    """This is a class.
+    I'm on a new line.
+    So am I."""
```
noanflaherty, line 15: How this should be rendered is pretty opinionated. Keeping it simple for now, but down to adjust based on feedback.

dvargas92495: doesn't this alter the string and add tabs to the value?

amckinney (approved):

PR #5739, opened 2025-01-23
feat(go): Generate wire tests

amckinney:
WIP


amckinney:
This proposal served its purpose - we'll come back to implementing this PR at a later time when it's needed!

PR #5743, opened 2025-01-24
fix(cli): Prefer application/json Content-Type

amckinney:
This updates the CLI to prefer the `application/json` content type whenever it's specified. Previously, Fern would choose to use non-JSON Content-Type header values if any existed (e.g. `application/x-www-form-urlencoded`).

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi-ir/vellum.json
+++ b/packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi-ir/vellum.json
@@ -4126,23 +4126,130 @@
       "headers": [],
       "generatedRequestName": "DocumentsUploadRequest",
       "request": {
-        "schema": {
-          "generatedName": "DocumentsUploadRequest",
-          "schema": "UploadDocumentBodyRequest",
-          "source": {
-            "file": "../openapi.yml",
-            "type": "openapi"
+        "name": "UploadDocumentBodyRequest",
+        "properties": [
+          {
+            "key": "add_to_index_names",
+            "schema": {
+              "value": {
+                "generatedName": "documentsUploadRequestAddToIndexNames",
+                "value": {
+                  "generatedName": "DocumentsUploadRequestAddToIndexNames",
+                  "value": {
+                    "generatedName": "DocumentsUploadRequestAddToIndexNames",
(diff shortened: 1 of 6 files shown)


packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi-ir/vellum.json
```diff
@@ -4126,130 +4126,23 @@
...
-              },
-              "type": "json"
-            }
-          },
-          {
-            "key": "contents",
-            "schema": {
-              "isOptional": false,
-              "isArray": false,
-              "type": "file"
```
dsinghvi, line 4211: @amckinney this will be a break for vellum and we won't do file uploads for their endpoint

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/merge.json
```diff
@@ -2279,80 +2214,6 @@ Fetch from the `LIST Employments` endpoint and filter by `ID` to show all employ
               "openapi": "../openapi.yml",
             },
           },
-          "EndUserDetailsRequest": {
```
dsinghvi, line 2282: @amckinney these would cause breaks for merge

dsinghvi:
To prevent breaks can we do json before application/url-encoded ? a little complex but will save a lot on the downstream support if merged as is

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi-ir/vellum.json
amckinney: Great call, fixed with the priority order you suggested.

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/merge.json
amckinney: Discussed offline - this is actually fixing a bug for Merge, so we can communicate that change to them after the CLI is released.

dsinghvi (approved):

PR #5762, opened 2025-01-26
chore(docs): clean up fern-docs extra folder

dannysheridan:
- change the location of files
- add redirects (best effort; hard to know if this works)


dannysheridan:
force merge. then fix the errors on main.

PR #5815, opened 2025-01-31
fix(cli): upgrade form-data to 4.0.1

zachkirsch:
## Description

Upgrade `form-data` to `4.0.1`. This gets rid of the deprecation warning that makes the CLI progress flicker during `fern generate`.

Specifically, the deprecation notice is for `form-data` relying on `util.isArray`, which was fixed in `4.0.1`: https://github.com/form-data/form-data/pull/564

## Before

https://github.com/user-attachments/assets/55e5e95f-b41d-4863-bf14-57eea72b0445

## After

https://github.com/user-attachments/assets/68d55c8b-1459-4232-99cc-05e4a4be7aa9

### DIFF
--- a/seed/ts-sdk/enum-query-params/package.json
+++ b/seed/ts-sdk/enum-query-params/package.json
@@ -13,7 +13,7 @@
     },
     "dependencies": {
         "url-join": "4.0.1",
-        "form-data": "4.0.1",
+        "form-data": "4.0.0",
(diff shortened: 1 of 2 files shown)


dsinghvi (approved):

zachkirsch:
@dsinghvi thoughts on me moving the `x-fern-ignore` lines to `debug` level?

image

PR #5831, opened 2025-02-02
[DNM] feat(docs): turn on http snippets by default

dannysheridan:
this PR proposes:
- HTTP snippets on by default
- a config in `docs.yml` to turn HTTP snippets off

when turned on by default, we can delete the edge-config flag for http-snippets.


fern/pages/docs/api-references/http-snippets.mdx
```diff
@@ -10,13 +10,16 @@ subtitle: HTTP snippets allow users to see API request examples using common HTT
...
-Currently, HTTP snippets are provided as an all-or-nothing set. You cannot configure which languages are displayed. If you would like this feature, please [open a GitHub issue](https://github.com/fern-api/fern/issues/new?template=docs-feature.yml).
-</Note>
+HTTP snippets are enabled by default for all Fern documentation sites. If you wish to disable them, you can do so in your `docs.yml` configuration file:
+
+<CodeBlock title="docs.yml">
+```yaml 
+navigation:
+  - api: API Reference
+    snippets:
+      http: false # <--- disable HTTP snippets; defaults to true
```
abvthecity, line 20: What if the customer wants to customize their own set of httpsnippet targets?

See:
https://github.com/Kong/httpsnippet/tree/master/src/targets
For supported targets.

Ie "I only want to enable python and php"

Or "I want to show php snippets using http2 and guzzle"

dannysheridan: good call. added this idea.

what if I want:
- SDK, typescript 
- HTTP, javascript fetch
- SDK, python
- HTTP, python requests

my current config doesn't allow this

dannysheridan:
relevant: https://github.com/fern-api/fern/blob/05b0e7b1dae51b2f136a257e64cb33071d5940b5/packages/cli/configuration/src/docs-yml/ParsedDocsConfiguration.ts#L266-L278

PR #5833, opened 2025-02-02
chore(docs): specify type: bytes isn't supported in a response

dannysheridan:
add a Note callout for a user to use `file` instead of `byte` in a response when writing a fern definition. also fix an incorrect code example with this same change.


sean-hacker:
+1 that we'd use this at Method as well!

chdeskur (approved):

PR #5838, opened 2025-02-03
chore(csharp): Misc improvements

Swimburger:
## Description
Miscellaneous improvement for the C# generator
- Call `.ConfigureAwait(false)` on Tasks
- Use `Enumerable<T>.Empty` instead of creating a new empty list
- Add PolySharp to test project and use C# 12
- Remove redundant `#nullable enable` directives
- Use `SystemTask` alias for `System.Threading.Tasks.Task` to avoid conflict with potentially named `Task` resources
- Improve C# syntax

## Testing

- [x] Seed

### DIFF
--- a/generators/csharp/codegen/src/asIs/test/StringEnumSerializerTests.Template.cs
+++ b/generators/csharp/codegen/src/asIs/test/StringEnumSerializerTests.Template.cs
@@ -38,7 +38,6 @@ public class StringEnumSerializerTests
         Assert.That(obj.EnumProperty, Is.EqualTo(KnownEnumValue2));
     }
 
-
     [Test]
     public void ShouldParseUnknownEnum()
     {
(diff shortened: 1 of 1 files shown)


generators/csharp/codegen/src/asIs/test/EnumSerializerTests.Template.cs
```diff
@@ -1,29 +1,24 @@
...
     private const DummyEnum KnownEnumValue2 = DummyEnum.KnownValue2;
     private const string KnownEnumValue2String = "known_value2";
 
-    private static readonly string JsonWithKnownEnum2 =
-        $$"""
-          {
-              "enum_property": "{{KnownEnumValue2String}}"
-          }
-          """;
+    private const string JsonWithKnownEnum2 = $$"""
```
amckinney, line 17: Is this string format preferred? I would have thought the previous nesting was better.

amckinney (approved):
Overall looks great, but a couple test definitions might need some attention (e.g. `pagination`).

generators/csharp/codegen/src/asIs/test/EnumSerializerTests.Template.cs
fern-support: This is csharpier I think

Swimburger: This must be csharpier, the source template doesn't format it like this

PR #5839, opened 2025-02-03
chore(internal): Improve docker build performance

Swimburger:
## Description
Improve Docker performance by:
- Re-ordering docker build instructions to increase caching hits
  (Docker builds layers on top of each other, so if a lower layer needs to be rebuilt because something is different, all layers on top also need to be rebuilt. The best strategy is to install/copy the least frequently changing things first, and the most frequently changing things last to maximize cache hits.)
- Use .dockerignore files to reduce the size of the Docker build context

## Testing
- [x] CI tests


generators/csharp/sdk/Dockerfile.dockerignore
```diff
@@ -0,0 +1,3 @@
+*
+!generators/csharp/sdk/features.yml
+!generators/csharp/sdk/dist
```
amckinney, line 3: Nice 👏

amckinney (approved):
As long as all the generators continue to run well, this lgtm

PR #5844, opened 2025-02-03
fix(java): Send only request body reference for body-only wrapped request

ajgateno:
Fixes an inline path params bug where we'd encode the entire request if there was a referenced body getter with inlined path parameters in the request. With this PR we only encode the body.


amckinney (approved):
Nice 👍

PR #5845, opened 2025-02-03
chore(php): Pin Guzzle to 7.4; update composer.json

amckinney:
This updates a variety of properties in the generated composer.json file, including:
  - Pin `guzzlehttp/guzzle` to version `^7.4`).
      - This improves compatibility for a wide range of customers, and is the minimum version the SDK requires.
  - Custom license support.
  - Set `memory-limit` to 1G in the `analyze` script.


ajgateno (approved):
LGTM 🚀

PR #5848, opened 2025-02-03
fix(java): Add accept header if endpoint has errors

ajgateno:
If an endpoint has errors, we need to pass in an accept header for application/json. In the future we'll want to delegate determining the content type to the IR but for now it's okay to have this here to get around customer issues.

### DIFF
--- a/.github/workflows/test-definitions.yml
+++ b/.github/workflows/test-definitions.yml
@@ -1,4 +1,4 @@
-name: test-definitions
+name: Validate Test Definitions
 
 on:
   push:
@@ -22,7 +22,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  check:
+  run:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
--- a/fern/docs.yml
+++ b/fern/docs.yml
@@ -96,7 +96,6 @@ navigation:
               - page: Parameter Names
                 icon: fa-regular fa-input-text
                 path: ./pages/api-definition/openapi/extensions/parameter-names.mdx
-                slug: parameter-names
               - page: Other
                 icon: fa-regular fa-ellipsis-h
                 path: ./pages/api-definition/openapi/extensions/others.mdx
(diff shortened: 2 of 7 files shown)


seed/java-sdk/accept-header/accept-header/src/main/java/com/seed/accept/resources/service/ServiceClient.java
```diff
@@ -0,0 +1,68 @@
...
+    public void endpoint(RequestOptions requestOptions) {
+        HttpUrl httpUrl = HttpUrl.parse(this.clientOptions.environment().getUrl())
+                .newBuilder()
+                .addPathSegments("container")
+                .build();
+        Request okhttpRequest = new Request.Builder()
+                .url(httpUrl)
+                .method("DELETE", null)
+                .headers(Headers.of(clientOptions.headers(requestOptions)))
+                .addHeader("Accept", "application/json")
```
ajgateno, line 41: Success!

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/AbstractEndpointWriter.java
```diff
@@ -1689,6 +1689,26 @@ public Boolean _visitUnknown(Object unknownType) {
         }
     }
 
+    public static Optional<CodeBlock> maybeAcceptsHeader(HttpEndpoint httpEndpoint, boolean withSemiColon) {
+        String ending = withSemiColon ? ";\n" : "\n";
+
+        Set<String> contentTypes = new HashSet<>();
+
+        // TODO: We'll need to get error content types from the IR once they're available.
```
amckinney, line 1697: Discussed offline - we can actually model this in a way where the `httpEndpoint` has an `accept` property that represents the set of `Accept` header values that should be sent in the request. Same idea that you're capturing here, just slightly different implementation.

amckinney (approved):

PR #5854, opened 2025-02-03
feat(cli): support parsing json and yaml from init

chdeskur:
this PR adds support for parsing both yaml and json files from external URLs. it also updates the ordering so that content warning is not thrown if we can reasonably assume the file is json or yaml.


packages/cli/init/src/utils/loadOpenApiFromUrl.ts
```diff
@@ -47,12 +47,15 @@ export async function loadOpenAPIFromUrl({ url, logger }: { url: string; logger:
...
-    if (contentType.includes("yaml")) {
-        return response.data;
+
+    if (contentType.includes("yaml") || url.endsWith(".yml") || url.endsWith(".yaml")) {
+        return typeof response.data === "string" ? response.data : dump(response.data);
     }
+
     logger.warn(
         `Unrecognized Content-Type "${contentType}" from endpoint ${url}. Please ensure you're pointing to a URL that returns JSON or YAML and not HTML (e.g. Swagger UI webpage)`
     );
```
dsinghvi, line 61: @chdeskur should we delete this logger.warn entirely?

dsinghvi (approved):

packages/cli/init/src/utils/loadOpenApiFromUrl.ts
chdeskur: i think it makes sense to leave it if they are pointing to something like swagger UI. we handle the URLs that could reasonably be parsed as YAML or JSON, so fair warning to throw afterward

PR #5869, opened 2025-02-04
chore(typescript): Remove node-fetch

Swimburger:
## Description
Remove node-fetch

- Breaking: Remove dependency on `node-fetch` which was used for versions older than Node.js 18.
  Users on old unsupported versions of Node.js can still polyfill fetch or pass in a custom fetch to the root client.

- Breaking: File download endpoints now return an object with properties and methods to get the file data.

  ```typescript
  const response = await client.downloadFile();
  const arrayBuffer: ArrayBuffer = await response.arrayBuffer();
  /**
   * const blob: Blob = await response.blob();
   * const bytes: Uint8Array = await response.bytes();
   * const body: ReadableStream<Uint8Array> | null = response.body;
   * const bodyUsed: boolean = response.bodyUsed;
  **/
  ```

## Testing

- [x] Seed testing completed


benmccann:
@Swimburger is there anything this PR is waiting on?

Swimburger:
@benmccann This was a very naive attempt at removing all the dependencies. Unfortunately, we have to be backward-compatible because some of our customers ask for Node 16 compatibility. I'm actively working on adding feature flags that can be enabled for specific features. 
These feature flags will become enabled by default in an upcoming breaking version of the generator so that these dependencies won't be included anymore by default.

Swimburger:
Great news, we have added an option to remove `node-fetch` and all other dependencies.
These options are now enabled by default in the [2.0.0 TypeScript SDK generator](https://github.com/fern-api/fern/pull/7763).

PR #5873, opened 2025-02-04
feat(cli): init docs with paginated api, title casing

chdeskur:
This PR updates the `fern init --docs` configuration to use a paginated API reference, and formats the docs title to use title casing for the organization name.


dsinghvi (approved):
@chdeskur do you think we should also default some layout things like horizontal tabs + searchbar in the header?

PR #5874, opened 2025-02-04
feat(python): Introduce `Lambda` AST Node for Python Generator

noanflaherty:
## Description
This PR introduces the ability to write `lambda` function in python. It fixes and improves a couple other things stumbled upon along the way.

## Changes Made
- Introduce new to-level `Lambda` Ast Node
- Don't require that a `type` be specified for Method parameters
- Add support for other common operators

## Testing

### DIFF
--- a/generators/python-v2/ast/src/__test__/Method.test.ts
+++ b/generators/python-v2/ast/src/__test__/Method.test.ts
@@ -56,7 +56,7 @@ describe("Method", () => {
         it("should generate a method with one argument without a type", async () => {
             const method = python.method({
                 name: "one_arg",
-                parameters: [python.parameter({ name: "arg1" })]
+                parameters: [python.parameter({ name: "arg1", type: undefined })]
             });
             method.write(writer);
             expect(await writer.toStringFormatted()).toMatchSnapshot();


generators/python-v2/ast/src/Lambda.ts
```diff
@@ -0,0 +1,41 @@
+import { LambdaParameter } from "./LambdaParameter";
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+
+export declare namespace Lambda {
+    interface Args {
+        /* The parameter names of the lambda */
+        parameters?: LambdaParameter[];
```
noanflaherty, line 8: Python lambdas don't support type hints for their args and are a simpler form of the existing `Parameter` AST Node, hence the introduce of the new `LambdaParameter` AST Node (below)

generators/python-v2/ast/src/LambdaParameter.ts
```diff
@@ -0,0 +1,17 @@
...
+        /* The initializer for the parameter */
+        initializer?: AstNode;
+    }
+}
+
+export class LambdaParameter extends Parameter {
+    constructor({ name, initializer }: LambdaParameter.Args) {
+        super({ name, initializer });
+    }
+}
```
noanflaherty, line 17: Notice we subclass `Parameter` and just always set `type` to undefined.

generators/python-v2/ast/src/Operator.ts
```diff
@@ -32,6 +32,16 @@ export class Operator extends AstNode {
                 return "or";
             case OperatorType.And:
                 return "and";
+            case OperatorType.Add:
```
noanflaherty, line 35: These new operators were useful for my testing and just generally a good thing to have.

generators/python-v2/ast/src/Parameter.ts
```diff
@@ -8,16 +8,16 @@ export declare namespace Parameter {
         /* The name of the parameter */
         name: string;
         /* The type of the parameter */
-        type: Type;
+        type?: Type;
```
noanflaherty, line 11: Types shouldn't be required when defining the arguments to a function

generators/python-v2/ast/src/Parameter.ts
```diff
@@ -8,16 +8,16 @@ export declare namespace Parameter {
...
         /* The initializer for the parameter */
-        initializer?: CodeBlock;
+        initializer?: AstNode;
```
noanflaherty, line 13: You should be able to use any Ast Node as a default value for a method argument – not just code blocks.

generators/python-v2/ast/src/Parameter.ts
amckinney: I loved the solution for `LambdaParameter`, but now that I see this change, it might be better if we just copied the code over. This otherwise makes it too easy for us to accidentally forget a type hint when defining parameters for function signatures.

Mind having `LambdaParameter` exist as a standalone implementation, even though the implementation will be nearly the same as `Parameter`?

amckinney (approved):
Nice - only concern is w/ the `Parameter` type property, but otherwise lgtm!

generators/python-v2/ast/src/Parameter.ts
noanflaherty: Yeah no problem 👍 will update before I merge.

as a nit: it does seem like the consumer should be allowed to opt out of specifying type hints, but not a hill I’d die on and even then, we could make it a required arg (like `type: Type | undefined`)

thoughts? Lmk if this changes your preferred implementation.

generators/python-v2/ast/src/Operator.ts
dvargas92495: Also enables us to implement these expressions in the near future

dvargas92495 (commented):
✅

generators/python-v2/ast/src/Parameter.ts
amckinney: Nice, the required parameter works for me - thanks for making that change!

PR #5876, opened 2025-02-04
feat(cli): update fern init behavior, logging

chdeskur:
This PR updates a few elements of the `fern init` behavior + logging:
- Distinguishes between creating a new `fern` folder and adding a single `api` to a pre-existing `fern` folder
- Prevents a user from overwriting their existing `docs.yml` if one already exists
- First checks for an existing `organization` name before prompting the user to enter one

### DIFF
--- a/packages/cli/init/src/initializeAPI.ts
+++ b/packages/cli/init/src/initializeAPI.ts
@@ -48,7 +48,7 @@ export async function initializeAPI({
     } else {
         await createFernWorkspace({ directoryOfWorkspace, cliVersion: versionOfCli, context });
 
-        context.logger.info(chalk.green("Created new Fern folder"));
+        context.logger.info(chalk.green("Created new fern folder"));
     }
 }
 


dsinghvi (approved):

packages/cli/cli/src/cli.ts
```diff
@@ -206,6 +206,17 @@ async function getIntendedVersionOfCli(cliContext: CliContext): Promise<string>
     return getLatestVersionOfCli({ cliEnvironment: cliContext.environment });
 }
 
+async function getOrganization(cliContext: CliContext): Promise<string | null> {
```
dsinghvi, line 209: nit: typically we return string | undefined (instead of `null`)

packages/cli/init/src/initializeAPI.ts
```diff
@@ -43,10 +43,13 @@ export async function initializeAPI({
             cliVersion: versionOfCli,
             context
         });
+
+        context.logger.info(chalk.green("Created new API: ./" + path.relative(process.cwd(), directoryOfWorkspace)));
     } else {
         await createFernWorkspace({ directoryOfWorkspace, cliVersion: versionOfCli, context });
+
+        context.logger.info(chalk.green("Created new Fern folder"));
```
dsinghvi, line 51: nit: `Created fern folder`

packages/cli/init/src/initializeDocs.ts
```diff
@@ -24,10 +26,41 @@ export async function initializeDocs({
...
-        join(createDirectoryResponse.absolutePathToFernDirectory, RelativeFilePath.of(DOCS_CONFIGURATION_FILENAME)),
-        yaml.dump(getDocsConfig(createDirectoryResponse.organization))
-    );
+    if (createDirectoryResponse.absolutePathToFernDirectory) {
+        const docsYmlPath = join(
+            createDirectoryResponse.absolutePathToFernDirectory,
+            RelativeFilePath.of(DOCS_CONFIGURATION_FILENAME)
+        );
+
+        try {
```
dsinghvi, line 35: we have a doesFileExist in `@fern-api/fs-utils` which may clean up this code since you dont have to try catch

packages/cli/init/src/initializeAPI.ts
chdeskur: we capitalize fern everywhere else in our own docs, why not here?

PR #5881, opened 2025-02-04
feat(php): Discriminated unions

ajgateno:
This PR implements the design from https://github.com/fern-api/fern/pull/5407 and the addenda from https://github.com/fern-api/fern/pull/5913

### DIFF
--- a/generators/php/sdk/versions.yml
+++ b/generators/php/sdk/versions.yml
@@ -3,6 +3,7 @@
     - type: feat
       summary: Add support for discriminated unions.
   irVersion: 55
+
 - version: 0.11.0
   changelogEntry:
     - type: feat


seed/php-sdk/unions/src/Union/Types/Shape.php
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * @var string $id
+     */
+    #[JsonProperty('id')]
+    public string $id;
+
+    /**
+     * @var string $type
+     */
+    public readonly string $type;
```
ajgateno, line 21: Having a union of string literals here will depend on the string literals work that comes right after. As it stands it's not a huge deal not to have it since we validate on (de)serialization; it's more of a nice-to-have here.

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+            types.push(this.getReturnType(variant));
+        }
+
+        types.push(php.Type.mixed());
+
+        return php.field({
+            // TODO(ajgateno): We'll want to disambiguate here if e.g. there's a "value" property
+            name: "value",
+            type: php.Type.union(types),
+            access: includeGetter ? "private" : "public",
```
amckinney, line 156: Same thing here w.r.t. the context method.

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+            readonly_: true
+        });
+    }
+
+    private toField({ property, inherited }: { property: ObjectProperty; inherited?: boolean }): php.Field {
+        const convertedType = this.context.phpTypeMapper.convert({ reference: property.valueType });
+        return php.field({
+            type: convertedType,
+            name: this.context.getPropertyName(property.name.name),
+            access: "public",
```
amckinney, line 166: We should use `this.context.getPropertyAccess` here, too.

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                type: php.Type.mixed()
+            })
+        );
+
+        const body = php.codeblock((writer) => {
+            const constructorArgs: php.Map.Entry[] = [];
+
+            for (const property of this.unionTypeDeclaration.baseProperties) {
+                constructorArgs.push({
+                    key: php.codeblock(`'${property.name.wireValue}'`),
```
amckinney: We should set the properties based on the `this.context.getPropertyName` (not the wire representation).
```suggestion
                    key: php.codeblock(`'${this.context.getPropertyName(property.name)}'`),
```

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                    value: php.codeblock(this.context.getVariableName(property.name.name))
+                });
+            }
+
+            constructorArgs.push({
+                key: php.codeblock(`'${this.unionTypeDeclaration.discriminant.wireValue}'`),
```
amckinney, line 206: Same thing here with `this.context.getPropertyName`.

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                parameters.push(
+                    php.parameter({
+                        name: this.context.getPropertyName(variant.discriminantValue.name),
+                        type: this.getReturnType(variant)
+                    })
+                );
+                break;
+            case "noProperties":
+            default:
+        }
```
amckinney, line 268: Let's use `assertNever` in the `default` case (like [this](https://github.com/fern-api/fern/blob/69b7daeadf4c29442b0bb49f18f48daab0df8346/generators/php/sdk/src/endpoint/http/HttpEndpointGenerator.ts#L285)). That way if we ever update the number of cases, we won't get a silent failure here.

```suggestion
        switch (variant.shape.propertiesType) {
            case "samePropertiesAsObject":
            case "singleProperty":
                parameters.push(
                    php.parameter({
                        name: this.context.getPropertyName(variant.discriminantValue.name),
                        type: this.getReturnType(variant)
                    })
                );
                break;
            case "noProperties":
            default:
                assertNever(variant.shape);
        }
```

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                        value: php.codeblock(this.context.getVariableName(variant.discriminantValue.name))
+                    });
+                    break;
+                case "noProperties":
+                default:
+                    constructorArgs.push({
+                        key: php.codeblock("'value'"),
+                        value: php.codeblock("null")
+                    });
+            }
```
amckinney, line 303: Same thing as above with `assertNever` in the `default` case - we can move this logic in `noProperties` on its own.
```suggestion
            switch (variant.shape.propertiesType) {
                case "samePropertiesAsObject":
                case "singleProperty":
                    constructorArgs.push({
                        key: php.codeblock("'value'"),
                        value: php.codeblock(this.context.getVariableName(variant.discriminantValue.name))
                    });
                    break;
                case "noProperties":
                    constructorArgs.push({
                        key: php.codeblock("'value'"),
                        value: php.codeblock("null")
                    });
                default:
                    assertNever(variant.shape);
            }
```

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                return php.Type.reference(this.context.phpTypeMapper.convertToClassReference(value));
+            },
+            singleProperty: (value: SingleUnionTypeProperty) => {
+                return this.context.phpTypeMapper.convert({ reference: value.type });
+            },
+            noProperties: () => {
+                return php.Type.null();
+            },
+            _other: (value) => {
+                throw new Error("Got unexpected union type: " + value);
```
amckinney, line 371: Let's use `assertNever` here instead - simply than throwing a custom error message.

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                    method: "get_debug_type",
+                    arguments_: [variableGetter],
+                    static_: true
+                })
+            );
+        });
+    }
+
+    private getDeserializationArrayKeyExistsErrorMessage(propertyName: NameAndWireValue): php.CodeBlock {
+        return php.codeblock(`"Json data is missing property '${propertyName.wireValue}'"`);
```
amckinney, line 458: nit: Let's use `JSON` here
```suggestion
        return php.codeblock(`"JSON data is missing property '${propertyName.wireValue}'"`);
```

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+    }
+
+    private getDeserliazationTypeCheckErrorMessage(propertyName: NameAndWireValue, typeName: string): php.CodeBlock {
+        return php.codeblock((writer) => {
+            writer.write(
+                `"Expected property '${this.context.getPropertyName(propertyName.name)}' in json data to be ${typeName}, instead received " . `
```
amckinney, line 464:
```suggestion
                `"Expected property '${this.context.getPropertyName(propertyName.name)}' in JSON data to be ${typeName}, instead received " . `
```

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                    // NOTE: Casting because LSP was complaining here
+                    writer.writeNode(underlyingTypeCheck as php.CodeBlock);
+                    writer.write(")");
+                });
+
+            case "typeDict":
+            case "enumString":
+            case "union":
+            default:
+                return null;
```
amckinney, line 587: We can use `assertNever` here which simplifies the function signature - we won't need to return a `null` anymore.

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                writer.controlFlow("if", arrayKeyDoesNotExist);
+                writer.writeNodeStatement(
+                    this.getErrorThrow(this.getDeserializationArrayKeyExistsErrorMessage(property.name))
+                );
+                writer.endControlFlow();
+
+                const type = this.context.phpTypeMapper.convert({ reference: property.valueType });
+                const typeCheck = this.getTypeCheck(php.codeblock(`$data['${property.name.wireValue}']`), type);
+
+                // TODO(ajgateno): We have to fix this or else we could be skipping base properties in deserialization.
```
amckinney, line 1049: Do we need to do this here, or are we tracking to come back in a follow-up?

generators/php/model/src/union/UnionGenerator.ts
```diff
@@ -0,0 +1,1329 @@
...
+                        php.invokeMethod({
+                            method: "is_null",
+                            arguments_: [php.codeblock("$" + asCastMethodName)],
+                            static_: true
+                        })
+                    );
+                    writer.write("$value = ");
+                    writer.writeTextStatement("$" + asCastMethodName);
+                    if (
+                        ["reference", "dateTime", "date", "optional"].includes(type.underlyingType().internalType.type)
```
amckinney, line 838: This line made me pause - is there any way we could write a helper here that more clearly describes what this is doing?

seed/php-sdk/unions/src/Union/Types/Shape.php
```diff
@@ -0,0 +1,233 @@
...
+        if (!is_array($decodedJson)) {
+            throw new Exception("Unexpected non-array decoded type: " . gettype($decodedJson));
+        }
+        return self::jsonDeserialize($decodedJson);
+    }
+
+    /**
+     * @param array<string, mixed> $data
+     */
+    public static function jsonDeserialize(array $data): static
```
amckinney, line 188: Niiice 👏

amckinney (approved):
Nice! Mainly nits and some light changes we can apply in general (e.g. `assertNever` in every `default` case).

Otherwise if this behaves well in your additional testing, then lgtm!

generators/php/model/src/union/UnionGenerator.ts
ajgateno: We can't use assertNever for the union, enumString, etc. because they are valid variants. And for them we have to return a null for now.

generators/php/model/src/union/UnionGenerator.ts
ajgateno: I'll do it here

PR #5883, opened 2025-02-04
chore(docs): bump openapi parser and add more robust endpoint resolution for config

RohinBhargava:
## Description
* Previously, there were errors in navigation config for endpoints. Now, with a parser version bump and wider support for endpoint locators, there are fewer ambiguous inputs.
* Additionally, request examples are now terse while response examples are verbose for example generation.

## Changes Made
- Bump OpenApi Parser v2 package.
- Locator changes to support a wider range of inputs

## Testing
Manual docs dev and preview generation for docs.


dsinghvi (approved):
Is it worth adding tests for the logic in `ApiReferenceNodeConverterLatest.ts` ?

Issue #5756, opened 2025-01-25
Support rendering local files in local development


dannysheridan:
### Feature Description

![Image](https://github.com/user-attachments/assets/a781e0e4-7970-48ee-a11b-a6dbca6c557b)

### Use Case

I am editing my documentation in local development (fern docs dev) and don't see the pretty screenshots I intend to add.

### Proposed Experience (Optional)

I'd like to see images in local development mode

dannysheridan:
Confirming that the image does appear in [a preview link](https://fern-preview-8dcaf2cc-4d29-4fa7-86c2-9faf5dd0f876.docs.buildwithfern.com/learn/docs/content/frontmatter#edit-this-page)

![Image](https://github.com/user-attachments/assets/b4915891-6838-419e-aef7-096c456e291b)

chdeskur:
This should be fixed for Mac users on the latest CLI

PR #5889, opened 2025-02-04
fix(cli): render images in local preview

Edsun457:
## Description
Fix ticket https://github.com/fern-api/fern/issues/5756

### Cause of the issue
In local environment when re-rendering markdown files, the image rendered to browser as file relative path. The browser can not parse the image unless it's the image ID matched in FileV2

For example: `Table of contents feature`

### What this PR fixed
- Make sure the image rendered to browser is image ID: example: `Edit this page feature`
- update the logic of path parser to take both relative path and absolute path

## Changes Made
Add logic to populate image file id so would render the right image to the browser for local env

## Testing

### DIFF
--- a/scripts/live-test.ps1
+++ b/scripts/live-test.ps1
@@ -25,75 +25,16 @@ try {
     Write-Host "Running Fern Commands!"
     $DebugPreference = "Continue"
     
-    # Initialize Fern project
-    Write-Host "Initializing Fern project..."
     & node $cli_path init --organization fern
     
-    # Create test docs structure
-    Write-Host "Creating test docs structure..."
-    New-Item -ItemType Directory -Path "docs\images" -Force
-    
-    # Create a test image
-    Write-Host "Creating test image..."
-    "Test Image Content" | Out-File -FilePath "docs\images\test.png"
-    
-    # Create test markdown with various Windows path formats
-    Write-Host "Creating test markdown..."
-    $testMarkdown = @"
-# Test Document
-
-## Relative Path Tests
-![Relative Image](.\images\test.png)
-![Parent Path](..\docs\images\test.png)
-
-## Absolute Path Tests
-![Absolute Path](${test_dir}\docs\images\test.png)
-
-## Mixed Slash Tests
-![Mixed Slashes](./images\test.png)
-"@
-    Set-Content -Path "docs\test.md" -Value $testMarkdown
-    
-    Write-Host "Created test markdown with Windows paths:"
-    Get-Content "docs\test.md"
-    
-    # Run normal commands
-    Write-Host "Adding SDK generators..."
     & node $cli_path add fern-java-sdk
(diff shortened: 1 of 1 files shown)


packages/cli/docs-markdown-utils/src/parseImagePaths.ts
```diff
@@ -227,18 +227,31 @@ export function replaceImagePathsAndUrls(
...
+        // Handle relative path
+        const resolvedPath = resolvePath(image, metadata);
+        if (resolvedPath) {
+            const fileId = fileIdsMap.get(resolvedPath);
+            context.logger.info(
+                "image - resolvedPath - fileId",
+                image || "undefined",
+                resolvedPath.toString() || "undefined",
+                fileId || "undefined"
+            );
```
dsinghvi, line 250: nit: im guessing you wanted to delete this before merging

packages/cli/docs-markdown-utils/src/parseImagePaths.ts
```diff
@@ -227,18 +227,31 @@ export function replaceImagePathsAndUrls(
...
-                // do nothing
-                return;
-            }
+        if (image == null || isExternalUrl(image) || isDataUrl(image)) {
+            return undefined;
         }
-        return;
+
+        // Check if it's an absolute path
+        if (image.startsWith("/")) {
```
dsinghvi, line 235: this won't work on windows, we'll want to make sure that our logic is compatible across different operating systems

packages/cli/docs-preview/src/previewDocs.ts
```diff
@@ -60,8 +60,27 @@ export async function getPreviewDocsDefinition({
                 continue;
             }
 
+            const fileIdsMap = new Map(
+                Object.entries(previousDocsDefinition.filesV2 ?? {}).map(([id, file]) => {
+                    const path = "/" + file.url.replace("/_local/", "");
+                    return [AbsoluteFilePath.of(path), id];
+                })
+            );
```
dsinghvi, line 68: what is happening in this transformation here?

packages/cli/docs-markdown-utils/src/parseImagePaths.ts
dsinghvi: https://github.com/fern-api/fern/blob/05fd0adfcbe44bb3ef2ff1637eeb331821b6d06a/.github/workflows/ci.yml#L143-L173 you could write a simple test that make ssure `fern docs dev` runs on windows with a sample docs fixture

dsinghvi:
Before you merge, you'll want to add a changelog to https://github.com/fern-api/fern/blob/main/packages/cli/cli/versions.yml -- that way a new version will get published to npm

dsinghvi (approved):
@Edsun457 can you fix `lint` and add a changelog? Other than that, we are good to go!

PR #5894, opened 2025-02-05
fix(python): Don't Write Imports for References w/o Module Paths

noanflaherty:
## Description
Sometimes you might want to instantiated a Reference AstNode for something that's either:
1. Defined in the current file; or
2. is an attribute of another AsttNode (e.g. `MyVar.my_method` – where `my_method` is the Reference) 

In this case, we shouldn't be attempting to generate an import path for that Reference.

Currently, we generate an invalid import that looks like `from  import my_method`.

## Changes Made
- Handle this case specifically
- Add a test case

## Testing


dvargas92495 (commented):
✅

amckinney (approved):

PR #5901, opened 2025-02-05
feat(php): Add HTTP retries to php-sdk generator

Swimburger:
## Description

Retry HTTP requests on status codes 408, 429, 5XX.
You can configure the maximum number of retries like this:
```php
$client = new FooClient("token", ['maxRetries' => 3])
$client->bar($request, ['maxRetries' => 5]);
```
The default for `maxRetries` is 2, meaning up to 3 HTTP requests may be sent.
Set `maxRetries` to 0 to disable retries.

## Changes Made
- Add `RetryMiddleware` to the default HTTP client. The `RetryMiddleware` needs to be manually added if the user passes in their own HTTP client.
- Pass options from root client to sub-package clients, and pass options from sub-package clients to sub-package clients.
- Merge client options with request options and pass options to the send HTTP request method.

## Testing

- [x] Seed testing completed

### DIFF
--- a/seed/php-sdk/accept-header/tests/Seed/Core/Client/RawClientTest.php
+++ b/seed/php-sdk/accept-header/tests/Seed/Core/Client/RawClientTest.php
@@ -165,6 +165,9 @@ class RawClientTest extends TestCase
         $this->assertEquals(0, $this->mockHandler->count());
     }
 
+    /**
+     * @throws ClientExceptionInterface
+     */
     public function testShouldRetryOnStatusCodes(): void
     {
         $this->mockHandler->append(
@@ -179,7 +182,7 @@ class RawClientTest extends TestCase
             new Response(599),
             new Response(200),
         );
-        $countOfErrorRequests = $this->mockHandler->count() - 2;
+        $countOfErrorRequests = $this->mockHandler->count() - 1;
 
         $request = new JsonApiRequest(
             $this->baseUrl,
@@ -187,13 +190,9 @@ class RawClientTest extends TestCase
             HttpMethod::GET
         );
 
-        try {
-            $this->rawClient->sendRequest($request, ['maxRetries' => $countOfErrorRequests]);
-            $this->fail("Request should've failed but succeeded.");
-        } catch (ClientExceptionInterface) {
-        }
+        $this->rawClient->sendRequest($request, ['maxRetries' => $countOfErrorRequests]);
 
-        $this->assertEquals(1, $this->mockHandler->count());
+        $this->assertEquals(0, $this->mockHandler->count());
     }
 
     public function testShouldFailOn400Response(): void
--- a/seed/php-sdk/alias-extends/tests/Seed/Core/Client/RawClientTest.php
+++ b/seed/php-sdk/alias-extends/tests/Seed/Core/Client/RawClientTest.php
@@ -165,6 +165,9 @@ class RawClientTest extends TestCase
         $this->assertEquals(0, $this->mockHandler->count());
     }
 
+    /**
+     * @throws ClientExceptionInterface
+     */
     public function testShouldRetryOnStatusCodes(): void
     {
         $this->mockHandler->append(
@@ -179,7 +182,7 @@ class RawClientTest extends TestCase
             new Response(599),
             new Response(200),
         );
-        $countOfErrorRequests = $this->mockHandler->count() - 2;
+        $countOfErrorRequests = $this->mockHandler->count() - 1;
 
         $request = new JsonApiRequest(
             $this->baseUrl,
@@ -187,13 +190,9 @@ class RawClientTest extends TestCase
             HttpMethod::GET
         );
 
-        try {
-            $this->rawClient->sendRequest($request, ['maxRetries' => $countOfErrorRequests]);
-            $this->fail("Request should've failed but succeeded.");
-        } catch (ClientExceptionInterface) {
-        }
+        $this->rawClient->sendRequest($request, ['maxRetries' => $countOfErrorRequests]);
 
-        $this->assertEquals(1, $this->mockHandler->count());
+        $this->assertEquals(0, $this->mockHandler->count());
     }
 
     public function testShouldFailOn400Response(): void
--- a/seed/php-sdk/alias/tests/Seed/Core/Client/RawClientTest.php
+++ b/seed/php-sdk/alias/tests/Seed/Core/Client/RawClientTest.php
@@ -165,6 +165,9 @@ class RawClientTest extends TestCase
         $this->assertEquals(0, $this->mockHandler->count());
     }
 
+    /**
+     * @throws ClientExceptionInterface
+     */
     public function testShouldRetryOnStatusCodes(): void
     {
         $this->mockHandler->append(
@@ -179,7 +182,7 @@ class RawClientTest extends TestCase
             new Response(599),
             new Response(200),
         );
-        $countOfErrorRequests = $this->mockHandler->count() - 2;
+        $countOfErrorRequests = $this->mockHandler->count() - 1;
 
         $request = new JsonApiRequest(
             $this->baseUrl,
@@ -187,13 +190,9 @@ class RawClientTest extends TestCase
             HttpMethod::GET
         );
 
-        try {
-            $this->rawClient->sendRequest($request, ['maxRetries' => $countOfErrorRequests]);
-            $this->fail("Request should've failed but succeeded.");
-        } catch (ClientExceptionInterface) {
-        }
+        $this->rawClient->sendRequest($request, ['maxRetries' => $countOfErrorRequests]);
 
-        $this->assertEquals(1, $this->mockHandler->count());
+        $this->assertEquals(0, $this->mockHandler->count());
     }
 
     public function testShouldFailOn400Response(): void
--- a/seed/php-sdk/any-auth/tests/Seed/Core/Client/RawClientTest.php
+++ b/seed/php-sdk/any-auth/tests/Seed/Core/Client/RawClientTest.php
@@ -165,6 +165,9 @@ class RawClientTest extends TestCase
         $this->assertEquals(0, $this->mockHandler->count());
     }
 
+    /**
+     * @throws ClientExceptionInterface
+     */
     public function testShouldRetryOnStatusCodes(): void
     {
         $this->mockHandler->append(
@@ -179,7 +182,7 @@ class RawClientTest extends TestCase
             new Response(599),
             new Response(200),
         );
-        $countOfErrorRequests = $this->mockHandler->count() - 2;
+        $countOfErrorRequests = $this->mockHandler->count() - 1;
 
         $request = new JsonApiRequest(
             $this->baseUrl,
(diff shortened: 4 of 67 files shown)


seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
```diff
@@ -0,0 +1,192 @@
...
+     *     baseDelay: int,
+     * }
+     */
+    private array $options;
+
+    /**
+     * @param callable $nextHandler
+     * @param array{
+     *     maxRetries?: int,
+     *  }|null $options
```
amckinney, line 34: nit: We can use an optional parameter instead of `|null` here.
```suggestion
     * @param array{
     *     maxRetries?: int,
     *  } ?$options
```

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
```diff
@@ -0,0 +1,192 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Throwable;
+
+class RetryMiddleware
+{
+    private const RETRY_STATUS_CODES = [408, 429];
+
+    /**
+     * @var callable(RequestInterface, array): PromiseInterface
+     * @phpstan-ignore missingType.iterableValue
```
amckinney, line 18: Can we add an explicit type here to prevent the `phpstan-ignore` comment? 

Can we just say `private callable $nextHandler` here?

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
```diff
@@ -0,0 +1,192 @@
...
+        int                $retries,
+        int                $maxRetries,
+        ?ResponseInterface $response = null,
+        ?Throwable         $exception = null
+    ): bool {
+        if ($retries >= $maxRetries) {
+            return false;
+        }
+
+        // Check for timeout
```
amckinney, line 103: nit: I'd remove these comments.

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
```diff
@@ -0,0 +1,192 @@
...
+    {
+        return static function (callable $handler) use ($options): RetryMiddleware {
+            return new RetryMiddleware($handler, $options);
+        };
+    }
+
+    /**
+     * @param RequestInterface $request
+     * @param array{
+     *      retries?: int,
```
amckinney, line 64: Can we update this to be `retryAttempt` so it's more clear it's the attempt #? Otherwise it's confusing how it relates to `maxRetries`.

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
```diff
@@ -0,0 +1,192 @@
...
+
+class RetryMiddleware
+{
+    private const RETRY_STATUS_CODES = [408, 429];
+
+    /**
+     * @var callable(RequestInterface, array): PromiseInterface
+     * @phpstan-ignore missingType.iterableValue
+     */
+    private $nextHandler;
```
amckinney, line 20: Thoughts on shortening this to just `$next`? I've seen this naming convention used elsewhere, but not married to it if you feel strongly otherwise.

seed/php-sdk/imdb/clientName/src/Core/Client/RawClient.php
```diff
@@ -31,23 +32,41 @@ class RawClient
...
 
     /**
+     * @return Client
+     */
+    private function createDefaultClient(): Client
+    {
+        $handler = HandlerStack::create();
+        $handler->push(RetryMiddleware::create());
+        return new Client(['handler' => $handler]);
+    }
```
amckinney, line 54: We'll want to wrap the user-defined `ClientInterface` with retry logic, too. Can we refactor this so that we resolve the `ClientInterface` then wrap it after the fact?

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
```diff
@@ -0,0 +1,192 @@
...
+     *  }|null $options
+     */
+    public function __construct(
+        callable $nextHandler,
+        ?array   $options = null,
+    ) {
+        $this->nextHandler = $nextHandler;
+        $this->options = array_merge([
+            'maxRetries' => 2,
+            'baseDelay' => 1000,
```
amckinney, line 43: Nice - thoughts on defining this as an un-exported constant at the top of the file? That way it's easy to see all the default retry configurations that we're using. Something like:

 ```php
const DEFAULT_RETRY_CONFIG = [
    'maxRetries' => 2,
    'baseDelay' => 1000
];
```

File: seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php [hunk 8]
```diff
@@ -0,0 +1,192 @@
...
+            return false;
+        }
+
+        // Check for timeout
+        if ($exception instanceof ConnectException) {
+            return true;
+        }
+
+        if ($response) {
+            // Check status codes
```
amckinney, line 109: I'd remove this comment, too.

generators/php/codegen/src/asIs/Client/RetryMiddleware.Template.php
```diff
@@ -0,0 +1,194 @@
...
+            $options['retries'] = 0;
+        }
+
+        $fn = $this->nextHandler;
+
+        return $fn($request, $options)
+            ->then(
+                $this->onFulfilled($request, $options),
+                $this->onRejected($request, $options)
+            );
```
amckinney, line 84: This is slick 👏

generators/php/codegen/src/ast/core/AstNode.ts
```diff
@@ -7,15 +7,18 @@ export abstract class AstNode extends AbstractAstNode {
...
+    public toString(param: {
         namespace: string;
         rootNamespace: string;
         customConfig: BasePhpCustomConfigSchema;
     }): string {
+        if (param == null) {
+            throw new Error(
+                "param is required. You are likely implicitly calling toString() inside a string interpolation or concatenation."
+            );
+        }
```
amckinney, line 19: Do we really need this given the type safety of this function? If so, can we prefix the error message so it's clear it's not the user's fault?

```
throw new Error(
  "Internal error; AstNode.toString method called incorrectly."
);
```

generators/php/codegen/src/ast/utils/convertToPhpVariableName.ts
```diff
@@ -3,5 +3,8 @@
  * (i.e. it includes the required '$' prefix).
  */
 export function convertToPhpVariableName(name: string): string {
-    return `${name.startsWith("$") ? "" : "$"}${name}`;
+    if (name.startsWith("$")) {
+        return name;
+    }
+    return `$${name}`;
```
amckinney, line 9: 👍

generators/php/codegen/src/ast/MergeArrays.ts
```diff
@@ -0,0 +1,53 @@
+import { AbstractAstNode } from "@fern-api/base-generator";
+
+import { codeblock } from "../php";
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+import { writeArguments } from "./utils/writeArguments";
+
+export declare namespace MergeArrays {
```
amckinney, line 8: I can see the value of this use case, but the types are a little difficult to follow. Is there any way we can reduce this down to only the use case that we're currently using, i.e.

```typescript
php.mergeArrays(`$this->${this.context.getClientOptionsName()}`, {
  ref: php.variable(this.context.getRequestOptionsName()),
  fallback: "[]"
})
```

If not, then this is likely fine as-is - just thinking of ways to reduce complexity and potentially remove the need for the new `writer.writeNodeOrString` function entirely.

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
Swimburger: A large amount of the code is copied from the built-in RetryMiddleware from the Guzzle library itself, so I think this is probably a good name

seed/php-sdk/imdb/clientName/src/Core/Client/RawClient.php
Swimburger: I would, but it's not possible. If the user provides their own client, they need to create their own stack and add the middleware themselves.

generators/php/codegen/src/ast/core/AstNode.ts
Swimburger: The problem is that there isn't type safety. We override the default `toString` method with additional parameters, but that doesn't mean TypeScript stops us from doing this:
```ts
const myVar = getAstNode();
const assignment = $`{myVar} = 10;`;
```

For string concatenation or string interpolation, our custom `toString` method is called with 0 parameters. 
The goal of the error message is to help ourselves understand the issue as the default error that would be thrown is not helpful.

Having said that, I will prefix with internal.

generators/php/codegen/src/ast/utils/convertToPhpVariableName.ts
Swimburger: Just a small perf improvement :)

generators/php/codegen/src/ast/MergeArrays.ts
Swimburger: If we don't accept strings, we'll have to wrap it in php.codeBlock() which is fine but I like reducing the friction.
I'm assuming we'll use this `mergeArrays` in a bunch of places.

I'm also down to just remove it and construct the merge call manually everywhere.

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
Swimburger: `callable` isn't allowed as a type, only in typedocs. 
The reference RetryMiddleware.php didn't type the array options out.
I think the point of the options is that it can contain anything, but not sure how to tell phpstan that's okay.
Without this ignore the `composer analyze` fails which is in our seed scripts.

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
Swimburger: Great call

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
Swimburger: Done

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
Swimburger: Done

seed/php-sdk/imdb/clientName/src/Core/Client/RetryMiddleware.php
Swimburger: Done

generators/php/codegen/src/asIs/Client/RawClientTest.Template.php
```diff
@@ -24,10 +24,18 @@ protected function setUp(): void
     {
         $this->mockHandler = new MockHandler();
         $handlerStack = HandlerStack::create($this->mockHandler);
+        // since the client is constructed manually, we need to add the retry middleware manually
```
amckinney, line 27: I'd drop this.

amckinney (approved):
🚀

PR #5903, opened 2025-02-05
feat(php): Generate __toString() methods

amckinney:
This adds to magic `__toString()` method implementation for all objects. The string representation is equivalent to the JSON representation of the type:

```php
<?php

namespace Imdb;

use Imdb\Core\Json\JsonSerializableType;
use Imdb\Core\Json\JsonProperty;

class Movie extends JsonSerializableType
{
    ...

    /**
     * @return string
     */
    public function __toString(): string
    {
        return $this->toJson();
    }
}
```

Note that we intentionally don't include this on the generated `enum` types because this is recommended against - users can simply just access the `string`-backed enums with the `$enum->value` accessor.

### DIFF
--- a/generators/browser-compatible-base/src/ast/AbstractWriter.ts
+++ b/generators/browser-compatible-base/src/ast/AbstractWriter.ts
@@ -53,6 +53,19 @@ export class AbstractWriter {
     }
 
     /**
+     * Writes a node or string
+     * @param input
+     */
+    public writeNodeOrString(input: AbstractAstNode | string): void {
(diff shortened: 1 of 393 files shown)


Swimburger:
LGTM, php-model is failing for some reason

Swimburger (approved):

amckinney:
> LGTM, php-model is failing for some reason

I think there are some occasional failures with `composer` failing to unzip (potentially w/ high concurrency). Will need to take a closer look at this later.

PR #5904, opened 2025-02-05
fix(internal): Replace broken link in README.md

hastyquickbeam:
## Description
Replaces a broken link with an updated link.

## Changes Made
See above

## Testing
None


Swimburger:
Thank you for catching that!

Swimburger (approved):

PR #5906, opened 2025-02-06
fix(csharp): Fix ambiguous pager methods by making unpaged method private

Swimburger:


generators/csharp/codegen/src/project/CsharpProject.ts
```diff
@@ -156,13 +156,13 @@ export class CsharpProject extends AbstractProject<AbstractCsharpGeneratorContex
...
-                DOTNET_CLI_TELEMETRY_OPTOUT: "1"
-            }
-        });
+        // await loggingExeca(this.context.logger, "dotnet", ["csharpier", "."], {
+        //     doNotPipeOutput: true,
+        //     cwd: absolutePathToSrcDirectory,
+        //     env: {
+        //         DOTNET_CLI_TELEMETRY_OPTOUT: "1"
+        //     }
+        // });
```
amckinney, line 165: Should we re-enable this before merging?

amckinney (approved):

PR #5922, opened 2025-02-07
fix(cli): fix in-line Alias, base properties, and extended properties example generation.

eyw520:
## Description
Help unblock dynamic snippets in testing.

## Changes Made
- Fix in-line Alias type example generation.
- Fix missing properties for unions and objects.

## Testing


generators/typescript-v2/dynamic-snippets/src/context/DynamicTypeLiteralMapper.ts
```diff
@@ -310,10 +310,10 @@ export class DynamicTypeLiteralMapper {
                     return [
                         ...baseFields,
                         {
-                            name: this.context.getPropertyName(unionVariant.discriminantValue.name),
+                            name: UNION_VALUE_KEY,
```
amckinney, line 313: I'll need to take a closer look at this one - we should be using the original solution in _most_ cases, so this might break other use cases, e.g.

**Before**

```typescript
shape: {
  type: "circle",
  circle: {
    radius: 10
  }
}
```

**After**

```typescript
shape: {
  type: "circle",
  value: 10
}
```

eyw520: You are totally correct. Made an update. I think the actual issue was that we weren't capturing the "value" key and matching it to record.value. Made an update that shows this.

amckinney (approved):
Nice 👏

PR #5924, opened 2025-02-07
chore(cli): bump OpenApi Parser v2, FDR packages and keep nulls in OpenApi specs

RohinBhargava:
## Description
This change upgrades the OpenApi parser v2 to version 0.0.53, and preserves `null` values in OpenApi specs.

## Changes Made
- Only filter undefined values from OpenApi specs.
- Types are now exported from the package, so there is no need to cast.
- FDR version is bumped.

## Testing

### DIFF
--- a/fern/pages/changelogs/cli/2025-02-07.mdx
+++ b/fern/pages/changelogs/cli/2025-02-07.mdx
@@ -1,0 +1,5 @@
+## 0.51.33
+**`(fix):`** The OpenAPI parser v2 now handles `null` values in examples, parses request and response bodies as unions if multiple, handles required 
+properties for global headers and improves upon example generation for APIs.
+
+


packages/commons/core-utils/src/mergeWithOverrides.ts
```diff
@@ -12,7 +12,7 @@ export function mergeWithOverrides<T extends object>({ data, overrides }: { data
             : undefined
     ) as T;
     // Remove any nullified values
-    const filtered = omitDeepBy(merged, isNull) as T;
+    const filtered = omitDeepBy(merged, (value) => value === undefined) as T;
```
RohinBhargava, line 15: critical change

dsinghvi (approved):

PR #5947, opened 2025-02-09
chore(internal): Reduce triggers of GHA workflows

Swimburger:
## Description
Reduce the amount of GHA workflows triggered when a file changes by adding path filters.

update-seed.yml used to get triggered on a CRON schedule and was eventually disabled because it ran too often. This change makes it only run when a relevant file is triggered and filters down to the specific programming languages that are changed.

### DIFF
--- a/.github/workflows/update-seed.yml
+++ b/.github/workflows/update-seed.yml
@@ -1,18 +1,6 @@
 name: Update Seed
 
 on:
-  pull_request:
-    branches:
-      - main
-    paths:
-      - '.github/workflows/update-seed.yml'
-      - 'packages/seed/**'
-      - 'packages/ir-sdk/fern/apis/**'
-      - 'packages/cli/generation/ir-generator/**'
-      - 'test-definitions/**'
-      - 'test-definitions-openapi/**'
-      - 'generators/**'
-      - 'seed/**'
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/test-events/update-seed/push.json
```diff
@@ -0,0 +1,11 @@
+{
+  "push": {
+    "paths": [
+      "packages/seed/src/cli.ts",
+      "generators/php/sdk/src/cli.ts"
```
amckinney, line 5: Should the other languages be included here too? Is this in a state that you were using for testing?

.github/workflows/validate-changelog.yml
```diff
@@ -60,6 +47,28 @@ jobs:
...
+            java-model
+            java-spring
+            ts-express
+            csharp-sdk
+            csharp-model
+            php-sdk
+            php-model
+          )
+          for generator in "${generators[@]}"; do
+            pnpm seed validate generator "$generator" --log-level debug
```
amckinney, line 73: Should this be `pnpm seed test --generator "$generator" --log-level debug`?

amckinney (commented):
The general premise makes sense to me - although we might need to edit a variety of the new `test-events` folder files to be less language-specific.

We can confirm behavior by iterating directly on this branch (e.g. update the `main` branch references to this branch name), then merge if things look good!

.github/test-events/update-seed/push.json
Swimburger: This is just for my own testing purposes; other languages could be tested like this, but not necessary. The thing that I'm testing is that it is triggered.

amckinney (approved):

Issue #5897, opened 2025-02-05
Enable additional headers to be passed with API calls through the `RequestOptions` in the Java SDK


a-ignatov-parc:
### Which Fern component?

SDK Generator

### How important is this?

P1 - High (Strongly needed)

### What's the feature?

Currently, the Java SDK doesn't allow to pass additional headers that are not defined in the OAS as part of requests, which limits its capabilities compared to other languages like TypeScript, Go, and Python.

### Any alternatives?

Another option is to add the necessary headers to the OAS specification, but this isn't always suitable for infrastructure or client-related features that fall outside the core business logic.

dannysheridan:
@a-ignatov-parc thanks for the feature request! A bit more info would be really helpful, specifically:

1. a minimum openapi spec for us to reproduce this
2. the actual output
3. the ideal output

Thanks!

ajgateno:
Hi @a-ignatov-parc, we're currently tracking this internally and will get back to you as soon as we have updates. Thanks for reporting this issue!

ajgateno:
Opened up a PR to fix this here https://github.com/fern-api/fern/pull/5949

ajgateno:
Hi @a-ignatov-parc, the fix should be available with version `2.12.0` of the Java SDK generator.

a-ignatov-parc:
@ajgateno, thank you!

PR #5949, opened 2025-02-10
feat(java): Let users define custom headers in request options

ajgateno:
Fixes https://github.com/fern-api/fern/issues/5897

Enables users to define custom request headers on the `RequestOptions` object, so that headers can be defined/overridden on a per-call basis if need be.


seed/java-sdk/exhaustive/custom-client-class-name/src/main/java/com/seed/exhaustive/core/RequestOptions.java
```diff
@@ -34,6 +46,10 @@ public Map<String, String> getHeaders() {
         if (this.token != null) {
             headers.put("Authorization", "Bearer " + this.token);
         }
+        this.headers.forEach(headers::put);
+        this.headerSuppliers.forEach((key, supplier) -> {
+            headers.put(key, supplier.get());
+        });
```
ajgateno, line 52: Don't love that you can't override a header supplier with a header since suppliers always take precedence. I'm doing it this way to preserve the same behavior as `ClientOptions`. If we change one, we should change both and have it be a breaking change.

dsinghvi (approved):

PR #5959, opened 2025-02-10
feat(php): Add getAdditionalProperties method

amckinney:
This updates the `JsonSerializableType` class to track additional properties not defined on the target class. These properties are retrievable like so:

```php
$user = $client->users->get(...);
$user.getAdditionalProperties();
```

Note that this also refactors the `jsonDeserialize` method to reduce nesting and make the implementation slightly more readable.

### DIFF
--- a/generators/php/sdk/versions.yml
+++ b/generators/php/sdk/versions.yml
@@ -1,3 +1,16 @@
+- version: 0.9.0
+  changelogEntry:
+    - type: feat
+      summary: >-
+        Add the ability to access alpha or undocumented response properties from every class.
+        Users can access the additional properties like so:
+
+        ```php
+        $response = $client->users->get(...);
+        $additionalProperties = $response->getAdditionalProperties();
+        ```
+  irVersion: 55
+
 - version: 0.8.0
   changelogEntry:
     - type: feat
@@ -19,6 +32,7 @@
         }
         ```
   irVersion: 55
+
 - version: 0.7.0
   changelogEntry:
     - type: feat


generators/php/base/src/asIs/Json/JsonSerializableType.Template.php
```diff
@@ -114,58 +115,82 @@ public static function jsonDeserialize(array $data): static
...
+    }
+
+    /**
+     * Get properties from JSON that weren't mapped to class fields
+     * @return array<string, mixed>
+     */
+    public function getAdditionalProperties(): array
+    {
+        return $this->__additionalProperties;
     }
```
Swimburger, line 194: Will this function be available to every type?
Can we limit it to only where the spec says additionalProperties: true?

Swimburger (changes requested):

generators/php/base/src/asIs/Json/JsonSerializableType.Template.php
amckinney: This will be available to every type, but that's the intention - this won't allow users to _set_ additional properties, only access them. This is useful for accessing alpha and/or undocumented properties (similar to the other SDKs).

amckinney: To be clear, when `additionalProperties: true`, we'll use that setting to determine whether or not the type is allowed to _set_ the properties (like you're imagining).

Swimburger (approved):

PR #5961, opened 2025-02-10
[DO NOT MERGE] chore(docs): document codeblock syncing

chdeskur:


dannysheridan:
@chdeskur would be helpful to see a simple example where there's no title. would also be helpful to see more than bash -- what are other example uses?

could see you documenting the params of a codeblock (language, title, for)

```bash title="Python" for="python"
pip install plant-store
```

PR #5970, opened 2025-02-11
feat(php): Add support for arbitrary parameters

amckinney:
This adds the ability for the user to specify the `bodyProperties` and `queryParameters` request options, which can be used to add arbitrary properties to the request. This is useful for interacting with alpha or undocumented functionality.

```php
$response = $client->users->list(
  new ListUsersRequest([
    ...
  ]),
  [
    'queryParameters' => [
      'limit' => 100,
    ],
  ]
);
```


Swimburger:
Let's make sure that the inheritance of headers overrides as expected because the $options parameter for sending a Guzzle HTTP request also allows headers, and we're passing those options there too:
https://docs.guzzlephp.org/en/stable/request-options.html#headers

generators/php/base/src/asIs/Client/RawClient.Template.php
```diff
@@ -87,60 +97,110 @@ private function buildRequest(
...
-
-        if (!empty($request->query)) {
-            $url .= '?' . $this->encodeQuery($request->query);
+        $query = array_merge(
+            $request->query,
+            $options['queryParameters'] ?? [],
+        );
+        if (!empty($query)) {
+            $url .= '?' . $this->encodeQuery($query);
         }
```
Swimburger, line 195: Instead of adding this to the URL ourselves, can we use the built-in query option?
We should be mindful of how this is passed to the request method, which may override or merge the query values from the client/request/request method.

https://docs.guzzlephp.org/en/stable/request-options.html#query

Swimburger:
I imagine that at some point we may add some custom logic to how we format query string or headers, but the values that we customize may be overridden by the options param we pass to the send HTTP request method. This may be hard to track down and debug.

Dummy code:

```php
$client = new FooClient("", []);
$client->getThing($request, ['headers' => ['X-Foo' => 'Bar']]);
```

As I understand, under the hood we'll do this:

```php
$headers = encodeHeaders($options);
$request = new Request(... , headers);
$guzzleClient->send($request, $options);
```
Since `headers` is in `$options` and is sent as part of `$guzzleClient->send($request, $options)`, the `encodeHeaders` headers may be overridden.

This doesn't really have a material change as the header keys and values are the same, but if we add more logic to `encodeHeaders`, this may cause bugs.

Swimburger:
This may not be relevant at this moment yet, but how is `bodyProperties` going to affect multi-part form bodies?

generators/php/base/src/asIs/Client/RawClient.Template.php
amckinney: Nice find + good call - I think it's best to update to use the built-in query option in a follow-up so that we have a commit focused on that potential behavioral change. The query parameters aren't passed in as options now, so we aren't susceptible to those side effects in the meantime.

amckinney:
> Since headers is in $options and is sent as part of $guzzleClient->send($request, $options), the encodeHeaders headers may be overridden.

Nice catch! I didn't even realize this side effect was happening in the `send` call - I updated to include a separate [toGuzzleOptions](https://github.com/fern-api/fern/pull/5970/commits/e2e376219c21dfba10d7d0506ee5048e92d5d23d) helper method so we explicitly prune off any unintentional options (like you suggested).

> This may not be relevant at this moment yet, but how is bodyProperties going to affect multi-part form bodies?

Good shout - I meant to handle this upfront. I refactored the generator to omit the `bodyProperties` request option entirely for multipart request options [here](https://github.com/fern-api/fern/pull/5970/commits/f819970024724c061bae8f05e55d0693053b19ff). That way we don't have any silent failures, and we can take our time to add support for multipart requests later (if ever). It's unclear how well that would work given that you need to serialize the JSON to a specific "part", so we may never add that request option there.

generators/php/base/src/asIs/Client/RawClient.Template.php
```diff
@@ -87,60 +112,110 @@ private function buildRequest(
...
+        if ($body instanceof JsonSerializable) {
+            $serialized = $body->jsonSerialize();
+            return is_array($serialized)
+                ? array_merge($serialized, $overrideProperties)
+                : $serialized;
+        }
+        return is_array($body)
+            ? array_merge($body, $overrideProperties)
+            : $body;
+    }
```
Swimburger, line 187: Can `bodyProperties` do deep properties?
If so, we should probably use `array_merge_recursive`

Swimburger (approved):

generators/php/base/src/asIs/Client/RawClient.Template.php
amckinney: Good call out - for now, I'd opt to keep recursive merging out to start since there might otherwise be some surprising behavior. I could imagine that this changes in the future though.

PR #5977, opened 2025-02-12
fix(csharp): support required literals in the SDK code gen

dsinghvi:
## Description
Support initializing literal properties with their value. 

## Changes Made
- Updated `ObjectGenerator.ts` to handle required literal properties.

## Testing

### DIFF
--- a/generators/csharp/model/src/object/ObjectGenerator.ts
+++ b/generators/csharp/model/src/object/ObjectGenerator.ts
@@ -6,8 +6,7 @@ import {
     NameAndWireValue,
     ObjectProperty,
     ObjectTypeDeclaration,
-    TypeDeclaration,
-    TypeReference
+    TypeDeclaration
 } from "@fern-fern/ir-sdk/api";
 
 import { ModelCustomConfigSchema } from "../ModelCustomConfig";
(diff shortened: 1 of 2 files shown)


generators/csharp/model/src/object/ObjectGenerator.ts
```diff
@@ -28,6 +29,37 @@ export class ObjectGenerator extends FileGenerator<CSharpFile, ModelCustomConfig
...
+                        return literal.string;
+                    case "boolean":
+                        return literal.boolean;
+                    default:
+                        return undefined;
+                }
+            }
+        }
+        return undefined;
+    }
```
amckinney, line 61: We can recursively call the method to prevent duplicative logic.
```suggestion
    private maybeGetLiteral({ typeReference }: { typeReference: TypeReference }): string | boolean | undefined {
        if (typeReference.type === "container" && typeReference.container.type === "literal") {
            const literal = typeReference.container.literal;
            switch (literal.type) {
                case "string":
                    return literal.string;
                case "boolean":
                    return literal.boolean;
                default:
                    return undefined;
            }
        }
        if (typeReference.type === "named") {
            const typeDeclaration = this.context.getTypeDeclarationOrThrow(typeReference.typeId);
            if (typeDeclaration.shape.type === "alias") {
                return this.maybeGetLiteral(typeDeclaration.shape.resolvedType);
            }
        }
        return undefined;
    }
```

amckinney (approved):

generators/csharp/model/src/object/ObjectGenerator.ts
```diff
@@ -40,16 +72,24 @@ export class ObjectGenerator extends FileGenerator<CSharpFile, ModelCustomConfig
...
-                    set: true,
+                    set: !isLiteral,
                     summary: property.docs,
                     jsonPropertyName: property.name.wireValue,
-                    useRequired: true
+                    useRequired: true,
+                    initializer: maybeLiteralInitializer != null 
+                        ? csharp.codeblock(typeof maybeLiteralInitializer === "boolean" 
+                            ? `${maybeLiteralInitializer.toString().toLowerCase()}`
+                            : `"${maybeLiteralInitializer}`)
```
Swimburger, line 90: I'm assuming this is a string, I don't see the closing quotes.

PR #5986, opened 2025-02-13
fix(php): JSON serialize empty arrays as empty object

Swimburger:
## Description
JSON serialize empty arrays as empty object

## Changes Made
- If incoming array is sequential, return as array and let PHP JSON serializer handle as is
- Else, continue as usual, but if result is an empty array, cast to object to force PHP JSON serializer to serialize it as an empty object.

## Testing

- [x] Seed

### DIFF
--- a/packages/cli/generation/ir-generator-tests/src/dynamic-snippets/__test__/test-definitions/bearer-token-environment-variable.json
+++ b/packages/cli/generation/ir-generator-tests/src/dynamic-snippets/__test__/test-definitions/bearer-token-environment-variable.json
@@ -1,7 +1,39 @@
 {
     "version": "1.0.0",
     "types": {},
-    "headers": [],
+    "headers": [
+        {
+            "name": {
+                "name": {
+                    "originalName": "version",
+                    "camelCase": {
+                        "unsafeName": "version",
+                        "safeName": "version"
+                    },
+                    "snakeCase": {
+                        "unsafeName": "version",
+                        "safeName": "version"
+                    },
+                    "screamingSnakeCase": {
+                        "unsafeName": "VERSION",
+                        "safeName": "VERSION"
+                    },
+                    "pascalCase": {
+                        "unsafeName": "Version",
+                        "safeName": "Version"
+                    }
+                },
+                "wireValue": "X-API-Version"
+            },
+            "typeReference": {
+                "type": "literal",
+                "value": {
+                    "type": "string",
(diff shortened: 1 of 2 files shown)


generators/php/base/src/asIs/Client/RawClient.Template.php
```diff
@@ -247,4 +261,21 @@ private function encodeQueryValue(mixed $value): string {
         // Unreachable, but included for a best effort.
         return urlencode(strval(json_encode($value)));
     }
+
+    /**
+     * Check if an array is sequential, not associative.
+     * @param array $arr
+     * @return bool
+     */
+    private static function isSequential(array $arr): bool { // @phpstan-ignore-line
```
amckinney, line 270: What are we ignoring here?

generators/php/base/src/asIs/Client/RawClient.Template.php
```diff
@@ -178,17 +179,28 @@ private function encodeRequestBody(
     private function buildJsonBody(
         mixed $body,
         array $options,
-    ): mixed {
+    ): mixed
+    {
         $overrideProperties = $options['bodyProperties'] ?? [];
+        if(is_array($body) && self::isSequential($body)){
```
amckinney, line 185: nit: Let's tidy up the spacing on these:
```suggestion
        if (is_array($body) && self::isSequential($body)) {
```

generators/php/base/src/asIs/Client/RawClient.Template.php
```diff
@@ -178,17 +179,28 @@ private function encodeRequestBody(
...
+        else{
+            $result = $body;
+        }
+        if(is_array($result)){
+            $result = array_merge($result, $overrideProperties);
+            if(empty($result)){
+                // force to be serialized as {} instead of []
+                return (object)($result);
+            }
+        }
```
amckinney, line 201:
```suggestion
        if (is_array($result)){
            $result = array_merge($result, $overrideProperties);
            if (empty($result)){
                // force to be serialized as {} instead of []
                return (object)($result);
            }
        }
```

amckinney (approved):
Only nits, but lgtm!

generators/php/base/src/asIs/Client/RawClient.Template.php
Swimburger: phpstan wants `array` to have more information, but we don't have more info

PR #5987, opened 2025-02-13
fix(cli): Make sure node.exe is executed.

Koubek:
## Description
It looks like the changes are not correct, `node.exe` is probably not executed on Windows.

## Changes Made
Executing `node.exe` explicitly.

## Testing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,7 @@ jobs:
             exit 1
           }
           cd windows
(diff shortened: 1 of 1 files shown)


Koubek:
You can see [here](https://github.com/Koubek/fern/actions/runs/13313036324/job/37180205927) the CLI is executed correctly (sure, I am not logged-in but that's fine). I think your execution doesn't have any output which means it's really not executed at all.

PR #5991, opened 2025-02-13
feat(php): Support literal types

ajgateno:


amckinney:
Discussed offline w/ @ajgateno - with the amount of one-off handling we need to accommodate setting literals under the hood, this feels bug prone and easy to mess up.

To start, we'll take on a design similar to the literal experience in TypeScript - strongly type each call-site in PHPDoc comments, but still require the user to specify them by hand in the constructor. We can always improve this experience later on if there's enough reason to, but this feels like the right middle ground to start.

generators/php/base/src/context/PhpAttributeMapper.ts
```diff
@@ -148,8 +148,25 @@ export class PhpAttributeMapper {
...
+                return this.getLiteralArgument(type);
             default:
                 assertNever(type.internalType);
         }
     }
+
+    private getLiteralArgument(type: php.Type) {
+        if (type.internalType.type !== "literal") {
+            throw new Error("Expected provided type to be a literal.");
+        }
```
amckinney, line 161: I don't think we need to throw here - we've already narrowed the type in the `switch` above, so we should be able to narrow the type (rather than use `php.Type`). That will prevent us from needing to `throw` here.

generators/php/codegen/src/ast/TypeLiteral.ts
```diff
@@ -143,7 +143,11 @@ export class TypeLiteral extends AstNode {
                     this.writeStringWithHeredoc({ writer, value: this.internalType.value });
                     break;
                 }
-                writer.write(`'${this.internalType.value.replaceAll("'", "\\'")}'`);
+                if (this.internalType.value.includes("'")) {
+                    writer.write(`"${this.internalType.value.replaceAll('"', '\\"')}"`);
+                    break;
+                }
+                writer.write(`'${this.internalType.value}'`);
```
amckinney, line 150: Nice 👏

seed/php-sdk/unions/no-custom-config/src/Types/Types/UnionWithLiteral.php
```diff
@@ -159,9 +165,9 @@ public static function jsonDeserialize(array $data): static
                 "JSON data is missing property 'base'",
             );
         }
-        if (!(is_string($data['base']))) {
+        if (!($data['base'] === 'base')) {
             throw new Exception(
-                "Expected property 'base' in JSON data to be string, instead received " . get_debug_type($data['base']),
+                "Expected property 'base' in JSON data to be literal, instead received " . get_debug_type($data['base']),
```
amckinney, line 170: Should this error message reference the `'base'` string instead? Otherwise the user won't know what's actually wrong.

```suggestion
                "Expected property 'base' in JSON data to be 'base', instead received " . get_debug_type($data['base']),
```

amckinney (commented):
Few nits + updates in the error messages we throw.

amckinney (approved):
Looks good - some of the large unions have _really_ large PHPDoc comments, but it feels safer to start in the more strict manner, and relax that requirement later on (if ever).

PR #5993, opened 2025-02-13
fix(php): Make constructor private for union types and remove static unknown constructor

ajgateno:


Swimburger (approved):

generators/php/codegen/src/ast/DataClass.ts
```diff
@@ -14,18 +14,23 @@ import { orderByAccess } from "./utils/orderByAccess";
 const CONSTRUCTOR_PARAMETER_NAME = "values";
 
 export declare namespace DataClass {
-    type Args = Class.Args;
+    interface DataClassArgs {
+        constructorAccess?: Access;
+    }
+    type Args = Class.Args & DataClassArgs;
```
amckinney, line 20: Ideally we could use `extends` here instead - does this work?

```typescript
type Args extends Class.Args {
    constructorAccess?: Access;
}
```

amckinney (approved):
Nice 👏

PR #5994, opened 2025-02-13
feat(csharp): support skipResponseValidation config in c#

dsinghvi:
## Description
Add new configuration `skipResponseValidation` to skip response validation. When set to true, the generated client will fail fast if the response schema does not match the generated code. This is particularly useful if you are concerned about your API Definition (e.g. OpenAPI, AsyncAPI) not being entirely correct. 

```
  version: 1.9.22
  config: 
    skip-response-validation: true
```

## Changes Made
- Updated `JsonUtils.Template.cs`
- Added a new method to the `SDKGeneratorContext` that returns `Deserialize` or `DeserializeWithFallbacks` depending on the feature flag

## Testing


amckinney (approved):

generators/csharp/codegen/src/asIs/JsonConfiguration.Template.cs
```diff
@@ -17,6 +18,24 @@ static JsonOptions()
...
+                    {
+                        foreach (var propertyInfo in typeInfo.Properties)
+                        {
+                            // Strip IsRequired constraint from every property.
+                            propertyInfo.IsRequired = false;
+                        }
+                    }
+                }
+            }
+        };
```
Swimburger, line 38: The `ConfigureJsonSerializerOptions` method is an extension point for SDK maintainers so they can customize some of the options.
I think that this would prevent them from modifying the type info resolvers and modifiers like we are doing.

Swimburger:
I'd rather see a single Serialize/Deserialize method. Is there a reason we would surface a validating and non-validating variant at the same time?

dsinghvi:
> I'd rather see a single Serialize/Deserialize method. Is there a reason we would surface a validating and non-validating variant at the same time?

Yes -- we would surface it so that SDK users can decide how they want validation to work. For example, internal consumers of the SDK would turn validation on and external ones would keep it off.

generators/csharp/codegen/src/asIs/JsonConfiguration.Template.cs
dsinghvi: yeah thats fair, i'm not sure how valuable it is to keep that extension method

generators/csharp/codegen/src/asIs/JsonConfiguration.Template.cs
```diff
@@ -39,4 +58,9 @@ public static T Deserialize<T>(string json)
     {
         return JsonSerializer.Deserialize<T>(json, JsonOptions.JsonSerializerOptions)!;
     }
+
+    public static T DeserializeWithFallback<T>(string json)
```
dsinghvi, line 62: swap out public -> internal

generators/csharp/codegen/src/asIs/JsonConfiguration.Template.cs
```diff
@@ -6,6 +6,7 @@ namespace <%= namespace%>;
 internal static partial class JsonOptions
 {
     public static readonly JsonSerializerOptions JsonSerializerOptions;
+    public static readonly JsonSerializerOptions JsonSerializerOptionsWithFallback;
```
dsinghvi, line 9: make it internal

PR #5995, opened 2025-02-14
fix(cli): Clean up ApiDefinitionLatest Nav Converter

RohinBhargava:
## Description
* Previously, there were issues with the nav converter resolving the right nav components, now we create a dedicated holder to resolve nav components, using the same generative logic for locators as the legacy holder
* Cleaned up ids for sensible id-ing and namespacing
* Upgrades the OpenApi v2 Parser to 0.0.57 
* Removes unnecessary merge step with legacy parser (improves performance)

## Changes Made
- Added `ApiDefinitionHolderLatest` class
- Increases `docs-parsers` package version
- Removes erroneous, useless merge step (moving into FDR)

## Testing
Unit tests and testing on customer specs


packages/cli/docs-resolver/src/utils/getLatestUrlSlug.ts
```diff
@@ -0,0 +1,22 @@
+import { kebabCase } from "lodash-es";
+import urlJoin from "url-join";
+
+import { FdrAPI } from "@fern-api/fdr-sdk";
+
+function getLatestUrlSlug<T extends { namespace?: string[]; id: string; operationId?: string }>(item: T): string {
```
dsinghvi, line 6: what does it mean to get a latest url slug?

dsinghvi (approved):

packages/cli/docs-resolver/src/utils/getLatestUrlSlug.ts
RohinBhargava: It's computing a slug for the latest nav structure. Changing the name!

PR #5997, opened 2025-02-14
feat(cli): AsyncAPI 3.0 Parser v1

eyw520:
## Description
This PR implements the `parseAsyncAPIV3` method, which enables conditional parsing based on the provided AsyncAPI spec version. 

## Changes Made
- Update `OpenAPI-IR` schema to support multiple channels for both v2 and v3 parsing.
- `buildFernDefinition` now writes several Websocket definition files, each of which corresponds to a single AsyncAPI channel (for both v2 and v3).
- Refactor out `AbstractAsyncAPIParserContext`, which shares common utils used in v2 and v3 parsing.
- Define updated v3 AsyncAPI types, with the most significant change being the decoupling of Channels and Operations.
- Custom parsing logic for Channels, Operations, and Channel Parameters.

## Upcoming Work
- Support adding AsyncAPI v3 example and parsing examples accordingly. 
- ... and more to come!

## Testing

This change has been tested on the AsyncAPI specs for Deepgram (v3), Paradex (v2), and Hume (v2).

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,5 +1,12 @@
 - changelogEntry:
     - summary: |
+        Introduce the AsyncAPI 3.0.0 parser, which enables conditional parsing of both 2.x and 3.x AsyncAPI specs.
+      type: fix
+  irVersion: 55
+  version: 0.52.0
+
+- changelogEntry:
+    - summary: |
         The OpenAPI v2 parser now respects object properties as optional, generates display names for undiscriminated union object 
         types, and fixes minor issues with docs dev when using specific navigation locators in docs.yml specification.
       type: fix


packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/AbstractAsyncAPIParserContext.ts
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public resolveSchemaReference(schema: OpenAPIV3.ReferenceObject): OpenAPIV3.SchemaObject {
+        if (!schema.$ref.startsWith(SCHEMA_REFERENCE_PREFIX)) {
+            throw new Error(`Failed to resolve schema reference: ${schema.$ref}`);
+        }
+
+        const schemaKey = schema.$ref.substring(SCHEMA_REFERENCE_PREFIX.length);
+        const splitSchemaKey = schemaKey.split("/");
+
+        const components = (this.document as any).components;
```
dsinghvi, line 83: why do we have to do a cast here? (generally try to stay away from `any` since it could create weird runtime issues)

packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/AbstractAsyncAPIParserContext.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * Resolves a schema reference of the form #/components/schemas/...
+     * Shared logic across v2 and v3 specs, so long as the shape matches:
+     *   this.document.components.schemas[...]
+     */
+    public resolveSchemaReference(schema: OpenAPIV3.ReferenceObject): OpenAPIV3.SchemaObject {
+        if (!schema.$ref.startsWith(SCHEMA_REFERENCE_PREFIX)) {
+            throw new Error(`Failed to resolve schema reference: ${schema.$ref}`);
```
dsinghvi, line 77: I think a better UX here is going to be to log an error and keep going? That way someone can still somewhat generate docs/SDKs even if their spec is slightly messed up

dsinghvi: @eyw520 might be doing a scan and reducing the number of errors thrown -- no issues if this is necessary though

packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/AbstractAsyncAPIParserContext.ts
eyw520: Resolved.

dsinghvi (approved):

PR #6036, opened 2025-02-18
feat(java): Flat package layout

ajgateno:
This PR adds a config flag `page-layout` that can be set to `nested` or `flat`, and defaults to `nested`. Setting it to `flat` causes us not to add any extra nesting portions to package paths meaning something like `com.seed.api.resources.service.requests` turns into `com.seed.api.service`


generators/java/generator-utils/src/main/java/com/fern/java/AbstractPoetClassNameFactory.java
```diff
@@ -34,13 +37,25 @@ public final ClassName getRootClassName(String className) {
     }
 
     public final String getCorePackage() {
-        List<String> tokens = new ArrayList<>(packagePrefixTokens);
-        tokens.add("core");
-        return String.join(".", tokens);
+        switch (packageLayout) {
+            case FLAT:
+                return String.join(".", packagePrefixTokens);
```
amckinney, line 42: Let's still use `core` here since those are our core utilities.

generators/java/generator-utils/src/main/java/com/fern/java/AbstractPoetClassNameFactory.java
```diff
@@ -34,13 +37,25 @@ public final ClassName getRootClassName(String className) {
...
+                tokens.add("core");
+                return String.join(".", tokens);
+        }
     }
 
     public final String getPaginationPackage() {
-        return getCorePackage() + ".pagination";
+        switch (packageLayout) {
+            case FLAT:
+                return getCorePackage();
```
amckinney, line 54: We _might_ want to keep the nested packages within `core`, but if this is compatible as-is, then this might be a nice improvement too.

seed/java-sdk/exhaustive/flat-package-layout/src/main/java/com/seed/exhaustive/types/object/NestedObjectWithOptionalField.java
```diff
@@ -0,0 +1,120 @@
+/**
+ * This file was auto-generated by Fern from our API Definition.
+ */
+package com.seed.exhaustive.types.object;
```
amckinney, line 4: Should this be in `com.seed.exhaustive.object`?

ajgateno: Just checked again, and this is fine because in the test definition it's in the path `definition/types/object.yml`

seed/java-sdk/exhaustive/flat-package-layout/src/main/java/com/seed/exhaustive/types/types/NestedObjectWithOptionalField.java
```diff
@@ -0,0 +1,120 @@
+/**
+ * This file was auto-generated by Fern from our API Definition.
+ */
+package com.seed.exhaustive.types.types;
```
ajgateno, line 4: This `types.types` here is because we have a folder called `types` and we append `types` for the model folder. We typically wouldn't have this otherwise.

amckinney (approved):
Nice looks good! Couple comments on implementation nits, but it'd also be great to configure the `flat-package-layout` on another simple test definition like `imdb` so that it's easy to see the impact on those SDKs too.

PR #6045, opened 2025-02-19
fix(cli): add async api support in specs & migrate generators.yml to api.specs[]

Swimburger:
## Changes Made
- Fix issue where the CLI would not parse asyncapi specs within the `api.specs` array in generators.yml.
- Migrate old spec configuration in generators.yml to the `api.specs[]` format.

## Testing

### DIFF
--- a/packages/cli/cli-migrations/src/migrations/0.54.0-rc0/use-generators-yml-specs/__test__/use-generators-yml-specs.test.ts
+++ b/packages/cli/cli-migrations/src/migrations/0.54.0-rc0/use-generators-yml-specs/__test__/use-generators-yml-specs.test.ts
@@ -1,6 +1,5 @@
 import { cp, readFile } from "fs/promises";
 import tmp from "tmp-promise";
-import { describe, expect, it } from "vitest";
 
 import { AbsoluteFilePath, RelativeFilePath, join } from "@fern-api/fs-utils";
 import { createMockTaskContext } from "@fern-api/task-context";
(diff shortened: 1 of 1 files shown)


packages/cli/cli/versions.yml
```diff
@@ -1,3 +1,13 @@
+- changelogEntry:
+    - summary: |
+        Fix issue where the CLI would not parse asyncapi specs within the `api.specs` array in generators.yml.
+      type: fix
+    - summary: |
+        Migrate old spec configuration in generators.yml to the `api.specs[]` format.
+      type: feat
+  irVersion: 55
+  version: 0.53.15
```
dsinghvi, line 9: lets make sure to release this as 0.54.0 since there is technically a break here

fern/apis/generators-yml/definition/generators.yml
```diff
@@ -290,30 +308,39 @@ types:
...
-      - AsyncAPISchema
\ No newline at end of file
+      - AsyncAPISpecSchema
+      - ProtobufSpecSchema
+
+  ProtobufSpecSchema:
+    extends:
+      - BaseProtobufDefinitionSchema
+    properties:
+      proto:
```
Swimburger, line 344: @amckinney Does this make sense, or should we discriminate another way?

amckinney (approved):
Implementation looks good to me. We'll need to adjust the implementation based on the decision to prioritize `target` vs. `root`, or require a discriminated union but otherwise good to on my end.

PR #6068, opened 2025-02-19
chore(cli): log docs violations summary at the end + include elapsed time

prb2:
## Description
This PR makes some minor improvements to violations summary displayed by `fern check`.

## Changes Made
- moved `logViolationsSummary` call to the end of `logViolations` to ensure it is visible even when there are a lot of violations
- added an optional `elapsedMillis` parameter; if provided, it is included in the summary
- moved oss validation into validateAPIWorkspaceAndLogIssues.ts so that we don't print multiple summary lines per workspace type

## Testing

**Before**
```
> fern check

[docs]: Found 0 errors and 2 warnings. Run fern check --warnings to print out the warnings not shown.
[sdks]: ✓ All checks passed
[v1]:   ✓ All checks passed
[v2]:   ✓ All checks passed
[v1]:   Found 0 errors and 8 warnings. Run fern check --warnings to print out the warnings not shown.
[v2]:   Found 0 errors and 8 warnings. Run fern check --warnings to print out the warnings not shown.
[sdks]: Found 0 errors and 8 warnings. Run fern check --warnings to print out the warnings not shown.
```

**After**
```
> pnpm fern-dev:build
> FERN_NO_VERSION_REDIRECTION=true node ~/fern/fern/packages/cli/cli/dist/dev/cli.cjs check

[docs]: Found 0 errors and 2 warnings in 5.730 seconds. Run fern check --warnings to print out the warnings not shown.
[v1]:   Found 0 errors and 8 warnings in 0.395 seconds. Run fern check --warnings to print out the warnings not shown.
[v2]:   Found 0 errors and 8 warnings in 0.112 seconds. Run fern check --warnings to print out the warnings not shown.
[sdks]: Found 0 errors and 8 warnings in 0.442 seconds. Run fern check --warnings to print out the warnings not shown.
```

### DIFF
--- a/packages/cli/ete-tests/src/tests/validate/__snapshots__/validate.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/validate/__snapshots__/validate.test.ts.snap
@@ -14,5 +14,5 @@ exports[`validate > simple 1`] = `
        error-discrimination is required because this API has declared errors.
 [api]: other.yml -> types -> MyType
        Type MissingType is not defined.
-[api]: Found 2 errors and 0 warnings in 0.009 seconds."
+[api]: Found 2 errors and 0 warnings in"
 `;
--- a/packages/cli/ete-tests/src/tests/validate/validate.test.ts
+++ b/packages/cli/ete-tests/src/tests/validate/validate.test.ts
@@ -27,11 +27,24 @@ function itFixture(fixtureName: string) {
                 reject: false
             });
 
-            expect(
-                stripAnsi(stdout)
-                    // for some reason, locally the output contains a newline that Circle doesn't
-                    .trim()
-            ).toMatchSnapshot();
+            if (fixtureName == "simple") {
+                expect(
+                    stripAnsi(stdout)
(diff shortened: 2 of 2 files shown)


dsinghvi:
@prb2 should we only show `All checks passed` if there are 0 errors and 0 warnings?

dsinghvi (approved):
To release this you'll want to add a changelog entry here https://github.com/fern-api/fern/blob/main/packages/cli/cli/versions.yml

prb2:
> @prb2 should we only show `All checks passed` if there are 0 errors and 0 warnings?

We should be doing that already:

https://github.com/fern-api/fern/blob/ecc1d3cc0cd8e29de93581c2e1b043816f4c8cf5/packages/cli/cli/src/commands/validate/logViolations.ts#L131-L137

My guess is that `fern check` is running multiple different validators that more than one of them use the same logger prefix, so that's why we see two lines.

I agree this is confusing but not sure of the best way to fix it. Should we try to use unique prefixes somehow? I think it's being derived from the workspace

fern-support (approved):

PR #6077, opened 2025-02-19
fix(cli): OpenApi Parser v2 Slug Generation when no Nav config is provided

RohinBhargava:
## Description
Previously, when no nav structure is provided in `docs.yml`, the output docs configuration generated incorrect slugs. Now, we generate the right slugs.

## Changes Made
- Changed slug generation strategy -- namespaced id when no operation id provided, or just base operation id if provided
- Refactored the logic to make it more readable/reusable

## Testing
Tested on 3 customer specs (one small w/ nav, one small w/out nav, one large with mixed nav/no nav)

### DIFF
--- a/fern/pages/changelogs/java-sdk/2025-02-19.mdx
+++ b/fern/pages/changelogs/java-sdk/2025-02-19.mdx
@@ -1,4 +1,3 @@
-## 2.14.0
-**`(feat):`** Support user-agent header
-
+## 2.15.0
+**`(feat):`** Add support for custom HTTP client in Java SDK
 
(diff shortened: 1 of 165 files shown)


Swimburger (approved):

packages/cli/docs-resolver/src/utils/getApiLatestToNavigationNodeUrlSlug.ts
```diff
@@ -1,26 +1,13 @@
...
-
-export function getApiLatestWebhookToNavigationNodeUrlSlug(
-    webhook: FdrAPI.api.latest.webhook.WebhookDefinition
-): string {
-    return getApiLatestToNavigationNodeUrlSlug(webhook);
+export function getApiLatestToNavigationNodeUrlSlug<T extends { id: string; operationId?: string }>(
+    item: T,
+    parentSlug: FernNavigation.V1.SlugGenerator,
+    namespacedSlug: FernNavigation.V1.SlugGenerator
+): FernNavigation.V1.Slug {
```
dsinghvi, line 9: @RohinBhargava can you make sure to use an object in these cases, the downside of swapping parentSlug and namespacedSlug is pretty high and too easy

PR #6082, opened 2025-02-19
fix(cli): improve broken-links output

prb2:
## Description
This PR improves the display of the `fern docs broken-links` command to make the output easier to understand.

## Changes Made
- added an optional `logBreadcrumbs` parameter to `logViolations`
  - defaults to `true`, but for the docs specific validators we set to `false` since the breadcrumbs aren't as useful in this context
- grouped violations by file and used the relative file path as a header to reduce clutter/improve scan-ability
- reworded violation message to minimize "wall of text"
- "fix here" file path with line/character position can be opened with CMD/CTRL + Click (opens in application configured for the filetype)

## Testing

Added new ete test and checked new output locally. Tested "fix here" links work in iTerm and VSCode's integrated terminal.

**Before**
Screenshot 2025-02-19 at 4 01 35 PM 

**After**

Screenshot 2025-02-21 at 3 08 43 PM


packages/cli/cli/src/commands/validate/logViolations.ts
```diff
@@ -52,13 +54,15 @@ function groupViolationsByNodePath(violations: ValidationViolation[]): Map<NodeP
 
 function logViolationsGroup({
     logWarnings,
-    relativeFilepath,
+    logBreadcrumbs,
+    configRelativeFilepath,
```
prb2, line 58: I renamed `relativeFilepath` to `configRelativeFilepath` because this value was always `docs.yaml` and not the actual filepath that these violations came from.

Let me know if there's a better name or I should change it back.

prb2:
Let me know if you'll have any suggestions for additional visual improvements or message wording!

dsinghvi:
@prb2 can you use terminal link for the `Fix here` ? https://github.com/sindresorhus/terminal-link

dsinghvi (approved):

PR #6092, opened 2025-02-20
chore(java): Upgrade IR dependency to v55

ajgateno:


seed/java-spring/object/types/Type.java
```diff
@@ -71,19 +71,19 @@ public final class Type {
 
   private final Name nineteen;
 
-  private final int twenty;
+  private final long twenty;
```
amckinney, line 74: Discussed offline - we should match the semantics of Protobuf [here](https://protobuf.dev/programming-guides/proto3/). In this case, let's define this as an `int` (since it's typed as a `uint32`).

PR #6103, opened 2025-02-20
fix(cli): improve image path parsing

abvthecity:
This PR applies some of the learnings from fern-platform around traversing MDX and estree for a slightly more succinct and robust implementation.

### DIFF
--- a/packages/cli/docs-markdown-utils/src/walk-estree-jsx-attributes.ts
+++ b/packages/cli/docs-markdown-utils/src/walk-estree-jsx-attributes.ts
@@ -7,7 +7,7 @@ export function walkEstreeJsxAttributes(
     handlers: Record<string, (attr: JSXAttribute, node: JSXElement) => void>
 ): void {
     walk(estree, {
-        enter (node) {
+        enter(node) {
             if (node.type === "JSXElement") {
                 node.openingElement.attributes.forEach((attr) => {
                     if (


dsinghvi:
@abvthecity this seems like a sensitive code path to change -- how do you want to test this/roll out to make sure there are no regressions?

abvthecity:
> @abvthecity this seems like a sensitive code path to change -- how do you want to test this/roll out to make sure there are no regressions?

@dsinghvi there's a pretty long list of tests that this change passes: https://github.com/fern-api/fern/blob/fa3134372a67a449c1b51fcf5291bf1076ef21b7/packages/cli/docs-markdown-utils/src/__test__/parseImagePaths.test.ts

i added 2 more tests (for video and the Download component) but otherwise I'm confident this approach is an improvement.

packages/cli/cli/versions.yml
```diff
@@ -11,6 +11,14 @@
   irVersion: 55
   version: 0.54.0-rc0
 
+- changelogEntry:
+    - summary: |
+        Update CLI-side markdown parsing to detect more hrefs and srcs (where src files are automatically uploaded),
+        enabling the <Download src="./path/to/file.pdf">Download me</Download> pattern to work.
+      type: fix
+  irVersion: 55
+  version: 0.53.19
```
dsinghvi, line 20: i dont think our automation is smart enough to release this -- youll need to release an rc after 0.54.0

dsinghvi (approved):

PR #6121, opened 2025-02-21
fix(cli): improve generators-yml-use-api-specs migration

zachkirsch:
## Description

Stumbled upon some issues with this migration when trying to hunt down a different bug. Figured I would PR!

Diff is a little nicer with whitespace changes hidden: https://github.com/fern-api/fern/pull/6121/files?diff=split&w=1

## Changes Made
- Handles case where the API spec lives in a folder (e.g. `fern/openapi/openapi.yml`) (added test that fails on main)
- Handles case where `apis/` folder is used (added test that fails on main)
- `api` is now modified in-place rather than moved to the bottom of `generators.yml`
- Don't add empty `settings: {}` key

## Testing

### DIFF
--- a/packages/cli/cli-migrations/src/migrations/0.54.0-rc0/use-generators-yml-specs/__test__/use-generators-yml-specs.test.ts
+++ b/packages/cli/cli-migrations/src/migrations/0.54.0-rc0/use-generators-yml-specs/__test__/use-generators-yml-specs.test.ts
@@ -48,7 +48,7 @@ describe("use-generators-yml-specs", () => {
         }
     });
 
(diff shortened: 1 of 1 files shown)


zachkirsch:
I think CI is failing because on the fork I don't have access to some/all of these secrets https://github.com/fern-api/fern/blob/17c2c83471910e9a9c1c6b90b86285b5857ac4ba/.github/workflows/ci.yml#L162-L164

dsinghvi (approved):

Swimburger:
@zachkirsch Thank you for helping here!

PR #6122, opened 2025-02-21
fix(cli): allow running `fern upgrade` in locally-built CLI

zachkirsch:
## Description

When running `fern upgrade`, I got this error:

```
ERROR 2025-02-21T18:13:32.192Z Failed.
ERROR 2025-02-21T18:13:32.192Z Cannot read properties of undefined (reading 'includes')
DEBUG 2025-02-21T18:13:32.192Z TypeError: Cannot read properties of undefined (reading 'includes')
                                   at eUi (/Users/zach/.npm/_npx/0915e03ab7b9b37f/node_modules/fern-api/bundle.cjs:2280:162)
                                   at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
                                   at async ojg (/Users/zach/.npm/_npx/0915e03ab7b9b37f/node_modules/fern-api/bundle.cjs:2280:1605)
                                   at async Object.handler (/Users/zach/.npm/_npx/0915e03ab7b9b37f/node_modules/fern-api/bundle.cjs:2280:12971)
```

It took me a while to trace down the bug because I couldn't run `fern upgrade` using the locally-built version of the CLI (it just always prints `No upgrade available.`).

With this change, you can now run `fern upgrade` using the locally-built version and it works just like the published CLI. Helps a lot with debugging upgrade issues.

(Turns out the original bug was actually fixed a while ago: https://github.com/fern-api/fern/pull/4675)

## Changes Made
- Change the return types of `loggingExeca` to match reality (i.e. `stdout` and `stderr` can be undefined
- Change the `upgrade` CLI command logic to actually perform an upgrade, even if we're in dev mode (i.e. version is `0.0.0`)
- Don't minify the CLI in production

## Testing

### DIFF
--- a/packages/cli/cli/build.prod.cjs
+++ b/packages/cli/cli/build.prod.cjs
@@ -9,7 +9,7 @@ async function main() {
     await tsup.build({
         entry: ['src/cli.ts'],
         format: ['cjs'],
-        minify: false,
+        minify: true,
         outDir: 'dist/prod',
         env: {
             AUTH0_DOMAIN: "fern-prod.us.auth0.com",


packages/cli/cli/build.prod.cjs
```diff
@@ -9,7 +9,7 @@ async function main() {
     await tsup.build({
         entry: ['src/cli.ts'],
         format: ['cjs'],
-        minify: true,
+        minify: false,
```
zachkirsch, line 12: Happy to revert this, it adds an extra 30mb to the CLI (~30mb -> ~60mb). But I imagine it would make understanding user's error messages a lot easier. (e.g. things like this: https://github.com/fern-api/fern/pull/4675)

packages/cli/cli/src/cli-context/upgrade-utils/getLatestVersionOfCli.ts
```diff
@@ -9,12 +9,7 @@ export async function getLatestVersionOfCli({
     cliEnvironment: CliEnvironment;
     includePreReleases?: boolean;
 }): Promise<string> {
-    // when running a non-published version of the CLI (e.g. in ETE tests),
-    // don't try to upgrade
-    if (cliEnvironment.packageVersion === "0.0.0") {
-        return cliEnvironment.packageVersion;
-    }
-    return latestVersion(cliEnvironment.packageName, {
+    return latestVersion("fern-api", {
```
zachkirsch, line 12: hardcoding `"fern-api"` because now when running `fern upgrade` with the dev-cli I get `Package `@fern-api/fern-api-dev` could not be found`

packages/commons/logging-execa/src/loggingExeca.ts
```diff
@@ -9,15 +9,17 @@ export declare namespace loggingExeca {
         substitutions?: Record<string, string>;
     }
 
-    export type ReturnValue = ExecaReturnValue;
+    export type Return = Omit<ExecaReturnValue, "stdout" | "stderr"> &
+        // the execa types are incorrect and sometimes stdout and stderr and not defined
+        Partial<Pick<ExecaReturnValue, "stdout" | "stderr">>;
```
zachkirsch, line 14: this was the root cause of https://github.com/fern-api/fern/pull/4675, so updating the types here to prevent it from happening elsewhere

packages/commons/logging-execa/src/loggingExeca.ts
```diff
@@ -37,5 +37,14 @@ export async function loggingExeca(
...
+    return command.then((result) => {
+        // the execa types are incorrect and sometimes stdout and stderr and not defined
+        if (result.stdout == null) {
+            result.stdout = "";
+        }
+        if (result.stderr == null) {
+            result.stderr = "";
+        }
+        return result;
+    });
```
zachkirsch, line 49: this was the root cause of https://github.com/fern-api/fern/pull/4675, so updating the types here to prevent it from happening elsewhere

zachkirsch:
I think CI is failing because on the fork I don't have access to some/all of these secrets https://github.com/fern-api/fern/blob/17c2c83471910e9a9c1c6b90b86285b5857ac4ba/.github/workflows/ci.yml#L162-L164

packages/cli/cli/src/rerunFernCliAtVersion.ts
```diff
@@ -37,7 +37,7 @@ export async function rerunFernCliAtVersion({
             [FERN_CWD_ENV_VAR]: process.env[FERN_CWD_ENV_VAR] ?? process.cwd()
         }
     });
-    if (stdout?.includes("code EEXIST") || stderr?.includes("code EEXIST")) {
+    if (stdout.includes("code EEXIST") || stderr.includes("code EEXIST")) {
```
zachkirsch, line 40: Fixed upstream in `loggingExeca.ts`

PR #6125, opened 2025-02-21
feat(java): Support deep object query parameters

ajgateno:
Allows callers to send objects as `application/x-www-form-urlencoded` strings with [deepObject](https://swagger.io/docs/specification/v3_0/serialization/#:~:text=b%7Cc.-,deepObject,-%E2%80%93%20simple%20non%2Dnested) notation in query strings


generators/java/generator-utils/src/main/resources/QueryStringMapper.java
```diff
@@ -0,0 +1,91 @@
...
+        ObjectNode flat;
+
+        if (nested.isObject()) {
+            flat = flattenObject((ObjectNode) nested);
+        } else if (nested.isArray()) {
+            flat = flattenArray((ArrayNode) nested, "");
+        } else {
+            throw new IllegalArgumentException("Expected value to serialize to JSON object or array. Instead, got "
+                    + (value == null ? null : value.toString()));
+        }
```
amckinney, line 25: Given that this is a core utility we'll include in the SDK, I'll include a _mega nit_ - I like to place the variable adjacent to the control flow where it's set (i.e. without a newline), so that it's more logically grouped.

```suggestion
        ObjectNode flat;
        if (nested.isObject()) {
            flat = flattenObject((ObjectNode) nested);
        } else if (nested.isArray()) {
            flat = flattenArray((ArrayNode) nested, "");
        } else {
            throw new IllegalArgumentException("Expected value to serialize to JSON object or array. Instead, got "
                    + (value == null ? null : value.toString()));
        }
```

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/HttpUrlBuilder.java
```diff
@@ -140,21 +143,38 @@ private GeneratedHttpUrl generateUnInlineableCodeBlock(List<EnrichedObjectProper
...
-                    httpUrlname,
-                    queryParamProperty.wireKey().get(),
-                    PoetTypeNameStringifier.stringify(
-                            CodeBlock.of(
-                                            "$L.$N()" + (isOptional ? ".get()" : ""),
-                                            requestName,
-                                            queryParamProperty.getterProperty())
-                                    .toString(),
-                            queryParamProperty.poetTypeName()));
+            if (isCollection || isObject) {
```
amckinney, line 152: This looks good as-is for now, but the `isCollection` check made me consider how `allow-multiple` query parameters are being handled in general. This made me dig into the implementation a bit, and I noticed that `allow-multiple` is completely ignored in Java - the [filter property](https://github.com/fern-api/fern/blob/c32c0131fd770465e73ae999e953dda878805c4e/test-definitions/fern/apis/query-parameters/definition/user.yml#L38) is [typed as a string](https://github.com/fern-api/fern/blob/c32c0131fd770465e73ae999e953dda878805c4e/seed/java-sdk/query-parameters/src/main/java/com/seed/queryParameters/resources/user/requests/GetUsersRequest.java#L160) instead of a list of strings, and we just directly serialize the single `string` value [here](https://github.com/fern-api/fern/blob/c32c0131fd770465e73ae999e953dda878805c4e/seed/java-sdk/query-parameters/src/main/java/com/seed/queryParameters/resources/user/UserClient.java#L58C53-L58C62).

@dsinghvi are you aware of this gap? Am I misreading this at all? This would be a breaking change for anyone that defines an `allow-multiple` query parameter. 

@ajgateno this is unrelated to this PR - we can follow-up in a separate change, but we might need to come back to this `isCollection` check here since we can't always assume a collection should be serialized as `deepObject` (it will often require `explode` instead).

amckinney (commented):
Implementation looks good! I know you plan to add that `QueryStringMapperTest.java` file, so I'll wait to stamp until everything is ready to go.

generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/HttpUrlBuilder.java
ajgateno: Off the top of your head do you know an example? If not I can look for it

amckinney (approved):
Nice, the implementation looks great!

PR #6129, opened 2025-02-21
feat(csharp): Add support for different date serialization and deserialization in JSON

Swimburger:
## Changes Made
- Add support for generating dates. Configure the type of dates you generating using the `date-type` configuration option.
  - `use-date-only-portable` (default): Generates native `DateOnly` properties on .NET 6 and above, and falls back to using the third party `Portable.System.DateTimeOnly` package.
  - `use-date-time`: Generates dates as `DateTime` properties.
  - `use-date-only-on-net6+`: Generates native `DateOnly` properties on .NET 6 and above, and falls back to using `DateTime` on older versions of .NET.
- Add test for serializing and deserializing date times.
- Add test for serializing and deserializing dates.

## Testing


generators/csharp/sdk/versions.yml
```diff
@@ -7,6 +7,20 @@
...
+  createdAt: "2025-02-21"
+  irVersion: 53
+  changelogEntry:
+    - type: feat
+      summary: |
+        Add support for generating dates. Configure the type of dates you generating using the `date-type` configuration option.
+        `date-type` has the following options:
+        - `use-date-only-portable` (default): Generates native `DateOnly` properties on .NET 6 and above, and falls back to using the third party `Portable.System.DateTimeOnly` package.
+        - `use-date-time`: Generates dates as `DateTime` properties.
+        - `use-date-only-on-net6+`: Generates native `DateOnly` properties on .NET 6 and above, and falls back to using `DateTime` on older versions of .NET.
```
amckinney, line 20: Couple thoughts on the naming of this configuration:
1. The `use` prefix is redundant across every setting - I vote we drop it.
2. We could make it more clear that there is a fallback behavior.

I'd say we should have:

```yaml
date-type: date-only-with-portable-fallback
```

```yaml
date-type: date-only-with-date-time-fallback
```

```yaml
date-type: date-time
```

It's still a mouthful, but I think it's a little more clear to the user. Let me know what you think though - open to ideas here.

amckinney (approved):
Implementation looks great - only nit is on configuration naming and the `versions.yml` entry.

generators/csharp/sdk/versions.yml
Swimburger: That sounds better.

Swimburger:
We're changing course here.
After discussing this with Rohin, the real problem here is that we can't trust the portable date-only library, hence, we're offering these alternatives.
If we were to fork, or embed the portable date-only libraries as source code or as DLLs, we could do stick to only supporting the date only class.
The main benefit being, `DateOnly` will be consistently used across .NET framework and .NET.

PR #6148, opened 2025-02-24
feat(csharp): Support C# snippet configuration in docs.yml

amckinney:
This updates the `docs.yml` configuration to support the `csharp` snippets configuration like so:

```yaml
# docs.yml
navigation:
  - tab: api
    layout:
      - api: API Reference
        snippets:
           typescript: acme-node
           python: acme-python
           csharp: acme-csharp
           go: https://github.com/acme/acme-go
```

With this, users can start to surface their C# snippets in the API reference.

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,4 +1,10 @@
 - changelogEntry:
+  - summary: Add support for specifying C# snippets in the `generators.yml` file.
+    type: feat
+  irVersion: 55
+  version: 0.54.0-rc12
+
+- changelogEntry:
   - summary: Improve error logging in AsyncAPI v3 parser when parsing location headers.
     type: fix
   irVersion: 55


fern/apis/docs-yml/definition/docs.yml
```diff
@@ -731,6 +731,7 @@ types:
       go: optional<SnippetLanguageConfiguration>
       java: optional<SnippetLanguageConfiguration>
       ruby: optional<SnippetLanguageConfiguration>
+      csharp: optional<SnippetLanguageConfiguration>
```
dsinghvi, line 734: https://github.com/fern-api/fern/blob/e778edf22a7a050cf2d20716a4c23d6a840de515/packages/cli/configuration-loader/src/docs-yml/parseDocsConfiguration.ts#L859-L872 guessing we will need some plumbing

dsinghvi (approved):

PR #6173, opened 2025-02-25
Async java clients proposal

ajgateno:
Right now the general format of a call seems to be

```java
public Response endpoint(Request request, RequestOptions requestOptions) {
    Request request = requestInitialization();
    Client client = clientPreparation();
    try (Response response = client.newCall(request).execute()) {
        if (response.success()) {
            return MAPPER.read(response, Response.class);
        } else {
            throw new ApiException("oops");
        }
    } catch (IOException e) {
        throw new ApiException("different oops");
    }
}
```
I'm thinking of adapting this into

```java
private Request prepareEndpoint(Request request, RequestOptions requestOptions) {
    return requestInitialization();
}

public Response endpoint(Request request, RequestOptions requestOptions) {
    Client client = clientPreparation();
    try (Response response = client.newCall(prepareEndpoint(request, requestOptions)).execute()) {
        if (response.success()) {
            return MAPPER.read(response, Response.class);
        } else {
            throw new ApiException("oops");
        }
    } catch (IOException e) {
        throw new ApiException("different oops", e);
    }
}

public CompletableFuture<Response> endpointAsync(Request request, RequestOptions requestOptions) {
    Client client = clientPreparation();
    CompletableFuture<Response> future = new CompletableFuture<>();
    client.newCall(prepareEndpoint(request, requestOptions)).enqueue(
        new Callback() {
            @Override
            public void onFailure(Call call, IOException e) {
                future.completeExceptionally(new ApiException("different oops", e));
            }

            @Override
            public void onResponse(Call call, Response response) throws IOException {
                if (response.success()) {
                    future.complete(MAPPER.read(response, Response.class));
                } else {
                    future.completeExceptionally(new ApiException("oops"));
                }
            }
        }
    );
}
```

### DIFF
--- a/seed/java-sdk/exhaustive/no-custom-config/src/main/java/com/seed/exhaustive/resources/endpoints/container/ContainerClient.java
+++ b/seed/java-sdk/exhaustive/no-custom-config/src/main/java/com/seed/exhaustive/resources/endpoints/container/ContainerClient.java
@@ -17,6 +17,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import okhttp3.Call;
+import okhttp3.Callback;
 import okhttp3.Headers;
 import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
@@ -24,6 +27,7 @@ import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
 import okhttp3.ResponseBody;
+import org.jetbrains.annotations.NotNull;
 
 public class ContainerClient {
     protected final ClientOptions clientOptions;
@@ -32,11 +36,7 @@ public class ContainerClient {
         this.clientOptions = clientOptions;
     }
 
-    public List<String> getAndReturnListOfPrimitives(List<String> request) {
-        return getAndReturnListOfPrimitives(request, null);
-    }
-
-    public List<String> getAndReturnListOfPrimitives(List<String> request, RequestOptions requestOptions) {
+    private Request prepareGetAndReturnListOfPrimitives(List<String> request, RequestOptions requestOptions) {
         HttpUrl httpUrl = HttpUrl.parse(this.clientOptions.environment().getUrl())
                 .newBuilder()
                 .addPathSegments("container")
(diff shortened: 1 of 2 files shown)


seed/java-sdk/exhaustive/no-custom-config/src/main/java/com/seed/exhaustive/resources/endpoints/container/ContainerClient.java
```diff
@@ -75,6 +83,44 @@ public List<String> getAndReturnListOfPrimitives(List<String> request, RequestOp
...
+                                    responseBody.string(), new TypeReference<List<String>>() {}));
+                            return;
+                        }
+                        String responseBodyString = responseBody != null ? responseBody.string() : "{}";
+                        future.completeExceptionally(new SeedExhaustiveApiException(
+                                "Error with status code " + response.code(),
+                                response.code(),
+                                ObjectMappers.JSON_MAPPER.readValue(responseBodyString, Object.class)));
+                    }
+                });
```
dsinghvi, line 120: is there some sort of core utility we could reuse here -- like a class that implements `Callback` and takes a generic to do the JSON serialization ?

seed/java-sdk/exhaustive/no-custom-config/src/main/java/com/seed/exhaustive/resources/endpoints/container/ContainerClient.java
```diff
@@ -75,6 +83,44 @@ public List<String> getAndReturnListOfPrimitives(List<String> request, RequestOp
...
+                .enqueue(new Callback() {
+                    @Override
+                    public void onFailure(@NotNull Call call, @NotNull IOException e) {
+                        future.completeExceptionally(
+                                new SeedExhaustiveException("Network error executing HTTP request", e));
+                    }
+
+                    @Override
+                    public void onResponse(@NotNull Call call, @NotNull Response response) throws IOException {
+                        ResponseBody responseBody = response.body();
```
dsinghvi, line 108: I believe the `ResponseBody` is closeable so we would want to implement `try` with `resources`

seed/java-sdk/exhaustive/no-custom-config/src/main/java/com/seed/exhaustive/resources/endpoints/container/ContainerClient.java
```diff
@@ -75,6 +83,44 @@ public List<String> getAndReturnListOfPrimitives(List<String> request, RequestOp
...
+                        if (response.isSuccessful()) {
+                            future.complete(ObjectMappers.JSON_MAPPER.readValue(
+                                    responseBody.string(), new TypeReference<List<String>>() {}));
```
dsinghvi, line 111: i think you can only call `responseBody.string()` a single time so you may need this in a temp variable

seed/java-sdk/exhaustive/no-custom-config/src/main/java/com/seed/exhaustive/resources/endpoints/container/ContainerClient.java
```diff
@@ -75,6 +83,44 @@ public List<String> getAndReturnListOfPrimitives(List<String> request, RequestOp
...
+    }
+
+    public CompletableFuture<List<String>> getAndReturnListOfPrimitivesAsync(
+            List<String> request, RequestOptions requestOptions) {
+
+        OkHttpClient client = clientOptions.httpClient();
+        if (requestOptions != null && requestOptions.getTimeout().isPresent()) {
+            client = clientOptions.httpClientWithTimeout(requestOptions);
+        }
+        CompletableFuture<List<String>> future = new CompletableFuture<>();
```
dsinghvi, line 97: if we instantiate a `CompletableFuture` do we need to override the `cancel` method to cancel the underlying okhttp call? we may want a generic utility that extends `CompletableFuture` if this is the case

PR #6177, opened 2025-02-26
feat(csharp): Add exported-client-class-name

amckinney:
This adds the `exported-client-class-name` configuration option so that anyone that customizes their exported client retains accurate snippets.

The most common use case for this feature is shown below:

```yaml
   version: 0.10.0
   config:
     client-class-name: BaseClient
     exported-client-class-name: AcmeClient
```

This will generate snippets like the following (including in the `README.md`):

```csharp
using Acme;

var client = new AcmeClient("TOKEN");
await client.Users.CreateAsync(new CreateUserRequest { Name = "john.doe" });
```

### DIFF
--- a/fern/pages/changelogs/csharp-sdk/2025-02-26.mdx
+++ b/fern/pages/changelogs/csharp-sdk/2025-02-26.mdx
@@ -1,0 +1,5 @@
+## 1.10.0
+**`(feat):`** Add support for `exported-client-class-name` to allow you to export the client class name.
+This is useful when you want to use a custom client class name for code snippets.
+
+


Swimburger (approved):

generators/csharp/sdk/src/SdkGeneratorContext.ts
```diff
@@ -277,16 +277,21 @@ export class SdkGeneratorContext extends AbstractCsharpGeneratorContext<SdkCusto
...
     }
 
-    public getRootClientClassReference(): csharp.ClassReference {
+    public getRootClientClassReference({ asSnippet }: { asSnippet?: boolean } = {}): csharp.ClassReference {
         return csharp.classReference({
-            name: this.getRootClientClassName(),
+            name: this.getRootClientClassName({ asSnippet }),
             namespace: this.getNamespace()
         });
     }
```
Swimburger, line 297: Having a separate function may be cleaner than passing this `asSnippet` argument. It'll have more duplicated code, but maybe cleaner.

amckinney: I actually started with that approach, then refactored to this afterwards. Given that you called it out, I'll refactor it back 👍

PR #6236, opened 2025-02-28
feat(java): Async clients

ajgateno:
Support async clients in Java


amckinney (approved):
Discussed offline w/ @ajgateno - got an overview of the refactors required to make the change. The generated output looks great!

PR #6242, opened 2025-02-28
feat(csharp): Add custom pager

Swimburger:
## Description
To do in follow-up PR, reconcile BiPager and Pager, and custom pager endpoints with normal pager endpoints (breaking change).

## Changes Made
- Split up creating HTTP request and sending HTTP request in RawClient
- Generate CustomPager and CustomPagerFactory
- Generate custom pager endpoints

## Testing


api-yml.schema.json
```diff
@@ -1341,6 +1341,16 @@
...
+        "type": {
+          "oneOf": [
+            {
+              "const": "cursor"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
```
amckinney, line 1353: The changes to the offset and cursor schemas should be removed now, right?

generators/csharp/codegen/src/asIs/RawClient.Template.cs
```diff
@@ -27,7 +26,16 @@ internal class RawClient(ClientOptions clientOptions)
...
 
-    public async Task<ApiResponse> MakeRequestAsync(
+    [Obsolete("Use SendRequestAsync instead.")]
+    public Task<ApiResponse> MakeRequestAsync(
+        BaseApiRequest request,
+        CancellationToken cancellationToken = default
+    )
+    {
+        return SendRequestAsync(request, cancellationToken);
+    }
```
amckinney, line 36: The `RawClient` is internal, so we can also just remove this entirely. We can also do that in a follow-up if you prefer.

generators/csharp/codegen/src/asIs/RawClient.Template.cs
```diff
@@ -85,37 +122,50 @@ public record ApiResponse
...
+    private static void AssertRetryableContent(HttpContent requestContent)
+    {
+        switch (requestContent)
+        {
+            case StreamContent:
+                throw new Exception("Cannot retry a request with StreamContent");
+            case MultipartContent:
+                throw new Exception("Cannot retry a request with MultipartContent");
+        }
+    }
```
amckinney, line 160: Will the user see these errors? They don't control the default retry logic, so would it be better to simply not retry for such endpoints?

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -383,6 +519,9 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
                             unpagedEndpointMethodName
                         });
                         break;
+                    case "custom":
+                        this.context.logger.error("Custom should already be handled elsewhere");
```
amckinney, line 523: Same here w.r.t. internal errors.

amckinney (approved):
Nice! Main thing to look into is removing the `type` for the offset and cursor pagination. All other comments are mostly nits.

generators/csharp/codegen/src/asIs/RawClient.Template.cs
Swimburger: Yes, fair!

generators/csharp/codegen/src/FileGenerator.ts
```diff
@@ -12,10 +12,20 @@ export abstract class FileGenerator<
     constructor(protected readonly context: Context) {}
 
     public generate(): GeneratedFile {
-        this.context.logger.debug(`Generating ${this.getFilepath()}`);
+        if (this.shouldGenerate()) {
```
Swimburger, line 15: @amckinney WDYT of this change?
I'm not forcing existing code to respect this new `shouldGenerate` method, so just logging it as a warning if it's called when not true.

This lets me remove the WireTest for custom pagination

amckinney: Interesting, I'm fine with this change since it's largely a no-op. 

Although it's unlikely to happen, the internal warning would be confusing to the user. I know it's more for us though.

Let's go with this for now, and we can always revisit later if it ends up causing confusion.

PR #6261, opened 2025-03-02
feat(csharp): Add exception handler

amckinney:
This adds support for the `include-exception-handler` configuration option, which generates an internal `ExceptionHandler` and user-facing `ExceptionInterceptor` interface for the SDK. 

This is useful when you want to act upon all exceptions thrown by the SDK (e.g. report them to a monitoring service). For example,

```csharp
class ExceptionInterceptor : IExceptionInterceptor
{
    public ExceptionInterceptor(...)
    {
    }

    public Exception Intercept(Exception exception)
    {
        // TODO: Log the exception to a monitoring service (e.g. Sentry)
        return exception;
    }
}
```

You can configure this feature with the following:

```yaml
# generators.yml
- name: fern-api/fern-csharp-sdk
  version: 1.12.0-rc1
  config:
    include-exception-handler: true
```

Note that this also fixes a small regression in the Protobuf mapper for `ReadOnlyMemory` types, which was verified using the gRPC test definition.

### DIFF
--- a/seed/csharp-sdk/imdb/include-exception-handler/src/SeedApi/Imdb/ImdbClient.cs
+++ b/seed/csharp-sdk/imdb/include-exception-handler/src/SeedApi/Imdb/ImdbClient.cs
@@ -31,43 +31,45 @@ public partial class ImdbClient
         CancellationToken cancellationToken = default
     )
     {
-        return await _exceptionHandler.TryCatchAsync(async () =>
-        {
-            var response = await _client
-                .SendRequestAsync(
-                    new RawClient.JsonApiRequest
+        return await _exceptionHandler
+            .TryCatchAsync(async () =>
+            {
+                var response = await _client
+                    .SendRequestAsync(
+                        new RawClient.JsonApiRequest
+                        {
+                            BaseUrl = _client.Options.BaseUrl,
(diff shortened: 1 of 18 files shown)


seed/csharp-sdk/imdb/include-exception-handler/src/SeedApi/Imdb/ImdbClient.cs
```diff
@@ -0,0 +1,135 @@
...
+                {
+                    // unable to map error response, throwing generic error
+                }
+                throw new SeedApiApiException(
+                    $"Error with status code {response.StatusCode}",
+                    response.StatusCode,
+                    responseBody
+                );
+            }
+        });
```
Swimburger, line 133: If we're `await`'ing anything, we should always add `ConfigureAwait(false)`

Swimburger (approved):

seed/csharp-sdk/imdb/include-exception-handler/src/SeedApi/Imdb/ImdbClient.cs
amckinney: Great catch - done!

PR #6262, opened 2025-03-02
feat(csharp): Make pagers consistent

Swimburger:
## Description
Pager and BiPager now have consistent properties and methods.
As part of this change, pagers are now generated asynchronously which is a breaking change for the endpoint methods.

```csharp
// Before:
var pager = client.GetItemsAsync(...);

// After:
var pager = await client.GetItemsAsync(...);
```

Also:
- Fixed an issue where generated code for setting UUIDs in a query string parameter would generate uncompilable code.
- Extend support for offset pagination to float, double, and decimal types.

## Testing

### DIFF
--- a/seed/csharp-sdk/pagination/custom-pager/src/SeedPagination/Complex/ComplexClient.cs
+++ b/seed/csharp-sdk/pagination/custom-pager/src/SeedPagination/Complex/ComplexClient.cs
@@ -53,6 +53,7 @@ public partial class ComplexClient
                 SearchInternalAsync,
                 (request, cursor) =>
                 {
+                    request.Pagination ??= new();
                     request.Pagination.StartingAfter = cursor;
                 },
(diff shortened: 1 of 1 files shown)


generators/csharp/sdk/versions.yml
```diff
@@ -6,6 +6,22 @@
 #         The C# SDK now uses forward-compatible enums which are not compatible with the previously generated enums.
 #         Set `enable-forward-compatible-enums` to `false` in the configuration to generate the old enums.
 #   irVersion: 53
+- version: 1.13.0-rc0
```
amckinney, line 9: Should we keep this in `1.12.0-rcX` until we're out of the RC state? Otherwise we'll have release candidates across multiple minor versions.

amckinney (approved):
Nice 👏

Issue #6263, opened 2025-03-03
"Hidden" Docs pages / Docs sections should be added to robots.txt excludes


danielchalef:
### Which Fern component?

Fern Docs

### How important is this?

P1 - High (Strongly needed)

### What's the feature?

Hidden pages are currently still indexed by Google. They're only supposed to be access directly by URL.

### Are you interested in contributing this feature?

No

chdeskur:
Hidden pages are now ensured to be `noindex,nofollow`, and will not be present in the sitemap.

Including the hidden URLs in the `robots.txt` `disallow` could make these pages more discoverable by bad actors, so closing the ticket out since the underlying goal has been achieved.

PR #6278, opened 2025-03-03
fix(cli): respect hidden inheritance

chdeskur:
This PR allows pages, APIs, changelogs, and sections to inherit the `hidden` property if their parent is hidden as well. 

Neede for https://github.com/fern-api/fern/issues/6263

Before: 
before-api
before

After: 
after-api
after


chdeskur:
`bad` arrows are pointing at sections/pages/endpoints that have been marked as `hidden` in the navigation, or should be hidden because their parent is hidden.

eyw520 (approved):

PR #6298, opened 2025-03-04
fix(csharp): Improve generated wire tests

Swimburger:
## Changes Made
**`(feat):`** Compare .NET objects instead of JSON strings when comparing objects in mock wire tests.
With this change, we can remove the dependency on the `FluentAssertions.Json` package (and implicitly on the `Newtonsoft.Json` package).

**`(feat):`** Update dependencies in generated test project:
- `Microsoft.NET.Test.Sdk`: 17.12.0 => 17.13.0
- `NUnit`: 4.2.2 => 4.3.2
- `NUnit3TestAdapter`: 4.6.0 => 5.0.0
- `NUnit.Analyzers`: 4.4.0 => 4.6.0
- `coverlet.collector`: 6.0.2 => 6.0.4
- `WireMock.Net`: 1.6.8 => 1.7.4

**`(fix):`** Compare .NET objects instead of JSON strings when comparing objects in mock wire tests,
which fixes bugs where JSON deserialization and serialization would cause differences in the JSON strings.

## Testing

### DIFF
--- a/fern/pages/changelogs/csharp-sdk/2025-03-04.mdx
+++ b/fern/pages/changelogs/csharp-sdk/2025-03-04.mdx
@@ -1,8 +1,9 @@
 ## 1.12.0-rc9
 **`(feat):`** Compare .NET objects instead of JSON strings when comparing objects in mock wire tests.
-With this change, we can remove the dependency on the `FluentAssertions.Json` package (and implicitly on the `Newtonsoft.Json` package).
+With this change, we can remove the dependency on the FluentAssertions.Json package (and implicitly on the Newtonsoft.Json package).
 
-**`(feat):`** Update dependencies in generated test project:
+
+**`(chore):`** Update dependencies in generated test project:
 - `Microsoft.NET.Test.Sdk`: 17.12.0 => 17.13.0
 - `NUnit`: 4.2.2 => 4.3.2
 - `NUnit3TestAdapter`: 4.6.0 => 5.0.0
@@ -13,3 +14,5 @@ With this change, we can remove the dependency on the `FluentAssertions.Json` pa
 
 **`(fix):`** Compare .NET objects instead of JSON strings when comparing objects in mock wire tests,
 which fixes bugs where JSON deserialization and serialization would cause differences in the JSON strings.
+
+


generators/csharp/codegen/src/context/convertExampleTypeReferenceToTypeReference.ts
```diff
@@ -0,0 +1,202 @@
...
+        case "uint":
+        case "uint64":
+        case "float":
+        case "double":
+        case "date":
+        case "datetime":
+        case "uuid":
+        case "base64":
+        case "bigInteger":
+            throw new Error("Not implemented");
```
amckinney, line 83: Although unlikely, we should add more context here - 
```suggestion
            throw new Error("Internal error; only boolean and string literals are permitted");
```

generators/csharp/sdk/src/test-generation/mock-server/MockServerTestGenerator.ts
```diff
@@ -272,4 +315,31 @@ export class MockServerTestGenerator extends FileGenerator<CSharpFile, SdkCustom
...
+                if (exampleResponse.value.type === "body") {
+                    if (exampleResponse.value.value) {
+                        const typeReference = convertExampleTypeReferenceToTypeReference(exampleResponse.value.value);
+                        const type = this.context.csharpTypeMapper.convert({
+                            reference: typeReference
+                        });
+                        return type;
+                    }
+                }
+                throw new Error("Internal error; Could not convert example response to C# type");
```
amckinney, line 331: nit:

```suggestion
                throw new Error("Internal error; could not convert example response to C# type");
```

generators/csharp/sdk/src/test-generation/mock-server/MockServerTestGenerator.ts
```diff
@@ -272,4 +315,31 @@ export class MockServerTestGenerator extends FileGenerator<CSharpFile, SdkCustom
...
+            case "error":
+                if (exampleResponse.body) {
+                    const typeReference = convertExampleTypeReferenceToTypeReference(exampleResponse.body);
+                    const type = this.context.csharpTypeMapper.convert({
+                        reference: typeReference
+                    });
+                    return type;
+                }
+                throw new Error("Internal error; Could not convert example response to C# type");
```
amckinney, line 340:
```suggestion
                throw new Error("Internal error; could not convert example response to C# type");
```

generators/csharp/codegen/src/context/convertExampleTypeReferenceToTypeReference.ts
```diff
@@ -0,0 +1,202 @@
+import { assertNever } from "@fern-api/core-utils";
```
amckinney, line 1: It looks like this file is duplicated in `ir-utils` - did you mean to import that version here instead?

amckinney (approved):
Looks great, only a few nits!

PR #6302, opened 2025-03-05
feat(java): Explicit nulls

ajgateno:
This PR handles `nullable<T>` as follows:

For a field `email: optional<Email>`, `getEmail` returns an `Optional<Email>` that is never null to the callsite. Since the underlying field can potentially be null and we use that to decide how to do serialization with Jackson, but the user never knows. `Nullable<T>` is used on the builder via a new method (non-breaking) to set an explicit null.

```java
User user = User.builder()
    .name("john.doe")
    .email(Nullable.ofNull())
    .profilePic(Nullable.empty())
    .build()

Optional<Email> johnsEmail = user.email(); // Optional.empty()

serialize(user) // { "name": "john.doe", "email": null }
```


fern/pages/changelogs/java-spring/2025-03-05.mdx
```diff
@@ -0,0 +1,4 @@
+## 1.6.0
+**`(feat):`** Add support for explicit nulls in serialization
```
amckinney, line 2: Discussed offline - 

It'd be _ideal_ if this didn't affect the spring generator at all. On the response end we'll always provide the user with a `Optional` type (not a `Nullable`) type, so the `Nullable` methods are effectively a no-op. 

We see a couple options here:
1. We add a way for the spring generator to tell the model generator that it should _not_ respect nullable types.
2. We add a feature flag to enable this feature, but _only_ expose it on the SDK generator.

We agree that (1) is more ideal because this feature should be enabled by default if the types are marked as `nullable` in OpenAPI and/or the Fern definition, but if it's sufficiently complex to make that happen (2) is a good compromise.

seed/java-sdk/nullable/no-custom-config/src/main/java/com/seed/nullable/resources/nullable/requests/GetUsersRequest.java
```diff
@@ -175,6 +193,17 @@ public Builder tags(String tags) {
             return this;
         }
 
+        public Builder tags(Nullable<String> tags) {
```
amckinney, line 196: Nice!

amckinney (approved):
Nice looks great! 🚀

PR #6308, opened 2025-03-05
fix(python): remove extraneous NameError-causing update_forward_refs

hughhan1:
## Description
54a7e15e1d6780dd978ffe348a89f1179832eb24 introduced a change that changes our Python generator to generate code that correctly imports transitive dependencies as ghost references at the bottom of Python modules containing Pydantic model definitions.

However, that change also causes our generator to generate extraneous `core.pydantic_utilities.update_forward_refs` on these transitive dependencies when such statements are not necessary due to those statements already running in the modules in which the transitive dependencies are defined.

In fact, these extraneous statements cause `NameError` exceptions due to attempting to invoke `typing.eval(...)` on variables that have not yet been defined.

Change the logic in our Python generator such that we no longer emit these extraneous `core.pydantic_utilities.update_forward_refs` statements, while still continuing to support generating the necessary ghost references that correspond to them.

## Changes Made

Modify [`_add_update_forward_ref_for_transitive_circular_dependencies`](https://github.com/fern-api/fern/blob/549b5a4b675af32b0d17ea4a45143966a9acbbe4/generators/python/src/fern_python/generators/pydantic_model/fern_aware_pydantic_model.py#L144-L172) in the following ways:
- rename to `_add_ghost_references_for_transitive_circular_dependencies`
- no longer emit `update_forward_refs` statements
- simplify conditional logic

## Testing

- [x] `test-definitions/fern/apis/circular-references/definition/ast.yml`
- [x] `test-definitions/fern/apis/circular-references-advanced/definition/ast.yml`

### DIFF
--- a/packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/circular-references-advanced/type_ast_FirstUnion.json
+++ b/packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/circular-references-advanced/type_ast_FirstUnion.json
@@ -1,0 +1,80 @@
+{
+  "anyOf": [
+    {
+      "$ref": "#/definitions/ast.FirstUnionFirstElement"
+    },
+    {
+      "$ref": "#/definitions/ast.FirstUnionSecondElement"
+    }
+  ],
+  "definitions": {
+    "ast.FirstUnionSecondElement": {
+      "type": "object",
+      "properties": {
+        "child": {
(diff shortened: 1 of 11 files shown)


amckinney (approved):
Nice, this looks great! I also confirmed that this doesn't introduce any additional `python-sdk` failures in our test definitions (re: [job](https://github.com/fern-api/fern/actions/runs/13682506191/job/38258147453)). 

We can ship this as a patch and continue to iterate from there!

eyw520 (approved):
snapshots look great

Issue #4817, opened 2024-10-07
Support generator cancellation for streaming requests


billytrend-cohere:
## Describe the Bug

A user reported that cancelling a generator with streaming isn't handled correctly https://github.com/cohere-ai/cohere-python/issues/578

PR #6319, opened 2025-03-06
fix(python): don't swallow BaseException in streaming responses

hughhan1:
## Description
2b75f8b0ef2d816c588abced29bff2b46c8418c8 introduced a change in the Python generator generates code that swallows all `BaseException` instances when streaming responses. This `except` block was intended to catch exceptions with respect to decoding malformed responses.

Using `except` has unintentional negative consequences because `GeneratorExit` subclasses `BaseException`, and swallowing such exceptions prevents the [`finally` block](https://github.com/encode/httpx/blob/4189b7f051c6c51ce74c3bee1a5f269f9c50c6b2/httpx/_client.py#L876-L877) of the [`httpx._client.Client.stream()`](https://github.com/encode/httpx/blob/4189b7f051c6c51ce74c3bee1a5f269f9c50c6b2/httpx/_client.py#L827-L877) from being executed, resulting in the [`httpx._models.Response.close()`] function from ever being invoked. 

This behavior ultimately leads to errors of the following form, as seen in https://github.com/cohere-ai/cohere-python/issues/578 and https://github.com/fern-api/fern/issues/4817.
```
RuntimeError: generator ignored GeneratorExit
```

## Changes Made

- Generate code that explicitly catches `Exception`, instead of all generic `BaseException`.

## Testing

- [x] Manually tested these changes against Cohere's Python SDK, as follows, and check that the expected exception is no longer raised.
```python
import cohere

co = cohere.Client(api_key=...)
s = co.chat_stream(message="How big is the world?")

i = 0
for chunk in s:
    if i < 2:
        print(chunk)
        i += 1
    else:
        s.close()
        break
```

### DIFF
--- a/generators/python/sdk/versions.yml
+++ b/generators/python/sdk/versions.yml
@@ -1,4 +1,11 @@
 # For unreleased changes, use unreleased.yml
+- version: 4.3.16
+  irVersion: 53
+  changelogEntry:
+    - type: fix
+      summary: |
+        Fix an issue where GeneratorExit was unintentionally being swallowed by code blocks within the HttpClient.stream() contextmanager.
+
 - version: 4.3.15
   irVersion: 53
   changelogEntry:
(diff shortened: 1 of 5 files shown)


eyw520 (approved):

amckinney (approved):
Nice!

PR #6322, opened 2025-03-06
fix(java): Line up protobuf type handling with our other generators

ajgateno:
We were using longs for uints, when we should have been using ints

### DIFF
--- a/fern/pages/changelogs/java-sdk/2025-03-06.mdx
+++ b/fern/pages/changelogs/java-sdk/2025-03-06.mdx
@@ -1,0 +1,4 @@
+## 2.23.0
+**`(fix):`** Represent protobuf uint as int rather than long
+
+


generators/java/generator-utils/src/main/java/com/fern/java/utils/TypeReferenceUtils.java
```diff
@@ -183,7 +183,7 @@ public String visitLong() {
 
         @Override
         public String visitUint() {
-            return "Long";
+            return "Integer";
```
amckinney, line 186: I wonder if we should name this distinct even though it still maps to an `int`. That way if a user has an undiscriminated union with both there's no conflict. Otherwise the following type would have two things named with `Integer`:

```yaml
types:
  Number:
    discriminated: false
    union:
      - uint
      - integer
```

amckinney (approved):
Nice!

PR #6326, opened 2025-03-07
feat(csharp): Add dynamic snippets options for C# tests

amckinney:
This updates `seed` to support generate dynamic snippet tests for the C# SDK. With this, we produce full usage examples that can be compiled alongside the generated SDK to confirm that the snippets are valid.

This also introduces a new `Options` configuration, which callers of the dynamic snippets can use to customize the style, formatting, and any language-specific configuration. The current options are shown below:

```typescript
import { AbstractFormatter } from "../ast/AbstractFormatter";

// Style of the generated snippets.
export enum Style {
    // Use a concise style for the snippet (top-level statements).
    Concise = "concise",

    // Use the full style for the snippet (a `main` function that can be run).
    Full = "full"
}

// Options used to customize the behavior of any dynamic snippets generator
export interface Options {
    config?: unknown;
    formatter?: AbstractFormatter;
    style?: Style;
}

```

### DIFF
--- a/seed/csharp-model/csharp-property-access/src/SeedCsharpAccess/User.cs
+++ b/seed/csharp-model/csharp-property-access/src/SeedCsharpAccess/User.cs
@@ -6,6 +6,7 @@ namespace SeedCsharpAccess;
 public record User
 {
     [JsonPropertyName("id")]
+    [JsonAccess(JsonAccessType.ReadOnly)]
     public required string Id { get; set; }
 
     [JsonPropertyName("name")]
@@ -15,6 +16,7 @@ public record User
     public required string Email { get; set; }
 
     [JsonPropertyName("password")]
+    [JsonAccess(JsonAccessType.WriteOnly)]
     public required string Password { get; set; }
 
     public override string ToString()
--- a/seed/csharp-model/pagination/src/SeedPagination/Core/SeedPaginationPager.cs
+++ b/seed/csharp-model/pagination/src/SeedPagination/Core/SeedPaginationPager.cs
@@ -1,0 +1,154 @@
+using System.Runtime.CompilerServices;
+using global::System.Net.Http;
+
+namespace SeedPagination.Core;
+
+internal static class SeedPaginationPagerFactory
+{
+    public static async Task<SeedPaginationPager<TItem>> CreateAsync<TItem>(
+        Func<HttpRequestMessage, CancellationToken, Task<HttpResponseMessage>> sendRequest,
+        HttpRequestMessage initialRequest,
+        CancellationToken cancellationToken = default
+    )
+    {
+        var response = await sendRequest(initialRequest, cancellationToken).ConfigureAwait(false);
+        var (nextPageRequest, hasNextPage, previousPageRequest, hasPreviousPage, page) =
+            await SeedPaginationPager<TItem>
+                .ParseHttpCallAsync(initialRequest, response, cancellationToken)
+                .ConfigureAwait(false);
+        return new SeedPaginationPager<TItem>(
+            sendRequest,
+            nextPageRequest,
+            hasNextPage,
+            previousPageRequest,
+            hasPreviousPage,
+            page
+        );
+    }
+}
+
+public class SeedPaginationPager<TItem> : BiPager<TItem>, IAsyncEnumerable<TItem>
+{
+    private HttpRequestMessage? _nextPageRequest;
+    private HttpRequestMessage? _previousPageRequest;
+
+    private readonly Func<
+        HttpRequestMessage,
+        CancellationToken,
+        Task<HttpResponseMessage>
+    > _sendRequest;
+
+    public bool HasNextPage { get; private set; }
+    public bool HasPreviousPage { get; private set; }
+    public Page<TItem> CurrentPage { get; private set; }
+
+    public SeedPaginationPager(
+        Func<HttpRequestMessage, CancellationToken, Task<HttpResponseMessage>> sendRequest,
+        HttpRequestMessage? nextPageRequest,
+        bool hasNextPage,
+        HttpRequestMessage? previousPageRequest,
+        bool hasPreviousPage,
+        Page<TItem> page
+    )
+    {
+        _sendRequest = sendRequest;
+        _nextPageRequest = nextPageRequest;
+        HasNextPage = hasNextPage;
+        _previousPageRequest = previousPageRequest;
+        HasPreviousPage = hasPreviousPage;
+        CurrentPage = page;
+    }
+
+    public async Task<Page<TItem>> GetNextPageAsync(CancellationToken cancellationToken = default)
+    {
+        if (_nextPageRequest == null)
+        {
+            return Page<TItem>.Empty;
+        }
+
+        return await SendRequestAndHandleResponse(_nextPageRequest, cancellationToken)
+            .ConfigureAwait(false);
+    }
+
+    public async Task<Page<TItem>> GetPreviousPageAsync(
+        CancellationToken cancellationToken = default
+    )
+    {
+        if (_previousPageRequest == null)
+        {
+            return Page<TItem>.Empty;
+        }
+
+        return await SendRequestAndHandleResponse(_previousPageRequest, cancellationToken)
+            .ConfigureAwait(false);
+    }
+
+    private async Task<Page<TItem>> SendRequestAndHandleResponse(
+        HttpRequestMessage request,
+        CancellationToken cancellationToken = default
+    )
+    {
+        var response = await _sendRequest(request, cancellationToken).ConfigureAwait(false);
+        var (nextPageRequest, hasNextPage, previousPageRequest, hasPreviousPage, page) =
+            await ParseHttpCallAsync(request, response, cancellationToken).ConfigureAwait(false);
+        _nextPageRequest = nextPageRequest;
+        HasNextPage = hasNextPage;
+        _previousPageRequest = previousPageRequest;
+        HasPreviousPage = hasPreviousPage;
+        CurrentPage = page;
+        return page;
+    }
+
+    internal static async Task<(
+        HttpRequestMessage? nextPageRequest,
+        bool hasNextPage,
+        HttpRequestMessage? previousPageRequest,
+        bool hasPreviousPage,
+        Page<TItem> page
+    )> ParseHttpCallAsync(
+        HttpRequestMessage request,
+        HttpResponseMessage response,
+        CancellationToken cancellationToken = default
+    )
+    {
+        throw new NotImplementedException();
+    }
+
+    public async IAsyncEnumerator<TItem> GetAsyncEnumerator(
+        CancellationToken cancellationToken = default
+    )
+    {
+        foreach (var item in CurrentPage)
+        {
+            yield return item;
+        }
+
+        await foreach (var page in GetNextPagesAsync(cancellationToken))
+        {
+            foreach (var item in page)
+            {
+                yield return item;
+            }
+        }
+    }
+
+    public async IAsyncEnumerable<Page<TItem>> GetNextPagesAsync(
+        [EnumeratorCancellation] CancellationToken cancellationToken = default
+    )
+    {
+        while (HasNextPage)
+        {
+            yield return await GetNextPageAsync(cancellationToken).ConfigureAwait(false);
+        }
+    }
+
+    public async IAsyncEnumerable<Page<TItem>> GetPreviousPagesAsync(
+        [EnumeratorCancellation] CancellationToken cancellationToken = default
+    )
+    {
+        while (HasPreviousPage)
+        {
+            yield return await GetPreviousPageAsync(cancellationToken).ConfigureAwait(false);
+        }
+    }
+}
--- a/seed/csharp-model/seed.yml
+++ b/seed/csharp-model/seed.yml
@@ -1,4 +1,4 @@
-irVersion: v53
+irVersion: v57
 displayName: C# Model
 image: fernapi/fern-csharp-model
 changelogLocation: ../../generators/csharp/model/versions.yml
--- a/seed/csharp-sdk/accept-header/src/SeedApi.DynamicSnippets/SeedApi.DynamicSnippets.csproj
+++ b/seed/csharp-sdk/accept-header/src/SeedApi.DynamicSnippets/SeedApi.DynamicSnippets.csproj
@@ -1,32 +1,11 @@
 
 <Project Sdk="Microsoft.NET.Sdk">
-
     <PropertyGroup>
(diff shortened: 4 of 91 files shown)


generators/csharp/codegen/src/ast/ClassInstantiation.ts
```diff
@@ -45,28 +45,44 @@ export class ClassInstantiation extends AstNode {
...
+        writer.indent();
         this.arguments_.forEach((argument, idx) => {
             if (isNamedArgument(argument)) {
-                writer.write(`${argument.name} = `);
+                writer.write(`${argument.name}`);
+                if (this.forceUseConstructor) {
+                    writer.write(": ");
+                } else {
+                    writer.write(" = ");
+                }
```
Swimburger, line 65: Ideally, we should disambiguate arguments from properties.
Arguments can be positional or named, but they are not properties.
Then any properties passed to the class instantiation should be used as property initializer.
This is fine for now tho, but this `this.forceUseConstructor` doesn't really make sense and is a workaround addressing the ambiguity.

generators/csharp/codegen/src/ast/TypeLiteral.ts
```diff
@@ -161,14 +167,30 @@ export class TypeLiteral extends AstNode {
...
+                break;
+            }
+            case "double": {
                 writer.write(this.internalType.value.toString());
                 break;
             }
+            case "float": {
+                writer.write(`${this.internalType.value}f`);
+                break;
+            }
```
Swimburger, line 193: Nice, lets add decimal here too

generators/csharp/dynamic-snippets/src/EndpointSnippetGenerator.ts
```diff
@@ -67,29 +68,40 @@ export class EndpointSnippetGenerator {
...
         snippet: FernIr.dynamic.EndpointSnippetRequest;
-        verbose?: boolean;
+        options: Options;
     }): csharp.AstNode {
         const body = csharp.codeblock((writer) => {
-            writer.writeNode(this.constructClient({ endpoint, snippet }));
-            writer.writeLine();
-            writer.writeNode(this.callMethod({ endpoint, snippet }));
+            writer.writeNodeStatement(this.constructClient({ endpoint, snippet }));
+            writer.newLine();
```
Swimburger, line 79: `newLine` is a nice addition :)

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
```diff
@@ -87,62 +99,70 @@ exports[`snippets (default) > examples > 'POST /movie (simple)' 1`] = `
...
+await client.Endpoints.Object.GetAndReturnWithOptionalFieldAsync(
+    new ObjectWithOptionalField{
+        String = "string",
+        Integer = 1,
+        Long = 1000000l,
+        Double = 1.1,
+        Bool = true,
+        Datetime = DateTime.Parse("2024-01-15T09:30:00Z", null, DateTimeStyles.AdjustToUniversal),
+        Date = DateOnly.Parse("2023-01-15"),
+        UUID = "d5e9c84f-c2b2-4bf4-b4b0-7ffd7a9ffc32",
```
Swimburger, line 150: This is probably a casing setting, but the default should be `Uuid`

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
```diff
@@ -178,109 +198,133 @@ exports[`snippets (default) > exhaustive > 'POST /container/list-of-objects (sim
...
-    
-)
-await client.Service.PostAsync(new MyRequest(
-    
-)).ConfigureAwait(false)"
+var client = new AcmeClient();
+
+await client.Service.PostAsync(
+    new MyRequest()
+);
```
Swimburger, line 246: I'm guessing this will be complicated, but it would be great if this was on one line. Maybe in the far future.

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
```diff
@@ -373,54 +431,67 @@ exports[`snippets (default) > nullable > 'Query parameters' 1`] = `
...
-var client = new AcmeClient{
-    Token = "<YOUR_API_KEY>", clientOptions = new ClientOptions{
+var client = new AcmeClient(
+    token: "<YOUR_API_KEY>",
+    clientOptions: new ClientOptions{
         BaseUrl = AcmeEnvironment.Staging
     }
-}
-await client.Dummy.GetDummyAsync().ConfigureAwait(false)"
+);
```
Swimburger, line 476: Should be 
```csharp
var client = new AcmeClient(
    token: "<YOUR_API_KEY>",
    clientOptions: new ClientOptions
    {
        BaseUrl = AcmeEnvironment.Staging
    }
);
```
or
```csharp
var client = new AcmeClient(
    token: "<YOUR_API_KEY>",
    clientOptions: new ClientOptions { BaseUrl = AcmeEnvironment.Staging }
);
```

generators/csharp/dynamic-snippets/src/context/DynamicTypeLiteralMapper.ts
```diff
@@ -73,7 +73,7 @@ export class DynamicTypeLiteralMapper {
             return csharp.TypeLiteral.nop();
         }
         return csharp.TypeLiteral.list({
-            valueType: this.context.dynamicTypeMapper.convert({ typeReference: list }),
+            valueType: this.context.dynamicTypeMapper.convert({ typeReference: list, unboxOptionals: true }),
```
Swimburger, line 76: Why do we unbox optionals here?

generators/csharp/dynamic-snippets/src/context/DynamicTypeMapper.ts
```diff
@@ -36,13 +38,16 @@ export class DynamicTypeMapper {
...
+                const value = this.convert({ typeReference: args.typeReference.value });
+                return args.unboxOptionals
+                    ? value
+                    : csharp.Type.optional(value);
+            }
             case "primitive":
                 return this.convertPrimitive({ primitive: args.typeReference.value });
             case "set":
-                return csharp.Type.set(this.convert({ typeReference: args.typeReference }));
+                return csharp.Type.set(this.convert({ typeReference: args.typeReference, unboxOptionals: true }));
```
Swimburger, line 50: Why unbox optionals here?

packages/cli/generation/local-generation/local-workspace-runner/src/dynamic-snippets/csharp/DynamicSnippetsCsharpTestGenerator.ts
```diff
@@ -0,0 +1,115 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
+        <LangVersion>12</LangVersion>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <IsPackable>false</IsPackable>
+        <IsTestProject>true</IsTestProject>
+        <PolySharpIncludeRuntimeSupportedAttributes>true</PolySharpIncludeRuntimeSupportedAttributes>
```
Swimburger, line 25: We should remove PolySharp and make it so that the language version is the one you want to target for compatibility.

packages/cli/generation/local-generation/local-workspace-runner/src/dynamic-snippets/csharp/DynamicSnippetsCsharpTestGenerator.ts
```diff
@@ -0,0 +1,115 @@
...
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.13.0"/>
+        <PackageReference Include="NUnit" Version="4.3.2"/>
+        <PackageReference Include="NUnit.Analyzers" Version="4.6.0">
+            <PrivateAssets>all</PrivateAssets>
+            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+        </PackageReference>
+        <PackageReference Include="coverlet.collector" Version="6.0.4">
+            <PrivateAssets>all</PrivateAssets>
+            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+        </PackageReference>
```
Swimburger, line 42: Are we running these as tests or just verify things compile?

packages/cli/generation/local-generation/local-workspace-runner/src/dynamic-snippets/csharp/DynamicSnippetsCsharpTestGenerator.ts
```diff
@@ -0,0 +1,115 @@
...
+                this.context.logger.error(
+                    `Failed to generate dynamic snippet for endpoint ${JSON.stringify(request.endpoint)}: ${error}`
+                );
+            }
+        }
+        this.context.logger.debug("Done generating dynamic snippet tests");
+    }
+
+    private async initializeProject(outputDir: AbsoluteFilePath): Promise<AbsoluteFilePath> {
+        const absolutePathToOutputDir = join(outputDir, RelativeFilePath.of("src/SeedApi.DynamicSnippets"));
```
Swimburger, line 97: `src/SeedApi.DynamicSnippets` this name is still hardcoded.

seed/csharp-sdk/auth-environment-variables/src/SeedApi.DynamicSnippets/Example0.cs
```diff
@@ -0,0 +1,20 @@
...
+        var client = new SeedAuthEnvironmentVariablesClient(
+            apiKey: "<value>",
+            clientOptions: new ClientOptions{
+                BaseUrl = "https://api.fern.com"
+            }
+        );
+
+        await client.Service.GetWithApiKeyAsync();
+    }
+
```
Swimburger, line 19: Extra newline not necessary here

Swimburger (approved):

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
amckinney: Yea that will be the case - this is just using an IR that was produced with the special casing convention (but this actually won't be used in practice for C#).

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
amckinney: Agreed - we can improve the formatting like this in other snippets too (e.g. Go). I'll leave this as-is for now since it's _correct_, but we can improve the style later on as we go.

generators/csharp/dynamic-snippets/src/context/DynamicTypeLiteralMapper.ts
amckinney: I previously wasn't, but this was required for when we construct a `list<optional<string>>`. By default, the generator produces this as `List<string>` because the optional is defined within the container. So without this we would have attemped to construct the type like so:

```csharp
// This was invalid for the generated type.
new List<string?>() {
  "one",
  "two"
}
```

 So this was updated to match the functionality of the C# SDK, but we may want to change this in the generator later.

generators/csharp/dynamic-snippets/src/context/DynamicTypeMapper.ts
amckinney: Same as the reason outlined [here](https://github.com/fern-api/fern/pull/6326#discussion_r1985475599).

packages/cli/generation/local-generation/local-workspace-runner/src/dynamic-snippets/csharp/DynamicSnippetsCsharpTestGenerator.ts
amckinney: That's actually OK in this context - we're just using this to verify that the project compiles so the project name doesn't actually need to match. This keeps the implementation a little simpler.

seed/csharp-sdk/auth-environment-variables/src/SeedApi.DynamicSnippets/Example0.cs
amckinney: Noted - I left this change out for now since I know you're editing some of the `Method` AST node in your change. I'll come back to this. Plus, this representation isn't actually shown to the user - the top-level statements form will be.

generators/csharp/codegen/src/ast/ClassInstantiation.ts
amckinney: Good call, I agree that the `forceUseConstructor` use case felt a little off. I ended up keeping things as-is for now, but I'm in favor of coming back to this to clean it up later.

generators/csharp/dynamic-snippets/src/context/DynamicTypeLiteralMapper.ts
Swimburger: This is valid code actually

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
amckinney: I made this change, but it impacts the `( ... )` use case since they're so tightly coupled as-is. I'd vote to come back to this improvement when we get around to resolving [this thread](https://github.com/fern-api/fern/pull/6326#discussion_r1985479592).

generators/csharp/codegen/src/ast/TypeLiteral.ts
amckinney: Done!

generators/csharp/dynamic-snippets/src/context/DynamicTypeLiteralMapper.ts
amckinney: Discussed offline - to be clear, it was due to the mismatch between `List<string>` and `List<string?>`. This makes the snippets consistent with what is produced by the C# SDK.

PR #6329, opened 2025-03-07
fix(docs): Clarify instructions for custom sections in docs

jacob-fm:
Makes it slightly more clear to use camelCase for tag names in code example

### DIFF
--- a/fern/pages/docs/api-references/customize-api-ref.mdx
+++ b/fern/pages/docs/api-references/customize-api-ref.mdx
@@ -195,7 +195,7 @@ You can add arbitrary folders in the sidebar by adding a `section` to your API R
           contents: 
             - PUT /user/{username}
             - plant
-            - plantInfo # convert tag names using camelCase convention
+            - plantInfo # tag names are converted to camelCase convention
   ```
   </Tab>
   <Tab title="Fern Definition">


fern/pages/docs/api-references/customize-api-ref.mdx
```diff
@@ -195,6 +195,7 @@ You can add arbitrary folders in the sidebar by adding a `section` to your API R
           contents: 
             - PUT /user/{username}
             - plant
+            - plantInfo # convert tag names using camelCase convention
```
chdeskur, line 198: how about "tag names are converted to camelCase convention" or "tag names will be converted using camelCase convention"

chdeskur (changes requested):

fern/pages/docs/api-references/customize-api-ref.mdx
dsinghvi: @chdeskur fyi that this camel casing is removed in the new OpenAPI parser (which respects tags)

chdeskur: huuuuuuuuge

chdeskur (approved):

PR #6334, opened 2025-03-08
feat(csharp): Add support for arbitrary parameters

amckinney:
The C# SDK now supports a `BodyProperties` and `QueryParameters` request option, which can be used to add arbitrary properties to the request. This is useful for interacting with alpha or undocumented functionality.

```csharp
var response = await client.Users.Create(
  ...,
  new RequestOptions {
    AdditionalBodyProperties = new Dictionary<string, object> { { "age", 42 } },
  }
);

### CODE REVIEW

**Comment by Swimburger** [comment]
I think query string parameters should be sent using a `NameValueCollection`. 
This type allows for duplicate query string parameter names, which is legal but not possible with a dictionary.

File: generators/csharp/codegen/src/asIs/JsonConfiguration.Template.cs [hunk 1]
```diff
@@ -94,6 +95,28 @@ internal static string SerializeAsString<T>(T obj)
...
+                    property.Value,
+                    JsonOptions.JsonSerializerOptions
+                );
+                jsonObject[property.Key] = propertyNode;
+            }
+            return jsonObject.ToJsonString(JsonOptions.JsonSerializerOptions);
+        }
+        throw new InvalidOperationException("The serialized object must be a JSON object to add properties.");
+    }
+
```
Swimburger, line 119: If I understand this correctly, this will replace the first level of properties, but not merge nested object, rather replace them.
Should we do a deep merge here?

generators/csharp/codegen/src/asIs/JsonConfiguration.Template.cs
```diff
@@ -94,6 +95,28 @@ internal static string SerializeAsString<T>(T obj)
         return json.Trim('"');
     }
 
+    internal static string SerializeWithAdditionalProperties<T>(T obj, Dictionary<string, object>? additionalProperties = null)
```
Swimburger, line 98: I'm in favor of making `additionalProperties` an object.
Since we don't have to worry about symmetric request/response type here.
That way, users can make anonymous objects or even create their own classes with JSON attributes to get it as they wish.
The only downside is, that someone could pass in primitive types likes int's which doesn't make sense, so I'd just check that when you serialize the object, the JSON node type is object.

generators/csharp/sdk/src/options/BaseOptionsGenerator.ts
```diff
@@ -189,4 +193,29 @@ export class BaseOptionsGenerator {
...
+            get: true,
+            init: true,
+            initializer: includeInitializer ? csharp.codeblock("null") : undefined
+        });
+    }
+
+    private getBodyPropertiesField({ includeInitializer }: OptionArgs): csharp.Field {
+        return csharp.field({
+            access: csharp.Access.Public,
+            name: "BodyProperties",
```
Swimburger, line 212: I think we should prefix it with `Additional` so it's clear it's an additive operation. It'll be long name but that's okay.

Swimburger: Same for QueryParameters

amckinney:
Discussed offline - few notes for posterity:

1. Use `NameValueCollection` for query parameters, which gives the user more control on the query parameter serialization strategy (e.g. `explode`, etc).
2. Update the `BodyProperties` option to use a deep object merge.
3. Keep the additional body properties typed as a `object`. Although users can shoot themselves in the foot with non-object types, this makes it possible to specify other custom classes that define their own JSON encoding attributes.
4. Update both request options to include the `Additional` prefix.
5. Add a test that includes a `null` property.

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
```diff
@@ -0,0 +1,300 @@
...
+    private WireMockServer _server;
+    private HttpClient _httpClient;
+    private RawClient _rawClient;
+    private string _baseUrl;
+
+    [SetUp]
+    public void SetUp()
+    {
+        _server = WireMockServer.Start();
+        _baseUrl = _server.Url ?? "";
```
Swimburger, line 24: ?? throw new

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
```diff
@@ -0,0 +1,300 @@
...
+        Assert.That(response.StatusCode, Is.EqualTo(200));
+
+        var content = await response.Raw.Content.ReadAsStringAsync();
+        Assert.That(content, Is.EqualTo("Success"));
+
+        Assert.That(_server.LogEntries.Count, Is.EqualTo(1));
+
+        var requestUrl = _server.LogEntries.First().RequestMessage.Url;
+        Assert.That(requestUrl, Does.Contain("foo=one"));
+        Assert.That(requestUrl, Does.Contain("foo=two"));
```
Swimburger, line 124: This does not verify whether `foo=baz` is removed

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
```diff
@@ -0,0 +1,300 @@
...
+
+        var request = new RawClient.JsonApiRequest
+        {
+            BaseUrl = _baseUrl,
+            Method = HttpMethod.Post,
+            Path = "/test",
+            Body = new Dictionary<string, object> { { "foo", "bar" } },
+            Options = new RequestOptions
+            {
+                AdditionalBodyProperties = new Dictionary<string, object> { { "foo", null! } },
```
Swimburger, line 187: make `Dictionary<string, object?>` to remove the `!`

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
```diff
@@ -0,0 +1,300 @@
...
+        const string expectedBody = """
+            {
+            "foo": {
+                "inner1": "original",
+                "inner2": "overridden",
+                "inner3": {
+                "deepProp1": "deep-override",
+                "deepProp2": "original",
+                "deepProp3": null,
+                "deepProp4": "new-value"
```
Swimburger, line 215: this indentation is off

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
```diff
@@ -0,0 +1,300 @@
...
+                }
+            },
+            "bar": "new-value"
```
Swimburger, line 218: This indentation is also off

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            }
+            """;
+
+        _server
+            .Given(
+                WireMockRequest
+                    .Create()
+                    .WithPath("/test-deep-merge")
+                    .UsingPost()
+                    .WithBody(new JsonMatcher(expectedBody))
```
Swimburger, line 228: This is cool, didn't know it could do this!

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
```diff
@@ -0,0 +1,300 @@
...
+                AdditionalBodyProperties = new Dictionary<string, object>
+                {
+                    {
+                        "foo",
+                        new Dictionary<string, object>
+                        {
+                            { "inner2", "overridden" },
+                            {
+                                "inner3",
+                                new Dictionary<string, object>
```
Swimburger, line 268: Dictionary<string, object?> to remove !

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
```diff
@@ -0,0 +1,300 @@
...
+        Assert.That(_server.LogEntries.Count, Is.EqualTo(1));
+
+        var requestBody = _server.LogEntries.First().RequestMessage.Body;
+        Assert.That(requestBody, Is.EqualTo(expectedBody));
+    }
+
+    [Test]
+    public async Task SendRequestAsync_AdditionalBodyProperties_DeepMerge()
+    {
+        const string expectedBody = """
```
Swimburger, line 206: we should update this JSON to include an array that you replace

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/RetriesTests.cs
```diff
@@ -1,6 +1,9 @@
+using System.Collections.Generic;
 using global::System.Net.Http;
 using NUnit.Framework;
+using SeedAliasExtends;
 using SeedAliasExtends.Core;
+using WireMock.Matchers;
```
Swimburger, line 6: These changes necessary?

seed/csharp-sdk/alias-extends/src/SeedAliasExtends/Core/IRequestOptions.cs
```diff
@@ -29,4 +30,15 @@ internal interface IRequestOptions
     /// The timeout for the request.
     /// </summary>
     public TimeSpan? Timeout { get; init; }
+
+    /// <summary>
+    /// Additional query parameters sent with the request.
+    /// </summary>
+    public IEnumerable<KeyValuePair<string, string>>? AdditionalQueryParameters { get; init; }
```
Swimburger, line 37: I'd make this not nullable and add a default initializer with `Enumerable.Empty<KeyValuePair<string, string>>()

seed/csharp-sdk/alias-extends/src/SeedAliasExtends/Core/JsonConfiguration.cs
```diff
@@ -100,6 +99,58 @@ internal static JsonDocument SerializeToDocument<T>(T obj) =>
...
+        }
+        MergeJsonObjects(jsonObject, additionalPropertiesJsonObject);
+        return jsonObject.ToJsonString(JsonOptions.JsonSerializerOptions);
+    }
+
+    private static void MergeJsonObjects(JsonObject baseObject, JsonObject overrideObject)
+    {
+        foreach (var property in overrideObject)
+        {
+            if (!baseObject.ContainsKey(property.Key))
```
Swimburger, line 133: In C# we use the TryX if available:
[TryGetPropertyValue(String, JsonNode)](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.nodes.jsonobject.trygetpropertyvalue?view=net-9.0#system-text-json-nodes-jsonobject-trygetpropertyvalue(system-string-system-text-json-nodes-jsonnode@))

seed/csharp-sdk/alias-extends/src/SeedAliasExtends/Core/IRequestOptions.cs
```diff
@@ -29,4 +30,15 @@ internal interface IRequestOptions
...
+
+    /// <summary>
+    /// Additional body properties sent with the request.
+    /// This is a no-op for multipart/form-data endpoints.
```
Swimburger, line 41: I'd change this to saying it is only applied to JSON endpoints

seed/csharp-sdk/accept-header/src/SeedAccept/Core/RawClient.cs
```diff
@@ -569,6 +579,53 @@ and not string
         return url;
     }
 
+    private static List<KeyValuePair<string, string>> GetQueryParameters(BaseApiRequest request)
+    {
+        var result = TransformToKeyValuePairs(request.Query);
+        if (
+            request.Options?.AdditionalQueryParameters != null
```
Swimburger, line 586: if `request.Options?.AdditionalQueryParameters` is null, I'd early return.
Also, the `is null` and `is not null` is preferred

Swimburger (changes requested):

Swimburger:
Looks great, small nits, but I'd include them before merging

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
amckinney: Discussed offline - this is left as-is for consistency with other tests.

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
amckinney: Good call - I'll add that in.

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/AdditionalParametersTests.cs
amckinney: Fixed!

seed/csharp-sdk/alias-extends/src/SeedAliasExtends.Test/Core/RawClientTests/RetriesTests.cs
amckinney: No - that came from a bad merge with your recent changes. Will update.

seed/csharp-sdk/accept-header/src/SeedAccept/Core/RawClient.cs
amckinney: Great call, definitely in favor of the early return.

generators/csharp/base/src/asIs/JsonConfiguration.Template.cs
```diff
@@ -103,6 +102,46 @@ internal static JsonDocument SerializeToDocument<T>(T obj) =>
...
+                continue;
+            }
+            if (existingValue is JsonObject nestedBaseObject && property.Value is JsonObject nestedOverrideObject)
+            {
+                // If both values are objects, recursively merge them.
+                MergeJsonObjects(nestedBaseObject, nestedOverrideObject);
+                continue;
+            }
+            // Otherwise, the overrideObject takes precedence.
+            baseObject[property.Key] = property.Value != null ? JsonNode.Parse(property.Value.ToJsonString()) : null;
```
Swimburger, line 141: Is `baseObject[property.Key] = property.Value != null ? JsonNode.Parse(property.Value.ToJsonString()) : null;` is duplicate, you can simplify this

Swimburger (approved):

generators/csharp/base/src/asIs/JsonConfiguration.Template.cs
amckinney: Discussed offline - going to leave as-is for now.

PR #6348, opened 2025-03-10
fix(csharp): consistent request wrapper with body params vs request reference behavior

Swimburger:
## Changes Made
- Make the behavior between a wrapped request with body properties and normal body request consistent.
  Previously, a wrapped request with body properties would not omit `null` values even if the JSON configuration is configured to omit `null` values.
- Fix a bug where required properties that were `[JsonIgnore]` threw an error during serialization.
- Add `AdditionalProperties` property to wrapped requests with body properties.
- Improve performance of query string value to string conversion by relying less on `JsonSerializer` + trim and more on `ToString()`.

## Testing

### DIFF
--- a/fern/pages/changelogs/csharp-sdk/2025-03-10.mdx
+++ b/fern/pages/changelogs/csharp-sdk/2025-03-10.mdx
@@ -6,7 +6,5 @@ Previously, a wrapped request with body properties would not omit `null` values 
 **`(fix):`** Fix a bug where required properties that were `[JsonIgnore]` threw an error during serialization.
 
 
-**`(fix):`** Add `AdditionalProperties` property to wrapped requests with body properties.
-
 **`(feat):`** Improve performance of query string value to string conversion by relying less on `JsonSerializer` and more on `ToString()`.
 


generators/csharp/sdk/versions.yml
```diff
@@ -6,6 +6,21 @@
...
+  changelogEntry:
+    - type: fix
+      summary: |
+        Make the behavior between a wrapped request with body properties and normal body request consistent.
+        Previously, a wrapped request with body properties would not omit `null` values even if the JSON configuration is configured to omit `null` values.
+    - type: fix
+      summary: |
+        Fix a bug where required properties that were `[JsonIgnore]` threw an error during serialization.
+    - type: fix
+      summary: Add `AdditionalProperties` property to wrapped requests with body properties.
```
amckinney, line 21: I think we'll want to revert this - we don't need to expose AdditionalProperties on inlined requests since they'll never be used as response objects. At least that was the original intention.

amckinney (approved):
Aside from the AdditionalProperties change, implementation looks great!

PR #6366, opened 2025-03-10
fix(cli): Only construct an enum when all parameter subtypes are literal string values.

eyw520:
## Description
This PR resolves a bug where parameter components would be incorrectly converted to enums, leading to dropped union values.

## Changes Made
- Update logic in `buildQueryParameter` during OpenAPI-IR to fern conversion.

## Testing

### DIFF
--- a/generators/ruby-v2/ast/package.json
+++ b/generators/ruby-v2/ast/package.json
@@ -17,8 +17,8 @@
   "scripts": {
     "clean": "rm -rf ./lib && tsc --build --clean",
     "compile": "tsc --build",
-    "test": "vitest --passWithNoTests --run",
-    "test:update": "vitest --passWithNoTests --run -u",
+    "test": "vitest --run",
+    "test:update": "vitest --run -u",
     "lint:eslint": "eslint --max-warnings 0 . --ignore-pattern=../../../.eslintignore",
     "lint:eslint:fix": "yarn lint:eslint --fix",
     "format": "prettier --write --ignore-unknown --ignore-path ../../../shared/.prettierignore \"**\"",
--- a/generators/ruby-v2/ast/src/ast/Comment.ts
+++ b/generators/ruby-v2/ast/src/ast/Comment.ts
@@ -1,0 +1,25 @@
+import { AstNode } from "./core/AstNode";
+import { Writer } from "./core/Writer";
+
+export declare namespace Comment {
+    interface Args {
+        docs?: string;
+    }
+}
+
+export class Comment extends AstNode {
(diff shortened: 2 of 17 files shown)


packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildQueryParameter.ts
```diff
@@ -125,7 +125,11 @@ function getQueryParameterTypeReference({
                 }),
                 allowMultiple: true
             };
-        } else if (resolvedSchema.type === "oneOf" && resolvedSchema.value.type === "undiscriminated") {
+        } else if (
+            resolvedSchema.type === "oneOf" &&
+            resolvedSchema.value.type === "undiscriminated" &&
+            resolvedSchema.value.schemas.every((s) => s.type === "literal" && s.value.type === "string")
+        ) {
```
amckinney, line 132: nit: It'd be nice if there was a helper method for this, e.g. `isUndiscriminatedUnionOfLiterals` or something similar.

amckinney (approved):

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/deepgram.json
```diff
@@ -10220,9 +10216,8 @@ channel:
                     "type": "optional<string>",
                   },
                   "keywords": {
-                    "allow-multiple": true,
```
amckinney, line 10223: Discussed offline and caught this after the fact - we should _not_ be dropping the `allow-multiple` key in this case. We should continue to specify the type as `string`.

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/deepgram.json
```diff
@@ -10519,103 +10579,139 @@ channel:
...
+              "literal<"meeting">",
+              "literal<"phonecall">",
+              "literal<"finance">",
+              "literal<"conversationalai">",
+              "literal<"voicemail">",
+              "literal<"video">",
+              {
+                "docs": "Custom string if you've had a model trained by Deepgram",
+                "type": "string",
               },
```
amckinney, line 10621: Discussed offline - this solution works really well for TypeScript, but it's terrible for other languages (e.g. Go, Java, etc). We should instead generate this type as an `enum` and a `string`. This will end up with a far better experience across the board.

PR #6368, opened 2025-03-10
feat(csharp): Multipart form data

Swimburger:
## Description
Add support for multipartform requests with file and non-file parameters.
This is useful when you want to send a file along with other form data in a single request.

For example, you can use the following code to upload a file with a description and some metadata:
```csharp
await client.UploadAsync(
    new UploadRequest
    {
        File = new FileParameter { Stream = fileStream, FileName = "file.txt" },
        Description = "This is a test file",
        Meta = new Meta {
            Key1 = "value1",
            Key2 = "value2"
        }
    }
);
```

## Testing


seed/csharp-sdk/file-upload/src/SeedFileUpload/Core/QueryString.cs
```diff
@@ -0,0 +1,82 @@
+using System.Text.Json;
+using SeedFileUpload.Core;
+
+internal static class QueryString
```
amckinney, line 4: Discussed offline - couple options here.

1. Redefine these methods as extensions on the `Convert` class.
2. Update the name of this to be something like `QueryStringConverter`.

seed/csharp-sdk/file-upload/src/SeedFileUpload/Service/ServiceClient.cs
```diff
@@ -20,22 +20,40 @@ internal ServiceClient(RawClient client)
...
+        multipartForm.AddStringPart("maybe_integer", request.MaybeInteger);
+        multipartForm.AddStringParts("optional_list_of_strings", request.OptionalListOfStrings);
+        multipartForm.AddJsonParts("list_of_objects", request.ListOfObjects);
+        multipartForm.AddJsonPart("optional_metadata", request.OptionalMetadata);
+        multipartForm.AddJsonPart("optional_object_type", request.OptionalMetadata);
+        multipartForm.AddStringPart("optional_id", request.OptionalId);
+        multipartForm.AddJsonParts("list_of_objects_with_optionals", request.ListOfObjectsWithOptionals);
+        multipartForm.AddJsonPart("alias_object", request.AliasObject);
+        multipartForm.AddJsonParts("list_of_alias_object", request.ListOfAliasObject);
+        multipartForm.AddJsonParts("alias_list_of_object", request.AliasListOfObject);
```
amckinney, line 45: Discussed offline - this is purely cosmetic so it's fine to start this way, but we might eventually refactor this so that the `null` checks and the `foreach` statements are more explicit in the generated method.

seed/csharp-sdk/file-upload/src/SeedFileUpload/Service/ServiceClient.cs
```diff
@@ -20,22 +20,40 @@ internal ServiceClient(RawClient client)
...
         var response = await _client
             .SendRequestAsync(
-                new RawClient.JsonApiRequest
-                {
-                    BaseUrl = _client.Options.BaseUrl,
-                    Method = HttpMethod.Post,
-                    Path = "",
-                    Options = options,
-                },
+                multipartForm,
```
amckinney, line 48: nit: I think we should name this `multipartFormRequest` like in the other files below.

test-definitions/fern/apis/file-upload/ir.json
```diff
@@ -3,30 +3,14 @@
   "apiVersion": null,
```
amckinney, line 3: Let's remove this file.

amckinney (commented):
Looks great so far - mostly just naming.

generators/csharp/base/src/asIs/QueryStringConverter.Template.cs
```diff
@@ -0,0 +1,230 @@
...
+        switch (element.ValueKind)
+        {
+            case JsonValueKind.Object:
+                foreach (var property in element.EnumerateObject())
+                {
+                    var newPrefix = string.IsNullOrEmpty(prefix)
+                        ? property.Name
+                        : $"{prefix}[{property.Name}]";
+
+                    JsonToForm(property.Value, newPrefix, parameters);
```
amckinney, line 90: Nice, this means we support deep object query params more broadly in C# now?

generators/csharp/base/src/asIs/test/RawClientTests/MultipartFormTests.Template.cs
```diff
@@ -0,0 +1,900 @@
...
+    {
+        var multipartFormRequest = CreateMultipartFormRequest();
+        multipartFormRequest.AddFormEncodedParts("objects", [_simpleObject, _simpleObject]);
+
+        var httpContent = multipartFormRequest.CreateContent();
+        Assert.That(httpContent, Is.InstanceOf<MultipartFormDataContent>());
+        var multipartContent = (MultipartFormDataContent)httpContent;
+
+        var boundary = GetBoundary(multipartContent);
+        var expected = $"""
```
amckinney, line 546: This level of testing is _awesome_ 🎉

generators/csharp/sdk/src/endpoint/http/RawClient.ts
```diff
@@ -130,22 +140,170 @@ export class RawClient {
...
+                break;
+            case "bodyProperty": {
+                propertyName = property.name.name.pascalCase.safeName;
+                partName = property.name.wireValue;
+                contentType = property.contentType;
+                encoding = property.style;
+                csharpType = this.context.csharpTypeMapper.convert({
+                    reference: property.valueType
+                });
+            }
```
amckinney, line 233: We should `assertNever` here.

amckinney (approved):
🚀

generators/csharp/base/src/asIs/QueryStringConverter.Template.cs
Swimburger: I didn't include it in our raw client because we don't support it in our IR. Deep object was the first thing I implemented as I assumed that was just normal form encoding, but it is not.

PR #6369, opened 2025-03-11
fix(cli): Revert only construct an enum when all parameters are literals

eyw520:
This reverts https://github.com/fern-api/fern/pull/6366 - we'll refactor this implementation with an undiscriminated union that contains an enum and a string (rather than a long list of `literal` values).


amckinney (approved):
Reverting re: https://github.com/fern-api/fern/pull/6366#discussion_r1988174359

PR #6378, opened 2025-03-11
feat(ruby): Add basic RBS types and generation of method signatures

ethnt:
## Description

Adds the ability to describe method signatures (parameters and return types) with RBS types.

## Changes Made
- Adds `Type`
- Adds `Type` and a return `Type` to `Method`

## Testing

### DIFF
--- a/generators/ruby-v2/ast/src/ast/Method.ts
+++ b/generators/ruby-v2/ast/src/ast/Method.ts
@@ -7,7 +7,6 @@ import { PositionalSplatParameter } from "./PositionalSplatParameter";
 import { Type } from "./Type";
 import { YieldParameter } from "./YieldParameter";
 import { AstNode } from "./core/AstNode";
-import { TypedAstNode } from "./core/TypedAstNode";
 import { Writer } from "./core/Writer";
 
 export type MethodKind = "instance" | "class";


generators/ruby-v2/ast/src/ast/core/TypedAstNode.ts
```diff
@@ -0,0 +1,28 @@
+import { AbstractWriter } from "@fern-api/browser-compatible-base-generator";
+import { AbstractFormatter } from "@fern-api/browser-compatible-base-generator";
+
+import { BaseRubyCustomConfigSchema } from "../../custom-config/BaseRubyCustomConfigSchema";
+import { AstNode } from "./AstNode";
+import { RubyFile } from "./RubyFile";
+
+export abstract class TypedAstNode extends AstNode {
```
dsinghvi, line 8: We chatted offline -- but its probably cleaner to rename this to `RubyAstNode` and make `writeTypeDefinition` a noop by default. Anything that has rbs implications can then override the `writeTypeDefinition` function

generators/ruby-v2/ast/src/ast/Type.ts
```diff
@@ -1,46 +1,341 @@
...
+
+interface Generic {
+    type: "generic";
+    baseKlass: string;
+    parameters: (Type | TypeParameter)[];
+}
+
+export type BaseType = Self | Class_ | Instance | Boolean_ | Nil | Top | Bot | Void;
+export type SingleType = Boolish | String_ | Integer | Union | Intersection | Singleton | Object_ | Generic;
+// TODO: These conceivably could just be Generics, at the cost of some specificity/usability
```
dsinghvi, line 95: i like them being specific as you have them since they are likely going to be used more often

dsinghvi (approved):
👏

PR #6403, opened 2025-03-12
feat(ruby): Add Ruby `Class` and `Module` AST nodes, flesh out `TypeParameter`

ethnt:
## Description

Provides the ability to create modules and classes in Ruby using the AST.

I have not added 

## Changes Made
- Add `Class_`
- Add `Module`
- Modify `TypeParameter` to allow for variances, unchecked, bounds, etc.
- Namespace tracking for modules and classes

## Testing

### DIFF
--- a/generators/ruby-v2/ast/src/ast/Module.ts
+++ b/generators/ruby-v2/ast/src/ast/Module.ts
@@ -9,7 +9,7 @@ export declare namespace Module {
         /* The module's name. */
         name: string;
         /* This module's namespace (i.e., the modules/classes it is wrapped in). */
-        namespace?: (Module | Class_)[];
+        namespace?: Set<Module | Class_>;
         /* If this module is generic, it takes a type parameter. */
         typeParameters?: TypeParameter[];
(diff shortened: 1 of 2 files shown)


generators/ruby-v2/ast/src/ast/Class.ts
```diff
@@ -0,0 +1,86 @@
...
+export declare namespace Class_ {
+    export interface Args extends Module.Args {
+        /* The superclass of this class. */
+        superclass?: Class_;
+    }
+}
+
+export class Class_ extends Module {
+    public readonly superclass: Class_ | undefined;
+    public readonly statements: AstNode[];
```
dsinghvi, line 15: not blocking, but do you think it makes sense to have a separate argument for fields? Or is the consumer going to have to specify that as a statement some how? (something like `@@no_of_customers = 0`, stolen from [here](https://www.tutorialspoint.com/ruby/ruby_classes.htm))

dsinghvi (approved):
This code is clean 🧑‍🍳

PR #6407, opened 2025-03-12
chore(go): Upgrade to IRv57

akallem:
- Treats nullable types as optional, and collapses `optional<nullable<...>>` or `nullable<optional<...>>` into a single optional
- Rejects (does not implement) IRv54 bytes responses
- Rejects (does not implement) IRv56 custom pagination

### DIFF
--- a/packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/nullable/type_nullable_User.json
+++ b/packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/nullable/type_nullable_User.json
@@ -42,6 +42,53 @@
     },
     "favorite-number": {
       "$ref": "#/definitions/nullable.WeirdNumber"
+    },
+    "numbers": {
+      "oneOf": [
+        {
+          "oneOf": [
+            {
+              "type": "array",
+              "items": {
+                "type": "integer"
+              }
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
+        {
+          "type": "null"
+        }
+      ]
+    },
+    "strings": {
+      "oneOf": [
+        {
+          "oneOf": [
+            {
+              "type": "object",
+              "additionalProperties": {
+                "type": [
+                  "string",
+                  "number",
+                  "boolean",
+                  "object",
+                  "array",
+                  "null"
+                ]
+              }
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
+        {
+          "type": "null"
+        }
+      ]
     }
   },
   "required": [
@@ -195,6 +242,23 @@
           "oneOf": [
             {
               "type": "number"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
+        {
+          "oneOf": [
+            {
+              "oneOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ]
             },
             {
               "type": "null"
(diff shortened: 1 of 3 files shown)


generators/go/internal/generator/snippet.go
```diff
@@ -865,6 +865,20 @@ func (e *exampleContainerVisitor) VisitMap(mapContainer *ir.ExampleMapContainer)
...
+	}
+	e.value = &ast.Optional{
+		Expr: exampleTypeReferenceShapeToGoType(
+			nullable.Nullable.Shape,
+			e.types,
+			e.baseImportPath,
+		),
+	}
+	return nil
+}
```
amckinney, line 880: Do we need to prevent this from becoming a doubly wrapped optional?

seed/go-sdk/nullable/snippet.json
```diff
@@ -1,37 +1 @@
-{
```
amckinney, line 1: It's worth looking into why this is failing now - there must have been some error in the snippet generator. I bet we could see what's going wrong by introspecting [here](https://github.com/fern-api/fern/blob/104e947574915a609894487e2e245a43f7c23e49/generators/go/internal/generator/generator.go#L960).

generators/go/internal/generator/model.go
```diff
@@ -982,8 +982,23 @@ func (u *undiscriminatedUnionContainerTypeVisitor) VisitMap(mapType *ir.MapType)
 	return nil
 }
 
-func (u *undiscriminatedUnionContainerTypeVisitor) VisitOptional(optional *ir.TypeReference) error {
-	u.value = fmt.Sprintf("%sOptional", typeReferenceToUndiscriminatedUnionField(optional, u.types, u.scope))
+func (u *undiscriminatedUnionContainerTypeVisitor) VisitNullable(nullable *ir.TypeReference) error {
+	// We treat nullable the same as optional, so we can just delegate to the visit implementation for optional.
+	u.VisitOptional(nullable)
```
amckinney, line 987: Nice 👍

seed/go-sdk/nullable/snippet.json
akallem: Ah apologies, I missed this in the diff.

Still not able to see the logs in the generated tmp file, nevertheless I've force pushed and addressed this. There is one change in the snippet diff which I'd like to discuss with you when you're back.

generators/go/internal/generator/snippet.go
akallem: Yes indeed! In fact I neglected to address snippet generation entirely -- I've pushed an update for that. (Force pushed, so the diff to snippet.json is clear.)

seed/go-sdk/nullable/snippet.json
amckinney: Discussed offline - the snippet changes look great! Looks like this actually resolved an issue from before.

generators/go/internal/generator/model.go
```diff
@@ -1245,21 +1260,46 @@ func (c *containerTypeVisitor) VisitMap(mapType *ir.MapType) error {
...
+	value := strings.TrimLeft(typeReferenceToGoType(optionalOrNullable, c.types, c.scope, c.baseImportPath, c.importPath, c.includeOptionals, c.packageLayout), "*")
 	if c.includeOptionals {
 		c.value = fmt.Sprintf("*core.Optional[%s]", value)
 		return nil
 	}
-	if optional.Unknown != nil || (optional.Container != nil && optional.Container.Literal == nil) {
+
+	isPointerRequired := func(typeReference *ir.TypeReference) bool {
+		return !(typeReference.Unknown != nil || (typeReference.Container != nil && typeReference.Container.Literal == nil))
+	}
```
amckinney, line 1283: Let's make a top-level function for this.

amckinney (approved):
Nice!!

PR #6412, opened 2025-03-13
feat(csharp): Implement union and file upload snippets

amckinney:
Now that the C# generator supports discriminated unions and multipart/form-data endpoints, this updates the snippet generator to match this behavior.

### DIFF
--- a/generators/browser-compatible-base/src/dynamic-snippets/Result.ts
+++ b/generators/browser-compatible-base/src/dynamic-snippets/Result.ts
@@ -36,7 +36,7 @@ export class Result {
     }
 
     private shouldUpdate({ snippet, reporter }: { snippet: string; reporter: ErrorReporter }): boolean {
-        if (this.reporter == null || reporter.size() === 0) {
+        if (this.reporter == null || (this.reporter.size() > 0 && reporter.size() === 0)) {
             return true;
         }
         return snippet.length > (this.snippet?.length ?? 0);


generators/csharp/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
```diff
@@ -164,9 +193,11 @@ export class DynamicSnippetsGeneratorContext extends AbstractDynamicSnippetsGene
...
 
     public getNamespaceForPublicCoreClasses(): string {
-        return this.customConfig?.["root-namespace-for-core-classes"]
-            ? this.getRootNamespace()
-            : this.getCoreNamespace();
+        const rootNamespaceForCoreClasses = this.customConfig?.["root-namespace-for-core-classes"];
+        if (rootNamespaceForCoreClasses == null) {
+            return this.getRootNamespace();
+        }
+        return rootNamespaceForCoreClasses ? this.getRootNamespace() : this.getCoreNamespace();
```
Swimburger, line 200: This code does the same thing as before AFAIK.

seed/csharp-sdk/unions/use-discriminated-unions/src/SeedApi.DynamicSnippets/Example1.cs
```diff
@@ -13,14 +12,8 @@ public class Example1
...
-        await client.Bigunion.UpdateAsync(
-            new Dictionary<string, object>() {
-                ["type"] = "normalSweet",
-                ["value"] = "value",
-                ["id"] = "id",
-                ["created-at"] = "2024-01-15T09:30:00Z",
-                ["archived-at"] = "2024-01-15T09:30:00Z",
-            }
+        await client.Union.UpdateAsync(
+
```
Swimburger, line 16: Not sure why this is empty

Swimburger (changes requested):

generators/csharp/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
amckinney: It's actually a little different - this now supports users that configure `root-namespace-for-core-classes: false` which wasn't used before. It also makes the new default behavior to be `getRootNamespace`.

There might be a better way to do this, but the intention is:
1. If the configuration is not specified at all, default to the root namespace.
2. If the configuration is specified, respect it.

seed/csharp-sdk/unions/use-discriminated-unions/src/SeedApi.DynamicSnippets/Example1.cs
amckinney: I dug in here - this was an interesting edge case resolved in [this commit](https://github.com/fern-api/fern/pull/6412/commits/bc30ffcd39c989fead336557c96e218f0e273efa). I'll break it down:

1. This test definition defines two endpoints with the same method and path (e.g. `PATCH /`).
2. The dynamic snippet generator handles duplicate endpoints by generating the _best_ snippet between both endpoints and selecting that.
    * This is especially relevant for streaming endpoints (which share the same method and path).
3. The original tie-breaking logic was based on the number of errors returned by the snippet.
4. Given that union examples are broken in some ways, there isn't a valid snippet being produced.
5. We fix this by updating the tie-breaking logic to _also_ consider the length of the snippet, and prefer the longer one.

Swimburger (approved):

PR #6438, opened 2025-03-13
fix(csharp): Rename Custom method to FromCustom in String Enums

Swimburger:
## Description

Forward compatible enums has a static method `Custom(string value)` that allows you to create a new instance of the enum with a custom value.
"Custom" is a commonly used enum value, and we want to avoid conflicts with the static method, so we're renaming the static method to `FromCustom(string value)`.
This feature is gated behind the `experimental-enable-forward-compatible-enums` configuration option, so we're accepting this as a breaking change without a major version bump.

---

Forward-compatible enums can be explicitly cast from and to strings.
```csharp
string myEnumAsString = (string)MyEnum.Enum1;
MyEnum myEnum = (MyEnum)"custom-value";
```
Note: We're not supporting implicit casts here because it could lead to behavior the user doesn't expect.

## Testing

### DIFF
--- a/packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/enum/type__EnumWithCustom.json
+++ b/packages/cli/fern-definition/ir-to-jsonschema/src/__test__/__snapshots__/enum/type__EnumWithCustom.json
@@ -1,0 +1,8 @@
+{
+  "type": "string",
+  "enum": [
+    "safe",
+    "Custom"
+  ],
+  "definitions": {}
+}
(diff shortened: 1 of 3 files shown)


generators/csharp/model/src/enum/StringEnumGenerator.ts
```diff
@@ -17,6 +18,12 @@ export class StringEnumGenerator extends FileGenerator<CSharpFile, ModelCustomCo
     ) {
         super(context);
         this.classReference = this.context.csharpTypeMapper.convertToClassReference(this.typeDeclaration.name);
+        this.customMethodName = this.getCustomMethodName(enumDeclaration);
+    }
+
+    private getCustomMethodName(enumDeclaration: EnumTypeDeclaration): string {
+        const d = "FromCustom";
+        return enumDeclaration.values.some((v) => v.name.name.pascalCase.safeName === d) ? "FromCustom_" : d;
```
amckinney, line 26: nice 👍

amckinney (approved):

PR #6443, opened 2025-03-13
feat(java): Add support for inline-file-properties in snippets

amckinney:
This adds support for the new `inline-file-properties` configuration (introduced in https://github.com/fern-api/fern/pull/6418) in dynamic snippets.


ajgateno (approved):
LGTM

PR #6454, opened 2025-03-14
feat(php): Support sending explicit nulls

akallem:


seed/php-sdk/nullable/src/Core/Json/JsonSerializableType.php
```diff
@@ -50,7 +63,20 @@ public function jsonSerialize(): array
...
+            if (!array_key_exists($property->getName(), $this->__explicitlySetProperties)) {
+                 continue;
+            }
+
             $value = $property->getValue($this);
+            $type = $property->getType();
+
+            // Omit properties that were explicitly set to null when the type isn't nullable,
+            // i.e., treat setting such properties to null as unsetting them.
+            if ($type && !$type->allowsNull() && $value === null) {
```
akallem, line 77: @amckinney Following up on our conversation, instead of adding a new param to this class's constructor to track non-nullable properties (or the inverse), I tried this alternative idea I speculated about, namely, using reflection to check whether a property is nullable.

I'm not sure I'll have enough time to test this new addition sufficiently before demo day followed by my EOD, so I wanted to flag it to make sure you saw it.

PR #6466, opened 2025-03-17
feat(python): Add support for response headers

amckinney:
## Problem

It's currently impossible for users to retrieve response headers from the Python SDK. We need to extend the generated SDK so that users can access these values as needed. This outlines a design for exposing raw HTTP response metadata (e.g. response headers) alongside the returned response. 

## Proposal

There's a few important details to call out with this feature:

1. The solution _must_ preserve backwards compatibility.
2. It should _not_ negatively impact the primary experience (i.e. endpoints that don't require response headers).
3. It should be widely reusable across every endpoint so that endpoints can freely add response headers in the future.

With all that said, we propose a separate client that returns an `APIResponse` wrapper that contains the HTTP response metadata alongside the strongly typed response body. Users will need to opt-in to these separate methods by calling the `with_raw_response` method on the client like so:

```python
client = AcmeClient()

# By default, we keep all the methods as-is and incur no breaking changes.
# This means we can access the standard response properties directly.
iterator = client.users.list(...)
for user in iterator:
    print(user)

# Alternatively, you can act upon the "with_raw_response" client that includes HTTP
# response headers.
#
# The caller needs to access the values with the `data` property.
response = client.users.with_raw_response.list(...)
for user in response.data:
    print(user)

# The caller can also access the response headers alongside it.
print(f"Got organization id: {response.headers.get('organization_id')}")
```

### Unary

This will require a few new components, including a generic `APIResponse` type as well as a "raw" client for each of the generated clients, both of which are shown below:

```python
class RawUserClient:
    def __init__(self, *, client_wrapper: SyncClientWrapper):
        self._client_wrapper = client_wrapper
    
    def list(
        self, *, request_options: typing.Optional[RequestOptions] = None
    ) -> APIResponse[typing.Iterator[User]]:
        """
        Equivalent to the `list` method on the UserClient class, but returns the raw APIResponse.
        """
        ...
        try:
            if 200 <= _response.status_code < 300:
                data = typing.cast(
                    User,
                    parse_obj_as(
                        type_=User,  # type: ignore
                        object_=_response.json(),
                    ),
                )
                return APIResponse(
                    headers=dict(_response.headers),
                    data=data,
                )
        ...


class UserClient:
    def __init__(self, *, client_wrapper: SyncClientWrapper):
        self._raw_client = RawUserClient(client_wrapper=client_wrapper)
        self._list = wrap_func_to_exclude_raw_response(self._raw_client.list)
    
    def with_raw_response(self) -> ImdbRawClient:
        """
        Access the raw client whenever HTTP response metadata is needed.
        """
        return self._raw_client

    def list(
        self, *, request_options: typing.Optional[RequestOptions] = None
    ) -> typing.Iterator[User]:
        return self._list(
            request_options=request_options,
        )

    ...
```

### Streaming

Streaming endpoints act similarly, but the method that supports response headers now requires the user to use _enter_ the generator context manager like so:

```python
with client.dummy.with_raw_response.generate_stream(
    num_events=42,
) as response:
    # Now that we've enterered the context manager, the response headers will be available.
    print(response.headers.get("X-Your-Header"))
    
    # We can iterate over the stream content the same as before, but with the `.data` property.
    for chunk in response.data:
        yield chunk
```


amckinney:
This proposal served its purpose - this fearture is being worked on in a separate PR, so I'll close this as-is.

PR #6500, opened 2025-03-20
fix(java): Fix environment import in dynamic snippets

amckinney:
This resolves a small issue where the environment was incorrectly imported in Java snippets.


ajgateno (approved):
Nice

PR #6501, opened 2025-03-20
feat(java): Add snippets to generated README.md

ajgateno:
This PR adds snippets to the generated README.md we produce in Java

### DIFF
--- a/generators/java-v2/sdk/features.yml
+++ b/generators/java-v2/sdk/features.yml
@@ -16,7 +16,7 @@ features:
       Paginated requests will return an Iterable<T>, which can be used to loop through the underlying items, or stream them. You can also call 
       `nextPage` to perform the pagination manually
 
-  - id: ERRORS
+  - id: EXCEPTION_HANDLING
     description: |
       When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/generators/java-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
+++ b/generators/java-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
@@ -16,6 +16,7 @@ interface EndpointWithFilepath {
 export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
     private static CLIENT_VARIABLE_NAME = "client";
     private static ENVIRONMENTS_FEATURE_ID: FernGeneratorCli.FeatureId = "ENVIRONMENTS";
+    private static EXCEPTION_HANDLING_FEATURE_ID: FernGeneratorCli.FeatureId = "EXCEPTION_HANDLING";
     private static BASE_URL_FEATURE_ID: FernGeneratorCli.FeatureId = "BASE_URL";
     private static SNIPPET_PACKAGE_NAME = "com.example.usage";
     private static ELLIPSES = java.codeblock("...");
@@ -71,7 +72,9 @@ export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
             [ReadmeSnippetBuilder.BASE_URL_FEATURE_ID]: {
                 renderer: this.renderBaseUrlSnippet.bind(this)
             },
-            [FernGeneratorCli.StructuredFeatureId.Errors]: { renderer: this.renderErrorsSnippet.bind(this) },
+            [ReadmeSnippetBuilder.EXCEPTION_HANDLING_FEATURE_ID]: {
+                renderer: this.renderExceptionHandlingSnippet.bind(this)
+            },
             [FernGeneratorCli.StructuredFeatureId.CustomClient]: {
                 renderer: this.renderCustomClientSnippet.bind(this)
             },
@@ -189,7 +192,7 @@ export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
         return this.renderSnippet(snippet);
     }
 
-    private renderErrorsSnippet(endpoint: EndpointWithFilepath): string {
+    private renderExceptionHandlingSnippet(endpoint: EndpointWithFilepath): string {
         const endpointMethodInvocation = this.getMethodCall(endpoint, [ReadmeSnippetBuilder.ELLIPSES]);
 
         const apiExceptionClassReference = this.context.getApiExceptionClassReference();
--- a/seed/java-sdk/multi-url-environment/README.md
+++ b/seed/java-sdk/multi-url-environment/README.md
@@ -58,7 +58,7 @@ SeedMultiUrlEnvironmentClient client = SeedMultiUrlEnvironmentClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/pagination/README.md
+++ b/seed/java-sdk/pagination/README.md
@@ -98,7 +98,7 @@ for (
 }
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/accept-header/accept-header/README.md
+++ b/seed/java-sdk/accept-header/accept-header/README.md
@@ -38,7 +38,7 @@ SeedAcceptClient client = SeedAcceptClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/alias/README.md
+++ b/seed/java-sdk/alias/README.md
@@ -37,7 +37,7 @@ SeedAliasClient client = SeedAliasClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/any-auth/README.md
+++ b/seed/java-sdk/any-auth/README.md
@@ -48,7 +48,7 @@ SeedAnyAuthClient client = SeedAnyAuthClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/api-wide-base-path/README.md
+++ b/seed/java-sdk/api-wide-base-path/README.md
@@ -37,7 +37,7 @@ SeedApiWideBasePathClient client = SeedApiWideBasePathClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/audiences/README.md
+++ b/seed/java-sdk/audiences/README.md
@@ -59,7 +59,7 @@ SeedAudiencesClient client = SeedAudiencesClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/auth-environment-variables/README.md
+++ b/seed/java-sdk/auth-environment-variables/README.md
@@ -38,7 +38,7 @@ SeedAuthEnvironmentVariablesClient client = SeedAuthEnvironmentVariablesClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/basic-auth-environment-variables/README.md
+++ b/seed/java-sdk/basic-auth-environment-variables/README.md
@@ -41,7 +41,7 @@ SeedBasicAuthEnvironmentVariablesClient client = SeedBasicAuthEnvironmentVariabl
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/basic-auth/README.md
+++ b/seed/java-sdk/basic-auth/README.md
@@ -41,7 +41,7 @@ SeedBasicAuthClient client = SeedBasicAuthClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/bearer-token-environment-variable/README.md
+++ b/seed/java-sdk/bearer-token-environment-variable/README.md
@@ -38,7 +38,7 @@ SeedBearerTokenEnvironmentVariableClient client = SeedBearerTokenEnvironmentVari
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
--- a/seed/java-sdk/cross-package-type-names/README.md
+++ b/seed/java-sdk/cross-package-type-names/README.md
@@ -45,7 +45,7 @@ SeedCrossPackageTypeNamesClient client = SeedCrossPackageTypeNamesClient
     .build();
 ```
 
-## Errors
+## Exception Handling
 
 When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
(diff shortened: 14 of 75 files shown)


generators/java-v2/ast/src/ast/Type.ts
```diff
@@ -84,6 +88,12 @@ interface Reference {
     value: ClassReference;
 }
 
+interface ParameterizedReference {
+    type: "parameterizedReference";
+    value: ClassReference;
+    parameters: Type[];
+}
```
amckinney, line 95: Is this equivalent to a generic, i.e. `Optional<T, U>`? Should we just call it `Generic`?

generators/java-v2/ast/src/ast/Type.ts
```diff
@@ -161,6 +196,15 @@ export class Type extends AstNode {
             case "uuid":
                 writer.writeNode(UUIDClassReference);
                 break;
+            case "void":
+                writer.write("Void");
```
amckinney, line 200: `void` is normally lowercase right?

```java
public static void main(String[] args) {
    ...
}
```

generators/java-v2/base/src/context/AbstractJavaGeneratorContext.ts
```diff
@@ -1,23 +1,58 @@
...
         public readonly customConfig: CustomConfig,
         public readonly generatorNotificationService: GeneratorNotificationService
     ) {
         super(config, generatorNotificationService);
+        this.javaTypeMapper = new JavaTypeMapper(this);
+    }
+
+    public abstract getRootPackageName(): string;
+
+    public abstract getTypesPackageName(fernFilePath: FernFilepath): string;
```
amckinney, line 25: Thoughts on implementing this methods here so that the same implementation is used across all contexts? The implementation should be the same everywhere, right?

generators/java-v2/sdk/features.yml
```diff
@@ -5,46 +5,43 @@ features:
 
   - id: ENVIRONMENTS
     description: |
-      You can choose between different environments by using the `option.WithBaseURL` option. You can configure any arbitrary base
-      URL, which is particularly useful in test environments.
+      This SDK allows you to configure different environments for API requests.
+
+  - id: BASE_URL
```
amckinney, line 10: Thoughts on consolidating the base URL setting in the environments section? I'm OK with this one here if you like the separation though - we can explore this distinction in other languages too.

generators/java-v2/sdk/features.yml
```diff
@@ -5,46 +5,43 @@ features:
...
+      `nextPage` to perform the pagination manually
 
   - id: ERRORS
     description: |
-      Structured error types are returned from API calls that return non-success status codes. These errors are compatible
-      with the `errors.Is` and `errors.As` APIs, so you can access the error like so:
+      When the API returns a non-success status code (4xx or 5xx response), an API exception will be thrown.
 
-  - id: REQUEST_OPTIONS
+  - id: CUSTOM_CLIENT
```
amckinney, line 23: Thoughts on including this in the Advanced section? We can place it alongside retries, etc.

generators/java-v2/sdk/features.yml
```diff
@@ -5,46 +5,43 @@ features:
...
-      > is used).
+      This SDK is built to work with any instance of `OkHttpClient`. By default, if no client is provided, the SDK will construct one. 
+      However, you can pass your own client like so:
 
   - id: RETRIES
     advanced: true
     description: |
       The SDK is instrumented with automatic retries with exponential backoff. A request will be retried as long
-      as the request is deemed retryable and the number of retry attempts has not grown larger than the configured
+      as the request is deemed retriable and the number of retry attempts has not grown larger than the configured
```
amckinney, line 32: Nice catch here 👍

generators/java-v2/sdk/src/SdkGeneratorCli.ts
```diff
@@ -60,13 +63,63 @@ export class SdkGeneratorCLI extends AbstractJavaGeneratorCli<SdkCustomConfigSch
...
+            const path = FernGeneratorExec.EndpointPath(endpoint.location.path);
+
+            for (const endpointExample of endpoint.examples ?? []) {
+                const syncClient =
+                    (await dynamicSnippetsGenerator.generate(convertDynamicEndpointSnippetRequest(endpointExample)))
+                        .snippet + "\n";
+
+                const asyncClient =
+                    (await dynamicSnippetsGenerator.generate(convertDynamicEndpointSnippetRequest(endpointExample)))
+                        .snippet + "\n";
```
amckinney, line 91: Given that they're equivalent, should we just create the snippet once and use it for both? Or is this setting things up for a future refactor?

generators/java-v2/sdk/src/SdkGeneratorContext.ts
```diff
@@ -26,4 +38,215 @@ export class SdkGeneratorContext extends AbstractJavaGeneratorContext<SdkCustomC
             readmeConfigBuilder: new ReadmeConfigBuilder()
         });
     }
+
+    public getReturnTypeForEndpoint(httpEndpoint: HttpEndpoint): java.Type {
+        const responseBody = httpEndpoint.response?.body;
+
+        if (responseBody == null) {
+            return java.Type.void();
```
amckinney, line 46: Ah I see now how you're using `Void` - if we need the capitlaized equivalent, thoughts on adding an optional parameter here? Something like the following:

```
java.Type.void() // Produces `void` for method signatures

java.Type.void({ captialized: true }) // Produces `Void` for variable type references.
```

generators/java-v2/sdk/src/SdkGeneratorContext.ts
```diff
@@ -26,4 +38,215 @@ export class SdkGeneratorContext extends AbstractJavaGeneratorContext<SdkCustomC
...
+                            this.javaTypeMapper.convert({ reference: responseBody.value.payload })
+                        );
+                    default:
+                        assertNever(responseBody.value);
+                }
+                break;
+            case "fileDownload":
+                return java.Type.inputStream();
+            case "streamParameter":
+                throw new Error("Returning stream parameter is not supported");
```
amckinney, line 75: is the reason for this error to match behavior with the Java SDK as-is?

generators/java-v2/sdk/src/SdkGeneratorContext.ts
```diff
@@ -26,4 +38,215 @@ export class SdkGeneratorContext extends AbstractJavaGeneratorContext<SdkCustomC
...
+    private getPackagePrefixTokens(): string[] {
+        if (this.customConfig?.["package-prefix"] != null) {
+            return this.customConfig["package-prefix"].split(".");
+        }
+        const prefix: string[] = [];
+        prefix.push("com");
+        prefix.push(...this.splitOnNonAlphaNumericChar(this.config.organization));
+        prefix.push(...this.splitOnNonAlphaNumericChar(this.getApiName()));
+        return prefix;
+    }
```
amckinney, line 205: Nice, great to add all this - we can follow-up to try and consolidate all these utilities between the dynamic snippets context and the SDK context. That's been a long running issue on my mind (but def out of scope from this change).

generators/java-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
```diff
@@ -0,0 +1,483 @@
...
+    endpoint: HttpEndpoint;
+    fernFilepath: FernFilepath;
+}
+
+export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
+    private static CLIENT_VARIABLE_NAME = "client";
+    private static ENVIRONMENTS_FEATURE_ID: FernGeneratorCli.FeatureId = "ENVIRONMENTS";
+    private static BASE_URL_FEATURE_ID: FernGeneratorCli.FeatureId = "BASE_URL";
+    private static SNIPPET_PACKAGE_NAME = "com.example.usage";
+    private static ELLIPSES = java.codeblock("...");
```
amckinney, line 21: 👏

generators/java-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
```diff
@@ -0,0 +1,483 @@
...
+        const templatedSnippetsConfig: Record<
+            FernGeneratorCli.FeatureId,
+            {
+                renderer: (endpoint: EndpointWithFilepath) => string;
+                predicate?: (endpoint: EndpointWithFilepath) => boolean;
+            }
+        > = {
+            [ReadmeSnippetBuilder.ENVIRONMENTS_FEATURE_ID]: {
+                renderer: this.renderEnvironmentsSnippet.bind(this),
+                predicate: (_endpoint: EndpointWithFilepath) => this.getDefaultEnvironmentId() != null
```
amckinney, line 69: We should still render the environments section even when there isn't a default environment - we can still render how the user can select one vs. the other. We can just demonstrate the first environment in the list.

seed/java-sdk/pagination/README.md
```diff
@@ -4,6 +4,147 @@
...
+import com.seed.pagination.resources.complex.types.PaginatedConversationResponse;
+
+SeedPaginationClient client = SeedPaginationClient.builder().build();
+
+SyncPagingIterable<PaginatedConversationResponse> response = client.complex().search(...);
+
+for (item : response) {
+    // Do something with item
+}
+```
```
amckinney, line 81: Nice 👏 🚀

seed/java-sdk/pagination/README.md
```diff
@@ -4,6 +4,147 @@
...
+### Custom Client
+
+This SDK is built to work with any instance of `OkHttpClient`. By default, if no client is provided, the SDK will construct one. 
+However, you can pass your own client like so:
+
+```java
+import com.seed.pagination.SeedPaginationClient;
+import okhttp3.OkHttpClient;
+
+OkHttpClient customClient = ...;
```
amckinney, line 108: Is there some common constructor we can reference here? Something like

```java
OkHttpClient customClient = new OkHttpClient(...);
```

seed/java-sdk/multi-url-environment/README.md
```diff
@@ -4,6 +4,120 @@
...
+
+This SDK allows you to configure different environments for API requests.
+
+```java
+import com.seed.multiUrlEnvironment.SeedMultiUrlEnvironmentClient;
+import com.seed.multiUrlEnvironment.core.Environment;
+
+// Using environment
+SeedMultiUrlEnvironmentClient client = SeedMultiUrlEnvironmentClient.builder().environment(Environment.Production).build();
+```
```
amckinney, line 44: Nice 👍 - I wonder if we should remove the comment though since it's already clear? We could also explore writing these builder calls across multiple lines for readability.

amckinney (commented):
Nice, this is looking great! Some feedback on the formatting / style of the output, but mainly minor details.

generators/java-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    private renderEnvironmentsSnippet(_endpoint: EndpointWithFilepath): string {
+        const clientClassReference = this.context.getRootClientClassReference();
+
+        const builderMethodInvocation = java.invokeMethod({
+            on: clientClassReference,
+            method: "builder",
+            arguments_: []
+        });
```
amckinney, line 134: Can you reuse the [Builder](https://github.com/fern-api/fern/blob/909f55f7846041e17440db296f535e562497ae59/generators/java-v2/ast/src/ast/TypeLiteral.ts#L43) TypeLiteral here? We use it in dynamic snippets [here](https://github.com/fern-api/fern/blob/909f55f7846041e17440db296f535e562497ae59/generators/java-v2/dynamic-snippets/src/EndpointSnippetGenerator.ts#L609):

```typescript
        return java.TypeLiteral.builder({
            classReference: java.classReference({
                name: this.context.getClassName(request.declaration.name),
                packageName: this.context.getRequestsPackageName(request.declaration.fernFilepath)
            }),
            parameters: [...pathParameterFields, ...headerFields, ...queryParameterFields, ...requestBodyFields]
        });
```

seed/java-sdk/circular-references/README.md
```diff
@@ -1,15 +0,0 @@
...
-
-## Contributing
-
-While we value open-source contributions to this SDK, this library is generated programmatically.
-Additions made directly to this library would have to be moved over to our generation code,
-otherwise they would be overwritten upon the next generated release. Feel free to open a PR as
-a proof of concept, but know that we will not be able to merge it as-is. We suggest opening
-an issue first to discuss with us!
-
-On the other hand, contributions to the README are always very welcome!
```
amckinney, line 15: Do we know why these `README.md` files are getting removed?

generators/java-v2/ast/src/ast/Type.ts
ajgateno: This matches the other primitives we have. I left a to-do to add unboxed versions.

seed/java-sdk/circular-references/README.md
amckinney: Ah ignore me, this is similar to [this comment](https://github.com/fern-api/fern/pull/6379#discussion_r1991500115). This isn't likely to happen in practice, but it'd be great to still render the non-`Usage` section whenever we don't have examples (if possible)

generators/java-v2/ast/src/ast/Type.ts
ajgateno: done

generators/java-v2/base/src/context/AbstractJavaGeneratorContext.ts
ajgateno: It's not the same in model. We can have them all be the same once we remove model

generators/java-v2/sdk/features.yml
ajgateno: Spoke offline, we're keeping them separate for now

generators/java-v2/sdk/src/SdkGeneratorContext.ts
ajgateno: Yes

seed/java-sdk/pagination/README.md
ajgateno: It'd be a builder, which would look kind of awkward with the ellipses

generators/java-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
ajgateno: It does that, but what if there are no environments configured? Is that possible? The field is optional.

seed/java-sdk/multi-url-environment/README.md
ajgateno: done

seed/java-sdk/circular-references/README.md
ajgateno: They don't have endpoints. I verified this behavior and it matches the Go one.

ajgateno: Let's do that in another PR for both Go and Java

ajgateno: Sure, I can do the non-usage oens

ajgateno: I'll try to do the non-usage ones

generators/java-v2/sdk/features.yml
ajgateno: Yeah this is supposed to be in there already, I must've missed this

ajgateno: done

generators/java-v2/sdk/src/SdkGeneratorCli.ts
ajgateno: They're supposed to be different in the future, but I have no problem with initializing it once and leaving a comment

generators/java-v2/sdk/features.yml
```diff
@@ -5,46 +5,44 @@ features:
...
+    description: |
+      You can set a custom base URL when constructing the client.
 
   - id: PAGINATION
     description: |
-      List endpoints are paginated. The SDK provides an iterator so that you can simply loop over the items. You can also iterate page-by-page.
+      Paginated requests will return an Iterable<T>, which can be used to loop through the underlying items, or stream them. You can also call 
+      `nextPage` to perform the pagination manually
 
   - id: ERRORS
```
amckinney, line 19: Discussed offline - how about `EXCEPTION_HANDLING`?

amckinney (approved):
Nice!! Discussed offline, but we can ship this as-is and continue to iterate on more improvements as we go (e.g. custom titles, better usage examples for specific sections, 
additional sections for things like streaming, etc).

PR #6504, opened 2025-03-20
fix(csharp): Improve generator performance

Swimburger:
## Description
Significantly improve the performance of the generator itself by asynchronously formatting snippets in parallel.
For example, Square's SDK generates in less than 20s as opposed to more than 3m previously.

~~Note: The bottleneck is now starting/stopping the processes. In theory, pooling them would help, but csharpier doesn't let you keep the process alive.~~ We're now doing all the formatting up front in one csharpier call and caching it.

## Testing

### DIFF
--- a/fern/pages/changelogs/csharp-sdk/2025-03-20.mdx
+++ b/fern/pages/changelogs/csharp-sdk/2025-03-20.mdx
@@ -1,5 +1,0 @@
-## 1.15.7
-**`(fix):`** Significantly improve performance of the generator itself by asynchronously formatting snippets in parallel.
-For example, Square's SDK generates in less than 20 seconds as opposed to more than 3 minutes previously.
-
-


dsinghvi (approved):
Smart idea on formatting snippets in parallel and not awaiting

Swimburger:
> Smart idea on formatting snippets in parallel and not awaiting

I think this would be the fastest if csharpier would let us reuse the process as the process creation/termination is now the bottleneck.
I rewrote the code to generate all the snippets without formatting, joined them all together, then formatted them all at once with one sharper call, and cached them upfront.

This reduced the generation for Square from 1m to 30s.

generators/csharp/base/src/findDotNetToolPath.ts
```diff
@@ -0,0 +1,32 @@
...
+    // Priority 3: Check DOTNET_CLI_HOME if set
+    if (process.env.DOTNET_CLI_HOME) {
+        const cliHomePath = path.join(process.env.DOTNET_CLI_HOME, ".dotnet", "tools", toolName);
+        return cliHomePath;
+    }
+
+    // Priority 4: Check standard location based on OS
+    const homeDir = os.homedir();
+    const standardPath = path.join(homeDir, ".dotnet", "tools", toolName);
+    return standardPath;
```
amckinney, line 31: Should we throw if we don't find it? Or leave that for when we actually go to call it?

generators/csharp/sdk/src/cli.ts
```diff
@@ -5,4 +5,5 @@ void runCli();
 export async function runCli(): Promise<void> {
     const cli = new SdkGeneratorCLI();
     await cli.run();
+    process.exit(0);
```
amckinney, line 8: Was this not the default behavior?

generators/csharp/sdk/src/endpoint/snippets/EndpointSnippetsGenerator.ts
```diff
@@ -100,12 +169,13 @@ export class EndpointSnippetsGenerator {
         if (snippet == null) {
             return undefined;
         }
-        const formattedAstNodeSnippet = snippet.toFormattedSnippet({
+        const formatter = format ? this.context.formatter : this.context.nopFormatter;
```
amckinney, line 172: Nice 👍

amckinney (approved):
A few nits on package structure, but otherwise looks great to me!

PR #6505, opened 2025-03-21
fix(csharp): Fix missing FileParameter bug in proto projects

Swimburger:
## Description
Fix missing FileParameter bug in proto projects

## Testing


amckinney (approved):

amckinney:
@Swimburger alternatively, we could only generate the `FileParameter` when there is at least one `multipart/form-data` endpoint.

Swimburger:
> @Swimburger alternatively, we could only generate the `FileParameter` when there is at least one `multipart/form-data` endpoint.

That makes sense, we can also remove the multipart tests when there are no multipart endpoints

Swimburger:
> > @Swimburger alternatively, we could only generate the `FileParameter` when there is at least one `multipart/form-data` endpoint.
> 
> That makes sense, we can also remove the multipart tests when there are no multipart endpoints

It's also referenced in the RawClient. We could split up RawClient into multiple parts and merge them with partials, but I think it's better to just leave them as is so SDK authors can use it if they want to.

PR #6514, opened 2025-03-22
chore(cli): opt in to blc for fern generate

chdeskur:
This PR updates the behavior of `fern generate`. Previously, the broken link checker would run on any `fern generate --docs` command. This PR introduces a new behavior, where the user needs to opt-in for broken link checking on docs generation.

### DIFF
--- a/packages/cli/cli/src/cli.ts
+++ b/packages/cli/cli/src/cli.ts
@@ -456,12 +456,6 @@ function addGenerateCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext)
                 return cliContext.failWithoutThrowing("The --local flag is incompatible with --preview.");
             }
             if (argv.api != null) {
-                if (argv.brokenLinks != null) {
-                    cliContext.logger.warn("--broken-links is ignored when generating docs");
-                }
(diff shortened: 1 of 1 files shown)


packages/cli/cli/src/cli.ts
```diff
@@ -445,6 +456,12 @@ function addGenerateCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext)
                 return cliContext.failWithoutThrowing("The --local flag is incompatible with --preview.");
             }
             if (argv.api != null) {
+                if (argv.brokenLinks != null) {
+                    cliContext.logger.warn("--broken-links is ignored when generating docs");
+                }
+                if (argv.strictBrokenLinks != null) {
+                    cliContext.logger.warn("--strict-broken-links is ignored when generating docs");
+                }
```
dsinghvi, line 464: Let's remove these log lines -- i think itll be annoying for folks to always see warnings on generate

chdeskur: this is my bad - the intent here was to warn users if they use the broken link flags with SDK generation, but it's pretty irrelevant. removed

dsinghvi (approved):

PR #6516, opened 2025-03-23
feat(docs): add products into schema

sbawabe:
DOCS-40

## Description
- goal of PR is to add `products` into the schema as a viable navigation option for users to include in their `docs.yml` files
- allows each product to be its own `yml` file

## Changes Made
- begins wiring up the addition of `products` into the schema

## Testing

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,5 +1,12 @@
 - changelogEntry: 
   - summary: |
+      Add support for products and versioned products in the docs config file.
+    type: feat
+  irVersion: 57
+  version: 0.60.30
+
+- changelogEntry: 
+  - summary: |
       Add support for `propertyAccess` in the IR -> FDR SDK conversion.
     type: feat
   irVersion: 57


sbawabe:
* **#6516** <a href="https://app.graphite.dev/github/pr/fern-api/fern/6516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/fern-api/fern/6516?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dsinghvi:
You'll want to add a release here so that the CLI gets published: https://github.com/fern-api/fern/blob/main/packages/cli/cli/versions.yml#L1-L6

fern/products/docs/latest.yml
```diff
@@ -0,0 +1,281 @@
+# THIS IS NOT READY -- We should use the product switcher for Fern, but that will be considered outside of the scope of this PR.
```
dsinghvi, line 1: nit: lets delete for now (and do a migration after)

packages/cli/yaml/docs-validator/src/docsAst/visitDocsConfigFileYamlAst.ts
```diff
@@ -178,6 +178,7 @@ export async function visitDocsConfigFileYamlAst({
                 context
             });
         },
+        products: noop,
```
dsinghvi, line 181: https://github.com/fern-api/fern/blob/fb89f62fc49f85319a94e3fd33ca04f96f4ccf3b/packages/cli/yaml/docs-validator/src/docsAst/visitDocsConfigFileYamlAst.ts#L224-L263

we should validate the keys inside the products file (just like we do versions!) for example, if a file does not exist, then we would want to give a user a heads up

dsinghvi (approved):
Added 2 required comments, but going to +1 to help unblock and get an alpha version of this functionality out

PR #6517, opened 2025-03-23
fix(cli): OpenAPI-IR recognizes list of channel messages instead of `publish` and `subscribe`

eyw520:
## Description
This PR updates the OpenAPI-IR to match the Fern Definition in enabling an arbitrary `messages` key, which maps to a list of `WebsocketMessage` objects, instead of a `publish` and `subscribe` key, which forces constructing an undiscriminated union during AsyncAPI parsing.

Note: This PR will need to be followed up with changes to the Typescript generator to support an arbitrary length `messages` in the IR.

## Changes Made
- Update `finalIr` in OpenAPI-IR.
- Remove undiscriminated union construction in v3 parser.
- Resolve a bug where updated message ids were not being used in reference resolution.

## Testing

### DIFF
--- a/packages/cli/api-importers/asyncapi-to-ir/src/AsyncAPIConverterContext.ts
+++ b/packages/cli/api-importers/asyncapi-to-ir/src/AsyncAPIConverterContext.ts
@@ -26,6 +26,10 @@ export class AsyncAPIConverterContext extends AbstractConverterContext<AsyncAPIV
         return parameter != null && "$ref" in parameter;
     }
 
+    public isMessageWithPayload(msg: unknown): msg is AsyncAPIV3.ChannelMessage {
+        return msg != null && typeof msg === "object" && "payload" in msg;
+    }
+
     public getTypeIdFromSchemaReference(reference: OpenAPIV3_1.ReferenceObject): string | undefined {
         const schemaMatch = reference.$ref.match(/\/schemas\/(.+)$/);
         if (!schemaMatch || !schemaMatch[1]) {
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v3/AsyncAPIV3ParserContext.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v3/AsyncAPIV3ParserContext.ts
@@ -53,4 +53,12 @@ export class AsyncAPIV3ParserContext extends AbstractAsyncAPIParserContext<Async
             ...resolvedInComponents
         };
     }
+
+    public isMessageWithPayload(msg: unknown): msg is AsyncAPIV3.ChannelMessage {
+        return msg != null && typeof msg === "object" && "payload" in msg;
+    }
+
+    public isReferenceObject(msg: unknown): msg is OpenAPIV3.ReferenceObject {
+        return msg != null && typeof msg === "object" && "$ref" in msg;
+    }
 }
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v3/parseAsyncAPIV3.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v3/parseAsyncAPIV3.ts
@@ -76,12 +76,12 @@ export function parseAsyncAPIV3({
         }
         if (channel.messages) {
             for (const [messageId, message] of Object.entries(channel.messages)) {
-                if (message && ("payload" in message || "$ref" in message)) {
+                if (context.isReferenceObject(message) || context.isMessageWithPayload(message)) {
                     if (!seenMessages[messageId]) {
                         seenMessages[messageId] = [];
                     }
-                    if ("$ref" in message) {
-                        const resolved = context.resolveMessageReference(message as OpenAPIV3.ReferenceObject);
+                    if (context.isReferenceObject(message)) {
+                        const resolved = context.resolveMessageReference(message);
                         seenMessages[messageId].push({
                             channelId,
                             payload: resolved.payload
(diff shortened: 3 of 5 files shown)


packages/cli/api-importers/asyncapi-to-ir/src/3.0/channel/ChannelConverter3_0.ts
```diff
@@ -21,6 +21,7 @@ export declare namespace ChannelConverter3_0 {
         channelPath: string;
         group: string[] | undefined;
         operations: Record<string, AsyncAPIV3.Operation>;
+        deduplicationMap: Record<string, Record<string, string>> | undefined;
```
dsinghvi, line 24: what is a `deduplicationMap`? Ideally whenever a developer looks at the args for a certain class they should be able to grok what something is. 

` Record<string, Record<string, string>>` is also tough because none would understand what each value is. You could either create a class called `ChannelDeduper` that has specific methods and stores the record as a private variable or you could use aliases like `Record<ChannelId, Record<MessageId, ...>>` where `MessageId = string`.

packages/cli/api-importers/asyncapi-to-ir/src/AsyncAPIConverterContext.ts
```diff
@@ -147,4 +160,15 @@ export class AsyncAPIConverterContext extends AbstractConverterContext<AsyncAPIV
 
         return undefined;
     }
+
+    private rebuildReference(
```
dsinghvi, line 164: nit: this method could be a bit more clearly named with `replaceReferencedTypeId`

packages/cli/api-importers/asyncapi-to-ir/src/3.0/channel/ChannelConverter3_0.ts
eyw520: fixed via aliases

packages/cli/api-importers/asyncapi-to-ir/src/AsyncAPIConverterContext.ts
eyw520: agreed. fixed.

packages/cli/api-importers/asyncapi-to-ir/src/AsyncAPIConverterContext.ts
```diff
@@ -42,19 +42,32 @@ export class AsyncAPIConverterContext extends AbstractConverterContext<AsyncAPIV
...
+    }: {
+        reference: OpenAPIV3_1.ReferenceObject;
+        channelPath?: string;
+        deduplicationMap?: Record<string, Record<string, string>>;
+    }): Promise<{ ok: true; reference: TypeReference } | { ok: false }> {
+        let updatedReference: OpenAPIV3_1.ReferenceObject = reference;
         let typeId: string | undefined;
         if (reference.$ref.includes("schemas")) {
             typeId = this.getTypeIdFromSchemaReference(reference);
         } else if (reference.$ref.includes("messages")) {
```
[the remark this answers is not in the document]
eyw520: fixed -- will use regexes to match on the ref structure

packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v3/AsyncAPIV3ParserContext.ts
```diff
@@ -27,10 +27,14 @@ export class AsyncAPIV3ParserContext extends AbstractAsyncAPIParserContext<Async
             if (resolvedInChannel == null) {
                 throw new Error(`${message.$ref} is undefined`);
             }
-            return {
-                name: messageKey,
-                ...resolvedInChannel
-            };
+            if ("$ref" in resolvedInChannel) {
```
dsinghvi, line 30: I'd love to get rid of any inlined `in` checks and just use type guarads wherever possible. Happy to do this moving forward if there are too many instances but ideally this would read something like `if(isReference(resolvedChannel))` ... (just more readable)

packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v3/parseAsyncAPIV3.ts
```diff
@@ -70,14 +76,22 @@ export function parseAsyncAPIV3({
         }
         if (channel.messages) {
             for (const [messageId, message] of Object.entries(channel.messages)) {
-                if (message && message.payload != null) {
+                if (message && ("payload" in message || "$ref" in message)) {
```
dsinghvi, line 79: nit: can we use a type guard here

packages/cli/ete-tests/src/tests/write-definition/fixtures/multiple-no-endpoint-async/fern/.definition/testChannel.yml
```diff
@@ -1,15 +1,12 @@
...
   auth: false
   examples:
-    - messages: []
+    - messages:
+        - body: string
+          type: publish
   messages:
-    publish:
-      body: TestChannelPublishEvent
+    testChannel_sendMessage:
```
dsinghvi, line 8: @eyw520 `testChannel_sendMessage` feels like an odd message id, do we need to prefix these?

eyw520: this message id is a result of the deduplication logic (since the fixture intentionally generates several naming conflicts).

eyw520: discussed offline

packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v3/parseAsyncAPIV3.ts
eyw520: done

packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v3/AsyncAPIV3ParserContext.ts
eyw520: done

dsinghvi (approved):

PR #6521, opened 2025-03-24
chore(cli): dont block example generation on unexpected example

dsinghvi:
## Description
Rather than throwing errors if an unexpected example is given, we now just default. Ideally `fern check` should catch errors if this is the case. 

## Changes Made
- See description

## Testing


eyw520 (approved):

amckinney (approved):
Discussed offline, but it'd be good to capture warnings here so that we don't swallow the issues entirely.

PR #6534, opened 2025-03-24
fix(java): Change main method name in Java snippets

ajgateno:
Generate
```java
public static void main(String[] args) {}
```
Instead of
```java
public static void run() {}
```
For Java snippets


amckinney (approved):
👏

PR #6557, opened 2025-03-26
chore(docs): add changelog for AI chat config block

dubwub:
## Description
Changelog update

## Changes Made
Only changelog

## Testing
N/A

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -11,7 +11,7 @@
         
     type: feat
   irVersion: 57
-  version: docs@0.12.5
+  version: 0.57.1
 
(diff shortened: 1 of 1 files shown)


packages/cli/cli/versions.yml
```diff
@@ -1,3 +1,10 @@
+- changelogEntry:
+  - summary: |
+      Add AI chat config block to docs.yml for model/system prompt customization
```
dsinghvi, line 3: can you add an example of what this might look like in `docs.yml` 

```
ai-chat: 
 ...
```

dsinghvi (approved):

PR #6561, opened 2025-03-26
fix(cli): add replace call for rid when used as key of a map when parsing conjure specs

drake-nominal:
## Description

Simple fix for https://github.com/fern-api/fern/issues/6559.

Essentially, if you mention `map<rid, ...>` (rid as key type for a map), fern fails to parse the conjure spec.
This is solved for all other permutations for `rid` by using regex to replace `rid` with `string`, however, fails to account for the edge case of using a map. This PR adds in an additional case to fix this.


dsinghvi (approved):

dsinghvi:
@drake-nominal would you be able to add an entry to https://github.com/fern-api/fern/blob/main/packages/cli/cli/versions.yml so we can release?

PR #6617, opened 2025-04-01
feat(python): Generate clients that support response headers for non-streaming methods

hughhan1:
## Description

See https://github.com/fern-api/fern/pull/6466 for the original design.

### Problem
It's currently impossible for users to retrieve response headers from the Python SDK. We need to extend the generated SDK so that users can access these values as needed. This change augments the Python SDK generator such that the clients it generates are capable of exposing raw HTTP response headers for non-streaming requests.

### New Features
Each generated client now exposes a `with_raw_response` property that returns a "raw" version of the client, which contains all of the same methods as the original client, but returns the expected type `T` wrapped in an `HttpResponse[T]` (or `AsyncHttpResponse[T]`) type. 

Streaming methods are not yet exposed on these "raw" clients, and will be supported in a future change.

Additionally, the unwrapping of fields in optional `Response` types is fixed.

##### Standard
```python
client = AcmeClient()

# By default, we keep all the methods as-is and incur no breaking changes.
# This means we can access the standard response properties directly.
iterator = client.users.list(...)
for user in iterator:
    print(user)

# Alternatively, you can act upon the "with_raw_response" client that includes HTTP
# response headers.
#
# The caller needs to access the values with the `data` property.
response = client.users.with_raw_response.list(...)
for user in response.data:
    print(user)

# The caller can also access the response headers alongside it.
print(f"Got organization id: {response.headers.get('organization_id')}")
```
##### Pagination
```python
client = AcmeClient()
pager = client.users.list_with_cursor_pagination(...)
for item in pager:
    ...

response = client.users.with_raw_response.list_with_cursor_pagination(...)
for item in response.data:
    ...

# alternatively, you can paginate page-by-page
for page in pager.iter_pages():
    ... 

for page in response.data.iter_pages():
    ...
```
##### File Downloads
```python
client = FileDownloadClient()

chunks = client.files.download_file(...)
for chunk in chunks:
    ...

response = client.files.with_raw_response.download_file(...)
for chunk in response.data:
    ...
```

### Additional Details
At the implementation level, the bulk of business logic contained in generated clients (`client.py`) is now moved into a corresponding "raw" client (`raw_client.py`). Each endpoint function of client calls into the underlying "raw" client.

## Changes Made

- Rename some methods for consistency (e.g. `socket_class` to `socket_client_class`).
- Implement shared `HttpResponse` and `AsyncHttpResponse` classes.
- Move some logic from `ClientGenerator` into a `BaseClientGenerator` abstract class.
- Implement a `RawClientGenerator` which contains most of previous logic of `ClientGenerator`, responsible for generating raw clients.
- Modify `ClientGenerator` to generate the proper code to call into the generated raw clients.
- Fix the unwrapping of fields in optional `Response` types (search the codebase for `# TODO: What is the case where you have a nested property as response, but no response property configured?`).

## Testing

- [x] Regenerate all fixtures for the Python SDK generator


seed/python-sdk/accept-header/README.md
```diff
@@ -26,7 +26,7 @@ client = SeedAccept(
     token="YOUR_TOKEN",
     base_url="https://yourhost.com/path/to/api",
 )
-client.service.endpoint()
+client.service.with_raw_response.endpoint()
```
amckinney, line 29: I'm not sure that we want to update all of the snippets to use the `with_raw_response` alternative. I'd expect none of the `snippet.json`, `reference.md`, or `README.md` examples (like this one) to be updated.

seed/python-sdk/audiences/src/seed/foo/client.py
```diff
@@ -56,36 +66,29 @@ def find(
...
-                    parse_obj_as(
-                        type_=ImportingType,  # type: ignore
-                        object_=_response.json(),
-                    ),
-                )
-            _response_json = _response.json()
-        except JSONDecodeError:
-            raise ApiError(status_code=_response.status_code, body=_response.text)
-        raise ApiError(status_code=_response.status_code, body=_response_json)
+        return response.data
```
amckinney, line 75: Nice, this looks great 👏

amckinney (commented):
Nice work! It doesn't need to happen in this PR, but we should add this to the generated `README.md` when we're done implementing the streaming portion. That would involve adding a block [here](https://github.com/fern-api/fern/blob/main/generators/python/sdk/features.yml), and wiring it up in the [ReadmeSnippetBuilder](https://github.com/fern-api/fern/blob/5f17140dca7f592c169649b36f59132c849818b9/generators/python/src/fern_python/generator_cli/readme_snippet_builder.py#L19).

We'll eventually move this functionality to the TypeScript-equivalent, but it should be relatively straightforward to document this like we document other features (e.g. the [timeout section](https://github.com/fern-api/fern/blob/5f17140dca7f592c169649b36f59132c849818b9/generators/python/src/fern_python/generator_cli/readme_snippet_builder.py#L120)).

seed/python-sdk/accept-header/README.md
hughhan1: Removed all of these changes.

amckinney (approved):

amckinney:
Discussed offline w/ @hughhan1 -

It looks like there are some `mypy` errors in the `examples` test definition, but the generated code looks fine -- this might be related to a `mypy` version that we can investigate in a follow-up (i.e. when we implement the streaming portion of this feature).

PR #6619, opened 2025-04-01
feat(java): Add fallback snippet renderer for readme generation

ajgateno:
Currently we don't generate the README.md if there's no endpoint examples. This PR makes it so that we can provide a fallback example in case the pre-rendered one can't be used.

### DIFF
--- a/fern/pages/changelogs/java-sdk/2025-04-01.mdx
+++ b/fern/pages/changelogs/java-sdk/2025-04-01.mdx
@@ -1,0 +1,4 @@
+## 2.32.2
+**`(fix):`** Add fallback snippet renderer for readme generation
+
+
--- a/fern/pages/changelogs/pydantic/2025-03-31.mdx
+++ b/fern/pages/changelogs/pydantic/2025-03-31.mdx
@@ -1,4 +1,0 @@
-## 1.4.8
-**`(internal):`** Update the IR to v57.
-
-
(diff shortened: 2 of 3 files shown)


generators/java-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
```diff
@@ -127,6 +128,26 @@ export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
...
+        const clientClassReference = this.context.getRootClientClassReference();
+        const endpointMethodInvocation = this.getMethodCall(endpoint, [ReadmeSnippetBuilder.ELLIPSES]);
+
+        const clientBuilder = java.TypeLiteral.builder({ classReference: clientClassReference, parameters: [] });
+
+        const snippet = java.codeblock((writer) => {
+            writer.writeNode(clientClassReference);
+            writer.write(" client = ");
+            writer.writeNodeStatement(clientBuilder);
+            writer.writeLine("\n");
```
amckinney, line 141: Hmm.. maybe confusing for future readers to write _two_ newlines this way - thoughts on rolling out separate calls like so?

```typescript 
writer.newLine();
writer.newLine();
```

amckinney (approved):

generators/java-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
ajgateno: done

PR #6626, opened 2025-04-01
chore(cli): Autogenerate example websocket sessions in the v3 parser

eyw520:
## Description
This PR introduces example websocket session generation in the v3 parser.

## Changes Made
- Implement `generateWebSocketExample`.
- `injectAutogeneratedExamples` now generates `ExampleWebSocketSession` examples.

## Testing

### DIFF
--- a/packages/commons/ir-utils/src/examples/generator/generateWebSocketExample.ts
+++ b/packages/commons/ir-utils/src/examples/generator/generateWebSocketExample.ts
@@ -45,7 +45,9 @@ export function generateWebSocketExample({
         skipOptionalRequestProperties,
         maxDepth: 1
     });
-    if (channelPathResult.type === "failure") {return channelPathResult;}
+    if (channelPathResult.type === "failure") {
+        return channelPathResult;
+    }
     result.pathParameters.push(...channelPathResult.example);
 
     const rootPathResult = generatePathParameterExamples(ir.pathParameters, {
@@ -53,7 +55,9 @@ export function generateWebSocketExample({
         skipOptionalRequestProperties,
         maxDepth: 1
     });
-    if (rootPathResult.type === "failure") {return rootPathResult;}
+    if (rootPathResult.type === "failure") {
+        return rootPathResult;
+    }
     result.pathParameters.push(...rootPathResult.example);
 
     // QUERY PARAMETERS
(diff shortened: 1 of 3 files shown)


packages/commons/ir-utils/src/examples/generator/generateWebSocketExample.ts
```diff
@@ -0,0 +1,225 @@
...
+        ir: Omit<IntermediateRepresentation, "sdkConfig" | "subpackages" | "rootPackage">;
+        channel: WebSocketChannel;
+        typeDeclarations: Record<TypeId, TypeDeclaration>;
+        skipOptionalRequestProperties: boolean;
+    }
+
+    interface ParameterGroup<T, K> {
+        params: T[];
+        add: (example: K) => void;
+    }
```
amckinney, line 31: nit: Thoughts on naming the generic types so that they're a little more clear at face value?
```suggestion
    interface ParameterGroup<ParamT, ExampleT> {
        params: ParamT[];
        add: (example: ExampleT) => void;
    }
```

packages/commons/ir-utils/src/examples/generator/generateWebSocketExample.ts
```diff
@@ -0,0 +1,225 @@
...
+            continue;
+        }
+        const generatedExample = generateTypeReferenceExample({
+            fieldName: queryParameter.name.name.originalName,
+            currentDepth: 0,
+            maxDepth: 10,
+            typeDeclarations,
+            typeReference: queryParameter.valueType,
+            skipOptionalProperties: skipOptionalRequestProperties
+        });
```
amckinney, line 99: nit: Not actionable here, but the repetition between each call makes me wonder if this would be better suited with a `class` that has its own `private` helper method that delegates to a simpler version of this function. The only things changes between each call are:
1. The fieldName
2. The typeReference
3. The currentDepth

The rest of the properties could be instance properties on a `WebsocketExampleGenerator` class (or something similar).

Like I said above, no need to change that here - just food for thought on another approach for future abstractions.

amckinney (approved):

packages/commons/ir-utils/src/examples/generator/generateWebSocketExample.ts
eyw520: Refactored!

PR #6629, opened 2025-04-01
feat(python): Add improved BaseModel

amckinney:
This proposes a new `BaseModel` class, which is meant to replace _both_ the `UniversalBaseModel` and the `UncheckedBaseModel`.

There's a fair amount of changes involved here, so we will eventually want to leverage a variety of serialization tests like the ones proposed in https://github.com/fern-api/fern/pull/6519. There are a couple things that need to happen before we can do so:
1. Fix all issues with example generation.
2. Implement dynamic snippets.

We'll leave this in a draft state for now, but we'll come back to this when we can actually finish the stroke.


amckinney:
Closing for now - we might restore this functionality later.

PR #6633, opened 2025-04-02
feat(python): Add PythonFormatter for browser compatibility

amckinney:
This updates the `@fern-api/python-ast` package so that it has zero browser incompatible dependencies (e.g. `base-generator`). This also adds the `@fern-api/python-formatter`, which abstracts away the `ruff` formatter.

This includes a single breaking change, where the `toStringFormatted` method actually requires an `AbstractFormatter` implementation. For example, the following uses the `PythonFormatter` implementation with a custom `ruff` configuration:

**Before**

```typescript
const node = AstNode(...)
const formatted = node.toStringFormatted({
  indent_width: 2
})
```

**After**

```typescript
const formatter = new PythonFormatter({
  config: {
    indent_width: 2
  }
})
const node = new AstNode(...)
const formatted = node.toStringFormatted(formatter)
```

Note that all test cases now use the simple `toString` method so that `@fern-api/ast` doesn't need to import `@fern-api/python-formatter` at all (even just for test cases).

### DIFF
--- a/generators/python-v2/ast/src/CodeBlock.ts
+++ b/generators/python-v2/ast/src/CodeBlock.ts
@@ -1,4 +1,4 @@
-import { CodeBlock as CommonCodeBlock } from "@fern-api/base-generator";
+import { CodeBlock as CommonCodeBlock } from "@fern-api/browser-compatible-base-generator";
 
 import { AstNode } from "./core/AstNode";
 import { Writer } from "./core/Writer";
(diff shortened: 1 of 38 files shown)


eyw520 (approved):

dvargas92495:
> Note that all test cases now use the simple toString method so that @fern-api/ast doesn't need to import @fern-api/python-formatter at all

Does this mean toString() formats under the hood?

We don't currently pass in parameters to toStringFormatted so I think we should be good - it still works with no arg passed in correct?

PR #6650, opened 2025-04-04
fix(typescript): Significantly improve performance of SDK generation when the `useLegacyExports` config is `false`.

Swimburger:
## Description
Significantly improve performance of SDK generation when the `useLegacyExports` config is `false`.

## Testing


amckinney (approved):
Nice! 👏

PR #6664, opened 2025-04-04
feat(python): Inline path params

ajgateno:
Adds a config flag that enables path parameters to become named in endpoint methods.

### DIFF
--- a/fern/pages/changelogs/csharp-sdk/2025-04-07.mdx
+++ b/fern/pages/changelogs/csharp-sdk/2025-04-07.mdx
@@ -1,0 +1,3 @@
+## 1.15.14
+**`(fix):`** Fix issue where read-only list properties wouldn't deserialize correctly.
+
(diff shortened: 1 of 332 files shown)


seed/python-sdk/exhaustive/inline-path-params/src/seed/endpoints/http_methods/client.py
```diff
@@ -0,0 +1,717 @@
...
+    def __init__(self, *, client_wrapper: SyncClientWrapper):
+        self._client_wrapper = client_wrapper
+
+    def test_get(self, *, id: str, request_options: typing.Optional[RequestOptions] = None) -> str:
+        """
+        Parameters
+        ----------
+        id : str
+
+        id : str
```
amckinney, line 29: Small detail here - we're probably just counting the parameter twice.

amckinney (approved):

PR #6665, opened 2025-04-04
feat(python): Implement dynamic snippets

amckinney:
This implements the bulk of the dynamic snippet generator for Python. For now, this is particularly tailored to SDKs that enable TypedDict requests. In a follow-up, this will be integrated into `seed`.


generators/python-v2/ast/src/BaseInvocation.ts
```diff
@@ -30,15 +34,32 @@ export class BaseInvocation extends AstNode {
...
+                } else {
+                    writer.write(" ");
+                }
             }
         });
-
+        if (this.multiline) {
+            writer.newLine();
+            writer.dedent();
+        }
```
hughhan1, line 62: Is this change purely to support code snippets, or do we plan on using this function for generating actual code as well? I'm wondering if we should delegate this type of logic to formatters, instead of owning logic that deals with formatting.

generators/python-v2/ast/src/ClassInstantiation.ts
```diff
@@ -1,15 +1,29 @@
...
         super({ reference: classReference, ...args });
+        this.assignTo = assignTo;
+    }
+
+    public write(writer: Writer): void {
+        if (this.assignTo) {
+            writer.write(this.assignTo);
+            writer.write(" = ");
+        }
+        super.write(writer);
```
hughhan1, line 27: Rather than having the `ClassInstantiation` own the assignee, should we think about designing this in a way such that we have an `Assign` node that has knowledge of the left and right operands?

amckinney: Absolutely, much better that way - I'll follow up with that change. I'll probably call it Assignment.

generators/python-v2/ast/src/BaseInvocation.ts
amckinney: For general SDK generation, that's exactly what we do. But given that dynamic snippets need to run in the browser, we need to include some basic level of formatting for that use case.

In the long run, it's possible that we can more generally depend on WASM formatters and not lean on this type of behavior.

generators/python-v2/ast/src/ClassInstantiation.ts
hughhan1: >  I'll probably call it Assignment.

Sure, I'm not too opinionated on this part. The standard library calls it [ast.Assign](https://docs.python.org/3/library/ast.html#ast.Assign).

generators/python-v2/ast/src/BaseInvocation.ts
hughhan1: > But given that dynamic snippets need to run in the browser, we need to include some basic level of formatting for that use case.

Makes sense.

> In the long run, it's possible that we can more generally depend on WASM formatters and not lean on this type of behavior.

Also makes sense.

I'm just worried about the scenario in which we end up putting a bunch of _new_ tech debt in the V2 generators.

generators/python-v2/ast/src/ClassInstantiation.ts
amckinney: Ah nice, let's go with `Assign` - I like maintaining consistency with the standard lib naming conventions.

ajgateno (approved):

PR #6678, opened 2025-04-07
fix(python): Fix NotRequired usage

ajgateno:
Currently, we're using `typing_extensions.NotRequired[T]` as a stand-in for `typing.Optional[T]` whenever `use_typeddict_requests` is enabled in custom config. This causes errors because `NotRequired` is strictly meant for expressing that a key may be absent from a TypedDict, and we're using it e.g. to express that union variants may be `None`, which we should be using `Optional` for.

This PR makes it so that we only use `NotRequired` whenever we've declared a field as `optional<T>` in an object property, which means it's going to end up as a maybe-absent field in a TypedDict. The following table illustrates the behavior after this change:

| class type (row)/property type (col) | `optional<T>`    | `nullable<T>`    | `optional<nullable<T>>`    | `nullable<optional<T>>` |
|--------------------------------------|------------------|------------------|----------------------------|-------------------------|
| object                               | `NotRequired[T]` | `Optional[T]`    | `NotRequired[Optional[T]]` | n/a                     |
| discriminated union                  | `Optional[T]`    | `Optional[T]`    | `NotRequired[Optional[T]]`              | n/a                     |
| undiscriminated union                | `Optional[T]`    | `Optional[T]`    | `Optional[T]`              | n/a                     |
| enum                                 | n/a              | n/a              | n/a                        | n/a                     |
| wrapped alias                        | `NotRequired[T]` | `Optional[T]` | `NotRequired[Optional[T]]` | n/a                     |


amckinney (approved):
lgtm -- thanks for including the thorough description in the summary 👏

PR #6697, opened 2025-04-08
fix(python): Gate `raw_client` method generation for paginated endpoints.

eyw520:


seed/python-sdk/pagination/src/seed/complex_/client.py
```diff
@@ -94,7 +157,7 @@ async def search(
 
         Returns
         -------
-        AsyncPager[Conversation]
+        AsyncPager[PaginatedConversationResponse]
```
amckinney, line 160: This looks off -- I think we need to return `AsyncPager[Conversation` here.

seed/python-sdk/pagination/README.md
```diff
@@ -1,198 +0,0 @@
-# Seed Python Library
```
amckinney, line 1: There might be some other issue here -- this file shouldn't be removed.

eyw520: This should be resolved.

seed/python-sdk/pagination/src/seed/complex_/client.py
eyw520: This regression was introduced at some point before this change, since even without the changes in this PR, the output code generation still looks off.

I'll look into a resolution to get the generator back to a stable place.

seed/python-sdk/pagination/snippet.json
```diff
@@ -1,19 +1,6 @@
...
-                "async_client": "import asyncio\n\nfrom seed import AsyncSeedPagination\nfrom seed.complex_ import SingleFilterSearchRequest, StartingAfterPaging\n\nclient = AsyncSeedPagination(\n    token=\"YOUR_TOKEN\",\n    base_url=\"https://yourhost.com/path/to/api\",\n)\n\n\nasync def main() -> None:\n    response = await client.complex_.search(\n        pagination=StartingAfterPaging(\n            per_page=1,\n            starting_after=\"starting_after\",\n        ),\n        query=SingleFilterSearchRequest(\n            field=\"field\",\n            operator=\"=\",\n            value=\"value\",\n        ),\n    )\n    async for item in response:\n        yield item\n    # alternatively, you can paginate page-by-page\n    async for page in response.iter_pages():\n        yield page\n\n\nasyncio.run(main())\n",
-                "type": "python"
-            }
-        },
```
amckinney, line 16: It looks like we might be missing the snippets here -- is it possible to restore these?

amckinney (approved):
Assuming that the snippets can be restored for these endpoints, this lgtm.

seed/python-sdk/pagination/snippet.json
eyw520: Restored.

PR #6701, opened 2025-04-09
fix(java): Fix getting the default version of an api versions enum to not require the CURRENT entry

ajgateno:
Currently, API version headers are represented using an enum in Java. If there's a default version, we add a `CURRENT` field, which is used to populate the default version in CustomConfig. This doesn't really make sense semantically, it's not done in other generators, and causes compilation failures when `enable-forward-compatible-enums` is set to true. This PR removes the need for this config field by populating the actual default API header set.


seed/java-sdk/version/no-custom-config/src/main/java/com/seed/version/core/ClientOptions.java
```diff
@@ -0,0 +1,220 @@
...
+     *     + "      \"snakeCase\" : {\n"
+     *     + "        \"unsafeName\" : \"v_2\",\n"
+     *     + "        \"safeName\" : \"v_2\"\n"
+     *     + "      },\n"
+     *     + "      \"screamingSnakeCase\" : {\n"
+     *     + "        \"unsafeName\" : \"V_2\",\n"
+     *     + "        \"safeName\" : \"V_2\"\n"
+     *     + "      }\n"
+     *     + "    }\n"
+     *     + "  }\n"
```
dsinghvi, line 52: we should get string value here

dsinghvi (approved):

PR #6710, opened 2025-04-09
fix(cli): Correctly parse primitive type arrays in the v3 OpenAPI parser.

eyw520:
## Description
This PR enables support for parsing schemas where schema.type is an array of primitive values. In effect, we enable parsing `type: ["string", "null"]` as an alternative pattern of expressing a nullable string, and `type: ["integer", "boolean"]` as an alternative way of specifying an undiscriminated union.

This PR also implements a fix where reference typeIds were being omitted during `requestBody` parsing.

## Changes Made
- Include `convertedSchema.inlinedTypes` during `requestBody` parsing.
- Implement `convertPrimitiveTypeArraySchema`, which handles `Array.isArray(schema.type)` cases.

## Testing

### DIFF
--- a/fern/pages/docs/getting-started/global-configuration.mdx
+++ b/fern/pages/docs/getting-started/global-configuration.mdx
@@ -423,7 +423,7 @@ layout:
 </ParamField>
 
 <ParamField path="layout.searchbar-placement" type="string" required={false}>
-  Sets the placement of the searchbar. Can be one of `header`, `sidebar` or `header_tabs` (places the searchbar in the header but on the tabs row). 
+  Sets the placement of the searchbar. Can be one of `header`, `sidebar` or `header-tabs` (places the searchbar in the header but on the tabs row). 
   Defaults to `sidebar`. 
 
   <Note>This setting is ignored when `disable-header` is set to true.</Note>


packages/cli/api-importers/v2-importer-commons/src/converters/schema/SchemaConverter.ts
```diff
@@ -61,6 +63,91 @@ export class SchemaConverter extends AbstractConverter<AbstractConverterContext<
...
+    private async convertPrimitiveTypeArraySchema({
+        context,
+        errorCollector
+    }: {
+        context: AbstractConverterContext<object>;
+        errorCollector: ErrorCollector;
+    }): Promise<SchemaConverter.Output | undefined> {
+        if (this.schema.type == null || !Array.isArray(this.schema.type) || this.schema.type.length === 0) {
+            errorCollector.collect({
+                message: "Received unexpected schema type array type with no types.",
```
dsinghvi, line 81: Ideally there is a different error message for each of these conditions and a sample of what the user provided: 

- `Received invalid array schema []` 
- `Received non-array schema ${this.schema}`
- `Received null schema ${schema}`

packages/cli/api-importers/v2-importer-commons/src/converters/schema/SchemaConverter.ts
```diff
@@ -61,6 +63,91 @@ export class SchemaConverter extends AbstractConverter<AbstractConverterContext<
...
+                path: this.breadcrumbs
+            });
+            return undefined;
+        }
+        const wrapAsNullable = this.schema.type.includes("null");
+        this.schema.type = this.schema.type.filter((type) => !this.SUPPORTED_PRIMITIVE_TYPES.includes(type));
+        if (this.schema.type.length === 0) {
+            errorCollector.collect({
+                message: "Received schema type array type with no supported primitive types.",
```
dsinghvi, line 90: `- Recevied schema ${schema} with unsupported primitive type ${type}`

PR #6723, opened 2025-04-10
feat(csharp): Support additional properties

Swimburger:
## Description
Add read-only and writeable additional properties for request and response types.

Example:
```csharp
var request = new FooRequest
{
    Id = "123",
    AdditionalProperties =
    {
        ["extra1"] = new { value = 42 },
        ["extra2"] = DateTime.Now,
        ["extra3"] = 99
    }
};

var response = await client.Foo.CreateAsync(request);

Dictionary<string, JsonElement> jsonElements = response.AdditionalProperties.ToJsonElementDictionary();
Dictionary<string, int> extra1 = jsonElements["extra1"].Deserialize<Dictionary<string, int>>();
DateTime extra2 = jsonElements["extra2"].GetDateTime();
int extra3 = jsonElements["extra3"].GetInt32();
````

To enable this feature, configure `experimental-additional-properties: true` in the `config` section of your C# generator in generators.yml.
This will become the default in the future.

Some of the thinking and reasoning behind the implementation of this feature can be found at the .NET runtime repo.
https://github.com/dotnet/runtime/issues/114106

## Testing

### DIFF
--- a/fern/pages/changelogs/csharp-sdk/2025-04-10.mdx
+++ b/fern/pages/changelogs/csharp-sdk/2025-04-10.mdx
@@ -1,0 +1,28 @@
+## 1.16.0
+**`(feat):`** Add read-only and writeable additional properties for request and response types.
+
+Example:
+```csharp
+var request = new FooRequest
+{
+    Id = "123",
+    AdditionalProperties =
+    {
+        ["extra1"] = new { value = 42 },
+        ["extra2"] = DateTime.Now,
+        ["extra3"] = 99
+    }
+};
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: generators/csharp/sdk/versions.yml [hunk 1]
```diff
@@ -7,6 +7,38 @@
 #         Set `enable-forward-compatible-enums` to `false` in the configuration to generate the old enums.
 #   irVersion: 53
 
+- version: 1.16.00
```
amckinney, line 10:
```suggestion
- version: 1.16.0
```

amckinney (approved):

PR #6764, opened 2025-04-16
chore(cli): Ensure schema and parameter `v2Examples` are validated.

eyw520:
## Description
This PR implements the `ExampleConverter` class, which takes in a Schema / Reference and an `unknown`-valued example, and 1) validates it against the underlying schema and 2) if invalid, returns a valid subschema example at each node. By default, we omit subschema generation for optional properties.

As part of this change, we ensure that `userSpecifiedExamples` are always valid upon reaching the `injectAutoGeneratedV2Examples` stage (to be implemented), which will greatly simplify the work required to construct valid v2 examples for the docs.

This PR ensures example validation for schemas and parameters. Support for validated `requestBody` and `response` objects coming shortly after.

## Changes Made
- Implement `ExampleConverter` class, which contains all the schema validation and autogeneration logic.
- Update `SchemaConverter` and `AbstractParameterConverter` to use the `ExampleConverter` to validate or autogenerate examples.

## Testing


packages/cli/api-importers/v2-importer-commons/src/AbstractConverterContext.ts
```diff
@@ -219,6 +224,31 @@ export abstract class AbstractConverterContext<Spec extends object> {
...
+                return example.value;
+            }
+            return example;
+        }
+        const resolved = await this.resolveReference(example);
+        if (resolved.resolved && this.isExampleWithValue(resolved.value)) {
+            return resolved.value.value;
+        }
+        return undefined;
+    }
```
amckinney, line 250: Thoughts on `resolveExampleWithValue`? That way it matches the semantics of the helper method we're calling (i.e. `isExampleWithValue`).

packages/cli/api-importers/v2-importer-commons/src/converters/schema/SchemaConverter.ts
```diff
@@ -307,40 +306,93 @@ export class SchemaConverter extends AbstractConverter<AbstractConverterContext<
...
-            const schemaExampleName = `${this.id}_example_${index}`;
-            if (context.isReferenceObject(example)) {
-                const resolved = await context.resolveReference(example);
-                if (resolved.resolved) {
-                    userSpecifiedExamples[schemaExampleName] = resolved.value;
-                }
-            } else {
-                userSpecifiedExamples[schemaExampleName] = example;
-            }
+            const resolvedExample = await context.resolveExample(example);
```
amckinney, line 358: Why do we not want to use `resolveExampleWithValue` here? Is there some reason why we need to be careful to use one vs. the other? If so, we might want to add some lightweight docs to these method declarations.

packages/cli/api-importers/v2-importer-commons/src/converters/schema/SchemaConverter.ts
```diff
@@ -307,40 +306,93 @@ export class SchemaConverter extends AbstractConverter<AbstractConverterContext<
...
+        const v2Examples = {
+            userSpecifiedExamples: {},
+            autogeneratedExamples: {}
+        };
+
+        const examples = this.getExamplesFromSchema(errorCollector);
+
+        if (examples.length === 0) {
+            const convertedExample = await this.generateOrValidateExample({
+                shouldCollectErrors: false,
```
amckinney, line 319: Why do we want to ignore the errors here? I see that we're passing in an `undefined` example value, but is it possible to refactor this so that the caller never has to decide whether or not they collect the errors?

packages/cli/api-importers/v2-importer-commons/src/converters/schema/SchemaConverter.ts
```diff
@@ -307,40 +306,93 @@ export class SchemaConverter extends AbstractConverter<AbstractConverterContext<
...
+        if (shouldCollectErrors) {
+            errors.forEach((error) => {
+                errorCollector.collect({
+                    message: error.message,
+                    path: error.path
+                });
+            });
+        }
+        return convertedExample;
+    }
```
amckinney, line 397: If we end up needing the `shouldCollectErrors` option after all (re: comment above), I wonder if we should invert this a bit.

It seems more common that we would want to collect the errors, so we don't have to require the caller to specify `true` every time. Instead we can mark this as an optional parameter and use it to _ignore_ the errors like so:

```suggestion
    private async generateOrValidateExample({
        example,
        context,
        errorCollector,
        ignoreErrors,
    }: {
        example: unknown;
        context: AbstractConverterContext<object>;
        errorCollector: ErrorCollector;
        ignoreErrors?: boolean;
    }): Promise<unknown> {
        const exampleConverter = new ExampleConverter({
            breadcrumbs: this.breadcrumbs,
            schema: this.schema,
            example
        });
        const { validExample: convertedExample, errors } = await exampleConverter.convert({ context, errorCollector });
        if (!ignoreErrors) {
            errors.forEach((error) => {
                errorCollector.collect({
                    message: error.message,
                    path: error.path
                });
            });
        }
        return convertedExample;
    }
```

That way, any time you call it by default, you can just do:

```typescript
const convertedExample = await this.generateOrValidateExample({
    example: resolvedExample,
    context,
    errorCollector
});
```

amckinney: Also, it looks like we have this function defined in the `AbstractParamterConverter`, too. Is there any way we can DRY that up, or do we need these implementations to be split?

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -0,0 +1,657 @@
...
+         *   coerced: false,  // not coerced
+         *   validExample: 42,
+         *   errors: []
+         * }
+         * ```
+         *
+         * In this case, the oneOf node will thus return the non-coerced example,
+         * treating the example as a number.
+         */
+        coerced: boolean;
```
amckinney, line 52: Should we emit a warning whenever we coerce a value?

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -0,0 +1,657 @@
...
+         * We can't collect errors along the way because for oneOf and anyOf, we don't know
+         * which branch will be valid until we have converted all of them.
+         */
+        errors: Error[];
+    }
+}
+
+export class ExampleConverter extends AbstractConverter<AbstractConverterContext<object>, ExampleConverter.Output> {
+    protected readonly MAX_DEPTH = 12;
+    protected readonly EXAMPLE_STRING = "foo";
```
amckinney, line 66: Should we still try to name the value based on the property name? Seeing `"foo"` everywhere _works_, but might not be ideal. I suppose we're doing this for `42` everywhere already, but something to consider.

We can always come back to this later, too.

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -0,0 +1,657 @@
...
+            return this.convertInteger();
+        }
+
+        if (resolvedSchema.type == "array" && "items" in resolvedSchema) {
+            return this.convertArray({
+                resolvedSchema,
+                context,
+                errorCollector
+            });
+        }
```
amckinney, line 154: Is there a `switch` we can use to make sure we've covered all the cases based on the `type` property? That way we can use `assertNever` so that if we ever add a new `type` we'll immediately get a compile time error.

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -0,0 +1,657 @@
...
+                  coerced: false,
+                  validExample: this.EXAMPLE_BOOLEAN,
+                  errors: [
+                      {
+                          message: `Example is not a boolean: ${JSON.stringify(this.example, null, 2)}`,
+                          path: this.breadcrumbs
+                      }
+                  ]
+              };
+    }
```
amckinney, line 260: These are slick 👏

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
eyw520: That's a great note. I'll implement that.

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
eyw520: I do agree -- `sample_{property_name}` might be better.

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
eyw520: I don't think so, given that it's quite common in practice, and it's probably better UX to assume the user intended the correct type.

packages/cli/api-importers/v2-importer-commons/src/converters/schema/SchemaConverter.ts
eyw520: I will aim to DRY that up in the next body of work on requestBody and response examples!

packages/cli/api-importers/v2-importer-commons/src/converters/schema/SchemaConverter.ts
eyw520: The idea here is to fully autogenerate the example when provided an `undefined` example, and when doing so, we wouldn't want to log errors that come up (since we know `undefined` is a faulty example).

packages/cli/api-importers/v2-importer-commons/src/converters/schema/SchemaConverter.ts
eyw520: Yep -- this is based on differences in the `examples` schema across schema examples and parameter, requestBody, and response examples.

packages/cli/api-importers/v2-importer-commons/src/AbstractConverterContext.ts
eyw520: Agreed!

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
eyw520: I made the `switch` `case` implementation but it's more confusing to read than this case. Given that we're depending on JsonSchema types, I don't think we'll run into errors with adding new `type` instances. 

I'm opting to keep the implementation as is for now.

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -0,0 +1,668 @@
...
+                coerced: true,
+                validExample: stringValue,
+                errors: []
+            };
+        }
+
+        return {
+            isValid: false,
+            coerced: false,
+            validExample: this.schemaId != null ? `sample_${this.schemaId}` : this.EXAMPLE_STRING,
```
amckinney, line 343: Hmm.. will `sample_${this.schemaId}` actually be pleasant in practice? Would we just see `sample_string` in many places? It might be best to start with `"foo"` (or something similar) for now, after all. But I imagine we'll want to come back and improve these placeholders later.

eyw520: It's often something like:

`sample_GenerateContentResponseBar`

Agreed. Can revert this and keep it as `foo` for the time being.

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -0,0 +1,668 @@
...
+        }
+        let resolvedSchema: OpenAPIV3_1.SchemaObject | undefined;
+        if (context.isReferenceObject(this.schema)) {
+            const resolved = await context.resolveReference<OpenAPIV3_1.SchemaObject>(this.schema);
+            if (resolved.resolved) {
+                resolvedSchema = resolved.value;
+            }
+        } else {
+            resolvedSchema = this.schema;
+        }
```
amckinney, line 100: nit: Anytime I see a `let` I think "this is a good place for a helper function". That way we can just call something like the following and avoid the `else` entirely with an early `return`.

```typescript
const resolvedSchema = this.resolveSchema(this.schema);
```

```typescript
private async resolveSchema(schema: <>): OpenAPIV3_1.SchemaObject | undefined {
  if (context.isReferenceObject(this.schema)) {
    const resolved = await context.resolveReference<OpenAPIV3_1.SchemaObject>(this.schema);
    if (resolved.resolved) {
      return resolved.value;
    }
    return undefined;
  }
  return this.schema;
}
```

packages/cli/api-importers/v2-importer-commons/src/converters/abstract/AbstractParameterConverter.ts
```diff
@@ -176,4 +183,35 @@ export abstract class AbstractParameterConverter<
                 : userSpecifiedExamples;
         return userSpecifiedExamples;
     }
+
+    private async generateOrValidateExample({
+        schema,
+        shouldCollectErrors,
```
amckinney, line 189: Thoughts on making that `optional` `ignoreErrors` parameter? I think that will clean up a lot of the call-sites.

packages/cli/api-importers/v2-importer-commons/src/converters/abstract/AbstractParameterConverter.ts
```diff
@@ -151,22 +152,28 @@ export abstract class AbstractParameterConverter<
...
-                const resolved = await this.context.resolveReference(example);
-                if (resolved.resolved) {
-                    if (this.context.isExampleWithValue(resolved.value)) {
-                        userSpecifiedExamples[key] = resolved.value.value;
-                    }
+            const resolvedExample = await this.context.resolveExampleWithValue(example);
+            if (resolvedExample != null) {
+                if (this.parameter.schema != null) {
+                    userSpecifiedExamples[key] = await this.generateOrValidateExample({
+                        shouldCollectErrors: true,
```
amckinney, line 161: We can drop this if we refactor that method above.

packages/cli/api-importers/v2-importer-tests/src/__test__/__snapshots__/openapi-v2/examples-endpoint-level.json
```diff
@@ -4508,16 +4539,12 @@
             "v2Examples": {
               "userSpecifiedExamples": {
                 "fernRequestExample": {
-                  "value": {
-                    "color": "red",
-                    "name": "RedFern"
-                  }
+                  "color": "red",
+                  "name": "RedFern"
```
amckinney, line 4543: Confirming that this diff is expected?

packages/cli/api-importers/v2-importer-tests/src/__test__/__snapshots__/openapi-v2/examples-parameter-level.json
```diff
@@ -479,16 +463,8 @@
...
-                  "first": {
-                    "value": "def",
-                    "summary": "A summary of def",
-                    "description": "A description of def"
-                  },
-                  "second": {
-                    "value": "ghi",
-                    "summary": "Another summary of ghi",
-                    "description": "Another description of ghi"
-                  }
```
amckinney, line 491: Confirming that this diff is expected?

packages/cli/api-importers/v2-importer-tests/src/__test__/__snapshots__/openapi-v2/examples-endpoint-level.json
eyw520: Yep!

packages/cli/api-importers/v2-importer-tests/src/__test__/__snapshots__/openapi-v2/examples-parameter-level.json
eyw520: Yep

PR #6801, opened 2025-04-21
chore(python): remove unnecessary poetry installs in docker builds

hughhan1:
This is done so that Docker builds can more frequently leverage the cache.

### DIFF
--- a/generators/python/fastapi/Dockerfile
+++ b/generators/python/fastapi/Dockerfile
@@ -1,4 +1,5 @@
-FROM python:3.9.14
+# Build base image with dependencies
+FROM python:3.9.14 AS base
 
 ENV PYTHONPATH=${PYTHONPATH}:${PWD}
 ENV _TYPER_STANDARD_TRACEBACK=1
@@ -6,6 +7,7 @@ ENV _TYPER_STANDARD_TRACEBACK=1
 RUN pip3 install poetry==1.8.5
 RUN poetry config virtualenvs.create false
 
+# Copy dependency files first to leverage caching
 COPY pyproject.toml ./pyproject.toml
 COPY poetry.lock ./poetry.lock
 COPY ./src/__init__.py ./src/__init__.py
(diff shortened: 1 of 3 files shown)


Swimburger (changes requested):
1. I don't believe making the build a multi-stage build doesn't enhance caching in this case.

2. Is `poetry install` not necessary at the end of the builds?
   I recall optimizing this somewhat but I couldn't remove those lines, the result would be broken seed checks.

hughhan1:
> 1. I don't believe making the build a multi-stage build doesn't enhance caching in this case.

You're right. It's just a habit that I formed. I don't think it hurts to have this. 

> 2. Is `poetry install` not necessary at the end of the builds?
>    I recall optimizing this somewhat but I couldn't remove those lines, the result would be broken seed checks.

I don't think it's necessary. This is the problematic part. Because `src/` changes quite often, it makes it such that we have to reinstall our Python dependencies every time we change our source code. But in reality this step could be cached. We should only need to reinstall our Python dependencies when `pyproject.toml` changes.

Swimburger:
> I don't think it's necessary. This is the problematic part. Because src/ changes quite often, it makes it such that we have to reinstall our Python dependencies every time we change our source code. But in reality this step could be cached. We should only need to reinstall our Python dependencies when pyproject.toml changes.

If you could verify that it isn't necessary. Does poetry not see what's already installed and only install additional things if added? Seems like that would be a big gap with other package managers I'm used to.

hughhan1:
> Does poetry not see what's already installed and only install additional things if added? 

Could you clarify what you mean?

The code in `src/` could change however we'd like, but changing the code in `src/` doesn't influence what packages need to get installed. Configuration of third-party packages happens in `pyproject.toml`. 

> Seems like that would be a big gap with other package managers I'm used to.

I'm not a Node expert so let me know if the following analogy I'm drawing isn't a fair analogy: we wouldn't need to re-run `npm install` when modifying JavaScript files in a Node project; we would only need to re-run it after modifying dependencies in the `package.json` file. Hence, we could cache the `npm install` step so long as the `package.json` file hasn't changed.

Also let me know if I misunderstood your comment!

Swimburger:
If you run npm (pnpm,yarn,etc) once, modify package.json, then install again, only the diff in the package.json would be (un)installed.

Since poetry install was called early in the build, then again later (no idea why), I'd assume it doesn't have any impact unless some file that is copied after the first install affects something related to poetry in the second install.
I don't know poetry, so I didn't dig into it when I ran into issues when optimizing Docker build.
But I trust you'll verify this works for local and CI actions.

Swimburger (approved):

hughhan1:
> If you run npm (pnpm,yarn,etc) once, modify package.json, then install again, only the diff in the package.json would be (un)installed.

Totally. And if you were to edit any JavaScript files in your project, it wouldn't affect your dependencies either, no?

PR #6809, opened 2025-04-21
chore(cli): Example generation recursively uses subschema examples when no value is provided.

eyw520:
## Description
This PR enables example generation to recursively check for subschema examples (when examples are missing, from referencing schemas or request / response schemas) during example generation.

## Changes Made
- Update `ExampleConverter` to take in a `generateOptionalProperties` parameter
- Primitive example generation first checks for any examples attached to the schema before falling back to default `EXAMPLE_VALUE`.
- Cleanup: Update `SchemaConverter` to resolve single-type type arrays as a non-array type.

## Testing


packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -5,6 +5,7 @@ import { AbstractConverter, AbstractConverterContext, type OpenApiError } from "
 export declare namespace ExampleConverter {
     export interface Args extends AbstractConverter.Args<AbstractConverterContext<object>> {
         schema: OpenAPIV3_1.SchemaObject | OpenAPIV3_1.ReferenceObject;
+        generateOptionalProperties?: boolean;
```
dsinghvi, line 8: @eyw520 can we make sure that `generateOptionalProperties` is true for responses and false for requests?

dsinghvi: e.g. developers can see a minimal request but a full response

PR #6717, opened 2025-04-10
feat(python): support response headers in streaming methods

hughhan1:
## Description
Support generating Python SDKs that include support for retrieving response headers from streaming endpoints. This is a continuation of https://github.com/fern-api/fern/pull/6617 and https://github.com/fern-api/fern/pull/6685.

**Note**: this reuses some of the functionality originally implemented by @eyw520 in https://github.com/fern-api/fern/pull/6695.

Specifically, we now support the following.
```python
client = AcmeClient()

with client.with_raw_response.generate_stream(
    num_events=42,
) as response:
    # Now that we've enterered the context manager, the response headers will be available.
    print(response.headers.get("X-Your-Header"))
    
    # We can iterate over the stream content the same as before, but with the `.data` property.
    for chunk in response.data:
        print(chunk)

    # We can retrieve the response headers after we've finished iterating over all data.
    print(response.headers.get("X-Your-Header"))

# We can still iterate over the stream without leveraging the raw response.
with client.generate_stream(
    num_events=42,
) as stream:
    for chunk in stream:
        print(chunk)
```

### Details
We leverage `contextlib.contextmanager` to allow downstream users to iterate over the response returned by "raw" clients. This makes it such that the response stays open until `__exit__` is invoked (which happens after the user exits the `with` block).
```python
class RawClient:
    @contextmanager
    def download_file(self, ...) -> Iterator[HttpResponse[Iterator[bytes]]]:
        with self._client_wrapper.httpx_client.stream(...) as _response:
            def stream() -> HttpResponse[Iterator[bytes]]:
                ...
            yield stream()

    @contextmanager
    def stream(self, ...) -> Iterator[HttpResponse[Iterator[T]]]:
        with self._client_wrapper.httpx_client.stream(...) as _response:
            def stream() -> HttpResponse[Iterator[T]]:
                ...
            yield stream()

class AsyncRawClient:
    @contextmanager
    async def download_file(self, ...) -> AsyncIterator[AsyncHttpResponse[AsyncIterator[bytes]]]:
        async with self._client_wrapper.httpx_client.stream(...) as _response:
            async def stream() -> AsyncHttpResponse[AsyncIterator[bytes]]:
                ...
            yield stream()

    @contextmanager
    async def stream(self, ...) -> AsyncIterator[AsyncHttpResponse[AsyncIterator[T]]]:
        async with self._client_wrapper.httpx_client.stream(...) as _response:
            async def stream() -> AsyncHttpResponse[AsyncIterator[T]]:
[...]

PR #6815, opened 2025-04-22
feat(python): support response headers for pagination endpoints

hughhan1:
## Description
Support generating Python SDKs that include support for retrieving response headers from pagination endpoints. This is a continuation of https://github.com/fern-api/fern/pull/6617 and https://github.com/fern-api/fern/pull/6717.

Specifically, we now support the following.
```python
client = AcmeClient()

response = client.users.with_raw_response.list_with_cursor_pagination(...)
for item in response.data:
    ...

for page in response.data.iter_pages():
    ...
```

Additionally, publish snippets for how the `with_raw_response` property can be used, now that raw responses are supported for unary, pagination, and streaming use cases.

Preliminary, fix many bugs (see the **Changes Made** section below).

## Changes Made
* Fix almost all type errors in `generators/python/core_utilities/sdk/pagination.py`.
* Refactor `generators/python/core_utilities/sdk/pagination.py` such that its logic is easier to follow.
* Fix a bug when generating paginated client methods in which `page = page if page is not None else 1` was being generated unconditionally, even if the `page` variable wasn't defined.
* Support raw responses for the pagination use case.
* Remove `pagination` from the list of `allowedFailures` for the Python SDK generator.
* Add the `ACCESS_RAW_RESPONSE_DATA` feature to the Python SDK generator, with example snippets for using `with_raw_response` for the unary, pagination, and streaming use cases,
* Bump the Python SDK generator to 4.14.3.

## Testing

- [x] Manually test against a generated Square API:
```python
from seed import Square
from seed.environment import SquareEnvironment
from seed.types.bank_account import BankAccount


def test_client() -> None:
    client = Square(environment=SquareEnvironment.SANDBOX)

    response = client.bank_accounts.list()

    for item in response:
        assert isinstance(item, BankAccount)
    
    for page in response.iter_pages():
        for item in page:
            assert isinstance(item, BankAccount)
    

def test_client_with_raw_response() -> None:
    client = Square(environment=SquareEnvironment.SANDBOX)

    response = client.bank_accounts.with_raw_response.list()

    assert response.headers.get("content-type") == "application/json"
    
    for item in response.data:
        assert isinstance(item, BankAccount)
    
    for page in response.data.iter_pages():
        for item in page:
            assert isinstance(item, BankAccount)
```

### DIFF
--- a/seed/pydantic/accept-header/poetry.lock
+++ b/seed/pydantic/accept-header/poetry.lock
@@ -91,13 +91,13 @@ reports = ["lxml"]
 
 [[package]]
 name = "mypy-extensions"
-version = "1.0.0"
+version = "1.1.0"
 description = "Type system extensions for programs checked with the mypy type checker."
 optional = false
-python-versions = ">=3.5"
+python-versions = ">=3.8"
 files = [
-    {file = "mypy_extensions-1.0.0-py3-none-any.whl", hash = "sha256:4392f6c0eb8a5668a69e23d168ffa70f0be9ccfd32b5cc2d26a34ae5b844552d"},
-    {file = "mypy_extensions-1.0.0.tar.gz", hash = "sha256:75dbf8955dc00442a438fc4d0666508a9a97b6bd41aa2f0ffe9d2f2725af0782"},
+    {file = "mypy_extensions-1.1.0-py3-none-any.whl", hash = "sha256:1be4cccdb0f2482337c4743e60421de3a356cd97508abadd57d47403e94f5505"},
+    {file = "mypy_extensions-1.1.0.tar.gz", hash = "sha256:52e68efc3284861e772bbcd66823fde5ae21fd2fdb51c62a211403730b916558"},
 ]
 
 [[package]]
--- a/seed/pydantic/alias-extends/poetry.lock
+++ b/seed/pydantic/alias-extends/poetry.lock
@@ -105,13 +105,13 @@ reports = ["lxml"]
 
 [[package]]
 name = "mypy-extensions"
-version = "1.0.0"
+version = "1.1.0"
 description = "Type system extensions for programs checked with the mypy type checker."
 optional = false
-python-versions = ">=3.5"
+python-versions = ">=3.8"
 files = [
-    {file = "mypy_extensions-1.0.0-py3-none-any.whl", hash = "sha256:4392f6c0eb8a5668a69e23d168ffa70f0be9ccfd32b5cc2d26a34ae5b844552d"},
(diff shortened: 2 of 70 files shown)


hughhan1:
I'd like for `ACCESS_RAW_RESPONSE_DATA` feature to be generated in the **Advanced**  section in the generated `README.md` files, but I don't know how to trigger that for the Python SDK.

Currently, it's being generated above the **Advanced** section.

generators/python/core_utilities/sdk/pagination.py
```diff
@@ -16,70 +17,57 @@
...
+    async def iter_pages(self) -> AsyncIterator[AsyncPager[T]]:
+        page: Optional[AsyncPager[T]] = self
+        while page is not None:
+            yield page
+
+            if not page.has_next or page.get_next is None:
+                return
+
+            page = await page.get_next()
+            if page is None or page.items is None or len(page.items) == 0:
```
amckinney, line 69: Great cleanups here 👏

generators/python/sdk/features.yml
```diff
@@ -19,6 +19,12 @@ features:
   - id: PAGINATION
     description: |
       Paginated requests will return a `SyncPager` or `AsyncPager`, which can be used as generators for the underlying object.
+  
+  - id: ACCESS_RAW_RESPONSE_DATA
+    advanced: true
+    description: |
+      The SDK provides access to raw response data, including headers, through the `.with_raw_response` property.
+      The `.with_raw_response` property returns a "raw" client that can be used to access the `.headers` and `.data` attributes.
```
amckinney, line 27: It might be worth adding a separate section tailored to streaming (for any API that defines a streaming endpoint), given that the UX is different / requires the context manager.

We might be able to use an `id` like `ACCESS_RAW_RESPONSE_DATA_(STREAMING)`, but we'll need to see if the title is rendered as I'd expect. If not, we've meant to add a separate `title` option for a while, so we can do that in a follow-up.

hughhan1: I actually wrote the code that generates these snippets to be dynamic. So for streaming endpoints, the generated `README.md` demonstrates an example for the streaming use case.

For example, if you look at `seed/python-sdk/streaming/no-custom-config/README.md`, it shows the following.
```python
from seed import SeedStreaming
client = SeedStreaming(..., )
with client.dummy.with_raw_response.generate_stream(...) as response:
    print(response.headers)  # access the response headers
    for chunk in response.data:
        print(chunk)  # access the underlying object(s)```

amckinney (approved):
Awesome work 🚀

PR #6857, opened 2025-04-24
feat(cli): Add x-fern-sdk-namespace extension

amckinney:
This adds the `x-fern-sdk-namespace` extension, which allows users to define schemas with the same name across different namespaces. This is particularly relevant if you have multiple API Definitions that have overlapping schema names or operation names.

For example,

```yaml
# ./v1/openapi.yml
components:
  schemas:
    Error:
      x-fern-sdk-namespace: v1
```

```yaml
# # ./v2/openapi.yml
components:
  schemas:
    Error:
      x-fern-sdk-namespace: v2
```


dsinghvi (approved):
skimmed the code but biggest thing for me was that none of the existing snapshots updated so this change feels safe 👌

PR #6859, opened 2025-04-24
feat(python): support retrieving response headers from ApiError

hughhan1:
## Description
Implement the `ApiError.headers` instance variable to support retrieving response headers from `ApiError` instances.

## Changes Made
- define the `ApiError.headers` instance variable
- propagate HTTP response headers when instantiating `ApiError`

## Testing


generators/python/src/fern_python/generators/sdk/client_generator/websocket_connect_response_code_writer.py
```diff
@@ -19,13 +19,15 @@ def write(self, writer: AST.NodeWriter) -> None:
...
                     self._context.core_utilities.instantiate_api_error(
+                        headers=None,
                         body=AST.Expression('"Websocket initialized with invalid credentials."'),
                         status_code=AST.Expression("status_code"),
                     )
                 )
             writer.write("raise ")
             writer.write_node(
                 self._context.core_utilities.instantiate_api_error(
+                    headers=None,
```
amckinney, line 30: Confirming we don't have access to headers in this case (even though we have the status code)?

amckinney (approved):

generators/python/src/fern_python/generators/sdk/client_generator/websocket_connect_response_code_writer.py
hughhan1: Actually, looking more carefully at the code, we _do_ have access to the headers. It wasn't obvious to me at first because of some code indirection in our generator.

Fixed in https://github.com/fern-api/fern/pull/6871.

PR #6861, opened 2025-04-24
chore(python): fix test_snippet_generator, test_generated_root_client_generator

hughhan1:
* Remove the trailing `\n` character in the `test_snipper_generator` test.
* Remove formatting assertions in the `test_generated_root_client_generator` test (taken care of by Ruff).

### DIFF
--- a/generators/python/tests/sdk/test_root_client_generator.py
+++ b/generators/python/tests/sdk/test_root_client_generator.py
@@ -1,3 +1,5 @@
+import textwrap
+
 from fern_python.codegen import AST
 from fern_python.generators.sdk.client_generator.root_client_generator import (
     RootClientGenerator,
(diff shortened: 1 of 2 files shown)


generators/python/tests/sdk/test_root_client_generator.py
```diff
@@ -52,28 +54,20 @@ def test_generated_root_client_builder() -> None:
...
-client = Acme(
-    base_url="acme.io",
-    environment=AcmeEnvironments.PRODUCTION,
-)
-"""
+    assert snippet.to_str() == textwrap.dedent(
+        """\
+        from acme.client import Acme
+        from acme.environments import AcmeEnvironments
+        client = Acme(base_url="acme.io", environment=AcmeEnvironments.PRODUCTION, )
```
amckinney, line 61: Hmm.. this looks off - why is there a `,` in the same line?

hughhan1: it’s how we generate the code. it’s different because previously we were hard coding the string during code generation, but now we’re leveraging the AST nodes.

Ruff figures out how to format it afterwards.

amckinney: Understood, thanks for that!

amckinney (approved):

PR #6871, opened 2025-04-25
fix(python): propagate headers to ApiError for websocket errors

hughhan1:
## Description
Previously, `headers` were not propagated to the generated `ApiError` for websocket errors. This change fixes that.

## Changes Made
- Implement a new `RaiseStatement` AST node type used for raising exceptions.
- Inline logic from `WebsocketConnectResponseCodeWriter` to remove indirection.
- Rewrite logic from `WebsocketConnectResponseCodeWriter` to leverage AST nodes.
- Propagate `headers` for websocket errors into `ApiError`.

## Testing

### DIFF
--- a/generators/python/sdk/versions.yml
+++ b/generators/python/sdk/versions.yml
@@ -2,7 +2,7 @@
 - version: 4.15.2
   irVersion: 57
   changelogEntry:
-    - type: feat
+    - type: fix
       summary: |
          Propagate response headers for websocket errors onto the generated `ApiError` class.
 


generators/python/sdk/versions.yml
```diff
@@ -1,4 +1,11 @@
 # For unreleased changes, use unreleased.yml
+- version: 4.15.2
```
amckinney, line 2: If we consider this a feature, let's bump the minor version. If it's a fix, then the version is good as-is.

amckinney (approved):
Test definition files look great, thanks for following up here!

PR #6887, opened 2025-04-28
feat(openapi): add support for parsing external refs in AbstractConverter/OpenAPI

sahil485:
## Description
- support for parsing definitions from external schema refs, specifically those that are publicly available over `http(s)://`

## Changes Made
- add `resolveExternalRefs()` in `AbstractConverter.ts` which takes OpenAPI and a context object as input, and uses a BFS traversal through its fields to find $refs
- after identifying refs, repeatedly calling `resolveReference` on them until either reaching an external ref or reaching a base type defined in the schema
- when an external ref is found, it is resolved and replaced in the spec just before the `convert()` calls in the `convertSchemas()` and `convertPaths()` functions of `OpenAPIConverter`
- added a function to recursively resolved schemas within external refs before changing them in the spec
- expanded url-reference test fixture with examples of external refs in requestBody, responseBody, schema, allOf, oneOf, etc.
## Testing
- tested using SPSCommerce API Spec
- test fixtures


packages/cli/api-importers/v2-importer-commons/src/AbstractConverter.ts
```diff
@@ -90,6 +92,66 @@ export abstract class AbstractConverter<Context extends AbstractConverterContext
...
+                            }
+                        }
+                    } else {
+                        queue.push(value);
+                    }
+                }
+            }
+        }
+        return spec;
+    }
```
dsinghvi, line 153: nit: this is a lot of nesting (is it possible to refactor into helper methods without more than ~3-4 levels of nesting for ease of readability

dsinghvi (commented):
@sahil485 would be great if you could some tests to the `v2-importer-tests` package with external refs

packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts
```diff
@@ -127,11 +128,12 @@ export class OpenAPIConverter extends AbstractConverter<OpenAPIConverterContext3
         });
 
         for (const [id, schema] of Object.entries(this.context.spec.components?.schemas ?? {})) {
+            const partiallyResolvedSchema = await this.resolveExternalRefs({ spec: schema, context: this.context });
```
dsinghvi, line 131: can we call `this.resolveExternalRefs` with the whole spec at the beginning of the convert function instead?

packages/cli/api-importers/v2-importer-commons/src/AbstractConverter.ts
```diff
@@ -90,6 +92,64 @@ export abstract class AbstractConverter<Context extends AbstractConverterContext
...
+            if (Array.isArray(current)) {
+                await this.processArray(current, context, queue);
+            } else if (typeof current === "object") {
+                await this.processObject(current as Record<string, unknown>, context, queue);
+            }
+        }
+        return spec;
+    }
+
+    private async processArray(arr: unknown[], context: Context, queue: unknown[]): Promise<void> {
```
dsinghvi, line 113: nit: resolveExternalRefsInArray

packages/cli/api-importers/v2-importer-commons/src/AbstractConverter.ts
```diff
@@ -90,6 +92,64 @@ export abstract class AbstractConverter<Context extends AbstractConverterContext
...
+        for (let i = 0; i < arr.length; i++) {
+            const resolvedRefVal = await this.resolveReferenceChain(arr[i], context, queue);
+            if (resolvedRefVal != null) {
+                arr[i] = resolvedRefVal;
+            }
+        }
+    }
+
+    private async processObject(obj: Record<string, unknown>, context: Context, queue: unknown[]): Promise<void> {
```
dsinghvi, line 122: nit: resolveExternalRefsInObject

dsinghvi (approved):
🚀 lets add a changelog to versions.yml https://github.com/fern-api/fern/blob/main/packages/cli/cli/versions.yml and release!

PR #6891, opened 2025-04-28
chore(docs): update changelog title

chdeskur:
this PR clarifies the changelog title to better describe the current state of the changelog. 

fast-follow: add an actual docs changelog

### DIFF
--- a/fern/pages/changelogs/docs-full/2025-04-21.mdx
+++ b/fern/pages/changelogs/docs-full/2025-04-21.mdx
@@ -1,7 +1,0 @@
-- fix(ci): Fix failing fern-generate CI job
-- fix: openrpc has json rpc wrapper even if no params
-- chore(deps-dev): bump @types/adm-zip from 0.5.5 to 0.5.7
-- chore: return body string with undefined body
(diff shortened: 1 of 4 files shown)


fern/pages/changelogs/docs-full/2025-04-21.mdx
```diff
@@ -0,0 +1,7 @@
+- fix(ci): Fix failing fern-generate CI job
+- fix: openrpc has json rpc wrapper even if no params
```
sbawabe, line 2: **[q/nit]** do we want these to copy our exact commit names like this, or do we want to edit these to conform to conventional commit styling so that this is easier to parse/read?

this could be a good opportunity to set ourselves up for maintaining better conventional commit messages, since the better we do at naming our commits, the less we have to change in these markdowns:)

chdeskur: yeah, we definitely should clean these up! 

right now, the SDK changelogs are added along with PRs, so we should follow a convention similar to this: https://buildwithfern.com/learn/sdks/introduction/changelog/csharp

dsinghvi (approved):

PR #6902, opened 2025-04-29
feat(php): Custom package path

ajgateno:
This PR adds the ability to configure a custom package path under which source files will be generated.


generators/php/base/src/project/PhpProject.ts
```diff
@@ -43,7 +44,17 @@ export class PhpProject extends AbstractProject<AbstractPhpGeneratorContext<Base
...
+        this.packagePathPrefix = "";
+        if (this.context.customConfig.packagePath != null) {
+            this.packagePathPrefix = this.context.customConfig.packagePath;
+            if (!this.packagePathPrefix.endsWith("/")) {
+                this.packagePathPrefix += "/";
+            }
+            if (!this.packagePathPrefix.startsWith("/")) {
+                this.packagePathPrefix = "/" + this.packagePathPrefix;
+            }
+        }
```
amckinney, line 56: Thoughts on turning this into a helper function so that we can just set the result directly?

```typescript
this.packagePathPrefix = getPackagePathPrefix(this.context.customConfig.packagePath);
```

The parameter is just pass by value so I can think we can just do something along the lines of:
```typescript
private getPackagePathPrefix(packagePath?: string): string {
  if (!packagePath) {
    return "";
  }
  if (!packagePath.endsWith("/")) {
    packagePath += "/";
  }
  if (!packagePath.startsWith("/")) {
    packagePath = "/" + packagePath;
  }
  return packagePath;
}
```

amckinney (approved):
Nice looks good! Just a nit for a helper function.

generators/php/base/src/project/PhpProject.ts
ajgateno: done

PR #6914, opened 2025-04-30
feat(cli): setup self hosted github option

dsinghvi:
## Description
```yml
# generators.yml

github: 
  uri: <some git url>
  token: <the token to push to>
```

## Changes Madex
- Adds github uri and token to generators.yml and ir 

## Testing

### DIFF
--- a/packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
+++ b/packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
@@ -182,6 +182,21 @@ function getPublishConfig({
 }: {
     generatorInvocation: generatorsYml.GeneratorInvocation;
 }): FernIr.PublishingConfig | undefined {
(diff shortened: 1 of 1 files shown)


fern/apis/generators-yml/definition/group.yml
```diff
@@ -63,9 +63,15 @@ types:
   GithubConfigurationSchema:
     discriminated: false
     union:
+      - GithubSelfhostedSchema
       - GithubCommitAndReleaseSchema
       - GithubPullRequestSchema
       - GithubPushSchema
+  
+  GithubSelfhostedSchema:
```
amckinney, line 71: nit: I'd consider this `GithubUriSchema` since it's not necessarily just used in self-hosted environments. The name is less important here though - the property names are what matter.

amckinney (approved):

fern/apis/generators-yml/definition/group.yml
dsinghvi: @amckinney in a bit of a rush, will follow up since a rename isnt a break for us

PR #6918, opened 2025-04-30
chore(docs): Add ideas for AI Search docs

dannysheridan:
Key ideas
- propose improved screenshot for citations.png
- show an example of a custom system prompt

Objections
- in `docs.yml` why are we using `ai-chat:` as the key?
- in `docs.yml` can I add a text file instead of pasting the prompt in `docs.yml`? I don't think that's the right location for it -- too messy.


ryan-fern (changes requested):
@dannysheridan thanks! All makes sense. 

The one change  that we should update (this is my original writing) is change `Edit the docs.yml file to include the ai-search key.` to be `ai-chat` instead. This is only in the interim while we change the key name.

dannysheridan:
> @dannysheridan thanks! All makes sense.
> 
> The one change that we should update (this is my original writing) is change `Edit the docs.yml file to include the ai-search key.` to be `ai-chat` as the key instead. This is only in the interim while we change the key name.

Edited to be `ai-search` per https://github.com/fern-api/fern/pull/6944

PR #6933, opened 2025-05-01
fix(cli): Correctly merge `allOf` schemas with parent `properties`.

eyw520:
## Description
Fix an issue where example generation would not correctly merge allOf schemas with parent object schemas.

## Changes Made
- Parse `object` and `allOf` in the same logic branch.

## Testing


packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -540,19 +534,67 @@ export class ExampleConverter extends AbstractConverter<AbstractConverterContext
...
+                    example: this.example,
+                    depth: this.depth + 1,
+                    generateOptionalProperties: this.generateOptionalProperties
+                });
+                return await exampleConverter.convert();
+            })
+        );
+
+        const isValid =
+            resultsByKey.every((entry) => entry.result.isValid) && allOfResults.every((result) => result.isValid);
```
dsinghvi, line 552: what happens when one entry is invalid? (wouldn't that generate an invalid example then?)

packages/cli/api-importers/v2-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -540,19 +534,67 @@ export class ExampleConverter extends AbstractConverter<AbstractConverterContext
...
+                    coerced: false,
+                    validExample: firstValidResult.validExample,
+                    errors: isValid
+                        ? []
+                        : [
+                              ...resultsByKey.flatMap(({ result }) => result.errors),
+                              ...allOfResults.flatMap((result) => result.errors)
+                          ]
+                };
+            }
```
dsinghvi, line 577: can you explain what happens here when we encounter a non object allOf example?

dsinghvi (approved):

PR #6936, opened 2025-05-02
fix(cli): openprc supports a top level `x-fern-parameters`

dsinghvi:
## Description
Adds a top-level `x-fern-parameters` to OpenRPC so that OpenRPC implementors can still document path params, query params, and headers. 

## Changes Made
- Add a ParameterConverter to OpenRPC conversion
- Add an extension converter for `x-fern-parameters` 

## Testing

### DIFF
--- a/fern/pages/changelogs/docs/2025-05-02.mdx
+++ b/fern/pages/changelogs/docs/2025-05-02.mdx
@@ -1,0 +1,3 @@
+- feat(cli): using `fern docs dev` on the latest CLI will now better reflect the docs in production
+- feat(search): the search UX now uses infinite scroll and allows for searching based on breadcrumb paths
+- fix(ai): small bug fixes to the AI chat experience


packages/cli/api-importers/openrpc-to-ir/src/1.x/OpenRPCConverter.ts
```diff
@@ -93,6 +95,48 @@ export class OpenRPCConverter extends AbstractConverter<OpenRPCConverterContext3
...
     private async convertMethods(): Promise<void> {
+        // Import the FernParametersExtension to handle custom parameters
+        const fernParametersExtension = new FernParametersExtension({
+            context: this.context,
+            breadcrumbs: ["methods"],
+            operation: this.context.spec
+        });
+        
+        const customParameters = fernParametersExtension.convert();
+        // Process custom parameters from x-fern-parameters extension
```
eyw520, line 106: Are we using this extension in OpenAPI and AsyncAPI, or is this only for OpenRPC?

dsinghvi: Only for OpenRPC, i dont think it makes sense for OpenAPI where you can declare them

eyw520 (approved):

PR #6937, opened 2025-05-02
fix(cli): v3 parser should not hard fail on receiving non-object schemas.

eyw520:
## Description
This PR ensures that the v3 parser does not hard fail upon receiving non-object schemas.

## Changes Made
- Collect an error when receiving a non-object schema.

## Testing


packages/cli/api-importers/v2-importer-commons/src/utils/ConvertProperties.ts
```diff
@@ -23,6 +23,13 @@ export async function convertProperties({
     let inlinedTypesFromProperties: Record<TypeId, TypeDeclaration> = {};
     for (const [propertyName, propertySchema] of Object.entries(properties ?? {})) {
         const propertyBreadcrumbs = [...breadcrumbs, "properties", propertyName];
+        if (typeof propertySchema !== "object") {
+            errorCollector.collect({
+                message: `Schema property ${propertyName} should be an object`,
+                path: propertyBreadcrumbs
+            });
+            continue;
+        }
```
dsinghvi, line 32: should we also add a try catch here in general, i.e. what if they make it an object without a `type` keyword?

dsinghvi (approved):

packages/cli/api-importers/v2-importer-commons/src/utils/ConvertProperties.ts
eyw520: Won't throw. E.g., we collect error:

`
       Failed to convert schema object: {
         "a": "string"
       }
`

PR #6951, opened 2025-05-04
feat(cli): support external refs in asyncapi specs on v3 parser, fix edge case for circular refs within external refs

sahil485:
## Description
- add external support for AsyncAPI specs on the v3 parser
- fix an edge case where external refs that contained circular refs would infinitely loop/crash the docs preview/CLI
## Changes Made
- add an extra schema parsing step in AsyncAPIConverter.convert() using a call to resolveExternalRefs() from AbstractConverter to replace external ref definitions in-place for AsyncAPI specs
- add a visited set to AbstractConverterContext.resolvedNestedExternalRefs() to prevent infinitely recursing into circular refs
- add fixtures for external ref testing in openapi, asyncapi-2.6.0, and asyncapi-3.0.0
## Caveat
- due to the recommended approach (use a visited set, don't add nested circular refs to IR directly, etc.), circular refs nested within external refs that are a property of some internal ref do not repeatedly expand as desired in documentation 

*Example affected case:*
```
components:
    schemas:
        Schema1:
            additionalProperties: false
            properties:
                prop1:
                    description: Property 1
                    example: Hello
                    type: string
                prop2:
                    $ref: 'https://<external-ref-url>#/components/schemas/Prop2'
```

If 'https://<external-ref-url>#/components/schemas/Prop2' contains a circular ref, it won't be repeatedly expandable because Prop2 itself is never added to IR, so the circular ref within it won't be accessible/expandable.

*Example unaffected case:*
```
components:
    schemas:
        Schema1:
             $ref: 'https://<external-ref-url>#/components/schemas/Schema1'
```

Since 'https://<external-ref-url>#/components/schemas/Schema1' is a top-level schema and is added to IR, if it contains a circular reference, it will be repeatedly expandable just like circular refs in internal schemas

## Testing
- plugged external refs into Paradex's AsyncAPI spec and verified that they resolved correctly
- created specs with custom external circular refs and verified they resolved correctly for SPSCommerce and Paradex in OpenAPI and AsyncAPI

### DIFF
--- a/generators/python/core_utilities/sdk/base_test_utilities.py
+++ b/generators/python/core_utilities/sdk/base_test_utilities.py
@@ -87,10 +87,10 @@ def validate_field(response: typing.Any, json_expectation: typing.Any, type_expe
 
     # When dealing with containers of models, etc. we're validating them implicitly, so no need to check the resultant list
     if not is_container_of_complex_type:
-        assert (
-            json_expectation == response
-        ), "Primitives found, expected: {0} (type: {1}), Actual: {2} (type: {3})".format(
-            json_expectation, type(json_expectation), response, type(response)
+        assert json_expectation == response, (
+            "Primitives found, expected: {0} (type: {1}), Actual: {2} (type: {3})".format(
+                json_expectation, type(json_expectation), response, type(response)
+            )
         )
 
 
--- a/generators/python/core_utilities/shared/serialization.py
+++ b/generators/python/core_utilities/shared/serialization.py
@@ -159,9 +159,7 @@ def _convert_mapping(
 ) -> typing.Mapping[str, object]:
     converted_object: typing.Dict[str, object] = {}
     try:
-        annotations = typing_extensions.get_type_hints(
-            expected_type, include_extras=True
-        )
+        annotations = typing_extensions.get_type_hints(expected_type, include_extras=True)
     except NameError:
         # The TypedDict contains a circular reference, so
         # we use the __annotations__ attribute directly.
--- a/generators/python/sdk/versions.yml
+++ b/generators/python/sdk/versions.yml
@@ -1,4 +1,20 @@
 # For unreleased changes, use unreleased.yml
+- version: 4.18.1
+  irVersion: 57
+  changelogEntry:
+    - type: fix
+      summary: |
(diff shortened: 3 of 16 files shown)


packages/cli/api-importers/v2-importer-commons/src/AbstractConverterContext.ts
```diff
@@ -185,6 +185,7 @@ export abstract class AbstractConverterContext<Spec extends object> {
         let fragment: string | undefined;
         let externalRef: boolean = false;
```
eyw520, line 186: Would rename to `isExternalRef` for clarity.

packages/cli/api-importers/v2-importer-commons/src/AbstractConverterContext.ts
```diff
@@ -185,6 +185,7 @@ export abstract class AbstractConverterContext<Spec extends object> {
         let fragment: string | undefined;
```
eyw520, line 185: Let's rename this to something that's a bit more appropriate, like `referenceSource`, given that we're indexing over this to resolve the final reference.

packages/cli/api-importers/v2-importer-commons/src/AbstractConverterContext.ts
```diff
@@ -252,27 +262,43 @@ export abstract class AbstractConverterContext<Spec extends object> {
...
      * @returns The resolved object
      */
-    private async resolveNestedExternalReferences(obj: unknown, rootDoc: unknown): Promise<unknown> {
+    private async resolveNestedExternalReferences(
+        obj: unknown,
+        rootDoc: unknown,
+        visitedRefs: Set<string>,
+        baseUrl: string | undefined
+    ): Promise<unknown> {
         if (obj === null || typeof obj !== "object") {
```
eyw520, line 271: Could we ever resolve to a non-object (e.g, string) reference that is still an external ref?

eyw520 (approved):
Left a few minor comments, but the logic looks sound.

packages/cli/api-importers/v2-importer-commons/src/AbstractConverterContext.ts
sahil485: Not in valid specs, but this also ensures we don't crash for invalid specs within external refs.

PR #6956, opened 2025-05-05
chore(ci): update action to commit lint fixes

davidkonigsberg:
This is just a test, I will need to open a separate PR to do this the way people want it to work.


davidkonigsberg:
Going a different direction with this

PR #6986, opened 2025-05-05
fix(python): Add headers to strongly typed errors

amckinney:
This updates the SDK to include response headers in strongly typed errors.

**Before**

```python
class MovieDoesNotExistError(ApiError):
    def __init__(self, body: MovieId):
        super().__init__(status_code=404, body=body)
```

**After**

```python
class MovieDoesNotExistError(ApiError):
    def __init__(self, body: MovieId, headers: typing.Optional[typing.Dict[str, str]] = None):
        super().__init__(status_code=404, headers=headers, body=body)
```

With this, users should be able to interact with the `ApiError`'s response headers whenever they're available.


dsinghvi (approved):
Blind +1

PR #7003, opened 2025-05-06
fix(cli): reduce rpc calls in local mode

dsinghvi:
## Description
Checks if the command is running in `local mode` and if so doesn't make an upgrade message or send posthog events. 

## Changes Made
See description. 

## Testing


mattblank11 (approved):
🚀

amckinney (approved):

PR #7017, opened 2025-05-07
chore(cli): all generators preinstall @fern-api/generator-cli

dsinghvi:
## Description
This PR updates all the SDK Dockerfiles to install `@fern-api/generator-cli`. It also updates the GeneratorAgent class to not fail hard if it cannot install a newer version of the generator CLI. 

## Changes Made
- See description

## Testing


generators/base/src/GeneratorAgentClient.ts
```diff
@@ -78,7 +78,12 @@ export class GeneratorAgentClient {
             logger: this.logger
         });
         this.logger.debug(`Installing ${GENERATOR_AGENT_NPM_PACKAGE} ...`);
-        await npm(["install", "-f", "-g", GENERATOR_AGENT_NPM_PACKAGE]);
+        try {
+            await npm(["install", "-f", "-g", GENERATOR_AGENT_NPM_PACKAGE]);
```
dsinghvi, line 82: @amckinney i think it would be better to just not make this call if selfHosted. What do you think?

amckinney (approved):

PR #7025, opened 2025-05-07
feat(python): generate standalone models instead of inheritance for extends

hughhan1:
## Description
There is a class of circular reference errors in Python related to inheritance of Pydantic models (see https://github.com/fern-api/fern/issues/7026). This change is done to resolve that class of problems.

Resolve this class of problems by generating class definitions for Pydantic models as standalone classes that inline all attributes of the Fern models that they extend (i.e. models defined via the [`extends`](https://buildwithfern.com/learn/api-definition/fern/types#objects) keyword), instead of generating these Pydantic models as subclasses of the models they extend.

For example, the following Fern definition would produce the following Python snippet, instead of producing Pydantic models that subclass each other.
```yml
types:
  NestedType:
    extends: JSON
    properties:
      name: string
    examples:
      - name: One
        value:
          docs: This is an example nested type.
          name: NestedExample
          raw: '{"nested": "example"}'

  Docs:
    properties:
      docs: string
    examples:
      - name: One
        value:
          docs: Types extend this type to include a docs property.

  JSON:
    extends: Docs
    properties:
      raw: string
    examples:
      - name: One
        value:
          docs: Types extend this type to include a docs and json property.
          raw: '{"docs": true, "json": true}'
```
```python
class NestedType(UniversalBaseModel):
    name: str
    raw: str
    docs: str
    ...

class Docs(UniversalBaseModel):
    docs: str
    ...

class Json(UniversalBaseModel):
    raw: str
    docs: str
    ...
```

This change is gated behind a new `use_inheritance_for_extended_models` configuration flag so that it can be rolled out slowly to different clients.
```yml
groups:
  python:
    generators:
        config:
          use_inheritance_for_extended_models: false
          ...
        ...
      ...
```

## Changes Made
- Update `circular-references-advanced` to include a test case that reproduces the problematic circular reference error described in https://github.com/fern-api/fern/issues/7026.
- Modify the Python generator to remove support for generating Pydantic models that subclass other Pydantic models.
- Modify the Python generator to inline attributes of models that are defined in API definitions to extend other models.

## Testing

### DIFF
--- a/fern/pages/changelogs/cli/2025-05-13.mdx
+++ b/fern/pages/changelogs/cli/2025-05-13.mdx
@@ -1,0 +1,4 @@
+## 0.60.30
+**`(feat):`** Add support for products and versioned products in the docs config file.
+
+


amckinney (approved):
Nice and simple, well done 👏

PR #7028, opened 2025-05-07
chore(cli): Exclude deprecated fields from examples in docs 

mattblank11:
## Description
We got [this issue](https://linear.app/buildwithfern/issue/FER-4523/docs-example-generation-should-skip-using-optional-deprecated) to exclude deprecated optional fields from docs examples

## Changes Made
Updated `ExampleConverter` to exclude deprecated + optional fields

## Testing
[Loom](https://www.loom.com/share/be8a28160074448898bf421d7eb77a37) + will run fixture tests before merging

### DIFF
--- a/packages/cli/register/src/ir-to-fdr-converter/__test__/__snapshots__/oas-ir-fdr/examples-endpoint-level.json
+++ b/packages/cli/register/src/ir-to-fdr-converter/__test__/__snapshots__/oas-ir-fdr/examples-endpoint-level.json
@@ -116,97 +116,6 @@
         ]
       }
     },
-    "PlantHealthStatus": {
-      "name": "PlantHealthStatus",
-      "shape": {
-        "type": "enum",
-        "values": [
-          {
-            "value": "healthy"
-          },
-          {
-            "value": "needs_water"
-          },
-          {
-            "value": "needs_sun"
-          },
-          {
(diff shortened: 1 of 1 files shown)


dsinghvi (approved):

eyw520 (changes requested):
LGTM, two minor comments:

1. I think @dsinghvi and I prefer to omit comments unless they're needed to understand the code -- in this case, the logic is pretty straightforward, so I'd remove the comments.

2. Were there any fixtures that contained deprecated properties that are now being removed? If not, it would be worth adding a new fixture for testing purposes.

packages/cli/api-importers/v2-importer-tests/src/__test__/fixtures/examples-endpoint-level/openapi.yml
```diff
@@ -181,22 +181,31 @@ components:
...
               format: float
-      
+      example:
+        species: "Quercus alba"
+        height: 25.4
+        age: 50
+        location:
+          latitude: 40.7128
+          longitude: -74.0060
+
```
mattblank11, line 205: @eyw520 Deprecated the age field to ensure that it won't show up in the example, which you can see above

dsinghvi (approved):

PR #7056, opened 2025-05-08
fix(cli): endpoint resolver should be more forgiving on special chars

dsinghvi:
## Description
Fern CLI fails on a user's oauth endoint if it contains "." (e.g. `/oauth2.token`). 

## Changes Made
- This updates the regex to be more forgiving. 

## Testing


mattblank11 (approved):
LGTM!

Issue #7063, opened 2025-05-08
Pydantic "not fully defined" errors for containerized forward references


hughhan1:
### Which Fern component?

SDK Generator

### How urgent is this?

P1 - High (Strongly needed)

### What's the issue?

Consider the following Fern definition.
```yml
types:
  Node:
    discriminated: false
    union:
      - BranchNode
      - LeafNode
  BranchNode:
    properties:
      children: list<Node>
  LeafNode:
    properties: {}
  NodesWrapper:
    properties:
      nodes: list<list<Node>>
```

Running the Python SDK generator produces the following code.
##### nodes_wrapper.py
```python
import typing

from ...core.pydantic_utilities import UniversalBaseModel
from .node import Node

class NodesWrapper(UniversalBaseModel):
    nodes: typing.List[typing.List[Node]]
    ...
```
##### node.py
```python
import typing

from .leaf_node import LeafNode

if typing.TYPE_CHECKING:
    from .branch_node import BranchNode
Node = typing.Union["BranchNode", LeafNode]
```
However, note that `Node` contains a forward reference, but no invocation to `update_forward_refs(NodesWrapper)` is generated after the `NodesWrapper` class definition. Hence, when attempting to instantiate `NodesWrapper(...)`, the following error is received.
```
pydantic.errors.PydanticUserError: `NodesWrapper` is not fully defined; you should define `BranchNode`, then call `NodesWrapper.model_rebuild()`.
```

The proper code to be generated would be the following.
##### nodes_wrapper.py
```python
import typing

from ...core.pydantic_utilities import UniversalBaseModel, update_forward_refs

class NodesWrapper(UniversalBaseModel):
    nodes: typing.List[typing.List["Node"]]
    ...

from .branch_node import BranchNode
from .node import Node

update_forward_refs(NodesWrapper)
```

### Fern CLI & Generator Versions

SDK Generator versions:
- fernapi/fern-python-sdk v4.18.4

### Workaround
The following are examples of workarounds for specific versions of this problem:
* https://github.com/elevenlabs/elevenlabs-python/pull/528
* https://github.com/vectara/python-sdk/pull/49

### Are you interested in contributing a fix?

Yes

PR #7064, opened 2025-05-08
feat(python): properly support forward references wrapped in containers

hughhan1:
## Description
Modify the Python generator such that it generates imports for Pydantic model attributes that are part of circular dependency chains to be imported as forward references. This change unfortunately causes some definitions that do not _need_ to be imported as forward references to also be imported as forward references, which increases some complexity, but is not incorrect.

This is done specifically to fix https://github.com/fern-api/fern/issues/7063, as the current behavior of the generator is not general enough to account for nested attributes (e.g. types within lists, unions, etc.).

## Changes Made

- Modify [`_must_import_after_current_declaration`](https://github.com/fern-api/fern/blob/23e8ea23c3a98f195dc6873bbf625158445ccb7b/generators/python/src/fern_python/generators/pydantic_model/fern_aware_pydantic_model.py#L202-L213) to also import definitions as forward references if any of the definitions are part of a circular reference chain.
- Release `v4.19.0` of the Python generator, which includes support for this fix.

## Testing

- [x] Augment `test-definitions/fern/apis/circular-references-advanced/definition/ast.yml` to include this test.


generators/python/src/fern_python/generators/pydantic_model/fern_aware_pydantic_model.py
```diff
@@ -201,16 +201,23 @@ def _type_id_for_forward_ref(self) -> Optional[ir_types.TypeId]:
...
+        if self._context.does_type_reference_other_type(type_id=type_name.type_id, other_type_id=type_id_to_reference):
+            # direct forward reference
             self._model_contains_forward_refs = True
+            return True
+
+        for referenced_type_id in self._context.get_referenced_types(type_id_to_reference):
+            if self._context.does_type_reference_other_type(
+                type_id=referenced_type_id, other_type_id=type_name.type_id
+            ):
+                # indirect forward reference, i.e. through a container (e.g. list, optional, etc.)
```
amckinney, line 216: Is this _always_ a container? From the implementation, it sounds like this could just be any of the other referenced types -- might these always just be named types instead?

amckinney: Discussed offline - we might actually look deeper into the `does_type_reference_other_type` function. I have a feeling we aren't visiting the nodes as we expect there.

hughhan1: Done. Much cleaner.

hughhan1: Went ahead and took a different approach instead, looking at if the thing we're importing is part of a circular dependency chain with itself.

dsinghvi (approved):

PR #7065, opened 2025-05-08
fix(cli): Implement v3 parser support for endpoint audiences filtering.

eyw520:
## Description
This PR implements endpoint filtering by audiences.

## Changes Made
- Track and filter endpointIds with IrGraph and the `x-fern-audiences` extension.
- Fix a bug where the populated baseUrl for endpoints without endpoint-level servers would map to invalid baseUrls.

## Testing

### DIFF
--- a/packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts
+++ b/packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts
@@ -308,10 +308,12 @@ export class OpenAPIConverter extends AbstractSpecConverter<OpenAPIConverterCont
                     }
                     this.ir.services[pkg.service]?.endpoints.push(endpoint.endpoint);
 
-                    const newService = this.ir.services[pkg.service]!;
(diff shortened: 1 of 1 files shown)


packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts
```diff
@@ -117,13 +135,17 @@ export class OpenAPIConverter extends AbstractSpecConverter<OpenAPIConverterCont
         return idToAuthScheme;
     }
 
-    private convertServers({ endpointLevelServers }: { endpointLevelServers?: OpenAPIV3_1.ServerObject[] }): void {
+    private convertServers({
+        endpointLevelServers
+    }: {
+        endpointLevelServers?: OpenAPIV3_1.ServerObject[];
+    }): string | undefined {
```
dsinghvi, line 142: nit: we should return an object so its easy to know what you are returning

dsinghvi (approved):

PR #7072, opened 2025-05-09
chore(cli): v3 IR construction methods are refactored into the `AbstractSpecConverter` class.

eyw520:
## Description
This is a staging PR that refactors all IR construction logic from the various SpecConverter nodes into the `AbstractSpecConverter` class.

A subsequent change will invoke IrGraph methods to eventually build the `FilteredIR`, which will be used in the `finalizeIR` step.

No snapshots are updated as part of this change.

## Changes Made
- Refactor out all direct `this.ir` construction logic into `AbstractSpecConverter` methods.
- Implement a `finalizeIr` method, which performs example injection and filtering.

## Testing

### DIFF
--- a/packages/cli/configuration-loader/src/generators-yml/index.ts
+++ b/packages/cli/configuration-loader/src/generators-yml/index.ts
@@ -10,4 +10,3 @@ export {
 } from "./loadGeneratorsConfiguration";
 export { updateGeneratorGroup } from "./updateGeneratorGroup";
 export { isGithubSelfhosted } from "./convertGeneratorsConfiguration";
-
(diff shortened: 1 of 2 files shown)


packages/cli/api-importers/v2-importer-commons/src/AbstractSpecConverter.ts
```diff
@@ -84,6 +89,193 @@ export abstract class AbstractSpecConverter<
         };
     }
 
+    protected finalizeIr(): IntermediateRepresentation {
```
dsinghvi, line 92: hell yeah

dsinghvi (approved):

PR #7093, opened 2025-05-09
feat(cli): upgrade fdr-cjs-sdk and add propertyAccess field for IR -> FDR

sahil485:
## Description
- upgrade fdr-cjs-sdk to 0.136.4-de2a61d1a to use propertyAccess field in IR -> FDR conversion
## Changes Made
- upgrade version in `packages/cli/register/package.json`
- add `propertyAccess` field to relevant conversion methods

## Testing
- [ X] Unit tests added/updated


packages/cli/register/src/ir-to-fdr-converter/convertPackage.ts
```diff
@@ -465,7 +465,8 @@ function convertRequestBody(irRequest: Ir.http.HttpRequestBody): FdrCjsSdk.api.v
                                 description: property.docs ?? undefined,
                                 key: FdrCjsSdk.PropertyKey(property.name.wireValue),
                                 valueType: convertTypeReference(property.valueType),
-                                availability: convertIrAvailability(property.availability)
+                                availability: convertIrAvailability(property.availability),
+                                propertyAccess: undefined
```
dsinghvi, line 469: are you sure there's no propertyAccess here that we can send up?

sahil485: not currently:
image

PR #7109, opened 2025-05-12
fix(cli): v3 parser fixes for Async API 2.x websockets

fern-support:
## Description
- fix issues with path parameter rendering, optional parameters, and headers for websocket 

## Changes Made
- entities defined in the `properties` section of an AsyncAPI spec are processed as path parameters (as per v1 parser)
- path parameters are required by default

## Testing
- local preview
- fixtures


ig-raiderler:
Great implementation of this feature. I particularly like the approach to error handling.

eyw520 (approved):

PR #7111, opened 2025-05-12
chore(cli): Fix broken `fileUpload` response example; insert path parameters into v2Examples.

eyw520:
## Changes Made
- Correctly set example responseBody when value is `null`.
- Update v2Examples to include a fully-formed path.

## Testing


ig-raiderler:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

dsinghvi (approved):

PR #7123, opened 2025-05-13
fix(python): Pydantic V2 models extend UniversalRootModel, not pydantic.RootModel

hughhan1:
## Description
Modify the Python generators such that Pydantic models that are generated to extend `pydantic.RootModel` are generated to instead extend `UniversalRootModel` when using Pydantic V2.

This is done because extending `pydantic.RootModel` without providing generic type arguments results in the following mypy error.
```
error: Missing type parameters for generic type "RootModel"  [type-arg]
```

Note that [`UniversalRootModel`](https://github.com/fern-api/fern/blob/fee0e56a169974fdae494edc481ac82a4b8b65db/generators/python/core_utilities/shared/pydantic_utilities.py#L180-L187) already [has these type-ignores defined](https://github.com/fern-api/fern/blob/fee0e56a169974fdae494edc481ac82a4b8b65db/generators/python/core_utilities/shared/pydantic_utilities.py#L182), and also already [handles V2 gating logic](https://github.com/fern-api/fern/blob/fee0e56a169974fdae494edc481ac82a4b8b65db/generators/python/core_utilities/shared/pydantic_utilities.py#L180-L187).

Release these fixes in new versions of the FastAPI, Pydantic, and SDK generators for Python.

## Changes Made

- Remove support for directly generating `pydantic.RootModel`, and replace it with generating `UniversalRootModel`.
- Release minor versions of the FastAPI and Pydantic generators that includes this fix along with a previous feature.
- Release patch version of the Python SDK generator that includes this fix.

### DIFF
--- a/generators/python/src/fern_python/generators/core_utilities/core_utilities.py
+++ b/generators/python/src/fern_python/generators/core_utilities/core_utilities.py
@@ -259,14 +259,15 @@ class CoreUtilities:
         )
 
     def get_universal_root_model(self) -> AST.ClassReference:
-        if self._pydantic_compatibility == PydanticVersionCompatibility.Both:
+        if (
+            self._pydantic_compatibility == PydanticVersionCompatibility.Both
+            or self._pydantic_compatibility == PydanticVersionCompatibility.V2
+        ):
             return AST.ClassReference(
                 qualified_name_excluding_import=(),
                 import_=AST.ReferenceImport(
                     module=AST.Module.local(*self._module_path, "pydantic_utilities"), named_import="UniversalRootModel"
                 ),
             )
-        elif self._pydantic_compatibility == PydanticVersionCompatibility.V2:
-            return Pydantic(self._pydantic_compatibility).RootModel()
         else:  # V1 or V1_ON_V2
             return Pydantic(self._pydantic_compatibility).BaseModel()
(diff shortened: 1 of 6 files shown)


generators/python/src/fern_python/generators/core_utilities/core_utilities.py
```diff
@@ -259,14 +259,15 @@ def universal_field_validator(self, field_name: str, pre: bool = False) -> AST.F
...
             return AST.ClassReference(
                 qualified_name_excluding_import=(),
                 import_=AST.ReferenceImport(
                     module=AST.Module.local(*self._module_path, "pydantic_utilities"), named_import="UniversalRootModel"
                 ),
             )
-        elif self._pydantic_compatibility == PydanticVersionCompatibility.V2:
-            return Pydantic(self._pydantic_compatibility).RootModel()
         else:  # V1 or V1_ON_V2
             return Pydantic(self._pydantic_compatibility).BaseModel()
```
amckinney, line 273: Discussed offline - the original implementation actually looks correct. My understanding of this is:
1. If you want to support both v1 and v2, we use our special `UniversalRootModel`.
2. If you only care about v2, we use `RootModel` (which superseded `BaseModel`).
3. If you only care about v1, we use `BaseModel`.

It sounds like the issue is actually that the `RootModel` inheritance is busted w.r.t. `mypy`, and we're using this setting to generate the IR. So instead of changing this implementation (and potentially breaking users), we should:

1. Try to move the Python IR package to `both`.
2. If (1) fails, fix the `RootModel` type hinting to make `mypy` happy.

PR #7160, opened 2025-05-14
chore(cli): check if headers in params are duplicates of authSchemes before adding

sahil485:
## Description
- don't add headers from endpoint params if they are duplicates of auth schemes   
## Changes Made
- check authSchemes before adding headers in ParameterConverter

## Testing

### DIFF
--- a/packages/cli/api-importers/openapi-to-ir/src/3.1/paths/operations/AbstractOperationConverter.ts
+++ b/packages/cli/api-importers/openapi-to-ir/src/3.1/paths/operations/AbstractOperationConverter.ts
@@ -124,6 +124,7 @@ export abstract class AbstractOperationConverter extends AbstractConverter<
                         break;
                     case "header": {
                         const headerName = convertedParameter.parameter.name.name.originalName;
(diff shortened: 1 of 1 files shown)


packages/cli/api-importers/openapi-to-ir/src/3.1/paths/operations/AbstractOperationConverter.ts
```diff
@@ -123,7 +123,25 @@ export abstract class AbstractOperationConverter extends AbstractConverter<
...
+                        let duplicateHeader = false;
+                        const authSchemes = this.context.authOverrides?.["auth-schemes"];
+                        if (authSchemes != null) {
+                            const headerWireValue = convertedParameter.parameter.name.wireValue;
+                            for (const authScheme of Object.values(authSchemes)) {
+                                if (
+                                    authScheme &&
+                                    typeof authScheme === "object" &&
+                                    "header" in authScheme &&
+                                    authScheme.header === headerWireValue
```
dsinghvi, line 136: i think this check should also be lower cased

eyw520 (approved):

PR #7172, opened 2025-05-15
fix(cli): Convert schemas of type `string` with format `date` to a Fern `date` type, instead of a `string` type

Swimburger:
## Description
Convert schemas of type `string` with format `date` to a Fern `date` type, instead of a `string` type

## Testing


dsinghvi (changes requested):
This could result in breaking changes in the SDK, we'll need to put this behind a feature flag

dsinghvi (approved):

PR #7183, opened 2025-05-15
feat(cli): V3 parser date string to date

Swimburger:
## Description
Make V3 parser respect the `type-dates-as-strings` flag.

## Changes Made
- Make V3 parser respect the `type-dates-as-strings` flag.
- Convert the parser options sooner in the program to reduce the amount of `| undefined`, `?:`, and `Partial`.
- 

## Testing


packages/cli/api-importers/openapi-to-ir/src/3.1/__test__/OpenAPIConverterContext3_1.test.ts
```diff
@@ -3,6 +3,7 @@ import { describe, expect, it } from "vitest";
 
 import { ErrorCollector } from "@fern-api/v2-importer-commons";
 
+import { getOpenAPISettings } from "../../../../../workspace/commons/src";
```
eyw520: Could we use a module import here vs a relative path import? If not, no worries.

eyw520 (approved):

packages/cli/api-importers/openapi-to-ir/src/3.1/__test__/OpenAPIConverterContext3_1.test.ts
Swimburger: Good call

PR #7195, opened 2025-05-16
fix(csharp): Mock the OAuth endpoint for the wire tests.

Swimburger:
## Description
Mock the OAuth endpoint for the wire tests.

## Testing


generators/csharp/sdk/src/test-generation/mock-server/BaseMockServerTestGenerator.ts
```diff
@@ -183,6 +205,47 @@ export class BaseMockServerTestGenerator extends FileGenerator<CSharpFile, SdkCu
...
+                if (typeof value === "string") {
+                    value = value.toLowerCase();
+                    if (value === "client_id") {
+                        jsonExample[prop] = "CLIENT_ID";
+                    }
+                    if (value === "client_secret") {
+                        jsonExample[prop] = "CLIENT_SECRET";
+                    }
+                }
+            }
```
amckinney, line 239: Does this work work if the user defines this API with `camelCase`? This feels too specific / prone to fail. What are your thoughts?

Swimburger: It'd be better if I can get the property name from IR, then use that to update the values in the examples.

Swimburger: I updated the code to pull the path from IR and deep set it into the example (if the property is present in the example).

PR #7211, opened 2025-05-19
fix(cli): Support reference types in AsyncAPI request params

amckinney:
This updates the AsyncAPI v2 parser to reference schemas in headers and query parameters rather than creating a duplicate type. This prevents a variety of naming collision issues that can arise.


dsinghvi (approved):
nit: we should follow up with some test cases so that we can commit the snapshots

Issue #6435, opened 2025-03-13
Document Tooltips for CodeBlocks


dannysheridan:
### What's the issue?

Missing content (not covered in the docs)

### Describe the gap or issue

https://tooltip-test.ferndocs.app/home

We’ve added in the app router deployment the ability to pass in tooltips. This isn't present in our docs at buildwithfern.com/learn.

Here's an example that a user would author:
```mdx
<Template
	data={{ API_KEY: "123456" }}
	tooltips={{
		API_KEY: (
			<p>
				This is a tooltip, you can use jsx here.
			</p>
		)
	}}
>
```

And the doc would render:
```ts
console.log(`{{API_KEY}}`);
```

</Template>

### Actions

1. Decide where this should go in our information architecture. Either a new page or a new section within an existing page.
2. Explain the feature, show an example, and show the markdown that underlies the example.
3. Tag @dannysheridan to review a PR when ready.

dannysheridan:
We’ve added in the app router deployment the ability to pass in tooltips.  

`````tsx
<Template
	data={{ API_KEY: "123456" }}
	tooltips={{
		API_KEY: (
			<p>
				This is a tooltip, you can use jsx here.
			</p>
		)
	}}
>

```ts
console.log(`{{API_KEY}}`);
```

</Template>
`````

PR #7215, opened 2025-05-19
feat(docs): Adding Document Tooltips for CodeBlocks

Ryan-Amirthan:
## Description
We’ve added in the app router deployment the ability to pass in tooltips. This isn't present in our docs at buildwithfern.com/learn. See [issue here](https://github.com/fern-api/fern/issues/6435) for more.

## Changes Made
- Created `tooltips.mdx` under Writing Content --> Components and added to `docs.yml` for navigation.
- Explained tooltips feature, showed an example as well as the markdown that underlies the example.

## Testing
I ran fern docs dev locally to check if my changes read well.

### DIFF
--- a/fern/pages/docs/components/tooltips.mdx
+++ b/fern/pages/docs/components/tooltips.mdx
@@ -105,7 +105,7 @@ const api = axios.create({
 		API_KEY: (<p>
 		Your API key is used to authenticate requests to our API. Keep this secure and never expose it in client-side code.
 		<br /><br />
-		You can find your API keys in the <a href="https://dashboard.example.com/settings/api-keys">Dashboard</a>.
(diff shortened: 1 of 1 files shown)


dannysheridan:
Feedback
1. Can we show a Basic Example and an Advanced Example? I want to see 2 or 3 Tooltips in a longer codeblock.
2. The description for `data` doesn't read clearly to me. I am confused as what {{KEY}} means here. Should I think of data as a string instead of an object? Is this what's visible to viewers vs tooltip is what's hidden and only available on hover?
3. We shouldn't saw `jsx` in the tooltip description. That feels more internal slang (relevant to us) than easily understood to viewers.

Up to take another revision here?

dannysheridan (approved):

PR #7217, opened 2025-05-19
feat(go): Release 1.0.0

amckinney:
This releases `1.0.0` of the Go SDK generator, and updates the default set of configuration flags. The generator is equivalent to the behavior of the previous minor version with the following configuration:

```yaml
groups:
  go-sdk:
    - name: fernapi/fern-go-sdk
      config:
        alwaysSendRequiredProperties: true
        inlineFileProperties: true
        inlinePathParameters: true
        useReaderForBytesRequest: true
        union: v1
```

The following set of configuration flags can be used to preserve behavior from the default behavior in the `v0.X.X` range:

```yaml
groups:
  go-sdk:
    - name: fernapi/fern-go-sdk
      config:
        alwaysSendRequiredProperties: false
        inlineFileProperties: false
        inlinePathParameters: false
        useReaderForBytesRequest: false
        union: v0
```

### DIFF
--- a/fern/pages/changelogs/go-sdk/2025-05-19.mdx
+++ b/fern/pages/changelogs/go-sdk/2025-05-19.mdx
@@ -1,4 +1,0 @@
-## 1.0.0
-**`(feat):`** Major version release to update the generator's default set of configuration options.
-
-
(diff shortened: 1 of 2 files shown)


generators/go/sdk/versions.yml
```diff
@@ -1,3 +1,10 @@
+- version: 1.0.0
+  changelogEntry:
+    - summary: |
+        Major version release to update the generator's default set of configuration options.
```
Swimburger, line 4: I would add the old and new options in the changelog so people can find it easily

Swimburger (approved):

generators/go/sdk/versions.yml
amckinney: Good call, added!

PR #7234, opened 2025-05-20
feat(cli): Report errors from `fern diff` and add --quiet flag

amckinney:
This updates the `fern diff` and `fern ir` commands with the following:

1. Add `--quiet` (-q) flag to `fern diff` command to suppress output written to stderr.
2. Add `--disable-examples` flag to `fern ir` command to disable automatic example generation.
3. Fix an issue where the `fern diff` command mistakenly reported breaking changes for API types that defined multiple errors with the same status code.
4. Fix an issue where the `fern diff` command mistakenly reported breaking changes for APIs that contain literal type references.


mattblank11 (approved):
🚀

PR #7248, opened 2025-05-21
feat(cli): add naming for discriminated unions based on title or discriminator key

sahil485:
## Description
- add naming for discriminated unions based on title or discriminator key

**Order of priority**
  1. top level titles
  2. schema-level titles
  3. discriminator key
## Changes Made
- check whether top-level title exists for reference objects
- use discriminator key when it doesn't exist
- add displayNameOverride to `convertTypeToTypeReference` method

## Testing


eyw520 (approved):
+1'd, but let's clean up the impl for the `convertReferenceToTypeReference` method.

PR #7257, opened 2025-05-21
feat(cli): Add support for HEAD method

amckinney:
This adds support for recognizing `HEAD` HTTP methods, which act like `GET` requests without a response body. This is useful for both verifying that some content exists and understanding how large the content is (e.g. via the `Content-Length` header), without actually incurring the cost to fetch it.

Fern SDKs will simply return the language-native HTTP header type in each of these endpoints (e.g. `http.Header` for Go).


dsinghvi:
@amckinney can you make sure to add it to the new parser as well https://github.com/fern-api/fern/blob/daef060a591185ae0de4855e53bde0b5a67ad266/packages/cli/api-importers/openapi-to-ir/src/3.1/paths/operations/AbstractOperationConverter.ts#L70-L84

Swimburger (approved):

PR #7282, opened 2025-05-22
feat(cli): Support rendering and parsing multiple endpoint baseUrls.

eyw520:
## Description
This PR updates the IR as well as endpoint-level server parsing to propagate endpoint-level servers to through to the FDR correctly.

## Changes Made
- Update logic with how endpoint level servers are collected.
- Add `v2BaseUrls` field to endpoints, which are parsed into multiple environments at the FDR level.
- Ensure `v2BaseUrls` are correctly remapped during `mergeIntermediateRepresentation`.
- Update how environment Ids are defined during FDR conversion.

## Testing

### DIFF
--- a/packages/cli/ete-tests/src/tests/dependencies/__snapshots__/dependencies.test.ts.snap
+++ b/packages/cli/ete-tests/src/tests/dependencies/__snapshots__/dependencies.test.ts.snap
@@ -482,6 +482,7 @@ exports[`dependencies > correctly incorporates dependencies 1`] = `
                     "auth": false,
                     "idempotent": false,
                     "baseUrl": null,
+                    "v2BaseUrls": null,
                     "method": "GET",
                     "basePath": null,
                     "path": {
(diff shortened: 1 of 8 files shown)


packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts
```diff
@@ -295,4 +304,20 @@ export class OpenAPIConverter extends AbstractSpecConverter<OpenAPIConverterCont
...
+        currentServers: OpenAPIV3_1.ServerObject[];
+    }): boolean {
+        return !currentServers.some(
+            (s) =>
+                s.url === server.url &&
+                "x-fern-server-name" in s &&
+                "x-fern-server-name" in server &&
+                s["x-fern-server-name"] === server["x-fern-server-name"]
+        );
+    }
```
dsinghvi, line 321: nit: can we put this in the `AbstractSpecConverter` or `context` since i think openrpc will make use of the same logic

PR #7284, opened 2025-05-22
feat(typescript): Reimplement wire tests without depending on an external mock server

Swimburger:
## Description
* Instead of relying on an external mock server, we use MSW which uses a service worker to intercept the HTTP request and return the HTTP response.
* Each test sets up an expected HTTP request that if not matched, will throw, if matched, will return the expected HTTP response.
  The parsed HTTP response is then compared against the expected parsed response.
* We are a little more lax with BigInts and with date times

A lot of small issues were found in the IR and SDK generation code which are patched along the way.

Currently, there are the following gaps/defects in the wire tests:
* No oauth
* No request bytes or fileUpload
* No response fileDownload, text, bytes, streaming, streamParameter
* No idempotent endpoint
* No pagination endpoint
* Unions with base properties generate correct JSON example in IR but base properties are not recorded in example IR, so generated examples (not generated from JSON example), are missing base properties.
* No global variables 
* ~Bug in example IR where URL that includes a path parameter is not URL encoded:
/path/>/red should be /path/%3E/red ([PR in progress](https://github.com/fern-api/fern/pull/7343))~

## Testing


generators/typescript/asIs/tests/bigint-setup.ts
```diff
@@ -0,0 +1,13 @@
+import { expect } from "@jest/globals";
```
amckinney: If we already use the `json.bigint.ts` filename convention, should we say this is `bigint.setup.ts` so that it's clear it augments the `setup.ts` file?

generators/typescript/asIs/tests/mock-server/MockServerPool.ts
```diff
@@ -0,0 +1,107 @@
+import { setupServer } from "msw/node";
+
+import { fromJson, toJson } from "../../src/core/json";
+import { MockServer } from "./MockServer";
+import { randomBaseUrl } from "./randomBaseUrl";
+
+const mswServer = setupServer();
```
amckinney, line 7: It looks like we only ever use this within the context of the `MockServerPool`. Why not just define this as an instance variable so that we don't rely on any global state?

generators/typescript/asIs/tests/mock-server/mockEndpointBuilder.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+export interface HttpHandlerBuilderOptions {
+    onBuild?: (handler: HttpHandler) => void;
+    once?: boolean;
+}
+
+class RequestBuilder implements MethodStage, RequestHeadersStage, RequestBodyStage, ResponseStage {
+    private method: HttpMethod = "get";
+    #baseUrl: string = "";
```
amckinney: Is this purely just to avoid the collision between the `baseUrl` method and the private variable? I normally don't see the two syntaxes intermixed like this, so I'm wondering we name them or use a `_` suffix to disambiguate. If you can't think of a better alternative, I'm down to leave as-is but figured I'd flag it since it stuck out.

generators/typescript/asIs/tests/mock-server/withHeaders.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        // Track mismatched headers
+        const mismatches: Record<
+            string,
+            { actual: string | null; expected: string | RegExp | ((value: string) => boolean) }
+        > = {};
+
+        // Check each expected header
+        for (const [key, expectedValue] of Object.entries(expectedHeaders)) {
+            // Get the header value (case-sensitive)
```
amckinney: I think we should remove a lot of these comments since they more-or-less explain the obvious. Given that this is checked into each SDK, I'd rather only include comments where totally necessary (like the function signature comment you included at the top).

generators/typescript/asIs/tests/mock-server/withJson.ts
```diff
@@ -0,0 +1,146 @@
...
+        let actualBody: unknown;
+        try {
+            clonedRequest = request.clone();
+            actualBody = fromJson(await clonedRequest.text());
+        } catch (error) {
+            console.error("Error processing request body:", error);
+            return passthrough();
+        }
+
+        const mismatches = findMismatches(actualBody, expectedBody);
```
amckinney, line 24: The `findMatches` implementation is entirely custom here, right? 

I wonder if we should instead largely reference `lodash`'s `isEqual` function and include a comment that the implementation is directly inspired by `lodash`. That way, we're far less likely to have any weird edge cases here and can more easily keep the code in-sync if the source of truth is ever updated.

Thoughts?

generators/typescript/asIs/tests/mock-server/mockEndpointBuilder.ts
```diff
@@ -0,0 +1,209 @@
...
+    constructor(options?: HttpHandlerBuilderOptions) {
+        this.handlerOptions = options;
+    }
+
+    baseUrl(baseUrl: string): MethodStage {
+        this.#baseUrl = baseUrl;
+        return this;
+    }
+
+    all(path: string): RequestHeadersStage {
```
amckinney, line 71: When is this required?

seed/ts-sdk/examples/examples-with-api-reference/tests/wire/health/service.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        const server = mockServerPool.createServer();
+        const client = new SeedExamplesClient({ token: "test", environment: server.baseUrl });
+
+        const rawResponseBody = true;
+        server.mockEndpoint().get("/ping").respondWith().statusCode(200).jsonBody(rawResponseBody).build();
+
+        const response = await client.health.service.ping();
+        expect(response).toEqual(true);
+    });
+});
```
amckinney, line 29: Nice, these look great.

seed/ts-sdk/exhaustive/allow-extra-fields/tests/wire/endpoints/object.test.ts
```diff
@@ -0,0 +1,450 @@
...
+            datetime: "2024-01-15T09:30:00Z",
+            date: "2023-01-15",
+            uuid: "d5e9c84f-c2b2-4bf4-b4b0-7ffd7a9ffc32",
+            base64: "SGVsbG8gd29ybGQh",
+            list: ["list", "list"],
+            set: ["set"],
+            map: { "1": "map" },
+            bigint: "1000000",
+        };
+        const rawResponseBody = {
```
amckinney, line 27: Why not just use the `jsonExample` value from the IR (represented as a `string`)? That would prevent us from needing to compute the concrete JSON object ourselves. It might also reduce the amount of noise in each of these test cases if we can just reference the JSON string in-line.

```typescript
        const server = mockServerPool.createServer();
        const client = new SeedExhaustiveClient({ token: "test", environment: server.baseUrl });
        server
            .mockEndpoint()
            .post("/object/get-and-return-with-optional-field")
            .jsonBody("{ ... }")
            .respondWith()
            .statusCode(200)
            .jsonBody("{ ... }")
            .build();
        const response = await client.endpoints.object.getAndReturnWithOptionalField({
            string: "string",
            integer: 1,
            long: 1000000,
            double: 1.1,
            bool: true,
            datetime: "2024-01-15T09:30:00Z",
            date: "2023-01-15",
            uuid: "d5e9c84f-c2b2-4bf4-b4b0-7ffd7a9ffc32",
            base64: "SGVsbG8gd29ybGQh",
            list: ["list", "list"],
            set: ["set"],
            map: {
                1: "map",
            },
            bigint: "1000000",
        });
        expect(response).toEqual({
            string: "string",
            integer: 1,
            long: 1000000,
            double: 1.1,
            bool: true,
            datetime: "2024-01-15T09:30:00Z",
            date: "2023-01-15",
            uuid: "d5e9c84f-c2b2-4bf4-b4b0-7ffd7a9ffc32",
            base64: "SGVsbG8gd29ybGQh",
            list: ["list", "list"],
            set: ["set"],
            map: {
                1: "map",
            },
            bigint: "1000000",
        });
```

It's still somewhat noisy with these large objects, but I can see why you're re-defining the object (a simple JSON string comparison won't work due to property ordering).

What matters most is that this is correct; we can change the implementation later as we go but it's important that we think about the cosmetics here.

amckinney (commented):
Primarily removed the as-is files and generated output; a few things to consider, but will do a separate pass over the generator implementation.

generators/typescript/asIs/tests/bigint-setup.ts
Swimburger: `json.bigint.ts` is written to the TS SDK as `json.ts`.
That's fine tho

generators/typescript/asIs/tests/mock-server/MockServerPool.ts
Swimburger: Either way that would be considered global state, whether it's at the module level, or module.mockServerPool level.
I think it's more clear as a singleton on the module level.
Hopefully, MSW will support external mockserver mode soon and then we can create multiple instances of this.

generators/typescript/asIs/tests/mock-server/mockEndpointBuilder.ts
Swimburger: correct

generators/typescript/asIs/tests/mock-server/mockEndpointBuilder.ts
Swimburger: We don't use it, I'm just mirroring the available methods on MSW, but we can reduce it to the ones we use ourselves.
However, if a SDK author (or us) wants to use this for handwritten tests, it could come in handy.

generators/typescript/asIs/tests/mock-server/withJson.ts
Swimburger: I'm down to adopt a library to help, but we should provide detailed errors when it doesn't match. It should tell us where the diff is, and it should allow us to customize the function with custom matchers.
isEqual allows us to add a custom matcher, but it doesn't give us any information why something was not equal.

seed/ts-sdk/exhaustive/allow-extra-fields/tests/wire/endpoints/object.test.ts
Swimburger: In JS we can't use the Raw JSON because of bigint.

generators/typescript/asIs/tests/mock-server/MockServerPool.ts
amckinney: Ah I didn't see the single `export` at the bottom - I was thinking we might construct multiple instances of the `class`. In that case either way is fine. I still like the idea of this being an instance variable, but you're right that it doesn't really matter.

generators/typescript/asIs/tests/mock-server/withJson.ts
amckinney: Understood - as long as we're confident this is _correct_. I imagine this could be a code smell to future SDK readers, but these are implementation details that we can change later on (if ever).

amckinney (approved):

PR #7362, opened 2025-05-29
fix(cli): Correctly parse `allOf` required properties.

eyw520:
## Description
`allOf` schemas that share properties with the parent schema's required properties will now be parsed as an inline `allOf` schema.

## Changes Made
When parsing object schemas, we now perform the following when the object contains `allOf` subschemas:
Case A: If an allOfSchema is a SchemaObject, we convert its properties per usual.
Case B: If an allOfSchema is a Reference, we will check the following:
    If there are no parent schema required properties, we will add it to the `extends` TypeReference array.
    We resolve the allOfSchema Reference. Then:
        If there are no overlapping properties with the parent schema's required properties, we add it to the `extends` TypeReference array.
        If there are overlapping properties with the parent schema's required properties, we will convert its properties as an inlined allOf schema.

## Testing


fern-support (approved):
🚀

PR #7370, opened 2025-05-31
fix(cli): Fix inline request body with file list parsing

Swimburger:
## Description
Implicitly interpert a request body as a multipart-form if it contains a property with a file, list of files, and optional file.

## Testing


packages/cli/fern-definition/schema/src/utils/parseFileUploadRequest.ts
```diff
@@ -119,3 +119,35 @@ function createRawFileUploadRequest(
...
+                return true;
+            }
+        } else {
+            if (isFileInGeneric(arg)) {
+                return true;
+            }
+        }
+    }
+    return false;
+}
```
amckinney: optional/nit: In case we ever add special handling for other keywords, it'd be good to define a function that acts upon arbitrary keywords (that we don't accidentally copy around more code between them.

Something like the following:

```typescript
function isFileInGeneric(input: ParseGenericNested): boolean {
  return isTypeInGeneric({ input, typeName: "file" });
}

function isTypeInGeneric({ input, typeName }: { input: ParseGenericNested; typeName: string }): boolean {
    for (const arg of input.arguments) {
        if (typeof arg === "string") {
            if (arg === typeName) {
                return true;
            }
        } else {
            if (isTypeInGeneric({ input: arg, typeName })) {
                return true;
            }
        }
    }
    return false;
}
```

packages/cli/fern-definition/schema/src/utils/generics/parseGenericNested.ts
```diff
@@ -0,0 +1,81 @@
...
+        while (index < input.length && /\s/.test(input[index]!)) {
+            index++;
+        }
+
+        // Check if there are generic arguments
+        if (index < input.length && input[index] === "<") {
+            index++; // consume '<'
+            depth++;
+
+            if (depth > 100) {
```
amckinney: If we really need this, we should try to define these recursive depth limits somewhere so that it's easy to spot and edit later on. We should start by making this number consistent to the depth number used in the example generation algorithm.

packages/cli/fern-definition/schema/src/utils/generics/parseGenericNested.ts
```diff
@@ -0,0 +1,81 @@
+export interface ParseGenericNested {
+    name: string;
+    arguments: (string | ParseGenericNested)[];
+}
+
+export function parseGenericNested(input: string): ParseGenericNested | undefined {
```
amckinney, line 7: Discussed offline - this function initially appears to be somewhat of a code smell (i.e. it'd be great if we could act upon more strongly typed structures rather than the free-form string at this stage), but it's actually necessary given the call-site of `parseFileUploadRequest`.

Let's make sure this is tested well with some test definitions so that we build confidence that it works (and is resistant to regressions).

amckinney (approved):
LGTM assuming this is tested well and resolves the issue at hand!

PR #7376, opened 2025-06-02
chore(cli): add fallback when pnpm fails to install

chdeskur:
improve fallback logic if any part of local dev fails to install


packages/cli/docs-preview/src/downloadLocalDocsBundle.ts
```diff
@@ -184,18 +184,66 @@ export async function downloadBundle({
...
-        });
+        // if pnpm still hasn't been installed, user should install themselves
+        try {
+            await loggingExeca(logger, process.platform === "win32" ? "where" : "which", ["pnpm"], {
+                cwd: absolutePathToBundleFolder,
+                doNotPipeOutput: true
+            });
+        } catch (error) {
+            logger.error(
+                "Requires [pnpm] to run local development. Please install and try again: https://pnpm.io/installation"
```
dsinghvi: Please run `npm install -g pnpm and then fern docs dev`

dsinghvi (approved):

PR #7378, opened 2025-06-03
feat(typescript): Add support for HTTP HEAD method

Swimburger:
## Description
Add support for HTTP HEAD method

## Changes Made
- Add support for HTTP HEAD method
- Content type for the request will fallback to `undefined`, unless the body type is inline body or reference, in which case it will fallback to `application/json`.
- For a default request (not upload, bytes, etc.), the request type will be `json` if inline body or reference, otherwise `undefined`. 

## Testing


seed/ts-sdk/accept-header/src/api/resources/service/client/Client.ts
```diff
@@ -60,8 +60,6 @@ export class Service {
                 "X-Fern-Runtime-Version": core.RUNTIME.version,
                 ...requestOptions?.headers,
             },
-            contentType: "application/json",
-            requestType: "json",
```
dsinghvi, line 64: @Swimburger this feels like a bug

Swimburger: This was a bug is fixed here. There is no request body in these endpoints, so the request content type and type should be omitted.

amckinney (approved):

amckinney (commented):
Assuming the Content-Type and requestType changes are functional, this lgtm

PR #7381, opened 2025-06-03
chore(cli): local dev improvements

chdeskur:
publishes a new version of the CLI to: 
- download the bundle using `tar.gz` format 
- log better errors when the bundle fails to install
- include more resilliant fallback logic
- kill more processes

### DIFF
--- a/packages/cli/docs-preview/src/runAppPreviewServer.ts
+++ b/packages/cli/docs-preview/src/runAppPreviewServer.ts
@@ -188,19 +188,26 @@ export async function runAppPreviewServer({
         }
     };
 
-    // Handle all possible termination scenarios
-    const shutdownSignals = ["SIGTERM", "SIGINT", "SIGQUIT"];
-    const failureSignals = ["SIGKILL", "SIGHUP"];
-    for (const shutSig in shutdownSignals) {
-        context.logger.debug("Shutting down server...");
-        process.on(shutSig, cleanup);
+    // handle termination signals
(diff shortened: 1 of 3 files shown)


packages/cli/docs-preview/src/runAppPreviewServer.ts
```diff
@@ -168,18 +168,60 @@ export async function runAppPreviewServer({
...
+    const failureSignals = ["SIGHUP"];
+    for (const shutSig of shutdownSignals) {
+        process.on(shutSig, () => {
+            context.logger.debug("Shutting down server...");
+            cleanup();
+        });
+    }
+    for (const failSig of failureSignals) {
+        process.on(failSig, () => {
+            context.logger.error("Server failed, shutting down process...");
```
dsinghvi: this isn't that useful to a user, lets make it a debug log

packages/cli/docs-preview/src/runAppPreviewServer.ts
```diff
@@ -168,18 +168,60 @@ export async function runAppPreviewServer({
...
+            cleanup();
+        });
+    }
+
+    // handle normal exit
     process.on("exit", cleanup);
 
+    // handle uncaught exits
+    process.on("uncaughtException", (err) => {
+        context.logger.error(`Uncaught exception: ${err}`);
```
dsinghvi: what does the message here end up looking like? also not that useful to a user or `${err}` could be pretty ugly

dsinghvi (approved):
couple comments on logging

PR #7395, opened 2025-06-03
chore(deps): Allow users to specify SDK output path in generators.yml

mattblank11:
## Description
Update the `generators.yml` file to let users specify the output path for the SDK + test files. The readme + all other root files will continue being generated in the root folder.

## Changes Made
Updated `generators.yml` file

### DIFF
--- a/fern/apis/generators-yml/definition/group.yml
+++ b/fern/apis/generators-yml/definition/group.yml
@@ -74,6 +74,7 @@ types:
     properties: 
       uri: string
       token: string
+      path: optional<string>
 
   GeneratorSnippetsSchema:
     properties:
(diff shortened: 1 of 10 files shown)


fern/apis/generators-yml/definition/group.yml
```diff
@@ -94,6 +94,7 @@ types:
       repository: string
       license: optional<license.GithubLicenseSchema>
       mode: optional<GithubCommitAndReleaseMode>
+      path: optional<string>
```
dsinghvi, line 98: can we add this to all the `GitHubConfigurationSchema`

```yml
  GithubConfigurationSchema:
    discriminated: false
    union:
      - GithubSelfhostedSchema
      - GithubCommitAndReleaseSchema
      - GithubPullRequestSchema
      - GithubPushSchema
```

mattblank11: Ah yep, missed that - updating!

mattblank11: Updated by adding `path` to each of those schemas

PR #7419, opened 2025-06-05
chore(typescript): Reduce duplicate code generation by passing headers from the root client down

Swimburger:
## Changes Made
- Reduce duplicate code generation by passing headers from the root client down to the subpackage clients.
- You can now pass in headers to the root client. These headers will be merged with service and endpoint-specific headers.

## Testing


dsinghvi (approved):
The code looks good tome!

PR #7432, opened 2025-06-05
feat(csharp): Add support for HTTP HEAD method

Swimburger:
## Description
Add support for HTTP HEAD method

## Testing


dsinghvi (approved):
This works -- typically for HEAD requests a user wants access to the response headers so i know we created a default return type in go

amckinney (approved):
Can we adjust the title / changelog so that it more accurately considers this a bug fix? It doesn't actually implement the `HEAD` endpoints correctly as-is (but I know that's planned in a follow-up).

amckinney (approved):

PR #7454, opened 2025-06-06
fix(express): prefix res.writableEnded check with '!'

sahil485:
## Description
prefix res.writableEnded check with '!'

## Testing


dsinghvi (approved):
missng versions.yml but we are g2g otherwise!

PR #7461, opened 2025-06-06
feat(java): Add support for HEAD methods

amckinney:
This adds support for HTTP `HEAD` methods by adapting the non-raw client to surface the OkHttp response headers as a `Map<string, List<string>>`, which matches the interface already used by the generated `HttpResponse` class.

### DIFF
--- a/seed/java-sdk/http-head/src/main/java/com/seed/httpHead/resources/user/AsyncRawUserClient.java
+++ b/seed/java-sdk/http-head/src/main/java/com/seed/httpHead/resources/user/AsyncRawUserClient.java
@@ -56,10 +56,7 @@ public class AsyncRawUserClient {
             @Override
             public void onResponse(@NotNull Call call, @NotNull Response response) throws IOException {
                 try (ResponseBody responseBody = response.body()) {
-                    if (response.isSuccessful()) {
-                        future.complete(new SeedHttpHeadHttpResponse<>(null, response));
-                        return;
-                    }
+                    if (response.isSuccessful()) {}
                     String responseBodyString = responseBody != null ? responseBody.string() : "{}";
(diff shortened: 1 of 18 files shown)


seed/java-sdk/http-head/src/main/java/com/seed/httpHead/resources/user/AsyncRawUserClient.java
```diff
@@ -56,10 +56,7 @@ public CompletableFuture<SeedHttpHeadHttpResponse<Void>> head(RequestOptions req
             @Override
             public void onResponse(@NotNull Call call, @NotNull Response response) throws IOException {
                 try (ResponseBody responseBody = response.body()) {
-                    if (response.isSuccessful()) {
-                        future.complete(new SeedHttpHeadHttpResponse<>(null, response));
-                        return;
-                    }
+                    if (response.isSuccessful()) {}
```
amckinney, line 59: This isn't right - we'll need to retain the original functionality in the raw client.

seed/java-sdk/http-head/src/main/java/com/seed/httpHead/resources/user/RawUserClient.java
```diff
@@ -49,9 +50,7 @@ public SeedHttpHeadHttpResponse<Void> head(RequestOptions requestOptions) {
         }
         try (Response response = client.newCall(okhttpRequest).execute()) {
             ResponseBody responseBody = response.body();
-            if (response.isSuccessful()) {
-                return new SeedHttpHeadHttpResponse<>(null, response);
-            }
+            if (response.isSuccessful()) {}
```
amckinney, line 53: Same thing here - we should still be returning `SeedHeadHttpResponse<Void>` here.

PR #7483, opened 2025-06-11
fix(cli): turn off posthog for self hosted CLI

ariel-langer:
## Description
For self-hosted docs we call the `generate --docs` command and don't want to make any outbound calls to 3rd parties. I turned off the wifi on my computer and ran our self hosted docker container and found that we were still trying to make this request to posthog from our CLI so i'm turning it off.

Screenshot 2025-06-11 at 10 11 22 AM

## Changes Made
Turn off posthog for the `generate --docs` command when we are running the CLI from the self hosted docker container. 

## Testing

Didn't test because this is pretty straightforward.


ariel-langer (commented):
@chdeskur can you review?

chdeskur (approved):

PR #7486, opened 2025-06-11
feat(typescript): Let SDK author choose streamResponseType to be wrapper or web

Swimburger:
## Changes Made
- You can now specify whether to return streams using the stream wrapper, or return the web standard stream.
        Change the type of stream returned by setting `streamResponseType` to `wrapper` or `web` in the `config` of your generator configuration.
        The default is `wrapper`.
- Core utilities have been simplified. They are now all in one folder, they match the file structure of how it would be in the output. Due to templating requirement, this is not a compilable/testable project anymore. To compile/test core utilities, iterate in a generated SDK and bring it back to the source.
- Any files with name matching `*.template.*` will be compiled against lodash's templating.

## Testing


dsinghvi (commented):
Screenshot 2025-06-16 at 10 24 32 AM

@Swimburger any idea why there is no readme generated in the `web-stream-wrapper` fixture?

Swimburger:
> Screenshot 2025-06-16 at 10 24 32 AM
> @Swimburger any idea why there is no readme generated in the `web-stream-wrapper` fixture?

Found it, the features.yml file wasn't copied any longer in the Docker build step.

generators/typescript/utils/commons/src/core-utilities/CoreUtilitiesManager.ts
```diff
@@ -75,86 +82,45 @@ export class CoreUtilitiesManager {
...
-                    }
-                }
-                if (utility.writeConditionalFiles != null) {
-                    await utility.writeConditionalFiles(toPath);
-                }
-                if (utility.repoInfoForTesting.ignoreGlob != null && process.env.NODE_ENV === "test") {
-                    const filesToDelete = await glob(utility.repoInfoForTesting.ignoreGlob, {
-                        cwd: toPath,
-                        absolute: true
+        const pathOnContainer = "/assets/core-utilities";
```
amckinney: Can we define this somewhere more discoverable?

seed/ts-sdk/basic-auth-environment-variables/tests/unit/fetcher/getResponseBody.test.ts
```diff
@@ -1,5 +1,6 @@
 import { RUNTIME } from "../../../src/core/runtime";
 import { getResponseBody } from "../../../src/core/fetcher/getResponseBody";
+
```
amckinney, line 3: Why are these files changing? Are we running a formatter that we previously weren't?

seed/ts-sdk/circular-references/src/core/runtime/runtime.ts
```diff
@@ -11,6 +11,9 @@ interface BunGlobal {
 declare const Deno: DenoGlobal | undefined;
 declare const Bun: BunGlobal | undefined;
 declare const EdgeRuntime: string | undefined;
+declare const self: typeof globalThis.self & {
+    importScripts?: unknown;
+};
```
amckinney, line 16: Is this related to the current change?

amckinney (approved):
It looks like we're changing _all_ of the core utility management along with this change. As long as you're confident this maintains behavior with the previous generator, LGTM.

seed/ts-sdk/circular-references/src/core/runtime/runtime.ts
fern-support: This is an eslint improvement, not related

seed/ts-sdk/basic-auth-environment-variables/tests/unit/fetcher/getResponseBody.test.ts
fern-support: Because of the .template.ts execution

PR #7491, opened 2025-06-11
Sd/ruby/add object codegen

sohdas:
## Changes Made

- 
- 
- 

## Testing


generators/ruby-v2/ast/src/ast/Field.ts
```diff
@@ -0,0 +1,41 @@
...
+        name: string;
+        type: Type;
+        optional?: boolean;
+    }
+}
+
+export class Field extends AstNode {
+    public readonly name: string;
+    public readonly type: Type;
+    public readonly optional: boolean;
```
sohdas: TODO: Field ought to support an arbitrary number of KeywordParameters

PR #7493, opened 2025-06-11
fix(seed): delete yarn / npm files

ariel-langer:
## Description
I was trying to figure out why we have a dependency on `"cross-spawn": "^7.0.3"` and it shows up in the package-lock.json file and the yarn.lock file. i dont believe we need these files because we are using pnpm for dependency management so i've deleted them.

## Changes Made
Delete unused files. Now i don't see any references to cross-spawn 7.0.3 in either `fern` or `fern-platform`

Screenshot 2025-06-11 at 6 56 13 PM

Screenshot 2025-06-11 at 6 56 04 PM

## Testing
N/A


ariel-langer (commented):
@dsinghvi is it possible that we are using these files to install packages somewhere and `"cross-spawn": "^7.0.3"` is getting installed from either `package-lock.json` or `yarn.lock`?

seems unlikely, i did build this version of the CLI + install it in the docker to check and i still see a vulnerability for cross-spawn which it did not. the vulnerability for axios disappeared but i made a few changes to the Dockerfile that probably explain that

dsinghvi (approved):

ariel-langer (commented):
ehh im not actually sure if we can delete this yarn.lock file; seems like the package manager for this file (`fern/seed/ts-mcp/imdb/sdk/package.json`) is yarn?

`"packageManager": "yarn@1.22.22"`

Screenshot 2025-06-12 at 10 16 04 AM

dsinghvi (approved):

PR #7506, opened 2025-06-13
fix(java): Array query params are treated as exploded: true

jsklan:
## Description
A regression of the Java SDK generator occurred at some point where query params formatting for array type params were malformed. By default the generator treats array query params as if they have  openapi's explode property set to true. In future we may want to modify this behavior to distinguish between these two specifications, but for now we're simply fixing the bug where the params were malformed.

### Examples
Open API Spec for param:
```
parameters:
  - name: foo
    required: true
    in: query
    schema:
      type: array
      items:
        type: string
```
Previous formatting:
```
https://api.example.com/stuff/things?req=1&req=2
```
Malformed formatting url encoded:
```
https://api.example.com/stuff/things?req=%5B1%2C%202%5D
```
Malformed formatting unencoded:
```
https://api.example.com/stuff/things?req=[1, 2]
```
Potential future formatting if we don't treat array types as `explode: true` (per open API) or `allowMultiple: true` (per fern-ar):
```
https://api.example.com/stuff/things?req=1%2C2
```
```
https://api.example.com/stuff/things?req=1,2
```

## Changes Made

- Changed call to `addQueryParameter` to not stringify the param before sending it to the method
- Passed `queryParamProperty.allowMultiple()` to `addQueryParameter` method's `arraysAsRepeats` parameter instead of always passing `false`

## Testing

Tested manually locally by creating a very simple fern config and openapi spec, running the java generator against that with seed, and then running a manually created test within the generated java sdk to verify that the url string looked correct after adding the query params.


seed/java-sdk/trace/src/main/java/com/seed/trace/resources/playlist/AsyncRawPlaylistClient.java
```diff
@@ -134,18 +130,18 @@ public CompletableFuture<SeedTraceHttpResponse<List<Playlist>>> getPlaylists(
                 .addPathSegments("all");
         if (request.getLimit().isPresent()) {
             QueryStringMapper.addQueryParameter(
-                    httpUrl, "limit", request.getLimit().get().toString(), false);
+                    httpUrl, "limit", request.getLimit().get(), false);
```
fern-support, line 133: @dsinghvi Given that this change to not stringify args will impact types other than array, should we test this further before pushing it out?

dsinghvi (approved):

PR #7510, opened 2025-06-14
fix(java): Remove Content-Type from headers when request has no body

jsklan:
## Description
Previously we were including the following header on all requests without a body:
```
{ 'Content-Type': 'application/json'}
```
This causes some issues for some APIs since it doesn't respect their API constraints.
Root cause was that the java sdk generator was adding a content type header whenever specific response types were specified in the request.

## Changes Made
- Removed consideration of the response type in the sendContent evaluation
- Ensured all places the Content-Type header was added were conditional on the sendContent boolean
- Manual file changes are limited to this file list. Remainder of changes were generated.
  - [AbstractEndpointWriter.java](https://github.com/fern-api/fern/pull/7510/files#diff-0d81fed66aede95aaaf2161a425cbbc4ec0720ecabef35792159b8228d260954)
  - [OnlyRequestEndpointWriter.java](https://github.com/fern-api/fern/pull/7510/files#diff-4d7d9f3606b4ab8bae59e2798a2011cf6587de6800a1b9712f69162b505c058d)

## Testing
Ran seed test --fixture exhaustive --skip-scripts and reviewed the changes manually.


dsinghvi (approved):

jsklan:
@dsinghvi Looks like these changes caused seed snapshot tests to fail. But in this case, the changes are correct per the new changes. How do I mark that the new values are expected so that these tests pass again?

PR #7511, opened 2025-06-14
chore(cli): remove extraneous log

chdeskur:
clean up log


dsinghvi (approved):
@chdeskur can you add a versions.yml entry?

PR #7531, opened 2025-06-16
Fix/add openrpc docs kg

kgowru:
## Description
- added OpenRPC docs to API Definition and Docs home 

## Testing
- Manually tested


Ryan-Amirthan (approved):
LGTM

PR #7532, opened 2025-06-16
fix(cli): Update protoc-gen-fern settings

amckinney:
This fixes the `settings` field in `protoc-gen-fern`, which previously caused a compile failure.


Swimburger (approved):
I'll approve this but we should give it the real settings. If we leave this here it may cause hard to find bugs in the future.

PR #7533, opened 2025-06-16
feat(cli): add exploded form parts to OpenAPI/FD/IR

Swimburger:
## Description
This change wires through the `explodeParts` flag from OpenAPI to FD to IR.

## Testing


packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/extensions/fernExtensions.ts
```diff
@@ -469,7 +469,14 @@ export const FernOpenAPIExtension = {
...
      */
-    TRANSPORT: "x-fern-transport"
+    TRANSPORT: "x-fern-transport",
+
+    /**
+     * Allows users to specify that a property should be exploded into multiple multipart form parts.
+     * This is currently only supported on JSON arrays. Each item in the array will be
+     * converted into a separate part in the multipart form data.
+     */
+    EXPLODE_PARTS: "x-fern-explode-parts"
```
amckinney, line 479: If this is a gap in OpenAPI, what are your thoughts on still naming this consistent with the convention used in OpenAPI?

I imagine we could just say this is `x-fern-explode: {false,true}`? Parts are implied in "explode" format.

Swimburger:
We decided to not move forward. Just default to serializing an array of complex object to a JSON array instead of multiple JSON parts.

PR #7535, opened 2025-06-16
Implementation of the first layer of the SSE Implementation for the Generator to consume

gupta11:
## Description

Fern SDKs support a variety of powerful features like [auto-pagination](https://github.com/fern-demo/square-java-sdk?tab=readme-ov-file#automatic-pagination), [retries](https://github.com/fern-demo/square-java-sdk?tab=readme-ov-file#retries), and [streaming](https://github.com/cohere-ai/cohere-go?tab=readme-ov-file#streaming). Java supports newline-delimited JSON streaming, but we haven't built out server-side event (SSE) streaming yet. This PR is the first step in enabling the generator to accept SSE on Java

## Changes Made

- 
- Added secondary constructor to Stream.java that enables the use for SSE. 
- Implemented a parser for SSE
- Added StreamOptions that are required for SSE stream type.

## Testing


generators/java/generator-utils/src/main/resources/Stream.java
```diff
@@ -18,10 +26,25 @@ public final class Stream<T> implements Iterable<T> {
...
 
+    private Options options;
+
+    private StreamType streamType;
+
+    /**
+     * The type of stream to be interpreted.
+     */
+    public enum StreamType {
+        SSE, NEWLINE
```
amckinney: nit: NDJSON might be better for NEWLINE since it's consistent with the Content-Type (e.g. [link](https://en.wikipedia.org/wiki/JSON_streaming))

generators/java/generator-utils/src/main/resources/Stream.java
```diff
@@ -69,9 +123,28 @@ public T next() {
...
+                        if (streamType == StreamType.SSE) {
+                            String line = scanner.next();
+                            while (true) {
+                                if (!line.contains(options.prefix)) {
+                                    line = scanner.next();
+                                } else {
+                                    break;
+                                }
+                            }
+                            if (line.contains(options.terminator)) {
```
amckinney: Should this be an exact match rather than "contains"? Or is it expected that there will be other text surrounding it?

generators/java/generator-utils/src/main/resources/Stream.java
```diff
@@ -69,9 +123,28 @@ public T next() {
...
-                        return parsedResponse;
+                        if (streamType == StreamType.SSE) {
+                            String line = scanner.next();
+                            while (true) {
+                                if (!line.contains(options.prefix)) {
+                                    line = scanner.next();
+                                } else {
+                                    break;
+                                }
+                            }
```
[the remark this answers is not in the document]
gupta11: done

generators/java/generator-utils/src/main/resources/Stream.java
```diff
@@ -69,9 +123,28 @@ public T next() {
                     throw new NoSuchElementException();
                 } else {
                     try {
-                        T parsedResponse =
-                                ObjectMappers.JSON_MAPPER.readValue(scanner.next().trim(), valueType);
-                        return parsedResponse;
+                        if (streamType == StreamType.SSE) {
```
amckinney: We might be able to more broadly clean up the control flow here. What are your thoughts on something like this, which leverages early `return` statements and reduces nesting?

```java
public T next() {
    if (!scanner.hasNext()) throw new NoSuchElementException();
    
    try {
        String line = scanner.next();
        if (streamType == StreamType.SSE) {
            while (!line.contains(options.prefix)) line = scanner.next();
            if (line.contains(options.terminator)) throw new NoSuchElementException();
            line = line.substring(options.prefix.length());
        } else {
            line = line.trim();
        }
        return ObjectMappers.JSON_MAPPER.readValue(line, valueType);
    } catch (Exception e) {
        throw new RuntimeException(e);
    }
}
```

We can also expand out those single line statements if we prefer:

```java
public T next() {
    if (!scanner.hasNext()) {
        throw new NoSuchElementException();
    }
    
    try {
        String line = scanner.next();
        if (streamType == StreamType.SSE) {
            while (!line.contains(options.prefix)) { 
                line = scanner.next();
            }
            if (line.contains(options.terminator)) {
                throw new NoSuchElementException();
            }
            line = line.substring(options.prefix.length());
        } else {
            line = line.trim();
        }
        return ObjectMappers.JSON_MAPPER.readValue(line, valueType);
    } catch (Exception e) {
        throw new RuntimeException(e);
    }
}
```

Thoughts?

generators/java/generator-utils/src/main/resources/Stream.java
gupta11: Changed to exact match.

generators/java/generator-utils/src/main/resources/Stream.java
gupta11: yeah makes sense. The former looks good. I will make the change.

PR #7542, opened 2025-06-17
fix(csharp): add serialization attribute

Swimburger:
## Description
Add `[Serializable]` attribute to generated schema, request, and error classes.
        This will make it easier for instances of these classes to be stored in places like `Exception.Data`.

## Testing


dsinghvi (approved):

generators/csharp/codegen/src/ast/core/Writer.ts
```diff
@@ -132,6 +132,7 @@ ${this.buffer}`;
         let result = referenceKeys
             // Filter out the current namespace.
             .filter((referenceNamespace) => referenceNamespace !== this.namespace)
+            .filter((referenceNamespace) => referenceNamespace !== "System") // System is implicitly imported
```
amckinney: nit: We might want to extract this out into a lightweight helper function. Something like:

```typescript
const IMPLICIT_NAMESPACES = new Set(["System"]);

private isInCurrentNamepace(referenceNamespace): boolean {
  return referenceNamespace === this.namespace || IMPLICIT_NAMESPACES.has(referenceNamespace);
}
```

That way we can filter with a single call:

```typescript
.filter((referenceNamespace) => !this.isInCurrentNamespace(referenceNamespace))
```

This will make it easier to add more implicit namespaces over time (if you think there will be more).

amckinney (approved):

generators/csharp/codegen/src/ast/core/Writer.ts
Swimburger: Only `System` is truly implicit, but I like your thinking

PR #7545, opened 2025-06-17
fix(typescript): jest issue with .js module mapping

Swimburger:
## Description
Fix jest issue with .js module mapping

## Testing


dsinghvi (approved):
should we add a `versions.yml` to trigger release

Swimburger:
> should we add a `versions.yml` to trigger release

I wrote a good update on versions.yml but lost it. Will write again.

PR #7546, opened 2025-06-17
fix(cli): Fix handling of legacy webhook examples snippets

jsklan:
## Description
Webhooks example snippets were malformed, showing a lot of metadata from the ir layer. This fix shows example snippets in docs as expected.

## Changes Made
- Changed how legacy examples were extracted from the webhook spec in the ir-to-fdr-converter

## Testing
Existing test fixtures don't seem to use these code paths, so I ran the locally generated cli against an example docs config provided by a customer. When I viewed the resultant docs in a browser, the snippets looked as expected.


dsinghvi (approved):
Up to add a changelog here so we can trigger release the CLI: https://github.com/fern-api/fern/blob/main/packages/cli/cli/versions.yml

dsinghvi (approved):

PR #7559, opened 2025-06-20
fix(python): Lazily validate model serialization

jsklan:
## Description
We've been seeing very large memory consumption (>2gb) for complex python sdks just by importing the lib.

## Changes Made
We've modified our UniversalBaseModel's @pydantic.model_serializer method to manually wrap the pydantic.BaseModel serializer. This allows us to do lazy schema validation instead of doing schema validation on import.

## Testing
  - Manually generated elevenlabs python sdk with seed
  - Manually added .fernignored files where necessary
  - Wrote a script to do their [developer quickstart](https://elevenlabs.io/docs/quickstart) with the generated sdk. the request worked, implying serialization is still working.

### DIFF
--- a/generators/python/sdk/versions.yml
+++ b/generators/python/sdk/versions.yml
@@ -1,5 +1,13 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 # For unreleased changes, use unreleased.yml
+- version: 4.22.1
+  changelogEntry:
+    - summary: |
+        Lazily validate model serialization.
+      type: feat
+  createdAt: '2025-06-20'
+  irVersion: 58
+
 - version: 4.22.0
   changelogEntry:
     - summary: |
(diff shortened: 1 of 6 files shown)


jsklan:
@dsinghvi I think I also need to see this change propagate to [this generated file](https://github.com/fern-api/fern/blob/main/generators/python/tests/utils/unaliased_models/types/core/pydantic_utilities.py), but I'm not sure what command we use to generate it.

generators/python/sdk/versions.yml
```diff
@@ -1,5 +1,13 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 # For unreleased changes, use unreleased.yml
+- version: 4.22.1-rc0
+  changelogEntry:
+    - summary: |
+        Lazily validate model serialization.
```
dsinghvi: @jsklan can we add some more details about the impact this may have around memory footprint related to 2.11.0 ? 

Basically things you want to cover: 
- Pydantic 2.11 increased memory footprint
- We now defer model serialization
- memory footprint is a lot less

dsinghvi (approved):

PR #7594, opened 2025-06-24
chore(cli): upload icon assets to s3

chdeskur:
upload locally hosted icon assets to s3 for rendering on the frontend

### DIFF
--- a/packages/cli/docs-resolver/src/utils/getImageFilepathsToUpload.ts
+++ b/packages/cli/docs-resolver/src/utils/getImageFilepathsToUpload.ts
@@ -129,7 +129,8 @@ function collectIconsFromNavigation(
                 if (version.landingPage != null) {
                     collectIconsFromNavigationItem(version.landingPage, filepaths, docsWorkspace);
                 }
-                collectIconsFromNavigation(version.navigation, docsWorkspace);
+                const nestedIcons = collectIconsFromNavigation(version.navigation, docsWorkspace);
+                nestedIcons.forEach((filepath) => filepaths.add(filepath));
             });
             break;
         case "productgroup":
@@ -137,7 +138,8 @@ function collectIconsFromNavigation(
                 if (product.landingPage != null) {
                     collectIconsFromNavigationItem(product.landingPage, filepaths, docsWorkspace);
                 }
-                collectIconsFromNavigation(product.navigation, docsWorkspace);
+                const nestedIcons = collectIconsFromNavigation(product.navigation, docsWorkspace);
+                nestedIcons.forEach((filepath) => filepaths.add(filepath));
             });
             break;
     }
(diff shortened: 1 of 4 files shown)


packages/cli/docs-resolver/src/utils/getImageFilepathsToUpload.ts
```diff
@@ -1,7 +1,22 @@
...
+    if (shouldProcessIconPath(iconPath)) {
+        const absoluteIconPath = resolve(docsWorkspace.absoluteFilePath, iconPath);
+        filepaths.add(absoluteIconPath);
+    }
+}
+
+export function collectFilesFromDocsConfig(
+    parsedDocsConfig: docsYml.ParsedDocsConfiguration,
+    docsWorkspace: DocsWorkspace
+): Set<AbsoluteFilePath> {
```
dsinghvi: nit: generally using objects (named args) with multi-parameter functions is better

dsinghvi: it makes the code that calls into the function more readable

chdeskur:
verified with an incorrect path

packages/cli/docs-resolver/src/utils/getImageFilepathsToUpload.ts
```diff
@@ -1,7 +1,66 @@
...
+    pathToImage: string | undefined;
+    absolutePathToFernFolder: AbsoluteFilePath;
+}): AbsoluteFilePath | undefined {
+    if (pathToImage == null || isExternalUrl(pathToImage)) {
+        return undefined;
+    }
+
+    const filepath = resolve(absolutePathToFernFolder, RelativeFilePath.of(pathToImage));
+
+    if (doesPathExistSync(filepath)) {
```
dsinghvi: can you use an async method and then handle all the `Promises`

packages/cli/docs-resolver/src/utils/getImageFilepathsToUpload.ts
```diff
@@ -78,3 +146,144 @@ export function collectFilesFromDocsConfig(parsedDocsConfig: docsYml.ParsedDocsC
...
+    docsWorkspace
+}: {
+    navigation: docsYml.DocsNavigationConfiguration;
+    docsWorkspace: DocsWorkspace;
+}): Promise<Set<AbsoluteFilePath>> {
+    const filepaths = new Set<AbsoluteFilePath>();
+
+    switch (navigation.type) {
+        case "untabbed":
+            for (const item of navigation.items) {
```
dsinghvi: can we some how do an await Promise.all() so that all of the icon file operations run in parallel, right now you are awaiting one-by-one as you loop through

dsinghvi (approved):

PR #7595, opened 2025-06-24
chore(cli): pin windows to legacy

chdeskur:
while we work on fixing local dev for windows users, we will pin them to the legacy dev environment


dsinghvi (approved):
do we care about `win64` ? not sure if that's a valid platform

PR #7640, opened 2025-06-24
fix(internal): ignore .preview dir

davidkonigsberg:
## Description
Sometimes stuff gets checked in from the .preview directory. We don't want that.

## Changes Made
- Added `.preview/` to the git ignore file

## Testing
Just a gitignore update

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,6 @@ seed/python-sdk/**/output.prof
 
 # Cursor
 .cursor
+
+# Preview
+.preview/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -70,4 +70,7 @@ seed/python-sdk/**/output.prof
 .php-cs-fixer.cache
 
 # Cursor
-.cursor
\ No newline at end of file
+.cursor
+
+# Preview
+.preview/
```
dsinghvi: can we make it every possible `.preview` folder (arbitrarily nested)

dsinghvi (approved):

.gitignore
davidkonigsberg: Yeah, good call, will do

PR #7660, opened 2025-06-25
fix(ci): Throw error (rather than only logging it) if we get a parsing error

davidkonigsberg:
## Description
Follow up from [this](https://buildwithfern.slack.com/archives/C05J55L8JES/p1750792102610959) slack conversation. Currently, these errors are logged but the script keeps running and thus people don't notice the failure. 

## Changes Made
- Throw parsing error, in addition to logging it, when encountered.

## Testing
Tested locally, will keep an eye on CI

### DIFF
--- a/packages/seed/src/utils/convertVersionsFileToReleases.ts
+++ b/packages/seed/src/utils/convertVersionsFileToReleases.ts
@@ -28,9 +28,9 @@ export async function parseGeneratorReleasesFile({
                 });
                 await action(release);
             } catch (e) {
-                context.logger.error(
-                    `Failed to parse and run action on release ${JSON.stringify(entry)}: ${(e as Error)?.message}`
-                );
+                const errorMessage = `Failed to parse and run action on release ${JSON.stringify(entry)}: ${(e as Error)?.message}`;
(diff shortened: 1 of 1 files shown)


packages/seed/src/utils/convertVersionsFileToReleases.ts
```diff
@@ -28,9 +28,9 @@ export async function parseGeneratorReleasesFile({
                 });
                 await action(release);
             } catch (e) {
-                context.logger.error(
-                    `Failed to parse and run action on release ${JSON.stringify(entry)}: ${(e as Error)?.message}`
-                );
+                const errorMessage = `Failed to parse and run action on release ${JSON.stringify(entry)}: ${(e as Error)?.message}`;
```
dsinghvi: can we format the JSON.stringify by doing `JSON.stringify(entry, undefined, 2)`

dsinghvi (approved):

packages/seed/src/utils/convertVersionsFileToReleases.ts
davidkonigsberg: Updated

PR #7665, opened 2025-06-25
feat(cli): support protobuf inputs for docs in generators.yml

sahil485:
## Description
- support protobuf inputs for docs in generators.yml
## Changes Made
- change protobuf parsing strategy to use entire folder from root when `target` is empty
- add `ProtobufIrGenerator` which invokes `buf generate`
- create a "proxy" plugin shell script that calls the compiled JS for the plugin

## Testing


packages/cli/workspace/lazy-fern-workspace/src/protobuf/ProtobufIRGenerator.ts
```diff
@@ -0,0 +1,123 @@
+import { execSync } from "child_process";
```
dsinghvi: can we use `@fern-api/logging-execa` instead?

dsinghvi: ah you already use it, why use execSync at all?

packages/cli/workspace/loader/src/loadAPIWorkspace.ts
```diff
@@ -47,27 +48,32 @@ export async function loadSingleNamespaceAPIWorkspace({
...
             }
 
-            const absoluteFilepathToProtobufTarget = join(
-                absolutePathToWorkspace,
-                RelativeFilePath.of(definition.schema.target)
-            );
+            // if the target is empty, don't specify a target because we are using 'strategy: all' from the root
+            const absoluteFilepathToTarget: AbsoluteFilePath =
+                definition.schema.target.length === 0
+                    ? ("" as AbsoluteFilePath)
```
dsinghvi: `"" as AbsoluteFilePath` feels like its going to come bite us -- is that accurate?

packages/cli/workspace/loader/src/loadAPIWorkspace.ts
```diff
@@ -260,7 +266,15 @@ export async function loadAPIWorkspace({
         return {
             didSucceed: true,
             workspace: new OSSWorkspace({
-                specs: specs.filter((spec) => spec.type !== "openrpc"),
+                specs: specs.filter((spec) => {
+                    if (spec.type === "openrpc") {
+                        return false;
+                    }
+                    if (spec.type === "protobuf" && spec.generateDocs) {
```
dsinghvi: what is `.spec.generateDocs` ? why do we need that field?

packages/cli/workspace/loader/src/loadAPIWorkspace.ts
sahil485: Making the type `AbsoluteFilePath | undefined`

dsinghvi (approved):

PR #7721, opened 2025-06-27
fix(csharp, typescript): Improve auto-pagination logic to also halt on empty string cursor

jsklan:
## Description
We saw some infinite loop paging behavior in csharp and typescipt sdks if the server response has an empty string instead of null for the cursor field.

## Changes Made
Modified csharp and typescipt generators to include check for empty string in addition to null to evaluate hasNextPage.

## Testing

Manually tested with seed against a live api using simple paginated list endpoints.
Observed halting behavior instead of infinite loop behavior.
Added template tests for C#.


Swimburger:
~~Can you use `string.IsNullOrEmpty` directly? The name for the helper method is misrepresenting what it's doing.
We don't need that helper method.~~

generators/csharp/base/src/asIs/Pager.Template.cs
```diff
@@ -455,6 +455,11 @@ SetCursor setCursor
         };
     }
 
+    private static bool isCursorNull(TCursor cursor)
+    {
+        return (cursor is null || (cursor is string s && string.IsNullOrEmpty(s)));
```
Swimburger: Since you already check for null, using `string.IsNullOrEmpty(s)` has some redundance.
You can simply check `s == ""`.
What I'd probably try go for is this:

```csharp
return request switch
        {
            null or "" => true,
            _ => false
        };
```

and rename `isCursorNull` to `isCursorNullOrEmpty`

Swimburger:
Can you push the seed changes so I can see the diff?

Swimburger (approved):

PR #7733, opened 2025-06-28
fix(openapi): add missing query params to converted octet requests

kafkas:
The returned object here is also missing a few other fields like `path-parameters` but `path-parameters` somehow makes its way through the converted definition. Didn't have time to investigate but probably worth looking into later..


dsinghvi (approved):
@kafkas missing 2 things: 
1. a changelog entry in `cli/cli/versions.yml` so that we can release the bug fix
2. updating snapshot tests `pnpm test:update`

PR #7734, opened 2025-06-28
feat(docs): Button Component

dannysheridan:
Add a `Button.mdx` page and add it to `docs.yml` to show how to use Button components.

Thanks, Manny, for the suggestion!

Minor: pick a different sparkle icon for `ask fern`

### DIFF
--- a/fern/pages/docs/components/button.mdx
+++ b/fern/pages/docs/components/button.mdx
@@ -1,0 +1,143 @@
+---
+title: 'Button'
+description: 'Interactive button component with multiple variants and intents'
+---
+
+The `Button` component provides a flexible way to create interactive buttons with various styles, sizes, and intents.
+
+## Example
+
+<Tabs>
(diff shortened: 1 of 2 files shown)


fern/pages/docs/components/button.mdx
```diff
@@ -0,0 +1,143 @@
...
+        <Button intent="warning">Warning Button</Button>
+        <Button intent="danger">Danger Button</Button>
+      </div>
+      
+      <div className="space-x-2">
+        <Button outlined>Outlined Button</Button>
+        <Button minimal>Minimal Button</Button>
+      </div>
+      
+      <div className="space-x-2">
```
chdeskur, line 25: Can you make these all primary?

dannysheridan: why? the default is `none`. if you think they should be `primary` in the examples, then we should change the default, no?

PR #7735, opened 2025-06-29
chore(typescript): Add default values to request parameters

mattblank11:
## Description
Clients were able to include `default` values in their specs, but we weren't doing anything with those values... until today!

Now, if clients specify a default value, we'll use that value as the default in request parameters. For example, if a field named `limit` has a default value of `50`, it will be specified like this in the SDK:

```
const request = {
   limit: 50,
}
```

## Changes Made
In `RequestWrapperParameter.ts`, we'll check if a default value was specified and, if so, we'll pass the `defaultValue` in to `createBindingElement`

## Testing

- [x] Unit tests added/updated - Added fixture tests (`request-parameters`)
- [x] Manual testing completed - Tested for Auth0 locally


seed/ts-sdk/request-parameters/src/api/resources/user/client/Client.ts
```diff
@@ -0,0 +1,192 @@
...
+    ): core.HttpResponsePromise<SeedRequestParameters.User> {
+        return core.HttpResponsePromise.fromPromise(this.__getUsername(request, requestOptions));
+    }
+
+    private async __getUsername(
+        request: SeedRequestParameters.GetUsersRequest,
+        requestOptions?: User.RequestOptions,
+    ): Promise<core.WithRawResponse<SeedRequestParameters.User>> {
+        const {
+            limit = 10,
```
mattblank11: @dsinghvi relevant update to look at

seed/ts-sdk/request-parameters/no-custom-config/src/api/resources/user/client/Client.ts
```diff
@@ -0,0 +1,192 @@
...
+    ): core.HttpResponsePromise<SeedRequestParameters.User> {
+        return core.HttpResponsePromise.fromPromise(this.__getUsername(request, requestOptions));
+    }
+
+    private async __getUsername(
+        request: SeedRequestParameters.GetUsersRequest,
+        requestOptions?: User.RequestOptions,
+    ): Promise<core.WithRawResponse<SeedRequestParameters.User>> {
+        const {
+            limit = 10,
```
mattblank11: @dsinghvi relevant code here

seed/ts-sdk/request-parameters/no-custom-config/src/api/resources/user/client/requests/GetUsersRequest.ts
```diff
@@ -0,0 +1,66 @@
...
+ *         excludeUser: {
+ *             name: "name",
+ *             tags: ["tags", "tags"]
+ *         },
+ *         filter: "filter",
+ *         searchTerm: "test"
+ *     }
+ */
+export interface GetUsersRequest {
+    limit: number;
```
dsinghvi, line 52: @mattblank11 you need to make `limit` optional here otherwise the user will be forced to pass it in

generators/typescript/utils/contexts/src/sdk-context/request-wrapper/types.ts
```diff
@@ -2,3 +2,10 @@ export interface RequestWrapperNonBodyProperty {
     propertyName: string;
     safeName: string;
 }
+
+export interface RequestWrapperNonBodyPropertyWithData extends RequestWrapperNonBodyProperty {
+    originalParameter?: {
+        type: "query" | "path" | "header" | "file";
+        parameter: any; // QueryParameter | PathParameter | HttpHeader | FileProperty
```
Swimburger: Is there a way to avoid `any`?

seed/ts-sdk/request-parameters/no-custom-config/src/api/resources/user/client/requests/GetUsersRequest.ts
mattblank11: Good point, fixing now!

seed/ts-sdk/request-parameters/no-custom-config/src/api/resources/user/client/Client.ts
Swimburger: limit = 10 is an example of a default I assume. Looks good

generators/typescript/sdk/client-class-generator/src/request-parameter/RequestWrapperParameter.ts
```diff
@@ -205,4 +221,60 @@ export class RequestWrapperParameter extends AbstractRequestParameter {
...
+                    if (booleanType.default != null) {
+                        return booleanType.default ? ts.factory.createTrue() : ts.factory.createFalse();
+                    }
+                    return undefined;
+                },
+                bigInteger: (bigIntegerType: BigIntegerType) => {
+                    if (bigIntegerType.default != null) {
+                        return ts.factory.createStringLiteral(bigIntegerType.default);
+                    }
+                    return undefined;
```
Swimburger: Depending on the custom SDK configuration, this actually needs to be a big int, not a string, same for long's.

generators/typescript/sdk/client-class-generator/src/request-parameter/RequestWrapperParameter.ts
```diff
@@ -205,4 +221,60 @@ export class RequestWrapperParameter extends AbstractRequestParameter {
...
+                },
+                uint: () => undefined,
+                uint64: () => undefined,
+                float: () => undefined,
+                date: () => undefined,
+                dateTime: () => undefined,
+                uuid: () => undefined,
+                base64: () => undefined,
```
Swimburger, line 296: Are we not given default values for these types?

mattblank11: @Swimburger if you look at these types under the hood (`FloatType`, `UUintType`, etc.) they don't have default value options, so I' just coercing these to undefined

mattblank11: image

generators/typescript/sdk/client-class-generator/src/request-parameter/RequestWrapperParameter.ts
mattblank11: Updated!

Swimburger: Unfortunately it depends on the custom configuration of the SDK, a combination of serde and `useBigInts`, and the logic for long is different than for big ints.

Swimburger: More info here: https://buildwithfern.com/learn/sdks/introduction/changelog/ts/2025/1/16

mattblank11: Updated!

generators/typescript/sdk/client-class-generator/src/request-parameter/RequestWrapperParameter.ts
```diff
@@ -205,4 +222,70 @@ export class RequestWrapperParameter extends AbstractRequestParameter {
...
     }
+
+    private extractDefaultValue(typeReference: TypeReference, context: SdkContext): ts.Expression | undefined {
+        const resolvedType = context.type.resolveTypeReference(typeReference);
+        
+        if (resolvedType.type === "container" && resolvedType.container.type === "optional") {
+            return this.extractDefaultValue(resolvedType.container.optional, context);
+        }
+        
+        const { useBigInt } = (context.config.customConfig as { useBigInt?: boolean }) ?? false;
```
Swimburger: This code pattern blows, but we pass configuration like that through the constructor, which needs to be passed from another constructor, and another, ...
Something we should change in the future. Either way, providing a `false` default is dangerous here. We should centralize the defaults so when we change a default, it's is changed everywhere.

generators/typescript/sdk/versions.yml
```diff
@@ -1,4 +1,11 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 1.10.5
+  changelogEntry:
+    - summary: |
+        Add default values to request parameters.
+        For example, if you have a query parameter `foo` with a default value of `bar`, the generated request parameter will have a default value of `bar`.
```
Swimburger, line 6: Can you explain how to enable this in here?

generators/typescript/sdk/client-class-generator/src/request-parameter/RequestWrapperParameter.ts
mattblank11: The core problem here is that the type of `config.customConfig` is unknown
image

There _is_ a default value for `useBigInt`, but this code unfortunately isn't aware of it, hence the type casting. This is a deeper issue we should fix in the TS generator, but is beyond the scope of this PR IMO. In the meantime, I can remove the default value and the code still works the same because `undefined` is effectively `false` when we hit line with `if (useBigInt) {`

generators/typescript/sdk/versions.yml
mattblank11: This will just be enabled out of the box for everyone. Are you suggesting that we feature flag it?

generators/typescript/sdk/client-class-generator/src/request-parameter/RequestWrapperParameter.ts
Swimburger: Yeah, my point is that the existing code pattern is to pass on the value with its default all the way from the root of the generator. It's very repetitive, but we should do it consistently until we refactor to a better way.
Screenshot 2025-07-01 at 4 49 46 PM

generators/typescript/sdk/versions.yml
Swimburger: Can we update this now that there's a feature flag?

seed/ts-sdk/request-parameters/use-big-int/src/api/resources/user/client/Client.ts
```diff
@@ -0,0 +1,269 @@
...
+            userList,
+            optionalDeadline,
+            keyValue,
+            optionalString,
+            nestedUser,
+            optionalUser,
+            excludeUser,
+            filter,
+            longParam = 9223372036854776000,
+            bigIntParam = 18446744073709551615,
```
Swimburger: This isn't a bigint at the moment, we should do `BigInt(9223372036854776000)`.
Looks like ts-morph doesn't do what would be expected.

generators/typescript/sdk/versions.yml
mattblank11: Added, just didn't push it up - updated now!

generators/typescript/sdk/client-class-generator/src/request-parameter/RequestWrapperParameter.ts
mattblank11: Yeah I just don't like the pattern 😅 I agree that the type casting isn't ideal, but I think it makes sense to start using `context` instead of passing all these variables through that live on context. Is this a hill you'll die on?

Swimburger (approved):

PR #7737, opened 2025-06-30
fix(typescript): Remove `url-join` dependency

Swimburger:
## Description
Remove `url-join` dependency

## Testing


generators/typescript/utils/core-utilities/tests/unit/url.test.ts
```diff
@@ -0,0 +1,101 @@
+import { joinUrl } from "../../src/core/url";
+
+describe("joinUrl", () => {
```
dsinghvi, line 3: 🚀

dsinghvi (approved):

PR #7758, opened 2025-07-01
feat(python): generate dependency files in pydantic-v2 generator

nevilgeorge:
## Description
Adds logic to generate both `requirements.txt` and `pyproject.toml` for the pydantic-v2 generator. 
Matches logic that exists for pydantic (v1). 

## Changes Made
 * Introduces new classes `PythonDependency` and `PyprojectToml`
 * `PythonProject` now generates and writes `requirements.txt`
 * `PythonProject` now creates a `PyprojectToml` instance to generate the text, and writes `pyproject.toml`
 * Generates a dependency list (currently hardcoded) within `AbstractPythonGeneratorContext`. 

## Testing

* Manual testing completed using seed imdb test.


generators/python-v2/base/src/project/PyprojectToml.ts
```diff
@@ -0,0 +1,210 @@
...
+        const keywords: string[] = [];
+        const projectUrls: string[] = []; 
+
+        lines.push(`description = "${description}"`);
+        lines.push(`readme = "${readme}"`);
+        lines.push(`authors = ${JSON.stringify(authors, null, 4)}`);
+        lines.push(`keywords = ${JSON.stringify(keywords, null, 4)}`);
+
+        const classifiers = [
+            "Intended Audience :: Developers",
```
nevilgeorge: Pulled directly from https://github.com/fern-api/fern/blob/main/generators/python/src/fern_python/codegen/pyproject_toml.py#L120

generators/python-v2/base/src/project/PythonDependency.ts
```diff
@@ -0,0 +1,25 @@
+
+/**
+ * Data class to represent a Python dependency.
+ * Supports writing as both pyproject.toml and requirements.txt formats.
+ */
+export class PythonDependency {
```
dsinghvi: i would make this an `interface` not a class (its more idiomatic in TypeScript and efficient at runtime, which somewhat matters when running this in the browser)

generators/python-v2/base/src/context/AbstractPythonGeneratorContext.ts
```diff
@@ -66,5 +82,19 @@ export abstract class AbstractPythonGeneratorContext<
...
+        // Include both dependencies and dev-dependencies.
+        return [
+            new PythonDependency("pydantic", ">=1.9.2"),
+            new PythonDependency("pydantic-core", ">=2.18.2"),
+            new PythonDependency("mypy", "==1.13.0", true),
+            new PythonDependency("pytest", "^7.4.0", true),
+            new PythonDependency("pytest-asyncio", "^0.23.5", true),
+            new PythonDependency("python-dateutil", "^2.9.0", true),
+            new PythonDependency("types-python-dateutil", "^2.9.0.20240316", true),
+            new PythonDependency("ruff", "==0.11.5", true),
```
dsinghvi: once you make this an interface then itll be something like 
```
return [
  { package: "pydantic", version: ">=1.9.2", type: "dev" } 
]
```

i would use an enum for dev vs non dev deps bc there are other types of deps that we can add to that enum



```

File: generators/python-v2/base/src/project/PyprojectToml.ts [hunk 4]
```diff
@@ -0,0 +1,210 @@
+import { PythonDependency } from "./PythonDependency";
+
+/**
+ * Represents a pyproject.toml file and generates a string representation of it. 
+ */
+export class PyprojectToml {
```
dsinghvi: can we add some tests for this in a test file called `PyProjectToml.test.ts` ? I think this class also might be better renamed as `PyProjectTomlGenerator.ts`

generators/python-v2/base/src/context/AbstractPythonGeneratorContext.ts
```diff
@@ -66,5 +82,19 @@ export abstract class AbstractPythonGeneratorContext<
         return [...fernFilepath.allParts.flatMap((part) => ["resources", this.getSnakeCaseSafeName(part)]), "types"];
     }
 
+    public generateDependencies(): PythonDependency[] {
```
dsinghvi: this function name feels a bit awkwardly named -- `getDependencies` maybe?

PR #7895, opened 2025-07-06
feat(swift): basic struct and property generation

kafkas:
## Description

Wipes the old code in Swift `codegen` directory and reinitializes it with some basic components for struct generation. The previous code had some mistakes like the term "inheritance" doesn't apply to structs. I think we should treat structs and classes separately as they're different concepts, which means we shouldn't have things like `type ClassLevel = 'struct' | 'class'`

Also wanted to add validation for constructor params but didn't wanna make the diff any longer so added TODOs instead.

## Changes Made

- Deleted all code in `codegen`
- Deleted the `openBlock()` method in the `AbstractWriter`. This change needs reviewing. I deleted this because the docs said it's there for Swift only and I suspect I won't be needing it (if I do, I can reimplement it somewhere in the swift directory). It's not used anywhere else.
- Initialized `Property`, `Struct`, and `Type` AST nodes
- Reserved keyword handling - Automatic backtick escaping for Swift reserved words in property names
- Added 2 test suites for behavior validation
- Also had to change a couple of lines of code in the TS generator to fix the lint errors

## Testing


kafkas:
I added `swift` to allowed scopes but `Lint PR title` is still failing...

generators/swift/codegen/src/ast/AccessLevel.ts
```diff
@@ -1,35 +1,36 @@
...
-    Public = "public",
+    Public: "public",
 
     /**
-     * Internal: Accessible within the same module.
-     * The default access level for most parts of a module’s internal implementation.
+     * Accessible within the same package.
      */
-    Internal = "internal",
+    Package: "package",
```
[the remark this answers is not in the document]
kafkas: lol

generators/swift/codegen/src/ast/syntax/reserved-keywords.ts
```diff
@@ -0,0 +1,65 @@
+const reservedKeywordsAsList = [
```
dsinghvi: https://github.com/fern-api/fern/blob/main/packages/commons/casings-generator/src/reserved.ts @kafkas can you add here? the IR has a concept of safe name and unsafe name that takes into account reserved keywords for you!

generators/browser-compatible-base/src/ast/AbstractWriter.ts
```diff
@@ -155,35 +155,6 @@ export class AbstractWriter {
...
-        }
-
-        try {
-            this.indent();
-            callback();
-            this.dedent();
-        } finally {
-            this.write(closingCharacter ?? "");
-        }
-    }
```
dsinghvi, line 185: thanks for deleting

dsinghvi (approved):

generators/swift/codegen/src/ast/syntax/reserved-keywords.ts
kafkas: Oh interesting. Looking into it shortly

kafkas: Sorry, not sure I understand this entirely. If I'm reading this right, this will suffix the relevant IR property with `_` is that right?

Just to be clear, reserved keywords are not actually banned from being used in struct properties. For example, `default` is a reserved word but this works absolutely fine:

```swift
let car = Car(default: "Toyota", model: "Camry")
print(car.default) // ✅ Works fine
```

as long as the struct is defined like so

```swift
public struct Car: Codable, Hashable, Sendable {
    public let `default`: String // must be escaped with backticks
    public let model: String

    public init(`default`: String, model: String) {
        self.default = `default`
        self.model = model
    }
}
```

I think this would be the idiomatic Swift way. But if you still think they should be handled at the IR level then please let me know. That approach would likely require `@fern-api/swift-codegen` to depend on `@fern-api/casings-generator` so that it knows how to render properties (which properties to escape etc.)

PR #7898, opened 2025-07-07
chore(internal): Use seed API def from src instead of copying

davidkonigsberg:
## Description
Resolves FER-5618. Use test API defs in situ instead of copying them.

## Changes Made
- Remove copy cmd
- Update test run cmd to use api def in situ

## Testing
Tested in CI [here](https://github.com/fern-api/fern/actions/runs/16119025038) and [here](https://github.com/fern-api/fern/actions/runs/16119230304)


dsinghvi (approved):
👍

PR #7905, opened 2025-07-08
fix(cli): don't use header name default for global header example values that share names with endpoint headers

sahil485:
## Description
- `globalHeaders` in v1 parser are inferred based on frequency of usage
- examples for these `globalHeaders` are created based on the header name, not the actual value 
## Changes Made
- deduplicate `globalHeaders` that share names with endpoint headers, and use the example value from the endpoint-level header definition/schema

## Testing


mattblank11 (approved):
🚀

PR #7908, opened 2025-07-08
fix(python): allow use of websockets>=14 via their legacy client

wodow:
## Description

Resolves https://linear.app/buildwithfern/issue/FER-5648/python-sdk-sarvam-support-websockets-package-before-and-after-v14

We recently allowed `websockets>=12`, which means we should be compatible with the changed API in `websockets` v14.

## Changes Made
- Imports are changed to use `websockets`' legacy client & client protocol clases when available (in v14) and the existing ones when not.

## Testing

- [x] Unit tests -- awaiting run of existing

### DIFF
--- a/generators/python/sdk/versions.yml
+++ b/generators/python/sdk/versions.yml
@@ -1,6 +1,14 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 # For unreleased changes, use unreleased.yml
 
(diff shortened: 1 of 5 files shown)


wodow:
Made `connect()` also a conditional import following comment from @dsinghvi

dsinghvi (approved):

PR #7909, opened 2025-07-08
chore(lint): Biome as new linter/formatter

davidkonigsberg:
This feels like a more complete solution, and they even offered tooling to help migrate both eslint and prettier rules. That said, running just with the migrated rules is turning up thousands of issues, many of which are trivial, but I haven’t had a chance to start trying to tweak the rules to resolve them.


davidkonigsberg:
Closing for now in favor of oxlint

PR #7951, opened 2025-07-09
fix(php): duplicate type array

sahilmob:
## Description
Solve the issue with rendering duplicate `array` types when dealing with types that are semantically the same
`list<string>` and `set<string>`
The PR addresses the issue reported in:
https://github.com/trycourier/courier-php/pull/48

## Changes Made
- Introduced a new method in the `Type` class to recursively flatten nested union types and ensure unique types.

## Testing

Manual verification:

Test case definition:
Screenshot 2025-07-13 at 2 20 39 PM

The generated code before the fix for the nested unions, which contains duplicate types:
Screenshot 2025-07-10 at 10 03 27 PM

The generated code after the fix:
Screenshot 2025-07-13 at 2 21 14 PM

Running the cli locally against courier's API definition before the fix:
Screenshot 2025-07-13 at 3 20 52 PM

After the fix:
Screenshot 2025-07-13 at 3 25 04 PM

- [x] Tested Courier php-sdk generation


dsinghvi (approved):
@sahilmob this looks good to me, in the future it would be great to make sure that the seed folder actually has a generated type for NestedUnionL2 or UnionWithDuplicateTypes

PR #7963, opened 2025-07-10
fix(typescript): Improve error response parsing

Swimburger:
## Changes Made

* Parse HTTP error bodies as JSON if the response content-type header is JSON, otherwise fallback to text.
* Fix `bytes` fetcher test.
* Fix Jest configuration when a `packagePath` is specified in _generators.yml_ config.

## Testing


dsinghvi (approved):
🚀

PR #8039, opened 2025-07-14
fix(python): handle non-multipleBaseUrls in WebsocketConnectMethodGenerator

wodow:
Unless the environment type is "multipleBaseUrls", a websocket connect method is generated as 

```python
class RealtimeClient:
    ...

    @contextmanager
    def connect(...
    ) -> typing.Iterator[RealtimeSocketClient]:
        ...
        ws_url = +"/realtime/"
```

Note the invalid `ws_url`.

It should default instead to using the base URL, i.e. `ws_url = <BASE URL> + "/realtime/"`

## Changes Made

- If no environment URL, use the base URL as the prefix of the websocket connection URL.

## Testing

## Release
- [x] Set version

### DIFF
--- a/generators/python/src/fern_python/generators/sdk/client_generator/websocket_connect_method_generator.py
+++ b/generators/python/src/fern_python/generators/sdk/client_generator/websocket_connect_method_generator.py
@@ -261,9 +261,13 @@ class WebsocketConnectMethodGenerator:
         parameters: List[AST.NamedFunctionParameter],
     ) -> AST.CodeWriter:
         def write(writer: AST.NodeWriter) -> None:
-            writer.write_line(
-                f'{self.WS_URL_VARIABLE} = {self._get_environment_as_str(websocket=websocket)} + "{websocket.path.head}"'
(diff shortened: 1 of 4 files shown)


generators/python/src/fern_python/generators/sdk/client_generator/websocket_connect_method_generator.py
```diff
@@ -615,7 +619,7 @@ def write_ternary(writer: AST.NodeWriter) -> None:
             context=self._context, object_=reference, type_reference=query_parameter.value_type
         )
 
-    def _get_environment_as_str(self, *, websocket: ir_types.WebSocketChannel) -> str:
+    def _get_environment_as_str(self, *, websocket: ir_types.WebSocketChannel) -> Optional[str]:
```
dsinghvi: nit: can we rename this method to `_get_ multiple_base_url_environment_as_string`

dsinghvi (approved):
We need a `versions.yml` but otherwise this is good to go

PR #8063, opened 2025-07-15
fix(python): updated _convert_undiscriminated_union_type logic to handle lists of classes

aditya-arolkar-swe:
### Context:

Merge was having issues casting permissions in File objects as the "Permission" class. It was coming out as the raw dictionary instead. 

### Solution:

Identified the issue as missing logic of parsing objects from lists. When a List of objects were given, we would previously just fail to parse it and instead return the raw data. Back propagated this fix into the python generators, and tested it by building on the 'exhaustive' fixture. 

### Sample code:

This should not error when ran on Merge's Fern generated filestorage SDK v1. 

```
merge_client = merge.Merge(
    api_key=api_key, account_token=account_token, environment=merge.MergeEnvironment.PRODUCTION_EU
)


response = merge_client.filestorage.files.list(page_size=1)
assert response.results
file = response.results[0]
first_permission = file.permissions[0]

assert isinstance(first_permission, Permission), type(first_permission)
```


generators/python/core_utilities/shared/unchecked_base_model.py
```diff
@@ -128,6 +128,15 @@ def _convert_undiscriminated_union_type(union_type: typing.Type[typing.Any], obj
         return object_
 
     for inner_type in inner_types:
+        # Handle lists of objects that need parsing
+        if get_origin(inner_type) == list and isinstance(object_, list):
```
wodow:
```suggestion
        if get_origin(inner_type) is list and isinstance(object_, list):
```

nit: type comparison by `is` (as with other uses of `get_origin()`)

aditya-arolkar-swe:
Note:

Sets or dicts of objects need hashable objects as their contents, so for the unit tests I added a hash and eq method to the test objects (Square/Circle). If someone tries to use a set/map with an unhashable object, the logic will default back to the raw file contents.

generators/python/tests/utils/example_models/types/core/unchecked_base_model.py
```diff
@@ -134,18 +122,73 @@ def construct(
...
+            try:
+                if inspect.isclass(list_inner_type) and issubclass(list_inner_type, pydantic.BaseModel):
+                    # Validate that all items in the list are compatible with the target type
+                    if _validate_collection_items_compatible(object_, list_inner_type):
+                        parsed_list = [construct_type(object_=item, type_=list_inner_type) for item in object_]
+                        return parsed_list
+            except Exception:
+                pass
+        # Handle sets of objects that need parsing
+        if get_origin(inner_type) is set and (isinstance(object_, set) or isinstance(object_, list)):
```
dsinghvi: lets delete set/map, bc the types that fern generates are not hashable to begin with

dsinghvi (approved):

PR #7638, opened 2025-06-24
chore(cli): Improve logging for subpackages not found by docs resolver

jsklan:
## Description
Existing subpackage errors weren't very useful in helping end users fix issues.

## Changes Made
Subpackage missing errors have been made more descriptive to aid users in resolving issues with their spec.

Example old message:
```
Subpackage undefined not found in 60cdd3a9-89f6-4b02-be89-3417dec15ab9
```
Example new message:
```
Unable to add subsection plantInfo to section Flora due to error: 
Cannot find identifier plantInfo in api definition. Did you mean plant_info?
```

## Testing
Non-exhaustive manual testing against locally malformed docs config.
Seeking advice on how to go about testing further.
- [ ] Unit tests added/updated
- [ ] Manual testing completed

PR #8065, opened 2025-07-15
feat(docs): Add `referenced-packages` key

udaykakade25:
## Description
For #7638 
This PR adds documentation for the `referenced-packages` key

## Changes Made
- Documented the `referenced-packages` key in `docs.yml`
- Created `fern/pages/docs/api-references/referenced-subpackages.mdx`

## Testing


udaykakade25:
Hello @dsinghvi and @chdeskur 
Documentation part seems fine but getting 3 warnings that are causing some CI checks to fail.
One warning in docs.yml: WCAG Contrast Warning
Two warnings in the .mdx file: Endpoint Path Conflicts

Fortunately, these are just warning and not the errors!
Should I work on solving these warnings, although it will be ofcourse from recent documentation part.

PR #8153, opened 2025-07-18
chore(java): exclude legacy-sdk tests

tstanmay13:
## Description
Excluding legacy-sdk tests to unblock Square CI. Would want this to be a temporary change while we figure out what is wrong. 

## Changes Made
- it will add this block to the build.gradle 
```
test {
    useJUnitPlatform()
    testLogging {
        showStandardStreams = true
    }
}

// Exclude legacy-sdk tests
subprojects {
    if (name == 'legacy-sdk') {
        test {
            enabled = false
        }
    }
}
```

Square build.gradle found here for comparison: https://github.com/square/square-java-sdk/blob/master/build.gradle

## Testing


tstanmay13:
Closing as legacy-sdk is a custom test that square has built.

PR #8163, opened 2025-07-18
feat(cli): Let users override the casing of type names in OpenAPI and Fern Definition

Swimburger:
## Description
Let users override the casing of type names in OpenAPI and Fern Definition

## Testing


Swimburger:
There are unexpected diffs in IR generation and I can't figure it out. Giving up on this for now.

PR #8190, opened 2025-07-21
fix(csharp): Fix request body cloning when retrying

windeous:
## Description
The change is just implementing a different way of copying the request that won't try to access the disposed object but make a seperate copy of the request content.

## Changes Made
- Changed the default case of CloneRequestAsync in RawClient in the csharp sdk generator
- Added a if null check at the start of the function
- Added a helper function for copying headers when cloning the request

## Testing


generators/csharp/base/src/asIs/RawClient.Template.cs
```diff
@@ -70,39 +70,48 @@ private static async Task<HttpRequestMessage> CloneRequestAsync(HttpRequestMessa
...
                     {
-                        newPart.Headers.TryAddWithoutValidation(header.Key, header.Value);
+                        var ms = new MemoryStream();
+                        await content.CopyToAsync(ms).ConfigureAwait(false);
+                        ms.Position = 0;
+                        var newPart = new StreamContent(ms);
+                        foreach (var header in oldMultipartFormContent.Headers)
+                        {
+                            newPart.Headers.TryAddWithoutValidation(header.Key, header.Value);
+                        }
```
Swimburger, line 99: Can we use `CopyContentHeaders` here too?

Swimburger (approved):

PR #8210, opened 2025-07-22
fix(go): stringify enum query parameters

dsinghvi:
## Description
Add logic to stringify an enum type into a string.  

## Changes Made
- Upstreams this manual PR into the go generator (https://github.com/anduril/lattice-sdk-go/pull/46).  

## Testing


generators/go-v2/ast/src/context/AbstractGoGeneratorContext.ts
```diff
@@ -389,6 +385,25 @@ export abstract class AbstractGoGeneratorContext<
...
+            }
+        }
+        if (typeReference.type === "container" && typeReference.container.type === "optional") {
+            return this.maybeEnum(typeReference.container.optional);
+        }
+        if (typeReference.type === "container" && typeReference.container.type === "nullable") {
+            return this.maybeEnum(typeReference.container.nullable);
+        }
+        return undefined;
+    }
```
amckinney: Fine as-is for now, but we might want to use the more verbose `assertNever` approach so that we make sure to catch all those cases.

generators/go-v2/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -531,6 +531,15 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
                 const headerField = `${this.getRequestParameterName({ endpoint })}.${this.context.getFieldName(header.name.name)}`;
+                if (this.context.maybeEnum(header.valueType) != null) {
+                    writer.writeNode(
+                        this.addHeaderValue({
+                            wireValue: header.name.wireValue,
+                            value: go.codeblock(`string(${headerField})`)
+                        })
+                    );
+                    continue;
+                }
```
amckinney: Nice catch 👏 -- I think we can actually update the `GoValueFormatter` to handle this case so that it applies for query parameters, too. I can look into that separately though.

amckinney (approved):
Looks good! I think we can refactor things a bit so this fix more widely applies to other use cases, but this fixes the problem at hand so stamping as-is.

PR #8212, opened 2025-07-22
feat(swift): subclient endpoint methods

kafkas:
## Description

With this PR, we are starting to generate subclient endpoint methods, likely the most critical part of the SDK as all request parameters are determined in these methods. We'll be implementing endpoint methods for the root client in a separate PR and will use the same code for that.

## Changes Made

- Adds the `multiline` option for callable expressions like `StructInitialization`,  `FunctionCall` etc. We likely won't need this once we have a proper formatter but let's see how far we get with this.
- Implements the `DictionaryLiteral`, `MethodCallWithTrailingClosure` and `Await` expressions.
- Implements `Optional` type container. With this change, we remove the `optional: bool` option from `FunctionParameter` and `Property` nodes as we can just wrap the type with the new Optional container, which mirrors Swift syntax.
- Implements `Void` type
- Major changes in `SubClientGenerator` to implement endpoint methods.
- Adds snapshot tests for `formatEndpointPathForSwift` where we're outputting the Swift-formatted paths for all Fern test definitions. This just helps us easily view all raw paths in one place but it can be removed if we wanna avoid having too many files.
- Updates seed snapshots
- Fixes a bug where the `persist()` method output would change sporadically due to name collisions which are currently not handled.

## Testing


seed/swift-sdk/literal/Sources/Resources/Headers.swift
```diff
@@ -5,7 +5,17 @@ public final class HeadersClient: Sendable {
...
 
-    public func send(requestOptions: RequestOptions? = nil) async throws -> SendResponse {
-        fatalError("Not implemented.")
+    public func send(endpointVersion: Any, async: Any, request: Any, requestOptions: RequestOptions? = nil) async throws -> SendResponse {
+        return try await httpClient.performRequest(
+            method: .post, 
+            path: "/headers", 
+            headers: [
+                "endpointVersion": endpointVersion, 
+                "async": async
```
dsinghvi, line 14: @kafkas this looks weird

seed/swift-sdk/multi-url-environment/Sources/Resources/Ec2.swift
```diff
@@ -5,7 +5,13 @@ public final class Ec2Client: Sendable {
         self.httpClient = HTTPClient(config: config)
     }
 
-    public func bootInstance(requestOptions: RequestOptions? = nil) async throws -> Any {
-        fatalError("Not implemented.")
+    public func bootInstance(request: Any, requestOptions: RequestOptions? = nil) async throws -> Any {
```
dsinghvi, line 8: nit: not required for now but this endpoint makes a request to a different base url than S3.swift aso ideally you should pass in the base url into `performRequest`

dsinghvi (approved):

seed/swift-sdk/literal/Sources/Resources/Headers.swift
kafkas: Yep, it looks a little weird :) But I think it's a Github rendering bug. I have confirmed it is syntactically valid -- `async` is not a reserved word so we don't need to escape it here.

seed/swift-sdk/multi-url-environment/Sources/Resources/Ec2.swift
kafkas: Ah thanks, I'll add a todo for that

PR #8219, opened 2025-07-22
feat(cli): Add API key OAuth support to IR

Swimburger:
## Description
API key-based OAuth is not a standardized method of implementing OAuth, but some customers do use it.
This PR does the following:
* Add `apiKey` discriminant for OAuth configuration
  * API key configuration accepts an API key, and maps the API key to the OAuth token request
  * The response and refresh mechanisms are essentially the same as client credentials OAuth.
* The API key auth uses the already used `RequestPropertyValue` type, but it is expanded to also support setting headers, as before, it only supported query strings and body properties.
  * We could move this logic more into API key configuration so it disrupts existing code less, but this change makes sense long term.

## Testing


dsinghvi (changes requested):
more context in slack, but we should be generalizing this much further!

Swimburger:
I added `InferredAuthScheme` which requires the minimal amount of info to do OAuth and OAuth like authentication.
- The majority of information will be extracted from `authEndpoint`.
  - Request parameters will be extracted from the endpoint and merged into the root client options
- We use sensible defaults to generate the `authenticatedRequestHeaders` from the response of the `authEndpoint`, while allowing users to specify this if necessary explicitly. Each header contains the information for retrieving the value from the authentication endpoint response and how to set the header on authenticated HTTP requests.
 

For future consideration:
- refresh endpoint
- env variables for auth request parameters

packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
```diff
@@ -24,6 +24,32 @@ types:
...
+        docs: The property to retrieve the header value from the auth response.
+      headerName: 
+        type: string
+        docs: |
+          The header name to put the token in for any authenticated HTTP request. 
+          Defaults to `Authorization`.
+      valuePrefix:
+        type: optional<string>
+        docs: |
+          Commonly used for setting the `Authorization` scheme, but could be used for other things.
```
dsinghvi, line 65: can we default to `Bearer`

packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
```diff
@@ -24,6 +24,32 @@ types:
       basic: BasicAuthScheme
       header: HeaderAuthScheme
       oauth: OAuthScheme
+      inferred: InferredAuthScheme
+
+  InferredAuthScheme:
+    extends: commons.WithDocs
+    properties:
+      authEndpoint: commons.EndpointReference
```
dsinghvi: rename to `getTokenEndpoint` (we will eventually need a refereshTokenEndpoint)

packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
```diff
@@ -24,6 +24,32 @@ types:
...
+      authenticatedRequestHeaders: list<InferredAuthenticatedRequestHeader>
+
+  InferredAuthenticatedRequestHeader:
+    docs:
+      A header that will be set on HTTP requests when the inferred auth scheme is applied to an endpoint.
+      These are not headers for the authorization endpoint.
+    properties:
+      authResponseProperty:
```
dsinghvi: can you add an `expiryProperty` which defaults to `expiry`

packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
```diff
@@ -24,6 +24,32 @@ types:
       basic: BasicAuthScheme
       header: HeaderAuthScheme
       oauth: OAuthScheme
+      inferred: InferredAuthScheme
+
+  InferredAuthScheme:
+    extends: commons.WithDocs
+    properties:
+      authEndpoint: commons.EndpointReference
+      authenticatedRequestHeaders: list<InferredAuthenticatedRequestHeader>
```
dsinghvi, line 39: does this need to be a list? im thinking we just make this a singleton

Swimburger:
Plan:
- publish IR version, not as latest, try it out in TS generator, add missing IR properties for generation
- Call Dynamic Snippets Generator to generate the snippet of the auth endpoint, let DSG handle positional vs named argument, path param vs body param, file param, etc.
- Fix DSG if necessary

dsinghvi (approved):

packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
Swimburger: Sometimes multiple headers are configured, like scopes

packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
Swimburger: Yes

PR #8279, opened 2025-07-23
fix(lint): allow fall through

davidkonigsberg:
## Description
Fix lint by allowing fall through

## Changes Made
- Add comments to allow fall through

## Testing
Tested in CI [here](https://github.com/fern-api/fern/actions/runs/16469986339/job/46556512902)


iamnamananand996 (approved):
This should do the job, have seen, this pattern in `swift-sdk` also

https://github.com/fern-api/fern/blob/c7c9de35ee82f893faebedd055f26f0bdc16bc08/generators/swift/model/src/object/ObjectGenerator.ts#L32

dsinghvi (approved):

PR #8286, opened 2025-07-23
feat(cli): add v2Responses field to `HttpEndpoint` IR type to store multiple successful responses

sahil485:
## Description
- add v2Responses field to `HttpEndpoint` IR type to store multiple successful responses
- append each 2XX response to `HttpEndpoint.v2Responses`
- don't stop processing 2XX responses after the first one is found

## Testing


packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
```diff
@@ -58,6 +58,7 @@ types:
       requestBody: optional<HttpRequestBody>
       sdkRequest: optional<SdkRequest>
       response: optional<HttpResponse>
+      v2Responses: optional<list<HttpResponse>>
```
dsinghvi: can we make this an object that contains the list of responses so that its easy to evolve over time

dsinghvi (approved):

PR #8290, opened 2025-07-23
feat(typescript): Additional query parameters

Swimburger:
## Description
Users can now pass in `additionalQueryParams` as part of the request options.
```ts
await client.foo.bar(..., {
  additionalQueryParams: {
    foo: "bar"
  }
});
```

## Testing


generators/typescript/sdk/client-class-generator/src/endpoints/utils/requestOptionsParameter.ts
```diff
@@ -2,6 +2,7 @@ import { getTextOfTsNode } from "@fern-typescript/commons";
 import { OptionalKind, ParameterDeclarationStructure, ts } from "ts-morph";
 
 export const REQUEST_OPTIONS_PARAMETER_NAME = "requestOptions";
+export const REQUEST_OPTIONS_ADDITIONAL_QUERY_PARAMETERS_PROPERTY_NAME = "additionalQueryParams";
```
dsinghvi: can we call this `queryParams` and add docs that it can be additional, thats what we do for headers for example

generators/typescript/sdk/versions.yml
```diff
@@ -1,4 +1,19 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.6.0
+  changelogEntry:
+    - summary: |
+        Users can now pass in `queryParams` as part of the request options.
+        ```ts
+        await client.foo.bar(..., {
+          queryParams: {
```
dsinghvi, line 8: nit: i would do `queryParameters` not `queryParams` but leave that decision to you! (to align with fetcher)

dsinghvi (approved):

PR #8353, opened 2025-07-24
feat(cli): add support for multiple 2XX responses and fix example generation naming

sahil485:
## Description
- add support for multiple 2XX responses and fix example generation naming
## Changes Made
-  previous approach iterated through user-specified request examples, then user-specified response examples, and then created auto-generated ones
- add a display name to examples to prevent type collisions in example keys while maintaining semantic naming

New approach: 

**Step 1**: Iterate through all 2XX responses, create examples for each one
- if there is a matching request/response pair, use that and mark the request as "used"
- if there is no matching request for a user-specified response, use an autogenerated or base request

**Step 2**: Iterate through all unused user-specified request, create examples for each one
- since we already checked responses, we don't have to check if there is a matching response example key in this step
- match autogenerated/base response examples with the remaining "unused" user-specified request examples

**Step 3**: If the endpoint examples object is empty, this means there were no non-empty responses and no user-specified request examples
- use an autogenerated/base request example with an autogenerated/base response example

image

## Testing


eyw520 (approved):
Reviewed changes in person. Excellent work.

PR #8409, opened 2025-07-28
fix(csharp): Update C# generator to support FileDownload and JSON Streaming 

fearthecowboy:
## Description
This adds support for the C# generator to create code to support FileDownload (returns a `Task<Stream>`) and streaming JSON responses (returns an `IAsyncEnumberable<...>`) 

## Changes Made
- [x] added support to declare a method as returning `IAsyncEnumerable<...>` (as opposed to `Task<...>`
- [x] emit endpoint implementation for FileDownload (trivial, returns a stream)
- [x] emit endpoint implementation for JSON Streaming (uses `ReadLineAsync()` to iterate over results and `IAsyncEnumerable` to return results as they are available)
- [x] Updated versions.yml for the c# generator 

## Testing


seed/csharp-sdk/server-sent-event-examples/src/SeedServerSentEvents/Completions/CompletionsClient.cs
```diff
@@ -36,6 +36,30 @@ internal CompletionsClient(RawClient client)
...
+                {
+                    throw new SeedServerSentEventsException(
+                        $"Unable to deserialize JSON response '{line}'"
+                    );
+                }
+                if (chunk is not null)
+                {
+                    yield return chunk;
+                }
+                yield break;
```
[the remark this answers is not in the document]
fearthecowboy: Ah, it's on the wrong side of the brace. Fixing...

generators/csharp/sdk/versions.yml
```diff
@@ -1,4 +1,11 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.1.2
+  changelogEntry:
+    - summary: |
+        Generate methods to support FileDownload (returns a `Task<Stream>`) and streaming JSON responses (returns an `IAsyncEnumberable<...>`) .
+      type: fix
```
Swimburger: I would create two items inside of `changelogEntry` for this.
One for file download and one for streaming, and make it `type: feat`.

seed/csharp-sdk/streaming/snippet.json
```diff
@@ -10,7 +10,7 @@
             },
             "snippet": {
                 "type": "csharp",
-                "client": "using SeedStreaming;\n\nvar client = new SeedStreamingClient();\nawait client.Dummy.GenerateStreamAsync(new GenerateStreamRequest { Stream = true, NumEvents = 1 });\n"
+                "client": "using SeedStreaming;\n\nvar client = new SeedStreamingClient();\nvar pager = await client.Dummy.GenerateStreamAsync(new GenerateStreamRequest { Stream = true, NumEvents = 1 });\n\nawait foreach (var item in pager)\n{\n    // do something with item\n}\n"
```
Swimburger: Can we update this snippet not to use the name `pager`?

generators/csharp/sdk/versions.yml
Swimburger: Does it support non-JSON streaming responses? I think it's worth calling this SSE too

fearthecowboy: I didn't actually implement support for non-json SSE - It should handle SSE style responses for JSON objects, but I don't see a test for non-JSON SSE.

seed/csharp-sdk/streaming/snippet.json
fearthecowboy: Sure, I'll change it to `items`.  It's reusing the same code path as the pager implementation, but both can be called `items`.

generators/csharp/sdk/versions.yml
```diff
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.1.2
+  changelogEntry:
+    - summary: Generate methods to support FileDownload (returns a `Task<Stream>`).
+      type: feat
+    - summary: Generate streaming/SSE JSON responses (returns an `IAsyncEnumberable<...>`).
```
Swimburger, line 6: Can we put a dummy code sample here too?

generators/csharp/sdk/versions.yml
```diff
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.1.2
+  changelogEntry:
+    - summary: Generate methods to support FileDownload (returns a `Task<Stream>`).
```
Swimburger, line 4: Can we put a dummy code sample here too?

Swimburger (approved):

PR #8411, opened 2025-07-28
feat(swift): nested subclients and root client endpoints

kafkas:
## Description

Implements several changes that let us generate nested subclients and the endpoint methods for the root client. Also includes several other code improvements.

## Changes Made

- Implements `ExistentialAny` type that lets us avoid build errors by using existential types like `any Encodable` instead of `Any`
- Adds `JSONValue` type
- Expands `SdkGeneratorContext` to allow extracting things like subclient names, environment names etc. from one place.
- Implements `ClientGeneratorContext` to reuse information between `RootClientGenerator` and `SubClientGenerator`
- Implements `EndpointMethodGenerator`
- Updates Seed snapshots

## Testing


generators/swift/codegen/src/ast/Struct.ts
```diff
@@ -67,7 +67,7 @@ export class Struct extends AstNode {
             property.write(writer);
             writer.newLine();
         });
-        if (this.initializers) {
+        if (this.initializers && this.initializers.length > 0) {
```
Swimburger: I would suggest always initializing `initializers` with an empty array, then you only have to check for the length, not whether it's undefined or not first.
Same for methods, nestedTypes, etc.

generators/swift/sdk/src/SdkGeneratorContext.ts
```diff
@@ -29,6 +30,30 @@ export class SdkGeneratorContext extends AbstractSwiftGeneratorContext<SdkCustom
...
+        return `${this.ir.apiName.pascalCase.unsafeName}Client`;
+    }
+
+    public get environmentEnumName(): string {
+        return `${this.ir.apiName.pascalCase.unsafeName}Environment`;
+    }
+
+    public getSubClientName(subpackage: Subpackage): string {
+        return `${subpackage.name.pascalCase.unsafeName}Client`;
+    }
```
Swimburger, line 55: Some of these we customers users to override using generators.yml config, I assume customers will want to do that for Swift too. Not necessary in this PR tho, just a note.

Swimburger (approved):

generators/swift/codegen/src/ast/Struct.ts
kafkas: Good call

generators/swift/sdk/src/SdkGeneratorContext.ts
kafkas: Yep, we can make them configurable once we start wiring up generator config

PR #8412, opened 2025-07-28
feat(cli): Add lfs-override param to fern cli to override output mode

jsklan:
## Description
Added lfs-override param to fern cli to override output mode

## Testing
Manually tested using fern local against sample generators.yml and api spec.


packages/cli/cli/src/cli.ts
```diff
@@ -520,6 +520,10 @@ function addGenerateCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext)
                     choices: ["docker", "podman"],
                     description: "Choose the container runtime to use for local generation.",
                     default: undefined
+                })
+                .option("lfs-override", {
```
dsinghvi, line 524: @jsklan can you hide this option in the CLI and then we are good to merge

dsinghvi (approved):

PR #8500, opened 2025-07-31
feat(cli): add support for proto --> openapi --> IR generation using from-openapi generators.yml option

sahil485:
## Description
- add support for proto --> openapi --> IR generation using from-openapi generators.yml option
## Changes Made
- add `from-openapi` option for proto in `generators.yml` 
- when `from-openapi: true` use `protoc-gen-openapi` using either the specified `target` or `root` dir
- update `buf.yaml` to include common `google` deps
## Testing


fern/apis/generators-yml/definition/generators.yml
```diff
@@ -193,6 +193,12 @@ types:
       local-generation:
         type: optional<boolean>
         docs: Whether to compile the `.proto` files locally. By default, we generate remotely.
+      from-openapi:
+        type: optional<boolean>
+        docs: Whether to convert to OpenAPI before generating
+      deps:
```
dsinghvi: nit: lets name it dependencies

dsinghvi (approved):

fern/apis/generators-yml/definition/generators.yml
sahil485: sure, will update later

PR #8505, opened 2025-08-01
feat(cli): allow custom pypi publishing targets

aditya-arolkar-swe:
## Description

Pypi publishing targets previously didn't propagate package names + versions to the generated wheel 

## Changes Made

- added the pypi target class


packages/ir-sdk/fern/apis/ir-types-latest/definition/publish.yml
```diff
@@ -54,3 +56,8 @@ types:
       usernameEnvironmentVariable: string
       passwordEnvironmentVariable: string
       mavenUrlEnvironmentVariable: string
+
+    PypiPublishTarget:
+      properties:
+        version: string
+        packageName: string
```
dsinghvi: make both `version` and `packageName` optional<string>

dsinghvi (approved):

PR #8521, opened 2025-08-01
fix(python): __iter__ on websockets should json.load messages

aditya-arolkar-swe:
## Description

Previously we weren't processing messages on the __iter__ methods of a websocket client with json.loads, causing some issues. 

## Changes Made

- added the json.loads() call, and some sample code from a client worked on that


eyw520 (approved):
LGTM

PR #8548, opened 2025-08-04
fix(cli): literal type headers should use literal value for example

aditya-arolkar-swe:
## Description

When headers of the "literal" type were added, the examples that were generated in the fern definition defaulted to using the name of the header itself. A validation rule down the line yelled about this and fern check failed. 

## Changes Made

Use the value of the literal as the example directly. Otherwise, fallback to using the header name.

## Testing

[X] Manual testing

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildEndpointExample.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildEndpointExample.ts
@@ -90,7 +90,7 @@ export function buildEndpointExample({
                         literal: (literal) => {
                             if (literal.type === "string") {
                                 return literal.string;
-                            } else if (literal.type == "boolean") {
+                            } else if (literal.type === "boolean") {
                                 return literal.boolean.toString();
                             }
                             return undefined;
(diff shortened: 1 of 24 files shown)


packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildEndpointExample.ts
```diff
@@ -35,24 +41,49 @@ export function buildEndpointExample({
...
+                    namedFullExamples.push(header);
+                }
+            }
+        }
+
+        // include global headers that are not already in the endpoint headers
+        for (const [header, info] of Object.entries(globalHeaders)) {
+            if (!endpointHeaderNames.has(header)) {
+                // get 'literal' headers and use the value as the example
+                if (typeof info === "object" && "type" in info && info.type.startsWith("literal")) {
```
dsinghvi: recursivelyVisitRawTypeReference and handle the literal case, otherwise just use `noop` from core utitlies on the other cases

packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildEndpointExample.ts
```diff
@@ -35,24 +44,68 @@ export function buildEndpointExample({
...
+                        optional: (valueType) => valueType,
+                        nullable: (valueType) => valueType,
+                        set: (valueType) => valueType,
+                        named: (name) => [name],
+                        literal: (name) => [name],
+                        unknown: () => []
+                    }
+                });
+
+                const exampleValue = valueToUse?.[0]?.string ?? 'default_example';
```
aditya-arolkar-swe: valueToUse looks something like this [ { string: '2.0', type: 'string', _visit: [Function: _visit] } ] object

So I thought this was a type safe way to query the value we want, but let me know if there's a preferred standard.

dsinghvi (approved):

PR #8573, opened 2025-08-04
feat(csharp): Trigger release to use new version of CSharpier formatter

Swimburger:
## Description
Trigger release to use new version of CSharpier formatter


fearthecowboy (approved):
Looks good

PR #8578, opened 2025-08-04
feat(swift): collision handling for symbol and file names

kafkas:
## Description

This PR adds a robust mechanism for detecting and resolving name collisions during codegen, covering both symbol and file names.

To handle symbol name collisions, we’ve implemented a `SymbolRegistry`, which we're adding to each `SwiftProject`. When the project is initialized, we register the built-in, `Foundation`, and "as-is" types used by our generator. We then traverse the IR to register symbol names for all nodes we intend to generate, such as subclient names, schema types, etc. Generators use this registry to retrieve symbol names that are guaranteed to be unique.

The registry automatically resolves naming conflicts during registration. If a collision is detected, it attempts to use a predefined list of fallback names based on the symbol type. As a last resort, it appends underscores (`_`) until the name becomes unique. For example, if the user defines an object type named `Int` (a Swift built-in), the registry will fall back to `IntType`. Similarly, if `PetstoreClient` is somehow taken, it will try `PetstoreApi`, then `PetstoreApiClient`, then `PetstoreClient_`, and so on.

For file name collisions, the approach is simpler: we append `_` until the conflict is resolved. This is acceptable for now, as file names don’t impact the generated SDK output significantly. We can revisit this later if needed.

After this PR, we should expect the generated SDKs for all definitions to compile successfully using `swift build`. Upcoming PRs will focus on adding more SDK features and generator config options.

## Changes Made

- Updates “as-is” file definitions to include symbol name metadata
- Updates `AbstractSwiftGeneratorContext` to initialize the project and register symbol names; moves `getSwiftTypeForTypeReference` into this class
- Introduces an `addSourceFile` method to `SwiftProject` to resolve file name collisions
- Parallelizes file writes in `persist()`, now that name uniqueness is guaranteed
- Implements `SymbolRegistry` in the `base` package
- Removes a now-redundant snapshot test that I added before Seed integration
- Removes `generateModels` and `generateInlinedRequestModels` functions, which previously returned arrays of `SwiftFile` instances. Since files are now created via `SwiftProject` to handle name collisions, this flow is inherently stateful. As a result, generators like `StringEnumGenerator` and `ObjectGenerator` now emit AST nodes instead of `SwiftFile` objects.
- Expands `SdkGeneratorCli` to generate source files
- Updates Seed snapshots

## Testing


generators/swift/base/src/project/SymbolRegistry.ts
```diff
@@ -0,0 +1,272 @@
+import { values } from "@fern-api/core-utils";
```
tstanmay13, line 1: This approach seems good for collision handling. I'm curious about the registration order - do you register all symbols upfront during IR traversal, or lazily needed? Any memory concerns we should be wary about with large API specs?

tstanmay13: Also can you have can you have User as both type name and a variable name in different scopes?

tstanmay13: Will the same API definition always produce the same code, or could registration order affect the output?

kafkas: > This approach seems good for collision handling. I'm curious about the registration order - do you register all symbols upfront during IR traversal, or lazily needed? Any memory concerns we should be wary about with large API specs?

Yep, we register every symbol upfront in a single IR walk. The registry is theoretically a small subset of the IR which we’re already holding in memory, so the extra footprint should be trivial compared to the IR itself.

kafkas: > Also can you have can you have User as both type name and a variable name in different scopes?

Yep, you can. The local variable name would just shadow the type name. That said, it’s pretty unlikely in our generated code. We follow Swift conventions (PascalCase for types, camelCase for vars/properties), and we control all the variable names.

kafkas: > Will the same API definition always produce the same code, or could registration order affect the output?

Registration order does affect the output but it only changes if the generator code changes. So the same API definition + same generator version -> identical output every run.

tstanmay13: Thanks! Do you think possible name changes could impact our customers and their users?

tstanmay13 (approved):

generators/swift/base/src/project/SymbolRegistry.ts
kafkas: Name changes can be breaking and may impact customers and their users. I think the probability of this will drop as we move from alpha to later versions, but we’re not at the optimal point yet. What we mentioned in the weekly about namespacing inline request structs is one thing that will mitigate this for example. What you mentioned about lazily registering also makes sense. We can pre-register only the built-ins we know are used by the generated code (e.g., if `Int` isn’t used, let the customer keep their `Int` struct). I haven’t pursued this yet since it’s error-prone and unlikely to matter during alpha.

Basically, the goal is to find the right balance between stability and aesthetics. If the latter was not a primary concern, we would just use fully qualified paths for all types and namespacing everything which would virtually eliminate collisions. My goal is to keep iterating with customer feedback to converge on the right point soon.

PR #8579, opened 2025-08-04
fix(csharp): address failing C# CI tests

fearthecowboy:
## Description
Addressing the c# test failures in CI.

## Changes Made
specifics:
 - Streaming APIs return `IAsyncEnumerable<T>` - snippets changed to use `async foreach`
 - String enums that contain a value called `Value` have their internal `Value` property renamed to `_Value` to avoid conflicts
 - UnionGenerator: Do not generate implicit cast from `object` or `IEnumerable<T>` to string enum
 - UnionGenerator: generate single null check at the end on deserialization - non nullable types failed to compile with multiple null checks inside the switch
 - HttpEndpointGenerator: refactor streaming to work for streaming parameter variants too.
 - MockServerTestGenerator: use `items` instead of `pager`
 - In `RawClient.cs` namespace references are prefixed with `global::`  to ensure that the class works correctly when an explicit namespace and client class are used that could cause it to collide.
 - Set csharpier version to `1.1.*` to avoid inadvertently picking up `1.0.3` which has an inconsistency in formatting. 
 
## Testing
- [x] regenerated seed tests


generators/csharp/codegen/src/ast/MethodInvocation.ts
```diff
@@ -49,11 +61,16 @@ export class MethodInvocation extends AstNode {
         this.configureAwait = configureAwait ?? false;
         this.generics = generics ?? [];
         this.multiline = multiline ?? false;
+        this.isAsyncEnumerable = isAsyncEnumerable ?? false;
     }
 
     public write(writer: Writer): void {
-        if (this.async) {
-            writer.write("await ");
+        if (this.isAsyncEnumerable) {
```
Swimburger, line 68: I think `await foreach(var item in` is a stretch for it being considered method-invocation. Let's create a different AST node for iteration or do something else, but not put it inside this.

generators/csharp/dynamic-snippets/src/EndpointSnippetGenerator.ts
```diff
@@ -129,13 +129,19 @@ export class EndpointSnippetGenerator {
         endpoint: FernIr.dynamic.Endpoint;
         snippet: FernIr.dynamic.EndpointSnippetRequest;
     }): csharp.MethodInvocation {
+        // if the example has *any* sample with stream set to true, then the method is an async enumerable
+        const isAsyncEnumerable = endpoint.examples?.some(
+            // biome-ignore lint/suspicious/noExplicitAny: allow explicit any
+            (example: Record<string, any>) => example.requestBody?.stream && example.requestBody?.stream === true
```
Swimburger: Can we use `unknown` here? It'll be the same but more code, but we avoid any

generators/csharp/dynamic-snippets/src/EndpointSnippetGenerator.ts
```diff
@@ -696,13 +702,15 @@ export class EndpointSnippetGenerator {
...
+        let str = this.context.getValueAsString({ value });
         if (str == null) {
             this.context.errors.add({
                 severity: Severity.Critical,
                 message: "The bytes request body must be provided in string format"
             });
-            return csharp.TypeLiteral.nop();
+
+            // if there is no value, then let's just use a random string
+            str = "Some random text to upload";
```
Swimburger: maybe `"[bytes]"`

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -1,5 +1,6 @@
 import { assertNever } from "@fern-api/core-utils";
 import { csharp } from "@fern-api/csharp-codegen";
+import { ClassReference } from "@fern-api/csharp-codegen/src/csharp";
```
Swimburger: Let's import from `@fern-api/csharp-codegen`, and update the necessary index.ts's to make this possible

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -130,16 +131,10 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
-            endpoint.response?.body?._visit({
-                streaming: () => true,
-                json: () => false,
-                fileDownload: () => false,
-                text: () => false,
-                bytes: () => false,
-                streamParameter: () => false,
-                _other: () => false
-            }) ?? false;
+            endpoint.response?.body?.type === "streaming" || endpoint.response?.body?.type === "streamParameter";
```
Swimburger: We prefer the visit pattern or exhaustive switch at Fern, so we're forced to consider a new discriminant if added.

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -305,10 +300,106 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                        yieldValue("line");
+                        yieldBreak();
+                    },
+                    sse: (sseChunk) => {
+                        // todo: implement SSE - this is a placeholder for now
+                        // const payloadType = this.context.csharpTypeMapper.convert({ reference: sseChunk.payload });
+                        writer.writeLine("/* SSE Not currently implemented */");
+                    },
+                    _other: () => {
+                        writer.write('/* "Other" Streaming not currently implemented */');
```
Swimburger, line 399: We usually throw an error when something is not supported

Swimburger (changes requested):

seed/csharp-model/trace/src/SeedTrace/V2/V3/Problem/TestCaseImplementationDescriptionBoard.cs
```diff
@@ -182,14 +182,14 @@ JsonSerializerOptions options
...
-                "paramId" => json.GetProperty("value").Deserialize<string>(options)
-                    ?? throw new JsonException("Failed to deserialize string"),
-                _ => json.Deserialize<object?>(options),
-            };
+            var value =
+                discriminator switch
+                {
+                    "html" => json.GetProperty("value").Deserialize<string>(options),
+                    "paramId" => json.GetProperty("value").Deserialize<string>(options),
+                    _ => json.Deserialize<object?>(options),
```
Swimburger: This worries me, some discriminants are nullable, so we shouldn't throw on null. Moving the `?? throw` to the bottom will be a breaking change/bug, unless we calculate up front that none of the discriminants can be null.

seed/csharp-model/undiscriminated-unions/src/SeedUndiscriminatedUnions/Union/KeyType.cs
```diff
@@ -13,13 +13,18 @@ namespace SeedUndiscriminatedUnions;
...
     /// <summary>
     /// The string value of the enum.
     /// </summary>
-    public string Value { get; }
+    string IStringEnum.Value => _Value;
+
+    /// <summary>
+    /// The string value of the enum.
+    /// </summary>
+    public string _Value { get; }
```
Swimburger: Good stuff!

seed/csharp-sdk/streaming-parameter/src/SeedApi.DynamicSnippets/Example0.cs
```diff
@@ -12,12 +12,12 @@ public class Example0
...
         );
 
-        await client.Dummy.GenerateAsync(
+        await foreach (var item in client.Dummy.GenerateAsync(
             new GenerateRequest{
                 Stream = false,
                 NumEvents = 5
             }
-        );
+        )) {/** consume each item */};
```
Swimburger, line 20: let's put the comment on its own line, unless this is undone by the formatter

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
fearthecowboy: oh. didn't even notice the editor did that.

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
fearthecowboy: The generated code still does throw an error (the throw follows that), but it's handy to see a note in the code when it's not implemented in there?

seed/csharp-sdk/streaming-parameter/src/SeedApi.DynamicSnippets/Example0.cs
fearthecowboy: I did! the formatter moved it back!

seed/csharp-model/trace/src/SeedTrace/V2/V3/Problem/TestCaseImplementationDescriptionBoard.cs
fearthecowboy: Yeah, I didn't catch that case. My Bad.

HMMMMM.

This is somewhat tricky. I **_can_** check if the individual ones are optional, but there isn't detail to identify when the type is guaranteed **_not_** to be null (in which case the compiler complains that the `?? throw ...` isn't permitted. (really, the compiler complaining about the `??` because it knows it isn't going to happen is pretty petty on it's part. 

I can check to see if any are optional and suppress the `?? ` entirely in that case since, that would be a valid result regardless of what's in the switch

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
fearthecowboy: Weird, it's not even necessary. `csharp.ClassReference` is ok. Silly editor. 😄

generators/csharp/sdk/src/endpoint/http/HttpEndpointGenerator.ts
Swimburger: Sounds good

generators/csharp/codegen/src/ast/MethodInvocation.ts
fearthecowboy: Sure, we'll followup later on this.

seed/csharp-model/trace/src/SeedTrace/V2/V3/Problem/TestCaseImplementationDescriptionBoard.cs
Swimburger: - If the type from IR is optional or nullable, we should not `?? throw`
- If the type from IR is primitive, we should not `?? throw`
- Else `?? throw`

I think we have some code for determining whether something is `primitive`; it's not the most reliable, but good enough.

Swimburger: It appears to respect nullable/optional types, but not the primitive case.

fearthecowboy: The trouble that I'm facing with the `??` is that there isn't any way to tell the difference between deserializing a `public readonly record struct` and a `class` - they both pass thru
``` typescript
                                const isReferenceType = csharpType.isReferenceType();
                                if (
                                    isReferenceType === true ||
                                    csharpType.internalType.type === "reference" ||
                                    csharpType.internalType.type === "coreReference"
                                ) {
```
this just fine.
I'm going to roll back the change for this, and add `circular-references` and `circular-references-advanced` to the acceptable failures list so we can get the rest of these changes in, and I can look deeper into it in a bit.

Swimburger (approved):

PR #8580, opened 2025-08-05
fix(python): nested folder creation for package names with "."

aditya-arolkar-swe:
## Description

If a package name had periods in it (e.g. adi.test.package), the SDK generated a folder like "src/adi.test.package" when what we want is "src/adi/test/package".

## Changes Made

- simple change in the naming of the relative path in the Project object. The call to `mkdir_p()` down the line should automatically create the nested folders as expected. 

## Testing


dsinghvi (approved):
@aditya-arolkar-swe this looks good to me, can you add a changelog?

PR #8587, opened 2025-08-05
feat(java): add extensible builder pattern with self-type support

tstanmay13:
Introduces an opt-in `enable-extensible-builder` flag that generates builders using the self-type pattern ([F-bounded polymorphism](https://stackoverflow.com/questions/59813323/advantages-of-f-bounded-polymorphism-over-typeclass-for-return-current-type-prob)). This allows users to create custom builders that extend the generated builders without losing type information during method chaining. This feature is for pipedream who have to do some jank setup currently to get this code to work. Copying my explanation of the issue here 

> So basically, the builder extension feature has a method chaining issue where all setter methods return the base builder type instead of the actual subclass type. Meaning:
> - when you call .token(abc) on your on your CustomApiClientBuilder, it returns BaseApiClientBuilder
> - and Then calling .build() returns BaseApiClient instead of your CustomApiClient
and this means you have to either cast or override evrey single setter method.

  ## Changes
  - Added `enable-extensible-builder` boolean flag to ICustomConfig base interface
  - Modified AbstractRootClientGenerator to conditionallygenerate:
    - Simple concrete builders (default) for backward
  compatibility
    - Abstract builders with self-type pattern when `enable-extensible-builder: true`
  - Extensible builders include:
    - Generic type parameter `<T extends Builder<T>>`
    - Abstract `self()` method returning `T`
    - All builder methods return `T` instead of concrete type
    - Default `Impl` inner class for direct usage
  - Added test fixtures to verify both modes work correctly

  ## Example Usage
  ```java
  // In generators.yml
  default-group: local
  groups:
    local:
      generators:
        - name: fernapi/fern-java-sdk
          config:
            extend: true  # Enable extensible builders

  //This enables users to create type-safe custom builders:
  public class CustomApiBuilder extends 
  BaseClientBuilder<CustomApiBuilder> {
      @Override
      protected CustomApiBuilder self() { return this; }

      public CustomApiBuilder tenantId(String id) {
          // Custom configuration
          return this;
      }
  }
  ```
### Testing
- Added tests and tested against Pipedreams definition by changing their generators.yml to include the identifier `enable-extensible-builder`. Everything worked and the code was using the extensible pattern. 

 ### Breaking Changes

  None. The feature is opt-in and maintains full backward
  compatibility when not enabled.


Swimburger:
I would make the feature flag name more explicit, `extend` could mean many things

generators/java/sdk/versions.yml
```diff
@@ -1,3 +1,11 @@
+- version: 2.39.6
+  changelogEntry:
+    - summary: |
+        Add extensible builder pattern with self-type support via opt-in `extend` flag
```
Swimburger: Add more details here, possibly a code sample

generators/java/sdk/versions.yml
```diff
@@ -1,3 +1,24 @@
+- version: 2.39.6
```
Swimburger, line 1: nit: since this is a feature, it should increment the minor version, not the patch version

Swimburger (approved):

PR #8601, opened 2025-08-06
feat(swift): implement doc comments

kafkas:
## Description

Implements doc comments across the entire SDK.

## Changes Made

- Docs for `Class`, `EnumWithRawValues`, `FunctionParameter`, `Method`, `Property`, `Statement` (`typealias-declaration` statements), `Struct` nodes
- Implements `DocComment` node
- Implements doc comments for endpoint methods

## Testing


generators/swift/codegen/src/ast/DocComment.ts
```diff
@@ -0,0 +1,98 @@
...
+                });
+            }
+        }
+    }
+
+    private writeMultilineText(writer: Writer, text: string) {
+        const lines = text.split("\n");
+        for (const line of lines) {
+            writer.write("/// ");
+            writer.write(line);
```
tstanmay13, line 97: Could there be any compilation errors if things aren't properly escaped?

generators/swift/codegen/src/ast/DocComment.ts
```diff
@@ -0,0 +1,98 @@
...
+                        writer.write("/// ");
+                        writer.write(line);
+                    }
+                    writer.newLine();
+                });
+            }
+        }
+    }
+
+    private writeMultilineText(writer: Writer, text: string) {
```
tstanmay13: Should add some sanitation here -- documenation from API spec flows directly into the code. If I'm understanding things correctly malicious documentation could potentially inject Swift code by breaking out of comment
blocks (e.g., using `*/` to close comments early)

kafkas: Good catch. Yes, sanitation is needed here. Will push some changes tomorrow

generators/swift/codegen/src/ast/DocComment.ts
kafkas: Yep will need to sanitize

generators/swift/codegen/src/ast/DocComment.ts
kafkas: Ok, I added input sanitization for DocComment with comprehensive testing against malicious input. Things like `*/` are actually safe because they're treated as literal text in doc comments as we're using the triple-slash syntax. But carriage returns (`\r`) and null bytes do break doc comment structure and cause compilation errors (or warnings) without sanitation so we're now removing them.

The current sanitation is surgical, we only remove what causes Swift compiler errors/warnings but we can be a bit more aggressive later if needed (e.g. if we want cosmetic improvements)

tstanmay13 (approved):

PR #8633, opened 2025-08-08
feat(php): Support for apiVersion header. Default accept header to */*

jsklan:
## Description
Php sdk generator didn't support the x-fern-version openapi flag previously.
Now it passes through the default header into the base client.
Php clients previously included no Accept header by default.
Now they include Accept */* by default, since that's commonly used in popular libraries like Node's fetch function.

## Changes Made
- Updated php sdk RootClientGenerator.ts to extract version header name and default value if present
- Updated php sdk RawClient.Template.php to include default accept header */*

## Testing

- ~~[ ] Unit tests added/updated~~
  - [x] seed test --fixture version --generator php-sdk
  - [x] seed run against customer lib
  - [x] verify requests with newly generated customer lib work and version is selectable


fern-support (approved):

dsinghvi (approved):
@jsklan not required for now, but would check the TS SDK for request options and client arguments on how it handles version

PR #8635, opened 2025-08-08
fix(csharp): Generate explicit types everywhere, and let `dotnet format` clean it up

fearthecowboy:
## Description
- C# generator was making some guesses about how explicit to generate namespace prefixes on things, but when you get a spec that has a bunch of endpoints with the same name, it was not being specific enough, and the compiler gets confused easily
- Rather than try to make decisions at generation time about that I have altered the generator to be far more explicit most of the time, and then introduced a step at the end of generation to use the dotnet format tool with a specially crafted `.editorconfig` to have it automagically trim down the explicit namespace prefixes as much as possible and remove unnecessary `using` statements from the generated code.
- (added that to the Dynamic Snippets generation too)
- The code that gets passed thru `dotnet format` is considerably cleaner-looking (ie, unnecessary `global::` prefixes and general namespace prefixes are removed (or truncated) when not necessary
- Regenerated the `csharp-model` and `csharp-sdk` and the same fixtures are still passing just fine (with cleaner-looking generated code)
- Properly handles nested types much cleaner now, since it knows about its enclosing type
- found/fixed a couple of incorrect namespaces in referenced classes.

## Testing
- [x] Regenerated the `csharp-model` and `csharp-sdk` and the same fixtures are still passing


generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
```diff
@@ -4,12 +4,12 @@ exports[`snippets (default) > examples > 'GET /metadata (allow-multiple)' 1`] =
 "using Acme;
 using Acme.Service;
 
-var client = new AcmeClient(
+var client = new Acme.AcmeClient(
```
Swimburger: Is there a conflict that specifying `Acme.` is necessary here? Can't it use `AcmeClient` from the `using Acme;`?

dsinghvi:
> (added that to the Dynamic Snippets generation too)

We can't leverage csharpier in the browser for dynamic snippets so this will dramatically affect the code snippets being generated. Can you confirm with @Swimburger that we won't be some how introducing regressions

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
fearthecowboy: It now generates much more explicit namespaces everywhere, and relies on the `dotnet format` tool to narrow it down. 

It's not reliable to try and detect when things can be narrowed down.

fearthecowboy: Now, I'm not applying `dotnet format` to the snippets, so they have to be more explicit.

Swimburger (approved):

PR #8668, opened 2025-08-11
feat(swift): explicitly import Foundation

kafkas:
## Description

A simple PR that removes the `Prelude.swift` “as is” file, which previously imported Foundation implicitly. This caused all client code to also import Foundation by default, which could lead to issues later on.

## Changes Made

- Removed `Prelude.swift` as is file
- Added explicit `import Foundation` statements to all source files
- Updated seed snapshots

## Testing


kafkas:
ouch, 'as is' files double-import Foundation. Fixing shortly

kafkas:
Done

dsinghvi (approved):

PR #8671, opened 2025-08-11
fix(csharp): Generate reference.md as long as there are *some* endpoints

fearthecowboy:
## Description
The generator would fail to generate a `reference.md` file if *any* endpoints are empty, which is sub optimal.

## Changes Made
- only throw (and hence, not generate a `reference.md` file if there are no valid endpoints.
- fixes a path bug with using `pnpm seed --local` where it wasn't finding the `features.yml` file

## Testing


generators/base/src/AbstractGeneratorAgent.ts
```diff
@@ -9,6 +9,7 @@ import { FernGeneratorCli } from "@fern-fern/generator-cli-sdk";
 
 import { GeneratorAgentClient } from "./GeneratorAgentClient";
 import { ReferenceConfigBuilder } from "./reference";
+import { existsSync } from "fs";
```
dsinghvi: nit: lets use `fs/promises` everywhere

generators/base/src/AbstractGeneratorAgent.ts
```diff
@@ -125,10 +125,24 @@ export abstract class AbstractGeneratorAgent<GeneratorContext extends AbstractGe
         return undefined;
     }
 
-    private getFeaturesConfigPath(): string {
-        if (process.env.NODE_ENV === "test") {
-            return path.join(__dirname, "../../features.yml");
+    private async getFeaturesConfigPath(): Promise<string> {
+        // try to find the features.yml file
+        // in local test mode, the features.yml is relative to the generator
+        for (const each of [
```
Swimburger: features.yml can also be at `./assets/features.yml`, which is the case for TypeScript SDK generator.

generators/base/src/AbstractGeneratorAgent.ts
```diff
@@ -125,10 +125,24 @@ export abstract class AbstractGeneratorAgent<GeneratorContext extends AbstractGe
...
+                const stats = await stat(each);
+                if (stats.isFile()) {
+                    return each;
+                }
+            } catch (error) {
+                // ignore
+            }
         }
+        // fallback to docker features config path (this is the original default)
         return DOCKER_FEATURES_CONFIG_PATH;
```
Swimburger: I know this var name already existed, let's rename it to not assume that it is in a container.

generators/base/src/AbstractGeneratorAgent.ts
```diff
@@ -125,10 +125,24 @@ export abstract class AbstractGeneratorAgent<GeneratorContext extends AbstractGe
         return undefined;
     }
 
-    private getFeaturesConfigPath(): string {
-        if (process.env.NODE_ENV === "test") {
-            return path.join(__dirname, "../../features.yml");
+    private async getFeaturesConfigPath(): Promise<string> {
```
Swimburger: Let's store all the possible paths in a const in the module scope, and for-loop and use `readFile` instead of `stat` in a try catch. If successful, return, otherwise try the next. If none are found, return undefined.

generators/csharp/sdk/src/reference/buildReference.ts
```diff
@@ -13,17 +13,20 @@ import { SingleEndpointSnippet } from "../endpoint/snippets/EndpointSnippetsGene
...
             : builder.addSection({ title: getSectionTitle({ service }) });
         const endpoints = getEndpointReferencesForService({ context, serviceId, service });
         for (const endpoint of endpoints) {
+            hasEndpoints = true;
             section.addEndpoint(endpoint);
         }
     });
-
+    if (!hasEndpoints) {
+        throw new Error(`No endpoint references found for any service while building reference.md.`);
```
Swimburger, line 28: Will this be caught somewhere else? If not, we should log this as a warning

generators/csharp/sdk/src/SdkGeneratorContext.ts
```diff
@@ -295,7 +295,7 @@ export class SdkGeneratorContext extends AbstractCsharpGeneratorContext<SdkCusto
         return Object.values(AsIsFiles.Test.Utils);
     }
 
-    public getExampleEndpointCallOrThrow(endpoint: HttpEndpoint): ExampleEndpointCall {
+    public getExampleEndpointCallIfExists(endpoint: HttpEndpoint): ExampleEndpointCall {
```
Swimburger: The return type should be `ExampleEndpointCall | undefined`, and then use `return undefined` instead of `return undefined as unknown as ExampleEndpointCall`

Swimburger (changes requested):
Small changes, but overall looks good

generators/base/src/AbstractGeneratorAgent.ts
fearthecowboy: OK

generators/csharp/sdk/src/reference/buildReference.ts
fearthecowboy: Yes, this gets caught where a different exception was previously, and a message gets printed out.

generators/base/src/AbstractGeneratorAgent.ts
fearthecowboy: I put in several options, should cover it.

generators/base/src/AbstractGeneratorAgent.ts
```diff
@@ -10,7 +10,13 @@ import { FernGeneratorCli } from "@fern-fern/generator-cli-sdk";
 import { GeneratorAgentClient } from "./GeneratorAgentClient";
 import { ReferenceConfigBuilder } from "./reference";
 
-const DOCKER_FEATURES_CONFIG_PATH = "/assets/features.yml";
+const FEATURES_CONFIG_PATHS = [
+  "/assets/features.yml",
+  path.join(__dirname, "../features.yml"),
+  path.join(__dirname, "../assets/features.yml"),
```
Swimburger: I'm not sure what the location is of the executable, but for the TS generator CLI, it'll be `./assets/features.yml` instead of `../assets/features.yml`

generators/base/src/AbstractGeneratorAgent.ts
```diff
@@ -1,4 +1,4 @@
-import { readFile } from "fs/promises";
+import { readFile, stat } from "fs/promises";
```
Swimburger: stat can be removed here

Swimburger (approved):

PR #8673, opened 2025-08-12
chore(cli): upload dynamic ir for snippets

chdeskur:
update the `fern generate --docs` command to upload the dynamic IR for `typescript` and `python` to an s3 bucket

this will allow the frontend to access the resource as we support dynamic snippets in the api explorer


packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -162,18 +167,81 @@ export async function publishDocs({
         },
         async ({ ir, snippetsConfig, playgroundConfig, apiName }) => {
             const apiDefinition = convertIrToFdrApi({ ir, snippetsConfig, playgroundConfig, context });
+            
+            let dynamicIr: Record<string, DynamicIntermediateRepresentation> = {};
+            const languages: GeneratorLanguage[] = ["typescript", "python"];
```
dsinghvi: this should be based on the SDKs that the customer actually has

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -162,18 +167,81 @@ export async function publishDocs({
...
+            for (const language of languages) {
+                try {
+                    // generate language-specific dynamic irs
+                    dynamicIr[language] = convertIrToDynamicSnippetsIr({
+                        ir,
+                        disableExamples: true,
+                        generationLanguage: language
+                    });
+                } catch (error) {
+                    context.logger.debug(`Failed to create dynamic IR for ${language}: ${error}`);
```
dsinghvi: If there are more than 1 APIs, you also want the name of the API in this log message

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -162,18 +167,81 @@ export async function publishDocs({
         },
         async ({ ir, snippetsConfig, playgroundConfig, apiName }) => {
             const apiDefinition = convertIrToFdrApi({ ir, snippetsConfig, playgroundConfig, context });
+            
+            let dynamicIr: Record<string, DynamicIntermediateRepresentation> = {};
```
dsinghvi: nit: would rename to something like `dynamicIRbyLangauge`

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -162,18 +167,81 @@ export async function publishDocs({
...
+                    });
+                } catch (error) {
+                    context.logger.debug(`Failed to create dynamic IR for ${language}: ${error}`);
+                }
+            }
+            
+            if (Object.keys(dynamicIr).length > 0) {
+                // register dynamic ir separately
+                // so we can link it to a specific api name
+                const uploadDynamicIrResponse = await fdr.api.v1.register.registerApiDefinition({
```
dsinghvi: Why are we re-registering the API definition twice? can we just register a single time with the dynamic IR?

chdeskur: we need to register the dynamic IR with the name of the API. the other registration just uses the IR api original name, which always ends up being `api`

fern-support (approved):

dsinghvi (approved):

PR #8675, opened 2025-08-12
feat(typescript): Implement inferred bearer auth

Swimburger:
## Description
Implement inferred bearer auth

## Changes Made
- Use `AuthProvider` interface to enforce a common abstraction for different auth providers
- Implement an `InferredAuthProvider` which is generated based upon the inferred bearer auth IR.
- Improve the wire test generation to include inferred auth options and mock the auth endpoint.

## Testing


generators/typescript/sdk/client-class-generator/src/auth-provider/InferredAuthProviderGenerator.ts
```diff
@@ -0,0 +1,464 @@
...
+                                          undefined
+                                      ),
+                                      ts.factory.createKeywordTypeNode(ts.SyntaxKind.UndefinedKeyword)
+                                  ])
+                              ),
+                              hasQuestionToken: false,
+                              scope: Scope.Private
+                          },
+                          {
+                              name: "authRequestPromise",
```
tstanmay13, line 157: In this file do we have to clear any auth promises for non expired tokens? Any memory leak potential?

generators/typescript/sdk/client-class-generator/src/auth-provider/InferredAuthProviderGenerator.ts
```diff
@@ -0,0 +1,464 @@
...
+                                          ),
+                                          ts.factory.createToken(ts.SyntaxKind.PlusToken),
+                                          ts.factory.createBinaryExpression(
+                                              getDeepProperty({
+                                                  variable: "response",
+                                                  property: this.authScheme.tokenEndpoint.expiryProperty,
+                                                  context
+                                              }),
+                                              ts.factory.createToken(ts.SyntaxKind.AsteriskToken),
+                                              ts.factory.createNumericLiteral("1000")
```
tstanmay13: Is it safe to assume seconds? Unsure with the pattern a bit but can we get milliseconds or absolute timestamps?

seed/ts-sdk/inferred-auth-explicit/src/auth/InferredAuthProvider.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    private async getCachedAuthRequest(): Promise<core.AuthRequest> {
+        if (this.expiresAt && this.expiresAt <= new Date()) {
+            // If the token has expired, reset the auth request promise
+            this.authRequestPromise = undefined;
+        }
+
+        if (!this.authRequestPromise) {
+            this.authRequestPromise = this.getAuthRequestFromTokenEndpoint();
```
tstanmay13, line 42: If we have two threads that go in this block because there are two token refreshes. 
Thread 1 - creates a promise at T and tries to return that authRequestPromise 
Thread 2 - creates a promise at T + 1 and returns that same authRequestPromise -- does this override thread 1?

Swimburger: JavaScript is single-threaded, so we don't have to worry about multi-threading and locking.

tstanmay13 (approved):

generators/typescript/sdk/client-class-generator/src/auth-provider/InferredAuthProviderGenerator.ts
Swimburger: We're only storing one promise at a time, and the recommendation is to just dereference it and not worry about cleanup.
I am updating the code to clear the promise and the expiry on error tho.

generators/typescript/sdk/client-class-generator/src/auth-provider/InferredAuthProviderGenerator.ts
Swimburger: I'm worried about this too, but this is how the existing OAuth generated auth works already, so I'm making the same assumption here.
We can expand the IR in the future when the assumption falls over.

PR #8734, opened 2025-08-13
feat(swift): implement string literal types

kafkas:
## Description

This PR adds support for string literal types in the Swift generator. Since Swift does not have a native concept of string literal types, the implementation works around this limitation to provide the closest possible behavior.

For inline literal types, we synthesize an enum with a single case inferred from the literal value and nest it inside the struct where it was defined. This avoids polluting the global namespace. For literal alias types, we generate an `enum` instead of a `typealias`, so all references to the alias point directly to the generated enum.

Synthesizing enums on the fly introduces potential name collisions:
1. Between a nested synthesized type and an existing global schema type (e.g. both named `Address`).
2. Between synthesized types themselves (e.g. literals `"usa"` and `"USA"` both producing `Usa`).

To resolve these issues, we introduce a local symbol registry, similar to our existing global registry, to track all synthesized nested types within a given struct. This registry detects both collisions with global schema types and conflicts between nested types themselves. When a type 1 collision occurs, we disambiguate by using fully qualified type names. For type 2 collisions, we rename the symbol names using the same approach we use for schema type name collisions.

## Changes Made

- Updated `getSwiftTypeForTypeReference` to accept a local type registry for nested type references.
- Added `LocalContext` to build struct-specific context during StructGenerator initialization.
- Added `LocalSymbolRegistry` to detect and resolve name collisions for synthesized nested types.
- Implemented several important tests for edge cases like an object type including a `coding-keys` literal value etc.
- Introduced several changes to `SdkGeneratorCli` and `StructGenerator` to synthesize nested enums on the fly
- Updated seed snapshots

## Testing


dsinghvi (approved):

seed/swift-sdk/simple-fhir/Sources/Schemas/Account.swift
```diff
@@ -55,6 +55,10 @@ public struct Account: Codable, Hashable, Sendable {
         try container.encodeIfPresent(self.practitioner, forKey: .practitioner)
     }
 
+    public enum Account: String, Codable, Hashable, CaseIterable, Sendable {
+        case account = "Account"
```
dsinghvi, line 59: @kafkas does it make more sense to do `case Value = "Account"` ?

kafkas: Good question. I went with derived case names for readability and intent.

For example:

```swift
let request = CreateAccountRequest(type: .account)
```

In Swift, when the type is already known from context, it’s idiomatic to use the shorthand member syntax `.account` instead of writing the full `Account.account` and presumably our customers will be doing this in the vast majority of cases. If we used `.value`, you’d see:

```swift
let request = CreateAccountRequest(type: .value)
```

which tells you nothing without jumping to the enum definition. Across a codebase, having many `.value` expressions each referring to completely different underlying strings would be confusing in my opinion.

I think using derived/different case labels keeps the code self-describing and aligns with standard Swift conventions, even though each of these enums will only ever have a single case.

PR #8745, opened 2025-08-14
chore(cli): upload generator config

chdeskur:
- name dynamic IR uploads based on API id
- revert api registration naming to previous behavior (since it is no longer used for naming dynamic IR)
- link dynamic IR to snippets configuration (so it only generates for snippets referenced in the docs)
- add generator configuration context to IR generation
- pass metadata alongside the dynamic IR for further context in `fern-platform` snippets package


packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -402,61 +403,102 @@ function parseBasePath(domain: string): string | undefined {
...
+                                language: generatorInvocation.language,
+                                githubRepo,
+                                publishInfo: {
+                                    ...publishInfo
+                                },
+                                dynamicIR: {
+                                    ...dynamicIR
+                                }
+                            }
+                        };
```
dsinghvi, line 475: can you add `publishInfo` and `githubRepo` to the dynamic IR definition directly as optional params? https://github.com/fern-api/fern/blob/main/packages/ir-sdk/fern/apis/ir-types-latest/definition/dynamic/ir.yml 

Right now you're sending up an untyped payload and the CLI + Docs will have to magically agree on this shape. If we can modify the schema of the dynamic IR directly, we'll be able to enforce the definition

dsinghvi (dismissed):
Requesting a change around storing metadata directly on the dynamic IR schema

aditya-arolkar-swe (approved):

PR #8749, opened 2025-08-14
feat(java): implement basic wire test generator

tstanmay13:
Added wire tests gated by a flag enable-wire-tests 

### Changes

  - Added constants for magic strings
  - Fixed null safety issues
  - Generate error tests (404/500) that compile and run
  - Handle different builder patterns for request objects
  - Clean up duplicated code

 ###  Limitations

  - Uses hardcoded patterns to detect required fields
  - Simple response generation based on endpoint names
  - Complex requests need manual customization


generators/java/sdk/versions.yml
```diff
@@ -1,3 +1,25 @@
...
+        using MockWebServer and are generated for all endpoints in a service.
+        
+        ```yaml
+        customConfig:
+          enable-wire-tests: true
+        ```
+        
+        Generated tests include:
+        - Working 404/500 error tests that compile immediately
+        - Success test templates with TODO comments for customization
```
Swimburger, line 14: Is there a way to detect whether customization is required? I would comment them out or skip generation.

generators/java/sdk/src/main/java/com/fern/java/client/generators/WireTestGenerator.java
```diff
@@ -0,0 +1,727 @@
...
+
+    // Response body templates
+    private static final String ARRAY_RESPONSE = "[{\"id\":\"test-id\",\"value\":\"test-value\"}]";
+    private static final String EMPTY_RESPONSE = "{}";
+    private static final String SUCCESS_RESPONSE =
+            "{\"id\":\"test-id\",\"status\":\"success\",\"createdAt\":\"2024-01-01T00:00:00Z\"}";
+    private static final String DEFAULT_RESPONSE =
+            "{\"id\":\"test-id\",\"name\":\"test-name\",\"status\":\"success\",\"data\":{}}";
+    private static final String NOT_FOUND_ERROR = "{\"error\":\"not_found\",\"message\":\"Resource not found\"}";
+    private static final String INTERNAL_ERROR = "{\"error\":\"internal_error\",\"message\":\"Internal server error\"}";
```
Swimburger: can we not generate these dynamically? I don't think the JSON shape if it even is JSON, can be guaranteed.

Swimburger (changes requested):

generators/java/sdk/src/main/java/com/fern/java/client/generators/WireTestGenerator.java
tstanmay13: In the latest changes:
  - Removed all dynamic response generation (ARRAY_RESPONSE, SUCCESS_RESPONSE, DEFAULT_RESPONSE)
  - Now using only {} empty response for success tests since we can't guarantee JSON shape
  - Error responses kept minimal with just error/message fields for 404/500 tests

  See the generated test - all success responses now use setBody("{}")

generators/java/sdk/versions.yml
tstanmay13: Going to this for now -- based on TypeScript's wire test patterns:
  1. Added shouldGenerateTest() method that detects complex endpoints:
    - Skips endpoints with path parameters, equest bodies, file upload/download endpoint, pagitation etc 
  2. Generates different test types based on complexity:
    - Simple endpoints: Get basic success test with empty response + working error tests
    - Complex endpoints: Get @Disabled skeleton tests with instructions

Swimburger (approved):

PR #8824, opened 2025-08-18
fix(csharp): Support `explicit-namespaces` - generate explicit types everywhere, and let `dotnet format` clean it up

fearthecowboy:
## Description
Added some support for options in c# generation:
Support for `explicit-interfaces`:
- C# generator was making some guesses about how explicit to generate namespace prefixes on things, but when you get a spec that has a bunch of endpoints with the same name, it was not being specific enough, and the compiler gets confused easily
- Rather than try to make decisions at generation time about that I have altered the generator to be far more explicit most of the time

Support for `experimental-dotnet-format`
-Introduced a step at the end of generation to use the dotnet format tool with a specially crafted `.editorconfig` to have it automagically trim down the explicit namespace prefixes as much as possible and remove unnecessary `using` statements from the generated code.
- The code that gets passed thru `dotnet format` is considerably cleaner-looking (ie, unnecessary `global::` prefixes and general namespace prefixes are removed (or truncated) when not necessary
- 
- Regenerated the `csharp-model` and `csharp-sdk` and the same fixtures are still passing just fine (with cleaner-looking generated code)
- Properly handles nested types much cleaner now, since it knows about its enclosing type
- found/fixed a couple of incorrect namespaces in referenced classes.

## Testing
- [x] Regenerated the `csharp-model` and `csharp-sdk` and the same fixtures are still passing


generators/csharp/base/src/project/CsharpProject.ts
```diff
@@ -161,16 +215,33 @@ export class CsharpProject extends AbstractProject<AbstractCsharpGeneratorContex
...
-        await loggingExeca(
-            this.context.logger,
-            csharpier,
-            ["format", ".", "--no-msbuild-check", "--skip-validation", "--compilation-errors-as-warnings"],
-            {
-                doNotPipeOutput: true,
-                cwd: absolutePathToSrcDirectory
-            }
-        );
+        if ((this.context.config.customConfig as BaseCsharpCustomConfigSchema)?.["experimental-dotnet-format"]) {
```
Swimburger: Is there a way to have this feature flag passed into the ctor?
We usually parse the feature flags at the root of the CLI and pass them through constructors.
So we only look at `.customConfig` in that spot.

fearthecowboy: ah, I was supposed to use `this.context.customConfig` .

And I'm moving the evaluation to the constructor.

generators/csharp/base/src/project/CsharpProject.ts
```diff
@@ -47,6 +49,7 @@ export class CsharpProject extends AbstractProject<AbstractCsharpGeneratorContex
             context: this.context,
             sourceConfig: this.context.ir.sourceConfig
         });
+        this.useDotnetFormat = this.context.customConfig["experimental-dotnet-format"] ?? false;
```
Swimburger: Add a function on the `AbstractCsharpGeneratorContext` or `SdkGeneratorContext` to retrieve the flag, keeping it consistent with the other flags

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
Swimburger, line 1: These snippets should not be using fully qualified names

generators/csharp/base/src/project/CsharpProject.ts
fearthecowboy: Ok, did that.

generators/csharp/sdk/versions.yml
```diff
@@ -1,4 +1,18 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.1.12
+  changelogEntry:
+    - type: fix
+      summary: |
+        Add support for configuration 'experimental-fully-qualified-namespaces' to 
+        generate explicit types everywhere.
```
Swimburger, line 7: I'd change `generate explicit types everywhere` to generate fully qualified type references or something similar.

Swimburger (approved):

PR #8835, opened 2025-08-18
fix(csharp): Correct usage of 'global::' in generated code

fearthecowboy:
## Description
- added a `global` flag to `ClassReference` to force the use of the `global::` qualifier.
- changed the `Method` class to use the `global` flag for the `System.Threading.Tasks`  namespace reference instead of using a literal `global::` qualifier.
- made the `Writer` class respect the `global` flag when writing namespace imports.
- `DynamicSnippetsGeneratorContext` and `DynamicTypeLiteralMapper` recognizes well known identifier names and switches to `fullyQualified` references when it would conflict.

## Testing
- added fixture for `csharp-system-collision` which verifies that words like `System` and `Task` can be used in a client without issue
- added fixture for `csharp-namespace-collision` which uses many names/namespaces that can collide, and uses `experimental-fully-qualified-namespaces` and `experimental-dotnet-foramt` to resolve them.

- [x] Seed tests regenerated


seed/csharp-sdk/seed.yml
```diff
@@ -166,6 +166,15 @@ fixtures:
...
         additional-properties: false
       outputFolder: no-additional-properties
+  csharp-namespace-collision:
+    - customConfig:
+        experimental-fully-qualified-namespaces: true
+        experimental-dotnet-format: true
+      outputFolder: fully-qualified-namespaces
+  csharp-system-collision:
+    - customConfig:
+        client-class-name: System
```
Swimburger, line 176: Can we add a type that's called `System` and leave the `client-class-name` default?

seed/csharp-sdk/seed.yml
```diff
@@ -166,6 +166,15 @@ fixtures:
     - customConfig:
         additional-properties: false
       outputFolder: no-additional-properties
+  csharp-namespace-collision:
+    - customConfig:
```
Swimburger, line 170: Let's add another variation called `no-custom-config` without any custom config and make sure that passes

Swimburger (approved):

PR #8838, opened 2025-08-18
feat(go): output Go SDK to custom path

patrickthornton:
## Description
Adds `config.packagePath` to the `config` attribute for Go SDK generation; setting this makes all Go files generate into `config.PackagePath/...` instead of the root-level directory `...`. This exclusively applies to Go files, so `go.mod` and `go.sum`, `snippet.json` and `snippet-templates.json`, `README.md` and `reference.md`, any github workflows, and any dynamic snippets will still generate outside of `config.PackagePath`.

## Changes Made
- Changes in `go/`: addition of `packagePath` to config, prepends `packagePath` to `importPath` just before generation, prepends `packagePath` to filenames of all Go files before returning from generation
- Changes in `go-v2/`: addition of `packagePath` to config, prepends `packagePath` to `importPath` upon reading config, prepends `packagePath` to filenames of all Go files just before persisting

## Testing
Added fixtures.


generators/go-v2/ast/src/utils/resolveRootImportPath.ts
```diff
@@ -24,13 +25,14 @@ function getImportPath({
     config: FernGeneratorExec.config.GeneratorConfig;
     customConfig: BaseGoCustomConfigSchema | undefined;
 }): string {
-    return (
+    let importPath = (
```
tjb9dc:
```suggestion
    const importPath = (
```

generators/go-v2/base/src/project/GoProject.ts
```diff
@@ -64,7 +64,14 @@ export class GoProject extends AbstractProject<AbstractGoGeneratorContext<BaseGo
 
     private async writeGoFiles({ files }: { files: GoFile[] }): Promise<AbsoluteFilePath> {
         await this.mkdir(this.absolutePathToOutputDirectory);
-        await Promise.all(files.map(async (file) => await file.write(this.absolutePathToOutputDirectory)));
+        let outputDir;
+        if (this.context.customConfig.packagePath) {
+            outputDir = path.join(this.absolutePathToOutputDirectory, this.context.customConfig.packagePath);
+        } else {
+            outputDir = this.absolutePathToOutputDirectory;
+        }
```
tjb9dc: can use const here
```suggestion
        const outputDir = this.context.customConfig.packagePath ? path.join(this.absolutePathToOutputDirectory, this.context.customConfig.packagePath) : this.absolutePathToOutputDirectory;
```

generators/go/internal/generator/v2/generator.go
```diff
@@ -50,6 +50,10 @@ func (g *Generator) Run() error {
 	stdout := bytes.NewBuffer(nil)
 	stderr := bytes.NewBuffer(nil)
 	cmd := exec.Command("node", "--enable-source-maps", v2BinPath, configFilepath)
+
+	// If logs are needed from the Go-V2 generator:
+	// cmd.Stdout = io.MultiWriter(stdout, os.Stdout)
+	// cmd.Stderr = io.MultiWriter(stderr, os.Stderr)
```
tjb9dc, line 56: Could you tell me more about this? Is this just a debugging step? Would we want this to always be on if e.g. a `--debug` flag was provided?

seed/go-sdk/seed.yml
```diff
@@ -81,6 +84,9 @@ fixtures:
         packageLayout: flat
         clientName: Acme
         clientConstructorName: New
+    - outputFolder: package-path
+      customConfig:
+        packagePath: inhereplease
```
tjb9dc, line 89: should we test a more `deeply/nested/` package too? Or however package-paths work? I'm not sure what the rules are for package paths

generators/go-v2/ast/src/custom-config/BaseGoCustomConfigSchema.ts
```diff
@@ -5,6 +5,7 @@ import { ModuleConfigSchema } from "./ModuleConfigSchema";
 export const BaseGoCustomConfigSchema = z.object({
     module: ModuleConfigSchema.optional(),
     packageName: z.string().optional(),
+    packagePath: z.string().optional(),
```
tjb9dc: Should we do some validation of the `packagePath` here? (Or wherever else we do that.) E.g. do we assume it's a relative path, then what would happen if someone put in `/` etc

generators/go/cmd/fern-go-sdk/main.go
```diff
@@ -38,8 +40,9 @@ func run(config *cmd.Config, coordinator *coordinator.Client) ([]*generator.File
 		config.SnippetFilepath,
 		config.ClientName,
 		config.ClientConstructorName,
-		config.ImportPath,
+		path.Join(config.ImportPath, config.PackagePath),
```
tjb9dc: this doesn't _really_ matter but doing the join here makes it feel to me like we're changing the meaning of importPath, but it's still called importPath in the config (maybe for back-compat)? Like I wonder if in our internal type we could rename it to make it more obvious that it's been joined and might be different from the importPath. Like even `resolvedImportPath` lol

Might be more helpful to chat about this one sync and it's not that important, but it did stand out to me

tjb9dc (approved):

generators/go-v2/ast/src/custom-config/BaseGoCustomConfigSchema.ts
tjb9dc: Caught up sync about the wide world of Zod validators, @patrickthornton had the great idea that we would want to look make sure there's no `../` funny-business

generators/go/internal/generator/v2/generator.go
patrickthornton: it's more of a note for later - currently the go-v1 generator calls the go-v2 generator to outsource some of its work, but the logs outputted by go-v2 get totally swallowed. something we gotta get back to and think about a good solution for

generators/go-v2/ast/src/custom-config/BaseGoCustomConfigSchema.ts
patrickthornton: agreed!

generators/go/cmd/fern-go-sdk/main.go
patrickthornton: agreed as well, changed to `FullImportPath`

generators/go/internal/cmd/cmd.go
```diff
@@ -289,6 +293,39 @@ type moduleConfig struct {
 	Imports map[string]string `json:"imports,omitempty"`
 }
 
+// ValidateRelativePath checks if a path is relative and valid
+func validateRelativePath(path string) error {
```
tjb9dc, line 297: Did this work better than doing it using Zod? I was wondering if we could validate directly on the Zod config shape and then it might be more discoverable (single place to do up-front validation) with nice error messages (Zod will tell you which field is bad, etc). I guess it all depends on the code paths tho. If you want, we could pair on it tomorrow morning

patrickthornton: actually, you're totally right - i was originally doing some checks that zod couldn't, but i've since binned those

generators/go-v2/ast/src/custom-config/RelativePathSchema.ts
```diff
@@ -0,0 +1,58 @@
...
+        (pathStr) => {
+            // check for overlong paths
+            if (pathStr.length > 260) {
+                return false;
+            }
+            return true;
+        },
+        {
+            message: `path too long (max 260 chars)`
+        }
```
tjb9dc, line 55: nit: Would prefer to use the built-in zod `max` method for consistent error messages (as we converge on zod validations over time), see https://zod.dev/api#strings

generators/go-v2/ast/src/custom-config/RelativePathSchema.ts
```diff
@@ -0,0 +1,58 @@
...
+            const invalidChars = ["<", ">", ":", "\"", "|", "?", "*"];
+            for (const char of invalidChars) {
+                if (pathStr.includes(char)) {
+                    return false;
+                }
+            }
+            return true;
+        },
+        {
+            message: `path contains invalid characters`
```
tjb9dc, line 42: This doesn't, like, _really_ matter but I'm surprised you're using backticks here instead of double quotes. But that's not really a you thing, just surprised we don't have a linter rule for it.

I don't think you should fix this though, until we have a lint rule for it

tjb9dc (approved):
still LGTM, mentioned a couple nits on zod thingz

PR #8878, opened 2025-08-20
fix(typescript): use type import for AuthProvider

ammubhave:
## Description
This PR fixes the following error
```
export 'AuthProvider' (reexported as 'AuthProvider') was not found in './AuthProvider.js' (module has no exports)
```
when SWC is used to build the source.

## Changes Made
- AuthProvider export now uses export type syntax.

## Testing


dsinghvi (approved):

ammubhave:
Looks like this also impacts builds that use Next.js 15 "--turbopack" beta build, which fail with "Export AuthProvider doesn't exist in target module"

PR #8882, opened 2025-08-20
chore(python): bump IR version to 59

aditya-arolkar-swe:
## Description

Updated the Python generator and all it's sub-generators (FastAPI servers and Pydantic models) to IR version 59.

## Changes Made

- Updated `pyproject.toml` (and subsequently auto-generated the changes in `poetry.lock`) to use v59
- Updated version tags in all python generators
- updated `security_file_generator.yml` to not break on the new inferred auth scheme
  - TODO: Python generator will need a refactor to actually implement inferred auth 

## Testing

  - [x] successfully generates python SDK
  - [x] successfully generates the FastAPI server code


generators/python/pyproject.toml
```diff
@@ -13,7 +13,7 @@ ordered-set = "^4.1.0"
 pydantic-core = "^2.18.2"
 fern-fern-fdr-sdk = { version = "0.98.20", source = "fern-prod" }
 fern-fern-generator-cli-sdk = { version = "0.0.18", source = "fern-prod" }
-fern_fern_ir_v58 = "58.3.0"
+fern_fern_ir_v58 = "59.0.0"
```
tjb9dc, line 16: I think this should be:
```suggestion
fern_fern_ir_v59 = "59.0.0"
```

Did this break something elsewhere in the tests? I would expect things to not work without the right package

generators/python/sdk/versions.yml
tjb9dc, line 1: yay glad to see we're formatting this now! Hopefully this is coming from biome and everyone has the same settings

generators/python/pyproject.toml
aditya-arolkar-swe: Yeah good question, I thought the same thing at first. However, it can't seem to find the new fern IR version when I did that:
```
aditya.arolkar@Mac python % poetry lock         
Resolving dependencies... (0.2s)

Because fern-python depends on fern_fern_ir_v59 (59.0.0) which doesn't match any versions, version solving failed.
```

I saw that the latest IR versions were documented here:  packages/ir-sdk/fern/apis/ir-types-latest/changelog/CHANGELOG.md

Which included the latest 59.0.0. Manual testing worked with my original change.

aditya-arolkar-swe: @dsinghvi would appreciate your input here - during testing it definitely accessed the new IR since it was trying to grab an inferred_auth from it. I think this _should_ be under a new `fern_fern_ir_v59` but we didn't migrate properly to that yet.

tjb9dc (approved):
lgtm!

PR #8885, opened 2025-08-20
chore(deps): Update pydantic to fix mypy

tjb9dc:
## Description
Fixes mypy (i.e. compile task) by upgrading pydantic

## Changes Made
- Upgrade pydantic to fix mypy crash, this was fixed in https://github.com/pydantic/pydantic/pull/11735

## Testing
Existing tests


generators/python/src/fern_python/generators/sdk/client_generator/base_wrapped_client_generator.py
```diff
@@ -88,6 +89,9 @@ def _create_wrapper_method(
             )
 
         # Use EndpointFunctionGenerator to create the wrapper method
+        if self._package.service is None:
+            raise ValueError("Package must have a service to generate endpoints")
```
tjb9dc, line 94: This was one place I couldn't figure out how to guarantee the narrowed / required type

seed/python-sdk/examples/legacy-wire-tests/src/seed/client.py
tjb9dc, line 1: Kinda interesting what happened in this directory. Looks like it got linted?

generators/python/core_utilities/shared/force_multipart.py
```diff
@@ -1,12 +1,15 @@
-class ForceMultipartDict(dict):
+from typing import Any, Dict
+
+
+class ForceMultipartDict(Dict[str, Any]):
```
tjb9dc, line 4: str, Any is probs fine for our usage but I'd also be down to rip `KT` and `VT` generics

dsinghvi (approved):

PR #8886, opened 2025-08-20
feat(typescript): Flag to allow TS SDK users to flatten body and query request params into a single object

mattblank11:
## Description
Adding a flag (`flattenRequestParameters`) that flattens query + body request parameters

## Changes Made
Updated the logic in `GeneratedRequestWrapperImpl` to flatten request + body parameters if `flattenRequestParameters` is true

## Testing
Added a fixture test (`request-parameters/flatten-request-parameters`) and updated the spec for that folder to compare changes. I was able to successfully build each folder.


generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
```diff
@@ -182,10 +194,9 @@ export class GeneratedRequestWrapperImpl implements GeneratedRequestWrapper {
         for (const pathParameter of this.getPathParamsForRequestWrapper()) {
             const type = context.type.getReferenceToType(pathParameter.valueType);
             const hasDefaultValue = this.hasDefaultValue(pathParameter.valueType, context);
-            const propertyName = this.getPropertyNameOfPathParameter(pathParameter);
```
Swimburger: Why are we not using `this.getPropertyNameOfPathParameter(pathParameter);`, I think the logic is more complicated than the code you replaced it with, considering the different feature flags there are for naming.

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
```diff
@@ -766,4 +813,186 @@ export class GeneratedRequestWrapperImpl implements GeneratedRequestWrapper {
...
+            case "list":
+                return container.list;
+            case "set":
+                return container.set;
+            case "optional":
+                return container.optional;
+            case "nullable":
+                return container.nullable;
+            default:
+                throw new Error(`Unsupported container type: ${container.type}`);
```
Swimburger: We usually use `assertNever` to force exhaustive switch statements

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
```diff
@@ -766,4 +813,186 @@ export class GeneratedRequestWrapperImpl implements GeneratedRequestWrapper {
...
+        
+        if (typeDeclaration?.shape.type === "object") {
+            this.addPropertiesFromTypeDeclaration(properties, typeDeclaration, context);
+        } else {
+            this.addUnflattenedReferencedRequestBodyProperties(properties, referenceToRequestBody, context);
+        }
+    }
+
+    private addUnflattenedInlinedRequestBodyProperties(
+        properties: GeneratedRequestWrapper.Property[],
```
Swimburger: Instead of passing in `properties`, return the computed property and add it to `properties` in the call to this function.

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
```diff
@@ -766,4 +813,186 @@ export class GeneratedRequestWrapperImpl implements GeneratedRequestWrapper {
...
+                name: BODY_PROPERTY_NAME,
+                safeName: BODY_PROPERTY_SAFE_NAME,
+                type: bodyType,
+                isOptional: false,
+                docs: undefined
+            });
+        }
+    }
+
+    private addUnflattenedReferencedRequestBodyProperties(
```
Swimburger: Instead of passing in `properties`, return the computed property and add it to `properties` in the call to this function.

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
```diff
@@ -766,4 +813,186 @@ export class GeneratedRequestWrapperImpl implements GeneratedRequestWrapper {
...
+    ) {
+        return context.type.getTypeDeclaration({
+            typeId: requestBodyType.typeId,
+            fernFilepath: requestBodyType.fernFilepath,
+            name: requestBodyType.name,
+            displayName: requestBodyType.displayName
+        });
+    }
+
+    private addPropertiesFromTypeDeclaration(
```
Swimburger: Instead of passing in `properties`, return the computed properties and add them to `properties` in the call to this function.

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
```diff
@@ -766,4 +813,186 @@ export class GeneratedRequestWrapperImpl implements GeneratedRequestWrapper {
...
+            default:
+                throw new Error(`Unsupported container type: ${container.type}`);
+        }
+    }
+
+    private createIndexedAccessType(parentTypeName: string, propertyName: string): ts.TypeNode {
+        return ts.factory.createIndexedAccessTypeNode(
+            ts.factory.createTypeReferenceNode(
+                ts.factory.createQualifiedName(
+                    ts.factory.createIdentifier("SeedRequestParameters"),
```
Swimburger: This hardcoded `SeedRequestParameters` should probably come from somewhere else

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
```diff
@@ -766,4 +813,186 @@ export class GeneratedRequestWrapperImpl implements GeneratedRequestWrapper {
...
+            const propertyType = context.type.getReferenceToType(property.valueType);
+            const hasDefaultValue = this.hasDefaultValue(property.valueType, context);
+            
+            const typeNode = this.isComplexType(property.valueType)
+                ? this.createIndexedAccessType(typeDeclaration.name.name.pascalCase.safeName, property.name.wireValue)
+                : propertyType.typeNodeWithoutUndefined;
+            
+            properties.push({
+                name: getPropertyKey(property.name.wireValue),
+                safeName: getPropertyKey(property.name.wireValue),
```
Swimburger: Is `wireValue` what we should use here? I think there's a reusable function somewhere to calculate the right name to use

seed/ts-sdk/request-parameters-flattening/flatten-request-parameters/src/api/resources/user/client/requests/CreateUsernameRequest.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * @example
+ *     {
+ *         tags: ["tags", "tags"],
+ *         username: "username",
+ *         password: "password",
+ *         name: "test"
+ *     }
+ */
+export interface CreateUsernameRequest {
```
Swimburger: 🎊

Swimburger (changes requested):

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
mattblank11: Good call, reverted

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
mattblank11: No longer applicable, but good to know going forward!

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
mattblank11: Done!

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
mattblank11: Done!

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
mattblank11: Done!

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
mattblank11: Removed

generators/typescript/sdk/request-wrapper-generator/src/GeneratedRequestWrapperImpl.ts
mattblank11: I didn't see a function anywhere, but I could have overlooked it... I added a helper function to this file but let me know if I overlooked something

seed/ts-sdk/request-parameters/use-default-request-parameter-values/src/api/resources/user/client/requests/CreateUsernameRequest.ts
```diff
@@ -5,12 +5,16 @@
...
  *     {
- *         username: "username",
- *         password: "password",
- *         name: "test"
+ *         tags: ["tags", "tags"],
+ *         body: {
+ *             username: "username",
+ *             password: "password",
+ *             name: "test"
+ *         }
```
Swimburger: The generated sample seems to disagree

seed/ts-sdk/request-parameters/flatten-request-parameters/src/api/resources/user/client/requests/CreateUsernameRequest.ts
```diff
@@ -0,0 +1,19 @@
...
+ *         password: "password",
+ *         name: "test"
+ *     }
+ */
+export interface CreateUsernameRequest {
+    tags: string[];
+    username: string;
+    password: string;
+    name: string;
+}
```
Swimburger, line 19: looks good

seed/ts-sdk/request-parameters/use-default-request-parameter-values/src/api/resources/user/client/requests/CreateUsernameRequest.ts
Swimburger: Should the generated type be like the sample here, since the flatten request is not enabled?

Swimburger (changes requested):

seed/ts-sdk/request-parameters/use-default-request-parameter-values/src/api/resources/user/client/requests/CreateUsernameRequest.ts
fern-support: Damn, looking again now

mattblank11: @Swimburger fixed this, thanks for flagging!

I ran the fixture tests on main with the updated `user.yml` file, which includes query params (main does not). The current types + examples match what we see with the legacy logic on main.

Swimburger (approved):

PR #8889, opened 2025-08-20
feat(java): implement basic wire test generation in java-v2

tstanmay13:
### Description
Implements wire test generation for Java SDKs following TypeScript's architectural approach - using the SDK's own example generation capabilities rather than manually constructing requests.

### Changes
- Wire test generator that leverages `DynamicSnippetsGenerator` for all request building
- JUnit tests with MockWebServer** for API endpoint testing
- Proper multi-line method call extraction** from generated snippets
- Test files organized by service** with proper package structure
- Feature gated behind `enable-wire-tests` configuration flag

### Current Limitations & Future Work
- Response validation: Currently only validates HTTP method, not response body content
- Test filtering: Generates tests for all endpoints (should skip OAuth, streaming, file upload/download)
- Auth support: Only handles bearer token auth (needs OAuth, Basic, Header, Inferred)
- Error testing: Only tests successful responses, not error cases
- Known bug: ArrayList initialization syntax issue in DynamicSnippetsGenerator needs fixing

### Testing
Tested with `examples` fixture which has `enable-wire-tests: true` in seed configuration.

### Next Steps
1. Add response body validation against expected values
2. Implement endpoint filtering for unsupported types
3. Enhance authentication scheme support
4. Add error response testing capabilities


generators/java-v2/sdk/src/wire-tests/WireTestGenerator.ts
```diff
@@ -0,0 +1,292 @@
...
+                if (firstExample) {
+                    const snippet = await this.generateSnippetForExample(firstExample);
+                    endpointSnippets.set(endpoint.id, snippet);
+                }
+            }
+        }
+
+        const testClass = java.codeblock((writer) => {
+            writer.writeLine("import static org.junit.jupiter.api.Assertions.*;");
+            writer.newLine();
```
tjb9dc: Do we typically add newlines manually? I'd expect that to be handled later by formatting and not try to worry about it

generators/java-v2/sdk/src/wire-tests/WireTestGenerator.ts
```diff
@@ -0,0 +1,292 @@
...
+            writer.writeLine("import com.fasterxml.jackson.databind.ObjectMapper;");
```
tjb9dc: Instead of manually writing the imports, should we use `writer.addImport`? Or even add a utility `addImports` that takes an iterable? IMO easier to read that way

generators/java-v2/sdk/src/wire-tests/WireTestGenerator.ts
```diff
@@ -0,0 +1,292 @@
...
+            writer.writeLine("import okhttp3.mockwebserver.MockResponse;");
+            writer.writeLine("import okhttp3.mockwebserver.MockWebServer;");
+            writer.writeLine("import okhttp3.mockwebserver.RecordedRequest;");
+            writer.writeLine("import org.junit.jupiter.api.AfterEach;");
+            writer.writeLine("import org.junit.jupiter.api.BeforeEach;");
+            writer.writeLine("import org.junit.jupiter.api.Test;");
+            writer.newLine();
+
+            writer.writeLine(`public class ${className} {`);
```
tjb9dc: Thoughts on using `.class_` and then `.addMethod()` to add the methods?

tjb9dc: You could probably pull this all up into some sort of factory function which will give you the `Ast.Class` with the setup/teardown methods installed, and then you could add the endpoints to it later

generators/java-v2/sdk/src/wire-tests/WireTestGenerator.ts
```diff
@@ -0,0 +1,292 @@
...
+            return "// Snippet generation failed";
+        }
+
+        try {
+            const snippetRequest = convertDynamicEndpointSnippetRequest(example);
+            const response = await this.dynamicSnippetsGenerator.generate(snippetRequest, { style: Style.Concise });
+            return response.snippet || "// No snippet generated";
+        } catch (error) {
+            this.context.logger.debug(`Failed to generate snippet: ${error}`);
+            return "// Snippet generation failed";
```
tjb9dc: Is this so fragile that it just fails? Are you planning to keep this in?

I'm not sure what we do elsewhere but generally I like to avoid this kind of sprawl and instead just know if the thing is reliable and throw a real error if it fails

generators/java-v2/sdk/src/wire-tests/WireTestGenerator.ts
```diff
@@ -0,0 +1,292 @@
...
+import { RelativeFilePath } from "@fern-api/fs-utils";
+import { java } from "@fern-api/java-ast";
+import { HttpEndpoint, dynamic } from "@fern-fern/ir-sdk/api";
+
+import { SdkGeneratorContext } from "../SdkGeneratorContext";
+import { convertIr } from "../utils/convertIr";
+import { convertDynamicEndpointSnippetRequest } from "../utils/convertEndpointSnippetRequest";
+
+export class WireTestGenerator {
+    private dynamicSnippetsGenerator: DynamicSnippetsGenerator | undefined;
```
tjb9dc: I think I probably prefer to not have this be a class field, and instead pass it through the generate call chain as a param. That way we can kill the check below `if (!this.dynamicSnippetsGenerator)` since we know it will always be defined at that point

generators/java-v2/sdk/src/wire-tests/WireTestGenerator.ts
```diff
@@ -0,0 +1,292 @@
...
+            const snippetRequest = convertDynamicEndpointSnippetRequest(example);
+            const response = await this.dynamicSnippetsGenerator.generate(snippetRequest, { style: Style.Concise });
+            return response.snippet || "// No snippet generated";
+        } catch (error) {
+            this.context.logger.debug(`Failed to generate snippet: ${error}`);
+            return "// Snippet generation failed";
+        }
+    }
+
+    private extractMethodCall(fullSnippet: string): string {
```
tjb9dc, line 215: I'm a bit wary of putting a dependency on snippet syntax for test generation. It may not be obvious to a dev improving snippet code that they are going to break test generation Does that make sense? Is this how it's done in the TS generator?

If it's ~~easy to get this information directly from the IR, I'd much prefer to do that. Like we could have some function `f(IR) -> {client, endpoint, blah}`, and then share it between this and the snippet generation

tjb9dc (commented):
A few things:

1. Some stylistic questions about using the writer class more specifically (for building class, method, imports) as opposed to writing lines
2. Whether we should raise an error if snippet gen fails
3. Would prefer to replace the dependency from snippet text -> tests with reading the IR (might need to share a utility between snippet generation and test generation)

tstanmay13:
Spoke offline and addressed the comments here: 

1. Removed manual newlines - Letting the formatter handle spacing
2. Using writer.addImport() - Proper import management instead of strings
3. Java AST limitations - Investigated java.class_() but it doesn't support fields/annotations yet, kept hybrid approach
4. Extracted factory functions - createTestClassBoilerplate() and createTestMethod() for better modularity
5. Dependency injection - DynamicSnippetsGenerator now passed as parameter
6. Error handling - Failed snippets log warnings and skip endpoints instead of adding placeholders
7. Improved extraction - More robust extractMethodCall() with better parsing for multi-line statements
8.

fern-support (approved):
Approving to unblock, caught up with Tanmay about:

- Should do pre-merge if possible
  - Fix double colon output bug
  - Run `gradle spotlessApply` at the end of v2 generation (or wherever else makes sense)
- Can do in FLUP
  - Using the AST classes directly (for building class and methods)
  - Not parsing the snippets, instead transforming the endpoint defns in the IR into the code (like what the snippet code is doing)

PR #8892, opened 2025-08-20
chore(cli): added support for api keys in query params

aditya-arolkar-swe:
## Description

A client introduced a new security schema:
```
components:
 securitySchemes:
    APIKeyQueryParam:
      type: apiKey
      name: apiKey
      in: query
```
However, we didn't support this and the error thrown wasn't clear. 
      
## Changes Made

- Added support for api keys in query param
- Added an improved error logging for when we fail to convert securitySchemas

## Testing

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/convertSecurityScheme.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/convertSecurityScheme.ts
@@ -1,6 +1,6 @@
+import { assertNever } from "@fern-api/core-utils";
 import { EnumSchema, SecurityScheme, Source } from "@fern-api/openapi-ir";
 import { OpenAPIV3 } from "openapi-types";
-
 import { getExtension } from "../../../getExtension";
 import { convertEnum } from "../../../schema/convertEnum";
 import { convertSchemaWithExampleToSchema } from "../../../schema/utils/convertSchemaWithExampleToSchema";
@@ -13,7 +13,6 @@ import {
     HeaderSecuritySchemeNames,
     SecuritySchemeNames
 } from "../extensions/getSecuritySchemeNameAndEnvvars";
-import { assertNever } from "@fern-api/core-utils";
 
 export function convertSecurityScheme(
     securityScheme: OpenAPIV3.SecuritySchemeObject | OpenAPIV3.ReferenceObject,
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/convertSecurityScheme.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/convertSecurityScheme.ts
@@ -28,19 +28,27 @@ function convertSecuritySchemeHelper(
     securityScheme: OpenAPIV3.SecuritySchemeObject,
     source: Source
 ): SecurityScheme | undefined {
-    if (securityScheme.type === "apiKey" && securityScheme.in === "header") {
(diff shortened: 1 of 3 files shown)


packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/generateIr.ts
```diff
@@ -68,6 +68,9 @@ export function generateIr({
     const securitySchemes: Record<string, SecurityScheme> = Object.fromEntries(
         Object.entries(openApi.components?.securitySchemes ?? {}).map(([key, securityScheme]) => {
             const convertedSecurityScheme = convertSecurityScheme(securityScheme, source);
+            if (convertedSecurityScheme == undefined) {
+                throw new Error(`Security scheme conversion failed for ${key}: ${JSON.stringify(securityScheme)}`);
+            }
             if (convertedSecurityScheme == null) {
                 return [];
```
tjb9dc: what's intended to be the difference between this and the `== null` check below?

aditya-arolkar-swe: Wasn't thinking about the null conditional block - was just trying to solve the undefined case. What happened was the returned schema came out as undefined, then when we try to access `.type` in the code just below we get the error in the CLI output as `Cannot read properties of undefined (reading 'type')`. This was just hoping to be an earlier error pitstop that would clarify that.

tjb9dc: caught up sync about the glorious null vs undefined check and == / === in js

packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/convertSecurityScheme.ts
```diff
@@ -16,52 +16,72 @@ import {
...
+            return SecurityScheme.basic({
+                usernameVariableName:
+                    basicSecuritySchemeNamingAndEnvvar?.username?.name ?? basicSecuritySchemeNaming.usernameVariable,
+                usernameEnvVar: basicSecuritySchemeNamingAndEnvvar?.username?.env,
+                passwordVariableName:
+                    basicSecuritySchemeNamingAndEnvvar?.password?.name ?? basicSecuritySchemeNaming.passwordVariable,
+                passwordEnvVar: basicSecuritySchemeNamingAndEnvvar?.password?.env
+            });
+        } else {
+            throw new Error(`Unsupported HTTP scheme: ${securityScheme.scheme}`);
```
tjb9dc: let's log the key in here too

tjb9dc (approved):

packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/convertSecurityScheme.ts
```diff
@@ -16,52 +16,72 @@ import {
...
+                headerEnvVar: headerNames?.env
+            });
+        } else if (securityScheme.in === "query") {
+            // Handle API key in query parameter
+            // Optionally, you could add support for FernOpenAPIExtension for query param naming/envvar in the future
+            return SecurityScheme.query({
+                queryParameterName: securityScheme.name
+            });
+        } else {
+            throw new Error(`Unsupported API key location ${key}: ${securityScheme.in}`);
```
tjb9dc, line 55: I need to dig in at some point about how we surface CLI errors, I feel like there's a juicier way to do this and make it nicely formatted. But this is okay for now I think

PR #8913, opened 2025-08-20
feat(go): support for `default`s in query parameters

patrickthornton:
## Description
Say you had an endpoint like
```
ListResources:
      path: ""
      method: GET
      request:
        name: ListResourcesRequest
        query-parameters:
          page_limit: 
            type: integer
            default: 25
```
which correctly defines a default value in OpenAPI. Before, we ignored this; now our Go SDK applies this default value to query parameters when it isn't specified by the user in requests.

## Changes Made
- On the `go-v1` side; added to internal file `query.go`, splitting `QueryValues()` into `QueryValues()` and `QueryValuesWithDefaults()`, the latter of which merges the request with a specified map of defaults.
- On the `go-v2` side; added the `useDefaultRequestParameterValues` config flag, which when set to true, directs endpoints to use the `QueryValuesWithDefaults()` call only when at least one relevant default has been specified. Constructs the defaults map by reading the relevant query parameters.

## Testing
An added configuration to the `mixed-case` fixture.


generators/go-v2/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -440,6 +446,71 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
         });
     }
 
+    // Extracts the default field from a TypeReference as a TypeInstantiation.
+    private extractDefaultValue(typeReference: TypeReference): go.TypeInstantiation | undefined {
+        switch (typeReference.type) {
```
tjb9dc, line 451: Is it possible to use a visitor here? Otherwise, I'd prefer we do an exhaustiveness check. It could look like:
```
switch (foo) {
  case "meaningful":
     return extractTheDefault(foo);
  case "union":
  case "alsoNotMeaningful":
    return undefined;
  default:
     assertNever(foo);
}
```

This makes the compiler help us remember to always be extracting if we add more types

generators/go-v2/sdk/src/endpoint/http/HttpEndpointGenerator.ts
```diff
@@ -440,6 +446,71 @@ export class HttpEndpointGenerator extends AbstractEndpointGenerator {
...
+                    },
+                    uint: () => undefined,
+                    uint64: () => undefined,
+                    float: () => undefined,
+                    date: () => undefined,
+                    dateTime: () => undefined,
+                    uuid: () => undefined,
+                    base64: () => undefined,
+                    bigInteger: () => undefined,
+                    _other: () => undefined
```
tjb9dc, line 508: I take it these types don't support defaults? Kinda surprising to me!

generators/go/internal/cmd/cmd.go
```diff
@@ -337,7 +337,7 @@ func customConfigFromConfig(c *generatorexec.GeneratorConfig) (*customConfig, er
 
 	// We use a custom decoder here to validate the custom configuration fields.
 	decoder := json.NewDecoder(bytes.NewReader(configBytes))
-	decoder.DisallowUnknownFields()
+	// decoder.DisallowUnknownFields(); allow v2-side-only config fields (zod will ensure they're valid)
```
tjb9dc: If you plan to remove this forever, could you just delete the `decoder.DisallowUnknownFields();` field? And then leave the comment.

If it's a particularly insightful comment, I like the `NOTE: Blah` and optionally you could do `NOTE(<github_username>): blah`

generators/go/internal/generator/sdk/internal/query_test.go
```diff
@@ -237,3 +237,103 @@ func TestQueryValues(t *testing.T) {
...
+		defaults := map[string]interface{}{
+			"name":    "default-name",
+			"age":     25,
+			"enabled": true,
+		}
+
+		values, err := QueryValuesWithDefaults(&example{
+			Name: "actual-name",
+			Age:  30,
+			// Enabled remains false (zero value), should get default
```
tjb9dc, line 276: do we have test coverage where the default is a non-zero value and we are trying to query a zero value? E.g. this test case but when we're providing `"enabled": false`? Bc in that case the zero-value should prevail

tjb9dc (commented):
Couple nits and then a real question about test/behavior

generators/go/internal/cmd/cmd.go
patrickthornton: that sounds like some syntax i could use!

generators/go-v2/sdk/src/endpoint/http/HttpEndpointGenerator.ts
patrickthornton: weird, right? this is at the level of `FernIr.InternalType`; i know the typescript generator behaves the same way

generators/go-v2/sdk/src/endpoint/http/HttpEndpointGenerator.ts
patrickthornton: agreed!

generators/go/internal/generator/sdk/internal/query_test.go
patrickthornton: this is a great call - i think this is the very reason that response parameter structs in our go sdks have pointers to values as every attribute rather than just the values themselves, because otherwise it would be literally impossible to differentiate unspecified vs. set-to-zero parameters. i'll add some more tests that feature more pointer-heavy and thus more realistic structs

tjb9dc (approved):
Approving pending Patrick pushing up the test he just showed me in person for overriding non-zero defaults with zero values

PR #8955, opened 2025-08-22
fix(cli): Nested protobuf types have correct type ids in converted ir

jsklan:
## Description
Nested protobuf types had invalid type ids when converted to ir.
A proto file like this:
```
syntax = "proto3";

package myPackage.mySubpackage;


service TestService {
  rpc DummyMessage(MyMessage) returns (MyMessage) {}
}

message MyMessage {
    enum MyNestedEnum {
        ...
    }
    message  MyNestedMessage {
        ...
    }
}
```
Would result in these type ids:
- `myPackage.myModule.MyMessage`
- `MyMessage.MyNestedEnum`
- `MyMessage.MyNestedMessage`

When they should be:
- `myPackage.myModule.myPackage.myModule.MyMessage`
- `myPackage.myModule.MyMessage.MyNestedEnum`
- `myPackage.myModule.MyMessage.MyNestedMessage`

## Changes Made
- `packages/cli/generation/protoc-gen-fern/src/converters/ProtofileConverter.ts`
  - Modified keys of nested type in the same we it updates non-nested type records before adding them to ir record list

## Testing
  - [ ] Tested against example


packages/cli/cli/versions.yml
```diff
@@ -1,6 +1,14 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+
 - changelogEntry:
-    - summary: | 
+    - summary: Nested protobuf types have correct type ids in converted ir.
+      type: fix
+  irVersion: 59
+  createdAt: "2025-08-22"
+  version: 0.66.21
```
sahil485: This should be 0.66.22

sahil485 (approved):
code + snapshots diffs look good!

PR #8987, opened 2025-08-25
fix(go): certain values being `Sprintf`ed incorrectly into pointers

patrickthornton:
## Description
Certain values would appear in a code chunk like the following in a `client.go` or `raw_client.go` file:
```go
if pageRequest.Cursor != nil {
	queryParams.Set("page", fmt.Sprintf("%v", pageRequest.Cursor))
}
```
But `pageRequest.Cursor` is a `*int`, not an `int`; this needs an extra `*`, which hopefully this PR provides.

## Changes Made
In `base/src/context/GoValueFormatter.ts`, there were several assignments to `prefix`, each of which would overwrite the last; this PR makes it so that the prefix is prepended to over these assignments rather than continually replaced.

## Testing
Updated the Go snapshots to reflect the change; you can see the only changes are in the places where a dereference would be correct.

### DIFF
--- a/generators/go-v2/base/src/context/GoValueFormatter.ts
+++ b/generators/go-v2/base/src/context/GoValueFormatter.ts
@@ -122,28 +122,26 @@ export class GoValueFormatter {
             isPrimitive = true;
         }
 
-        // merge the two
-        if (unwrapPrefix) {
-            const origPrefix = prefix;
-            prefix = origPrefix
+        // merge prefix and suffix with unwrap logic
+        const mergedPrefix = unwrapPrefix != undefined
+            ? (prefix != undefined
                 ? go.codeblock((writer) => {
-                      writer.writeNode(origPrefix);
-                      writer.writeNode(unwrapPrefix);
-                  })
-                : unwrapPrefix;
-        }
-        if (unwrapSuffix) {
-            const origSuffix = suffix;
-            suffix = origSuffix
+                    writer.writeNode(prefix);
+                    writer.writeNode(unwrapPrefix);
+                })
+                : unwrapPrefix)
+            : prefix;
+        const mergedSuffix = unwrapSuffix != undefined
+            ? (suffix != undefined
                 ? go.codeblock((writer) => {
-                      writer.writeNode(unwrapSuffix);
-                      writer.writeNode(origSuffix);
-                  })
-                : unwrapSuffix;
-        }
+                    writer.writeNode(unwrapSuffix);
+                    writer.writeNode(suffix);
+                })
+                : unwrapSuffix)
+            : suffix;
 
         return {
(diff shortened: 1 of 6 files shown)


generators/go-v2/base/src/context/GoValueFormatter.ts
```diff
@@ -120,6 +122,26 @@ export class GoValueFormatter {
...
+        // merge the two
+        if (unwrapPrefix) {
+            const origPrefix = prefix;
+            prefix = origPrefix
+                ? go.codeblock((writer) => {
+                      writer.writeNode(origPrefix);
+                      writer.writeNode(unwrapPrefix);
+                  })
+                : unwrapPrefix;
+        }
```
tjb9dc: the variable switching here is pretty hard for me to follow, if we could I would strongly prefer to use `const` for everything. E.g. maybe
```suggestion
            const mergedPrefix = prefix
                ? go.codeblock((writer) => {
                      writer.writeNode(prefix);
                      writer.writeNode(unwrapPrefix);
                  })
                : unwrapPrefix;
        }
        // Then use mergedPrefix accordingly
```

I'd love for this whole file to use `const` instead of `let`, but I understand if that's out of scope for this change. You'd probably have to refactor some stuff into functions to do that

generators/go-v2/base/src/context/GoValueFormatter.ts
```diff
@@ -120,6 +122,26 @@ export class GoValueFormatter {
             isPrimitive = true;
         }
 
+        // merge the two
+        if (unwrapPrefix) {
```
tjb9dc: nit: The strictest typescript fellows will prefer the below check for explicitness, unless you're deliberately trying to include falsey values (like "" or 0).

```suggestion
        if (unwrapPrefix != null) {
```

generators/go-v2/base/src/context/GoValueFormatter.ts
tjb9dc: Your code is a pretty tasteful reassignment, but the strictest typescript fellows always prefer immutability even if it means...lots of variables and the chance to use the incorrect one

tjb9dc (approved):
Looks good! Left a couple of typescript tips/nits

generators/go-v2/base/src/context/GoValueFormatter.ts
patrickthornton: i thought you'd mention that - this whole issue arose because of `let` shenanigans, honestly. i can definitely do the mergedPrefix change, that feels strong. i'll keep the other `let`s for now just because i want to get this back to auth0, but definitely a note for later

tjb9dc: 100% that makes a ton of sense

PR #8991, opened 2025-08-25
fix(internal): Fix formatting handlers. Prettier for yaml. Biome for everything else.

jsklan:
## Description
Auto-format on save was causing issues because it was using prettier instead of biome for yaml files.
Biome doesn't support yaml formatting, so we've added a prettier config and a ci check to ensure this stays consistent.
Also updated formatting rules so biome formatting is applied to json files (except those used in tests for snapshot diffs)


package.json
```diff
@@ -29,6 +29,8 @@
     "format": "biome format --fix",
     "format:fix": "biome format --fix",
     "format:check": "biome format",
+    "format:yaml:fix": "prettier --write \"**/*.{yml,yaml}\"",
```
tjb9dc, line 32: hmm this thing takes just under 30s to run on my machine, whereas biome is ~5s. I kinda have mixed feelings about making this part of the core dev workflow. Maybe there's a way to make it a lot faster? Or if we use another tool that's smart about looking just at changed files.

Even if we decide it's too slow, we could always merge this PR and remove it as a required check (for now) and just do this great linting periodically

biome.json
tjb9dc, line 1: wow so meta to lint the biome.json

.prettierignore
```diff
@@ -0,0 +1,61 @@
...
+.vscode/
+__snapshots__/
+*.tempate.*
+*.template.*
+
+# Ignore .definition directories for snapshot tests
+.definition/
+**/__test__/outputs/**
+
+# Ignore all other file types (but allow YAML)
```
tjb9dc, line 31: Is this part of the file necessary since `.prettierrc` is told to only look at yaml files?

tjb9dc (approved):
Looks great and IMO 100% worthwhile to merge this for the spring cleaning formatting and using biome for JSON.

Folks may complain that the YAML formatter is too slow locally to be a part of the core workflow, but we can deal with that later! Separately, I was having some issues getting this working on my machine, so may be worth pairing on that.

PR #8992, opened 2025-08-25
feat(java): add nullable vs optional type distinction with @Nullable annotations

tstanmay13:
## Description
Add support for semantic distinction between `nullable<T>` and `optional<T>` types in Java SDK and Spring generators gated by the `use-nullable-annotation` feature flag.

## Changes Made
- Added feature flag: `use-nullable-annotation` with backward compatibility (default: false)
- Fixed root cause: Updated `PoetTypeNameMapper.visitNullable()` to conditionally return raw type when flag enabled
- Created utility class: `NullableAnnotationUtils` for consistent `@Nullable` annotation handling anywhere in the future
- Updated object generators: Added `@Nullable` annotations to getter methods for nullable fields
- Fixed some bugs I noticed 
  - Removed incorrect `@NotNull` annotations from nullable field builder methods

Currently there is a Spring limitation that i will address in a future PR. The limitation is that the Spring generator's staged builders incorrectly require non-null values for nullable fields. This makes it so developers can't construct objects with null values for fields defined as nullable<T> in their spec. This isn't the case for the SDK generator. 

## Testing

- added `nullable-optional` fixture for both SDK and Spring generators -- you can view this in the file UserProfile.java
- Also tested against old fixtures to make sure there was no regression


generators/java/generator-utils/src/main/java/com/fern/java/utils/NullableAnnotationUtils.java
```diff
@@ -0,0 +1,28 @@
+package com.fern.java.utils;
+
+import com.fern.ir.model.types.ContainerType;
+import com.fern.ir.model.types.TypeReference;
+import com.squareup.javapoet.AnnotationSpec;
+import com.squareup.javapoet.ClassName;
+
+public final class NullableAnnotationUtils {
+
+    private static final ClassName NULLABLE = ClassName.get("org.jetbrains.annotations", "Nullable");
```
tjb9dc, line 10: hm it looks like we already have a dependency on `org.jetbrains.annotations` for this. I'm kind of surprised we chose that (?) would have expected us to roll with the Spring one for the Spring generator, and the most popular one possible for the SDKs.

No need to address in this PR!

tjb9dc (approved):

test-definitions/fern/apis/nullable-optional/definition/nullable-optional.yml
tjb9dc, line 1: can you regen seed for the other languages? Also, are you sure we need this new definition? Any chance we have one floating around that already comprehensively does nullable stuff?

tstanmay13: True! There is a nullable fixture already as well but not as comprehensive.

generators/java/generator-utils/src/main/java/com/fern/java/utils/NullableAnnotationUtils.java
tstanmay13: Will keep in mind I was debating between `org.jetbrains.annotations` as I saw it in our repo already but was thinking of javax.annotation.Nullable - Part of JSR-305 which is widely supported as well.

PR #9018, opened 2025-08-25
chore(ci): remove exhaustive overflow tests

fearthecowboy:
## Description
Removing the integer overflow test from the exhaustive fixture

## Changes Made
Removing the exhaustive integer overflow tests

## Testing

- [x] tests regenerated


tjb9dc (approved):
@fearthecowboy while you're here, could you delete the corresponding test directory from the other generators? I'm not sure if the seed updater job does that..

Alternatively, we could just see what happens 😄 and then improve the tool to handle that? Maybe that would be better

PR #9028, opened 2025-08-25
feat(cli): Add whiteLabel setting to FernIr.ReadmeConfig based on org properties

jsklan:
## Description
Adds whiteLabel setting to FernIr.ReadmeConfig based on org properties

## Changes Made
- added whiteLabel field to FernIr.ReadmeConfig
- assigned value to that field based in LocalWorkspace generation based on org properties fetched from venus

## Testing


packages/cli/generation/ir-generator/src/converters/convertReadmeConfig.ts
```diff
@@ -16,6 +16,7 @@ export function convertReadmeConfig({
         apiReferenceLink: readme.apiReferenceLink,
         bannerLink: readme.bannerLink,
         introduction: readme.introduction,
+        whiteLabel: undefined,
```
dsinghvi: can we default to `false` here

dsinghvi (approved):

PR #9048, opened 2025-08-26
fix(go): fix ignored default values for `page`

patrickthornton:
## Description
The addition of default values clashed with the setting of `page` directly from the `request` struct in `client.go` code for pagination endpoints; now the value is pulled from `queryParams`, which should respect all default-setting logic.

## Changes Made
Changes to `generators/go-v2/sdk/src/endpoint/utils/getPaginationInfo.ts`'s `getPagePropertyInitializer()` function.

## Testing
Updated seed snapshots; only changes are to the pagination codeblocks in the correct spaces; manual testing confirms the `page` default is respected.

### DIFF
--- a/seed/go-sdk/pagination/users/client.go
+++ b/seed/go-sdk/pagination/users/client.go
@@ -11,6 +11,7 @@ import (
 	internal "github.com/pagination/fern/internal"
 	option "github.com/pagination/fern/option"
 	http "net/http"
+	strconv "strconv"
 )
 
 type Client struct {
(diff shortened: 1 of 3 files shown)


seed/go-sdk/pagination/users/client.go
```diff
@@ -635,8 +648,11 @@ func (c *Client) ListWithGlobalConfig(
 		}
 	}
 	next := 1
-	if request.Offset != nil {
-		next = *request.Offset
+	if queryParams.Has("offset") {
```
musicpulpite, line 651: This doesn't look right... overwriting `next` with the value for `offset`...

seed/go-sdk/pagination/users/client.go
```diff
@@ -341,8 +348,11 @@ func (c *Client) ListWithOffsetStepPagination(
 		}
 	}
 	next := 1
-	if request.Page != nil {
-		next = *request.Page
+	if queryParams.Has("page") {
+		var err error
```
musicpulpite, line 352: I see: becuase you are reassigning next not declaring it with `:=` then you have to also predeclare `err`

musicpulpite (changes requested):

seed/go-sdk/pagination/users/client.go
musicpulpite: ^^ This is just a comment not a request for changes

musicpulpite (approved):

PR #9053, opened 2025-08-26
fix(go): regression in client generation code for config case that uses GoV1 generator

musicpulpite:
## Description
Without fully understanding all of the code paths, there is a particular config option that forces client generation using the go v1 generator (line [here](https://github.com/fern-api/fern/blob/8062163a50342b133d4ac6436a4d63ed02750248/generators/go/internal/generator/generator.go#L205)) therefore we still need to apply the logic changes from https://github.com/fern-api/fern/pull/8956 into the go v1 generator also.

> [!NOTE]  
> Open Question: Should I invest the time now into porting over support for the `PackageLayoutFlat` config option to the v2 generator so we don't have core logic split over both generators?

## Changes Made
- Added the new client/subclient structure to the go v1 generator

## Testing

### DIFF
--- a/generators/go/internal/generator/generator.go
+++ b/generators/go/internal/generator/generator.go
@@ -134,15 +134,15 @@ func (g *Generator) Generate(mode Mode) ([]*File, error) {
 	if err != nil {
 		return nil, err
 	}
-    if g.config.PackagePath == "" {
-        return files, nil
-    }
+	if g.config.PackagePath == "" {
+		return files, nil
+	}
 
 	// Somewhat hacky fix; prefix packagePath to all .go file paths after the fact
 	for i := range files {
-	    if strings.HasSuffix(files[i].Path, ".go") {
-	        files[i].Path = path.Join(g.config.PackagePath, files[i].Path)
-	    }
+		if strings.HasSuffix(files[i].Path, ".go") {
+			files[i].Path = path.Join(g.config.PackagePath, files[i].Path)
+		}
 	}
 
 	return files, nil
--- a/generators/go/internal/generator/sdk.go
+++ b/generators/go/internal/generator/sdk.go
@@ -971,7 +971,7 @@ func (f *fileWriter) WriteClient(
 	f.P("type ", clientName, " struct {")
 	f.P("baseURL string")
 	f.P("caller *internal.Caller")
-	f.P("header http.Header")
+	f.P("options *core.RequestOptions")
 	f.P()
 	if len(endpoints) > 0 {
 		f.P("WithRawResponse *", rawClientName)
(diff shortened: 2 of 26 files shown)


generators/go/internal/generator/generator.go
```diff
@@ -134,15 +134,15 @@ func (g *Generator) Generate(mode Mode) ([]*File, error) {
...
+	}
 
 	// Somewhat hacky fix; prefix packagePath to all .go file paths after the fact
 	for i := range files {
-	    if strings.HasSuffix(files[i].Path, ".go") {
-	        files[i].Path = path.Join(g.config.PackagePath, files[i].Path)
-	    }
+		if strings.HasSuffix(files[i].Path, ".go") {
+			files[i].Path = path.Join(g.config.PackagePath, files[i].Path)
+		}
```
patrickthornton, line 145: wow, i need to make sure my IDE is tab-only

generators/go/internal/generator/sdk.go
```diff
@@ -991,8 +991,8 @@ func (f *fileWriter) WriteClient(
 	f.P()
 
 	// Generate the client constructor.
-	f.P("func ", clientConstructorName, "(opts ...option.RequestOption) *", clientName, " {")
-	f.P("options := core.NewRequestOptions(opts...)")
+	f.P("func ", clientConstructorName, "(options *core.RequestOptions) *", clientName, " {")
+	// TODO(williammcadams) are there any code paths where this creates the root client? And therefore needs to include "options := core.NewRequestOptions(opts...)"
```
patrickthornton, line 995: i certainly don't think so now with the v2 generator. this is maybe another reason why we should eventually just try and get the v2 generator to pick up this slack 100%

seed/go-sdk/examples/package-path/dynamic-snippets/example0/snippet.go
```diff
@@ -1,8 +1,8 @@
 package example
 
 import (
-    client "github.com/examples/fern/pleaseinhere/client"
-    option "github.com/examples/fern/pleaseinhere/option"
+    client "github.com/examples/fern/client"
+    option "github.com/examples/fern/option"
```
patrickthornton, line 5: ? is this intentional? i think they should be as they were before in the examples/package-path fixture specifically

seed/go-sdk/imdb/deep-package-path/README.md
```diff
@@ -2,7 +2,7 @@
 
 [![fern shield](https://img.shields.io/badge/%F0%9F%8C%BF-Built%20with%20Fern-brightgreen)](https://buildwithfern.com?utm_source=github&utm_medium=github&utm_campaign=readme&utm_source=Seed%2FGo)
 
-The Seed Go library provides convenient access to the Seed API from Go.
+The Seed Go library provides convenient access to the Seed APIs from Go.
```
patrickthornton: true

seed/go-sdk/examples/package-path/dynamic-snippets/example0/snippet.go
musicpulpite: They absolutely should good catch I'm not sure what's going on here!

seed/go-sdk/imdb/deep-package-path/README.md
musicpulpite: I'm actually not sure where this change is coming from...

musicpulpite:
Realized that this config option is no longer necessary so will fully remove this option rather than adding this retroactive fix

PR #9054, opened 2025-08-26
feat(python): upgrade docker python version to 3.11.13

dsinghvi:
## Description
This PR upgrades python generators to use python version `3.11.13`

## Changes Made
- See files changed

## Testing
Ran seed tests to ensure that the dockers still worked


generators/python/sdk/versions.yml
```diff
@@ -1,5 +1,13 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 # For unreleased changes, use unreleased.yml
+- version: 4.26.2
+  changelogEntry:
+    - summary: |
+        Upgrade docker image to use python 3.11.13 so that users can depend on packages that require python 3.11.
```
tjb9dc, line 6: For bonus points, what would you think about adding a runtime validation that `pyproject_python_version` is not > this version? I think it would just be a bit of parsing and looking at `sys.version_info` (like you could do it dynamically and not have to remember to update the check).

tjb9dc (approved):

PR #9058, opened 2025-08-26
feat(swift): generate readme with dynamic usage snippet

kafkas:
## Description

Adds README generation to the Swift SDK pipeline. The README includes Requirements, Installation, Usage and Contributing sections. Usage snippets are built using our existing generators to ensure consistency with emitted code.

## Changes Made

- Expanded `SwiftProject` to track schema and request types used for synthesizing example values
- Added `TupleLiteral` expression type (needed for example construction)
- Upgraded `@fern-fern/generator-cli-sdk` to 0.0.41 in `base` and `swift-sdk` packages
- Adjusted `SdkGeneratorCli`, `SdkGeneratorContext`, `SwiftGeneratorAgent`, `RootClientGenerator` to support readme gen
- Initialized `ReadmeSnippetBuilder` which contains usage snippet generation
- Updated seed snapshots

## Testing


generators/swift/sdk/src/SdkGeneratorCli.ts
```diff
@@ -66,18 +67,36 @@ export class SdkGeneratorCLI extends AbstractSwiftGeneratorCli<SdkCustomConfigSc
...
+        try {
+            const snippets = await new SnippetJsonGenerator({ context }).generate();
+            const content = await context.generatorAgent.generateReadme({
+                context,
+                endpointSnippets: snippets.endpoints
+            });
+            context.project.addRootFiles(new File("README.md", RelativeFilePath.of(""), content));
+        } catch (e) {
+            context.logger.warn("Failed to generate README.md, this is OK.");
+        }
```
tjb9dc: haha is this okay? Why would it fail to generate?

tjb9dc: Regardless, can we log the exception

generators/swift/codegen/src/ast/Type.ts
```diff
@@ -123,7 +123,7 @@ type InternalType =
     | JsonValue;
 
 export class Type extends AstNode {
-    private internalType: InternalType;
+    public internalType: InternalType;
```
tjb9dc, line 126: ah yeah I mean it doesn't feel too bad how you're using it?

generators/swift/sdk/src/generators/endpoint/snippets/SnippetJsonGenerator.ts
```diff
@@ -0,0 +1,42 @@
...
+                    },
+                    snippet: {
+                        type: "typescript",
+                        client: "some code content for snippet",
+                        _visit: (res) => {
+                            return res.typescript({ client: "ts some code content for snippet" });
+                        }
+                    },
+                    exampleIdentifier: "abc123"
+                }
```
tjb9dc, line 38: is this behind a feature flag or anything? It's not gonna generate this hardcoded stuff in a user-facing way yet right?

tjb9dc (approved):

generators/swift/sdk/src/generators/endpoint/snippets/SnippetJsonGenerator.ts
kafkas: Yeah based on my tests it doesn't. So the problem here is that this `endpoints` array cannot be empty, otherwise readme gen fails. The abstract classes all rely on these `FernGeneratorExec` types which currently doesn't support Swift and I think Deep doesn't wanna extend it since it's being deprecated... So, based on my understanding, we need to add something to that array for the time being...

generators/swift/sdk/src/SdkGeneratorCli.ts
kafkas: Haha yeah this is actually ported as is from the C# generator, but yeah I'll log the exception

generators/swift/codegen/src/ast/Type.ts
kafkas: Had to do this as I just needed to switch over its cases from an external function. Will find a better way.. Maybe this `internalType` shouldn't be internal.... This is also ported from C# btw

generators/swift/sdk/src/SdkGeneratorCli.ts
dsinghvi: Would actually change this log line -- lets omit the `this is OK` at the very least since at this point its no longer okay for fern generators to not produce a readme

kafkas: Oh, saw this after merge. Will update in the next PR

PR #9068, opened 2025-08-27
fix(cli): fix client code not being generated when full project mode is enabled

tstanmay13:
## Description
Fixed critical bug preventing client code generation for enterprise customers when full project mode is enabled. Enterprise customers like Bloomberg were unable to get functional Java SDKs with `fern generate --local`

The issue only affected Java because:
  1. Python has special-case handling for `local-file-system` output that already sets `generateFullProject`
  2. Java relies on the `downloadFiles` mode which was returning `undefined` instead of a proper filesystem config

## Changes 
- Fixed `downloadFiles` output mode to set `generateFullProject` based on organization's `selfHostedSdKs`flag
- Corrected variable name in TypeScript workspace runner (`context` → `interactiveTaskContext`) that was causing some ts failure 

## Testing
- [X] Verified Java SDK generation for Bloomberg includes both Gradle files and client code works locally

### DIFF
--- a/packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
+++ b/packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
@@ -233,7 +233,12 @@ function getPublishConfig({
     }
 
     return generatorInvocation.outputMode._visit({
-        downloadFiles: () => undefined,
+        downloadFiles: () => {
+            return FernIr.PublishingConfig.filesystem({
+                generateFullProject: org?.selfHostedSdKs ?? false,
+                publishTarget: undefined
+            });
+        },
         github: () => undefined,
         githubV2: () => undefined,
         publish: () => undefined,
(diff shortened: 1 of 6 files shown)


packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
```diff
@@ -233,7 +233,12 @@ function getPublishConfig({
     }
 
     return generatorInvocation.outputMode._visit({
-        downloadFiles: () => undefined,
+        downloadFiles: () => {
+            return FernIr.PublishingConfig.filesystem({
+                generateFullProject: org?.selfHostedSdKs ?? false,
+                publishTarget: undefined
```
dsinghvi, line 239: @tstanmay13 we'll eventually want to add a publishTarget for Java here. Can you define that as well? Ideally would be be built near line 220

dsinghvi (approved):
+1 to unblock but we need to support passing in java publish target for the generator to use

packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
tstanmay13: Ack! Will track in another ticket as you have approved and we should unblock first

tstanmay13: https://linear.app/buildwithfern/issue/FER-6325/java-add-publishtarget-to-java-similar-to-how-other-generators-do-it

PR #9074, opened 2025-08-27
fix(internal): Improve changelog validation

jsklan:
## Description
Changelog validation previously only verified that the release was parseable. It now asserts that the new version is valid relative to the previous version.
It also enforces that if a feature type changelog is found, the version must have a change to a level other than patch.

## Testing


generators/rust/model/versions.yml
tjb9dc, line 1: Do we actually want to release these SDKs? Or just make the CI improvement

packages/seed/src/commands/validate/semVerUtils.ts
```diff
@@ -0,0 +1,101 @@
...
+            prerelease ? parseInt(prerelease) : undefined
+        );
+    }
+}
+
+export function hasFeatureLevelSemVerChange(currentVersion: SemVer, previousVersion: SemVer): boolean {
+    return currentVersion.major !== previousVersion.major || currentVersion.minor !== previousVersion.minor;
+}
+
+export function isValidSemVerChange(currentVersion: SemVer, previousVersion: SemVer): boolean {
```
tjb9dc, line 103: can you write some unit tests for this? It's so testable!

tjb9dc (approved):

packages/seed/src/commands/validate/semVerUtils.ts
```diff
@@ -0,0 +1,101 @@
...
+            (currentVersion.prerelease === undefined || currentVersion.prerelease === 0)
+        );
+    }
+    if (currentVersion.prerelease === undefined) {
+        return false;
+    }
+    if (previousVersion.prerelease === undefined) {
+        return true;
+    }
+    return currentVersion.prerelease === previousVersion.prerelease + 1;
```
tjb9dc: nit: can you add a comment for each `if` case? Would make it easier to tell what you're intending to check

generators/rust/model/versions.yml
jsklan: I only changed the ones where CI checks will fail because of the new rules if we don't.

PR #9105, opened 2025-08-28
fix(typescript): grab overrideable root header value from Client as default

aditya-arolkar-swe:
## Description

If a value wasn't provided for an overrideable root header in the request options param, we'd default to null and then exclude it entirely in the headers.

## Changes Made

- This checks the request options first but if it's null will use the header set on initialization 

## Testing


generators/typescript/sdk/versions.yml
```diff
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.9.4
+  changelogEntry:
+    - summary: |
+        Grab overrideable root header value from Client as default.
+      type: chore
```
tjb9dc: IMO this is a fix

I took the liberty of upgrading your PR description

tjb9dc (approved):

seed/ts-sdk/auth-environment-variables/src/api/resources/service/client/Client.ts
```diff
@@ -62,7 +62,7 @@ export class Service {
         let _headers: core.Fetcher.Args["headers"] = mergeHeaders(
             this._options?.headers,
             mergeOnlyDefinedHeaders({
-                "X-Another-Header": requestOptions?.xAnotherHeader,
+                "X-Another-Header": requestOptions?.xAnotherHeader ?? this._options?.xAnotherHeader,
```
Swimburger, line 65: This increases the amount of code generated again. I think we should make sure this is already copied into `this._options?.headers`, and it'll be overwritten if `requestOptions?.xAnotherHeader` is defined.

PR #9109, opened 2025-08-28
feat(ruby): Adding README.md generation to ruby-v2

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6291/ruby-setup-readme-generation
Updated the ruby-v2 generator so it now outputs readme files with generated snippets in Ruby

## Changes Made
- Updated README.md generator for ruby-v2
- Updated CLI to generate readme files for each fixture

## Testing
- Manual inspection of generated seed files


chrismallinson:
Marking draft: generated snippets in fixtures are ruby code, but readme is being generated with go code for some reason. Also leaving draft until https://github.com/fern-api/fern/pull/9108 is merged and this is rebased

seed/ruby-sdk-v2/imdb/README.md
```diff
@@ -0,0 +1,148 @@
...
+        timeout: 5
+    )
+)
+
+# Specify options for an individual request.
+response = client.Imdb.CreateMovie(
+    ...,
+    token: "<YOUR_API_KEY>"
+)
+```
```
dsinghvi: are you sure this snippet is correct? I would test this with an existing Ruby SDK

seed/ruby-sdk-v2/imdb/README.md
```diff
@@ -0,0 +1,148 @@
...
+```ruby
+require "client"
+
+response = client.Imdb.WithRawResponse.CreateMovie(...)
+rescue => error
+    raise error
+end
+
+puts "Got response headers: #{response.headers}"
+```
```
dsinghvi: The Ruby SDK doesnt support response headers so lets delete this section for now

seed/ruby-sdk-v2/imdb/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+client = client.new(
+    max_retries: 1
+)
+
+response = client.foo.find(
+    ...,
+    max_retries: 1
+)
+```
```
dsinghvi: Let's also delete this section since this snippet is not entirely correct and we don't support this in the SDK today

seed/ruby-sdk-v2/imdb/README.md
```diff
@@ -0,0 +1,148 @@
...
+You can choose between different environments by using the `base_url` option. You can configure any arbitrary base
+URL, which is particularly useful in test environments.
+
+```ruby
+require "client"
+
+client = client::Client.new(
+    base_url: "https://example.com"
+)
+```
```
dsinghvi, line 37: Can you format this more like https://github.com/fern-demo/square-ruby-sdk/tree/main?tab=readme-ov-file#environment-and-custom-urls 

You'll need to change the verbiage based on if there are environments configured or not

seed/ruby-sdk-v2/imdb/README.md
chrismallinson: will be removed in next push 👍

seed/ruby-sdk-v2/imdb/README.md
chrismallinson: will be removed in next push 👍

seed/ruby-sdk-v2/imdb/README.md
chrismallinson: @dsinghvi  Still need to test a lot of these per the final comment, but I updated this section in my recent push. It feels a little hacky, but it works. Let me know if this flow and the output text in the case of is acceptable. If you take a look at the commit `Providing examples` then you can see an example of each case. `Multi-url-environment` which has an environment option and `content-type` which does not

seed/ruby-sdk-v2/imdb/README.md
chrismallinson: as discussed on slack, will handle this after this merge and the reference.md merge

fern-support (approved):

PR #9110, opened 2025-08-28
fix(cli): skip OpenAPI 3.1 type: "null" properties instead of failing

tstanmay13:
## Description
OpenAPI 3.1 allows explicit null type definition using `type: "null"` for fields that are always null. The importer was throwing "Failed to convert schema" errors when encountering this valid schema type.

## Changes
Based on team discussion, we're skipping properties with type: "null" entirely rather than including them in the generated SDKs. This approach:
  - Fixes the immediate parsing issue
  - Produces cleaner SDKs (no fields that are always null)
  - Works well for response schemas where null fields are mostly informational
  - Serves a  as a temporary solution until we implement proper null type support the IR

## Testing
Verified fix by testing with and without the change:
- Conversion fails without the fix
- Conversion succeeds with the fix on customer's folderit OpenAPI spec

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertSchemas.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertSchemas.ts
@@ -522,20 +522,6 @@ export function convertSchemaObject(
         });
     }
 
-    // null type (OpenAPI 3.1 spec)
-    if ((schema.type as string) === "null") {
-        return SchemaWithExample.unknown({
-            nameOverride,
-            generatedName,
-            title,
-            description,
-            availability,
-            namespace,
-            groupName,
-            example: null
-        });
-    }
-
     // arrays
     if (schema.type === "array") {
         return convertArray({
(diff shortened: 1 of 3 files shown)


packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertSchemas.ts
```diff
@@ -522,6 +522,20 @@ export function convertSchemaObject(
...
+            nameOverride,
+            generatedName,
+            title,
+            description,
+            availability,
+            namespace,
+            groupName,
+            example: null
+        });
+    }
```
dsinghvi: can you add a fixture in `openapi-ir-to-fern-tests` ? so we can have a test that covers this code path?

dsinghvi (approved):

packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertSchemas.ts
tstanmay13: Will fix in future PR! will merge to get Chris unblocked

tjb9dc (approved):

PR #9113, opened 2025-08-29
fix(ci): Skip seed (other expensive CI jobs) for PR's in "Draft" state

musicpulpite:
## Description
Linear ticket: N/A

Based on this Slack thread we should embrace the habit of preemptively opening PR's as "Draft" PR's and skipping expensive jobs. Thus far we are skipping:

- The seed test jobs (but not the changes prerequisite job)

### DIFF
--- a/.github/workflows/seed.yml
+++ b/.github/workflows/seed.yml
@@ -14,6 +14,7 @@ on:
       - "generators/**"
       - "seed/**"
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - main
     paths:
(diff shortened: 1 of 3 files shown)


.github/workflows/seed.yml
```diff
@@ -14,6 +14,7 @@ on:
       - "generators/**"
       - "seed/**"
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
musicpulpite, line 17: Note that I have to add this condition in here in order to include the "ready_for_review" event since the default triggering events are just the first three ([source](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request))

>By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. To trigger workflows by different activity types, use the types keyword. For more information, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onevent_nametypes).

tjb9dc (approved):

PR #9115, opened 2025-08-29
fix(ci): attempt to fix turbo comile

musicpulpite:
## Description
Linear ticket:

## Changes Made

- 
- 

## Testing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
         uses: ./.github/actions/install
 
       - name: Compile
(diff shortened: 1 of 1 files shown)


musicpulpite:
Closing this as I determined that the only problem with the existing turbo remote caching functionality was that the token was invalidated when the original creator of the token was removed from our Vercel organization. I created a new token with no expiry (still associated to my vercel user account as it appears you CANNOT create tokens tied to the organization at large), updated the secret in Github Actions, and added it to 1Password in the "fern Employees" as "TURBO_TOKEN for Github Actions"

PR #9123, opened 2025-09-01
feat(swift): implement `dynamic-snippets` package

kafkas:
## Description
Linear ticket: [FER-6328](https://linear.app/buildwithfern/issue/FER-6328/implement-dynamic-snippets-package)

Implements dynamic snippet generation for Swift and extends `codegen` with the expression types required to power the new snippet generators.

## Changes Made

- Added `StringLiteral`, `NumberLiteral`, `BoolLiteral`, `DateLiteral`, `UUIDLiteral`, `DictionaryLiteral`, `ArrayLiteral`, `DataLiteral` and `Nop` expression types to `codegen`.
- Introduced a new `dynamic-snippets` package with core components like `DynamicTypeLiteralMapper`, `DynamicTypeMapper`, `EndpointSnippetGenerator` and more.
- Added `eventInfo` and `migration` keys to the `POST /big-entity` test case in `DynamicSnippetsTestRunner` in `browser-compatible-base` to test discriminated unions and enums and updated the relevant snapshots for all languages.
- Updated `CodeBlock` to allow `LineBreak` nodes.
- Implemented `FunctionDeclaration` statement type.
- Moved literal enum name generation/sanitization to `codegen` to allow reuse between `dynamic-snippets` and other packages.
- Implemented changes to generate dynamic Swift snippets for Seed fixtures
- Updated Seed snapshots

## Testing


generators/browser-compatible-base/src/dynamic-snippets/test-utils/DynamicSnippetsTestRunner.ts
```diff
@@ -171,6 +171,19 @@ export class DynamicSnippetsTestRunner {
...
+                                id: "event-12345",
+                                data: {
+                                    key1: "val1",
+                                    key2: "val2"
+                                },
+                                jsonString: "abc"
+                            },
+                            migration: {
+                                name: "Migration 31 Aug",
+                                status: "RUNNING"
```
tjb9dc, line 186: kind of a weird movie, not sure I'd want to go see this one...

generators/swift/codegen/src/ast/Expression.ts
```diff
@@ -103,28 +91,81 @@ type Await = {
...
+
+type BoolLiteral = {
+    type: "bool-literal";
+    value: boolean;
+};
+
+type DateLiteral = {
+    type: "date-literal";
+    timestampSeconds: number;
+};
```
tjb9dc: is `timestampSeconds` an epoch? That would make sense for DatetimeLiteral (depending on what Fern is capable of supporting), but for the `date` primitive (like `1996-06-11`) I think we would want to use a string

tjb9dc: would also probably also recommend using a non-epoch for datetime literal, probably some built-in ISO8601 swift thing and then over the wire it would be a string

generators/swift/codegen/src/helpers/literal-enum/literal-enum.ts
```diff
@@ -0,0 +1,45 @@
...
+    return pascalCase(sanitizedLiteralValue);
+}
+
+/**
+ * Generates a safe enum case label for a string literal enum from the literal value.
+ */
+export function generateEnumCaseLabel(literalValue: string): string {
+    const sanitizedLiteralValue = sanitizeLiteralValue(literalValue);
+    return camelCase(sanitizedLiteralValue);
+}
```
tjb9dc, line 19: do we need to check against keywords at all here?

generators/swift/dynamic-snippets/src/context/DynamicTypeMapper.ts
```diff
@@ -0,0 +1,116 @@
...
+            case "DOUBLE":
+                return swift.Type.double();
+            case "BOOLEAN":
+                return swift.Type.bool();
+            case "STRING":
+                return swift.Type.string();
+            case "DATE":
+                return swift.Type.date();
+            case "DATE_TIME":
+                return swift.Type.date();
```
tjb9dc, line 104: does this work for both of these types?

generators/swift/dynamic-snippets/src/context/DynamicTypeLiteralMapper.ts
```diff
@@ -0,0 +1,554 @@
...
+        as?: DynamicTypeLiteralMapper.ConvertedAs;
+    }): boolean | undefined {
+        const bool =
+            as === "mapKey"
+                ? typeof value === "string"
+                    ? value === "true"
+                    : value === "false"
+                      ? false
+                      : value
+                : value;
```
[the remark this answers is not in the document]
tjb9dc: can we write this with case statements instead

generators/swift/dynamic-snippets/src/context/DynamicTypeMapper.ts
```diff
@@ -0,0 +1,116 @@
...
+    private context: DynamicSnippetsGeneratorContext;
+
+    constructor({ context }: { context: DynamicSnippetsGeneratorContext }) {
+        this.context = context;
+    }
+
+    public convert(args: DynamicTypeMapper.Args): swift.Type {
+        switch (args.typeReference.type) {
+            case "list":
+                return swift.Type.array(this.convert({ typeReference: args.typeReference }));
```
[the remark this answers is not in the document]
tjb9dc: worth a look at this one I think

generators/swift/codegen/src/ast/Expression.ts
kafkas: > is `timestampSeconds` an epoch? That would make sense for DatetimeLiteral (depending on what Fern is capable of supporting), but for the `date` primitive (like `1996-06-11`) I think we would want to use a string

Yep, it is -- and that’s a good point. Storing it as a string makes more sense here.

kafkas: > would also probably also recommend using a non-epoch for datetime literal, probably some built-in ISO8601 swift thing and then over the wire it would be a string

Good point. I wasn’t using the built-in ISO-8601 decoder because handling fractional seconds makes the snippet a bit wordy, and I also wanted to avoid a `try!` statement in the examples. I can just round the fractional seconds instead, which simplifies things. Over the wire it's always a string btw. It'll look like `try! Date("2023-03-10T08:45:30Z", strategy: .iso8601)` in the snippet, which is fine, so I’ll make it work that way.

generators/swift/codegen/src/helpers/literal-enum/literal-enum.ts
kafkas: Nope, Swift allows reserved words as case labels. They just need to be escaped with backticks which we're already doing when rendering the enum: https://github.com/fern-api/fern/blob/6c774c340e57d0c0148c1f82abf313bc05904570/generators/swift/codegen/src/ast/EnumWithRawValues.ts#L68-L70

generators/swift/dynamic-snippets/src/context/DynamicTypeMapper.ts
kafkas: At the moment both map to Date as I was trying to mirror JS behavior, where "YYYY-MM-DD" gets parsed as midnight UTC. But yeah you're right, that's actually a mistake since it turns a calendar date into an instant. I’ll look at splitting `DATE` and `DATE_TIME` properly in a follow-up PR.

generators/browser-compatible-base/src/dynamic-snippets/test-utils/DynamicSnippetsTestRunner.ts
kafkas: Not a movie haha just a "big entity", previously defined here https://github.com/fern-api/fern/blob/main/packages/cli/generation/ir-generator-tests/src/ir/__test__/test-definitions/examples.json

seed/swift-sdk/seed.yml
```diff
@@ -18,6 +23,7 @@ test:
   local:
     workingDirectory: generators/swift
     buildCommand:
+      - chmod +x install-swift-toolchain.sh && ./install-swift-toolchain.sh
```
tjb9dc: could we pull this into installScript? And would probably recommend making the `install-swift-toolchain.sh` file that's checked into the repo executable

kafkas: > could we pull this into installScript?

Yep, I can try that. I tried to mimic `java-v2`, which had `install-gradle.sh` under `buildCommand` for local tests. Re executability, looks like it already is so we can probably drop the runtime `chmod +x`

dsinghvi (approved):

PR #9130, opened 2025-09-02
chore(ci): Turn Off Seed Caching

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6367/turn-off-seed-caching
Added flag to disable caching of seed outputs in CI. Can debug and then turn the flag back to false

## Changes Made
Added cache-disabled variable in cached-seed/action.yaml to force actions to run regardless of seed cache state

## Testing
Pushing up and testing CI outputs, will monitor steps run and check if caching is used


chrismallinson:
replaced by https://github.com/fern-api/fern/pull/9147

PR #9151, opened 2025-09-02
feat(cli): Fern diff command can consider from and to generator versions

jsklan:
## Description
Fern diff command previously only considered a from ir file and a to ir file when calculating the diff.
Now it's possible to input a `from` and a `to` generator version with the intended use to pass in the generator version that was used with the corresponding ir version to generate the sdk.

## Changes Made
- Added params to cli.diff command
- Added utilities to commons/core-utils for handling semvers and handling nullish values.

## Testing
- [x] Manually tested with sample ir definitions

### DIFF
--- a/packages/cli/cli/src/cli.ts
+++ b/packages/cli/cli/src/cli.ts
@@ -361,6 +361,14 @@ function addDiffCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
                     string: true,
                     description: "The previous version of the API (e.g. 1.1.0)"
                 })
+                .option("from-generator-version", {
+                    string: true,
+                    description: "The previous version of the generator (e.g. 1.1.0)"
+                })
+                .option("to-generator-version", {
+                    string: true,
+                    description: "The next version of the generator (e.g. 1.1.0)"
+                })
                 .option("quiet", {
                     boolean: true,
                     default: false,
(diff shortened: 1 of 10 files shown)


packages/cli/cli/src/cli.ts
```diff
@@ -361,19 +360,40 @@ function addDiffCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
...
+                .middleware((argv) => {
+                    if (!haveSameNullishness(argv.fromGeneratorVersion, argv.toGeneratorVersion)) {
+                        throw new Error(
+                            "Both --from-generator-version and --to-generator-version must be provided together, or neither should be provided"
+                        );
+                    }
                 }),
         async (argv) => {
-            const fromVersion = argv.fromVersion != null ? argv.fromVersion : undefined;
+            const fromVersion = undefinedIfNullish(argv.fromVersion);
```
tjb9dc, line 385: call me a boomer but I'm more of a `argv.fromVersion != null ? argv.fromVersion : undefined`

But maybe it's nice to use the function since you use the other function right below

packages/cli/cli/src/cli.ts
```diff
@@ -361,19 +360,40 @@ function addDiffCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
...
+            const generatorVersions = undefinedIfSomeNullish({
```
tjb9dc, line 386: definitely get the appeal of using it here

packages/cli/cli/src/commands/diff/diff.ts
```diff
@@ -70,3 +78,73 @@ async function readIr({
...
+}
+
+// export for testing
+export function diffGeneratorVersions(
+    context: CliContext,
+    generatorVersions: { from: string; to: string } | undefined
+): Result {
+    if (generatorVersions === undefined) {
+        return {
+            bump: "patch",
```
tjb9dc: could this just be `undefined`?

tjb9dc (approved):

packages/cli/cli/src/cli.ts
jsklan: Yeah the idea here sortof went backwards. I wanted the undefinedIfSomeNullish function and then once I saw that, I figured why not have one for the single variable case.

packages/cli/cli/src/commands/diff/diff.ts
jsklan: Yeah... I think that could work. I as basing this off how the ir diff checker worked, but I think I don't need to repeat that pattern. I'll update this in a bit.

PR #9158, opened 2025-09-03
fix(cli): do not fail sdk generation on missing websocket header

dsinghvi:
## Description
Unblocks customer trying to upgrade to the latest version of the TypeScript SDK generator .

## Changes Made
- Don't fail hard when converting websocket header to IR. 

## Testing

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 - changelogEntry:
-    - summary: Example generation ignores headers that don't exist in the channel instead of faling hard.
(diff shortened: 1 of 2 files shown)


packages/cli/cli/versions.yml
```diff
@@ -1,4 +1,11 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- changelogEntry:
+    - summary: Example generation ignores headers that don't exist in the channel instead of faling hard.
```
tjb9dc:
```suggestion
    - summary: Example generation ignores headers that don't exist in the channel instead of failing hard.
```

tjb9dc (approved):

PR #9172, opened 2025-09-03
Niels/ts/generate error wire tests no snapshots

Swimburger:
## Description
Linear ticket:

## Changes Made

- 
- 

## Testing


generators/typescript/sdk/generator/src/test-generator/JestTestGenerator.ts
```diff
@@ -1080,8 +1159,27 @@ function getExpectedResponseBody({
...
+            if (!value.body) {
+                if (neverThrowErrors) {
+                    return code`undefined`;
+                }
+                return code`${getTextOfTsNode(errorReference)}()`;
+            }
+            const needsToJson = false; // doesTypeRequireToJsonForErrorBody({
+            //     body: value.body.shape,
+            //     context
+            // });
```
tjb9dc: should this use `doesTypeRequireToJsonForErrorBody`? Or should we delete

generators/typescript/asIs/tests/mock-server/mockEndpointBuilder.ts
```diff
@@ -191,10 +191,15 @@ class ResponseBuilder implements ResponseStatusStage, ResponseHeaderStage, Respo
...
     public build(): HttpHandler {
         const responseResolver: HttpResponseResolver = () => {
-            return new HttpResponse(this.responseBody, {
+            const response = new HttpResponse(this.responseBody, {
                 status: this.responseStatusCode,
                 headers: this.responseHeaders
             });
+            // if no Content-Type header is set, delete the default text content type that is set
+            if (Object.keys(this.responseHeaders).some((key) => key.toLowerCase() === "content-type") === false) {
+                response.headers.delete("Content-Type");
```
tjb9dc, line 200: will this delete the content-type header regardless of case?

Not that it really matters, just a bit weird to me that we do a case insensitive lookup above

tjb9dc (approved):

generators/typescript/sdk/generator/src/test-generator/JestTestGenerator.ts
Swimburger: Yeah we should, this was no longer necessary

generators/typescript/asIs/tests/mock-server/mockEndpointBuilder.ts
Swimburger: `this.responseHeaders` are the headers that are provided by the spec, so we don't wanna touch the `content-type` if it was explicitly provided. However, if not, it means the default content type is set which is `text/plain` which we don't want.

PR #9211, opened 2025-09-04
fix(cli): Fix nullable wrapping to not be self referential

tjb9dc:
## Description
There is currently a bug where an OpenAPI schema like:
```yml
    MyObjectNullable:
      type:
        - object
        - "null"
      properties:
        prop:
          type: string
```

would produce the following IR:
```json
    "MyObjectNullable": {
      "shape": {
        "type": "alias",
        "value": {
          "type": "nullable",
          "itemType": {
            "type": "id",
            "value": "MyObjectNullable"
          }
        }
      }
    }
```

This is because the inlined inner type would have the same ID as the outer type, and would get overwritten when adding types to the IR. Properly bread-crumbing is one way to fix this, at the cost of a more verbose inner type. I'm open to feedback on other ways to do this.

## Changes Made
- Properly breadcrumb the anyOf nullable extraction to wrap the inner type

## Testing
- Covered by existing IR snapshot tests


packages/cli/api-importers/v3-importer-commons/src/converters/schema/OneOfSchemaConverter.ts
```diff
@@ -370,7 +370,7 @@ export class OneOfSchemaConverter extends AbstractConverter<
 
         const schemaOrReferenceConverter = new SchemaOrReferenceConverter({
             context: this.context,
-            breadcrumbs: this.breadcrumbs,
+            breadcrumbs: [...this.breadcrumbs, "inner"],
```
fern-support, line 373: @tjb9dc `Inner` could potentially cause renames in schemas in SDKs causing breaks

dsinghvi: For example, the Hume fixture is a bit concerning.

dsinghvi: Anything we can do to prevent that?

dsinghvi (approved):

tjb9dc:
succeeded by https://github.com/fern-api/fern/pull/9217

PR #9261, opened 2025-09-08
fix(ci): Adding to allowed go-model failures

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6439/fixci-add-to-go-model-allowed-failures
New, consistent failure popped up. Adding to allowed failures list

## Changes Made
Added to allowed failures list for go-model

## Testing
Saw issue on other PRs, pushing here and if CI passes then this can be considered a pass

### DIFF
--- a/seed/go-model/seed.yml
+++ b/seed/go-model/seed.yml
@@ -54,5 +54,6 @@ allowedFailures:
   - exhaustive
   - literal
   - mixed-case
+  - nullable-optional
   - reserved-keywords
   - trace


dsinghvi (approved):

musicpulpite:
Hope this one wasn't caused by 😢 : https://github.com/fern-api/fern/pull/9107

chrismallinson:
> Hope this one wasn't caused by 😢 : #9107

@musicpulpite the strange thing is that by this time, it should have been required that everything was passing before merge so it had to have passed when I made the updates at the end of the week. Only thing I can think of is that there was a state issue between when my checks ran and when they included your changes here. Something else for us to think about during the CI updates this week

musicpulpite:
> > Hope this one wasn't caused by 😢 : #9107
> 
> @musicpulpite the strange thing is that by this time, it should have been required that everything was passing before merge so it had to have passed when I made the updates at the end of the week. Only thing I can think of is that there was a state issue between when my checks ran and when they included your changes here. Something else for us to think about during the CI updates this week

Yeah good point. I just ran it locally and took a look and it appears unrelated to my PR but that doesn't explain why it was passing for your previously. Better to just ignore it for now!

PR #9264, opened 2025-09-08
fix(ci): Temporarily skip typescript, go and ruby-v2 SDKs seed validation to get checks passing

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6420/ci-remove-final-check-for-diffs-in-seed
Handful of generators have non-deterministic outputs. Add these to an exceptions list to get all checks passing again (and later come back to fix root issue)
Also updating filter for seed tests to run if we change our installation or cached-seed actions (since testing is dependent on those things running)
One more update to poetry lock file to get past seed comparison validation before PR to fix that more permanently

## Changes Made
Added to cached-seed workflow file to skip generators causing issues

## Testing
Pushed to CI and verified

### DIFF
--- a/.github/actions/cached-seed/action.yaml
+++ b/.github/actions/cached-seed/action.yaml
@@ -86,7 +86,7 @@ runs:
       # Note: generators added here are currently generating non-deterministic results.
       # To get us to a passing state in CI, add them as exceptions here. This is not
       # a long-term solution.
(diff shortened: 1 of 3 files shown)


dsinghvi (approved):

chrismallinson:
Posting verification - C# model for example is running validate step after running tests:
image
but ruby-sdk-v2 is not running that step:
image

PR #9265, opened 2025-09-08
feat(cli): respect min/max constraints in auto-generated number examples

aditya-arolkar-swe:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6409/openapi-examples-dont-respect-maxmin

Our auto-generated openapi examples for numbers would default to 1.1 when an example wasn't given. However, this won't respect the min/max constraints if given (e.g. if min = 0 and max = 1)

## Changes Made

- Respects number constraints of `minimum`/`maximum`/`exclusiveMinimum`/`exclusiveMaximum`
- If the default 1.1 doesn't match the constraints, we will randomly generate a number within the allowable range
- if an example is given, that example is used except when it is outside the constraints - then we will generate a new number

## Testing


packages/cli/api-importers/v3-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -391,6 +393,77 @@ export class ExampleConverter extends AbstractConverter<AbstractConverterContext
         };
     }
 
+    /**
+     * Adjusts a number to respect the min/max constraints defined in the schema
+     */
+    private adjustNumberToConstraints(number: number): number {
```
tjb9dc: can you add some unit tests for this?

You'll probably want to refactor the function to take in an optional schemaObj (that's pre-resolved) to make it much easier to test

packages/cli/api-importers/v3-importer-commons/src/converters/ExampleConverter.ts
```diff
@@ -391,6 +393,77 @@ export class ExampleConverter extends AbstractConverter<AbstractConverterContext
...
+                "number:",
+                number.toString()
+            );
+            return number;
+        }
+
+        const { minimum, maximum, exclusiveMinimum, exclusiveMaximum } = schemaObj;
+
+        // Calculate lower bound
+        let lowerBound: number | undefined = undefined;
```
tjb9dc, line 416: nit: if you can, always best to prefer `const` to `let` and avoid the type hinting.

But it's really not that big of a deal if the code works and is tested

tjb9dc (approved):

PR #9273, opened 2025-09-08
fix(swift): distinguish between `date` and `datetime` field decoding

kafkas:
## Description
Linear ticket: [FER-6387](https://linear.app/buildwithfern/issue/FER-6387/decode-date-and-datetime-values-differently)

Implements distinct handling of `date` vs `datetime` fields in the Swift SDK. `date` values are now decoded into a new `CalendarDate` type (RFC 3339 `YYYY-MM-DD`), while `datetime` continues to map to Foundation's `Date`.

## Changes Made

- Implements `CalendarDate` "as is" type with RFC 3339 compliant `YYYY-MM-DD` format
- Updated `QueryParameter` "as is" type to handle `CalendarDate`
- Implemented generator changes to map `date` to `CalendarDate` and `datetime` to `Date`
- Updated dynamic snippets code/tests to use the new calendar date type
- Updated Seed snapshots

## Testing

### DIFF
--- a/generators/swift/base/src/asIs/CalendarDate.swift
+++ b/generators/swift/base/src/asIs/CalendarDate.swift
@@ -69,36 +69,21 @@ public struct CalendarDate: Codable, Hashable, Sendable, CustomStringConvertible
 
     // MARK: - Private Helpers
 
-    /// Validates that the given year, month, and day form a valid calendar date.
+    /// Validates that the given year, month, and day form a valid calendar date using Foundation's Calendar APIs.
     private static func isValidDate(year: Int, month: Int, day: Int) -> Bool {
-        guard year >= 1, year <= 9999,
-            month >= 1, month <= 12,
-            day >= 1
-        else {
+        let calendar = Calendar(identifier: .gregorian)
+        let components = DateComponents(year: year, month: month, day: day)
+
+        guard let date = calendar.date(from: components) else {
             return false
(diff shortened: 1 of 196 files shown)


generators/swift/base/src/asIs/CalendarDate.swift
```diff
@@ -0,0 +1,117 @@
...
+            31,  // December
+        ]
+
+        return day <= daysInMonth[month - 1]
+    }
+
+    /// Determines if the given year is a leap year
+    private static func isLeapYear(_ year: Int) -> Bool {
+        return (year % 4 == 0 && year % 100 != 0) || (year % 400 == 0)
+    }
```
tjb9dc: FWIW for this stuff I could see it being reasonable to convert to the foundation date internally and let it handle some of this. I just get paranoid that we forgot some weird leap-thing that only happens every 163 years

Or if we wanted to have getWeekday functionality etc

tjb9dc (approved):

generators/swift/base/src/asIs/CalendarDate.swift
kafkas: Good call! Foundation's `Date` doesn't help us much here but turns out it has `Calendar` and `DateComponents` APIs to which we can defer this logic.. Pushing the changes shortly

PR #9281, opened 2025-09-09
feat(typescript): try retry-after and x-ratelimit-reset before exponential backoff

patrickthornton:
## Description
Linear ticket:
https://linear.app/buildwithfern/issue/FER-6416/auth0-use-ready-after-etc-instead-of-starting-with-exponential-backoff

Retries for requests now try to use the value in `retry-after` (RFC 7231) or otherwise in `x-ratelimit-reset` (standard practice) before defaulting to exponential backoff.

## Changes Made
Changes to as-is utilities in the TypeScript generator.

## Testing
Four new tests to `requestWithRetries.test.ts`.

### DIFF
--- a/generators/typescript/utils/core-utilities/src/core/fetcher/requestWithRetries.ts
+++ b/generators/typescript/utils/core-utilities/src/core/fetcher/requestWithRetries.ts
@@ -9,6 +9,42 @@ function addJitter(delay: number): number {
     return delay * jitterMultiplier;
 }
 
+function getRetryDelayFromHeaders(response: Response, retryAttempt: number): number {
+    // Check for retry-after header first (RFC 7231)
+    const retryAfter = response.headers.get("retry-after");
+    if (retryAfter) {
+        // Parse as number of seconds...
+        const retryAfterSeconds = parseInt(retryAfter, 10);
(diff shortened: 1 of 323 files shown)


generators/typescript/utils/core-utilities/src/core/fetcher/requestWithRetries.ts
```diff
@@ -9,6 +9,42 @@ function addJitter(delay: number): number {
     return delay * jitterMultiplier;
 }
 
+function getRetryDelayFromHeaders(response: Response, retryAttempt: number): number {
+    // Check for retry-after header first (RFC 7231)
+    const retryAfter = response.headers.get("retry-after");
```
Swimburger: Pretty sure this API is case-insensitive, but headers are capitalized, so I'd use `"Retry-After"`

Swimburger (approved):

generators/typescript/utils/core-utilities/src/core/fetcher/requestWithRetries.ts
patrickthornton: gotcha

PR #9285, opened 2025-09-09
chore(internal): Sparse checkout where possible in CI

tjb9dc:
## Description
Checking out the repo is quite expensive, usually taking about 2 minutes in CI (!)

This is mostly due to two classes of files with different characteristics:

- Snapshot files, with a reasonable number of unreasonably large files (e.g. IRs upwards of 100MB)
- Seed test files, with an unreasonable number (>100k files) of reasonably small files

We can take advantage of [sparse checkout](https://git-scm.com/docs/git-sparse-checkout), especially the new "cone mode" for much faster git operations. But using cone mode requires explicitly providing the names of the folders to check out, so it was helpful to reorganize the snapshot files in a root directory.

## Changes Made
- Move snapshot files to a root level directory
  -  Note that I only did JSON files for now, as .snap files actually can't be moved (the vitest API doesn't support a custom path)
- Fix references to the files (in tests and IDE configuration files)
- Sparse checkout the relevant directories in each GHA workflow

## Testing
- Existing tests


tjb9dc:
Closing this until we decide to revisit

PR #9286, opened 2025-09-09
feat(go): try retry-after and x-ratelimit-reset before exponential backoff

patrickthornton:
## Description
Linear ticket:
https://linear.app/buildwithfern/issue/FER-6416/auth0-use-ready-after-etc-instead-of-starting-with-exponential-backoff

The equivalent of https://github.com/fern-api/fern/pull/9281 for Go.

## Changes Made
Changes to as-is utilities in the Go generator.

## Testing
Four new tests in `retrier_test.go`.


musicpulpite (approved):
Looks great regardless of whether or not you want to address: https://buildwithfern.slack.com/archives/C05J55L8JES/p1757439753625199?thread_ts=1757436328.408959&cid=C05J55L8JES

PR #9291, opened 2025-09-09
fix(ci): Ignore package manager files for seed comparison checks

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6476/ci-ignore-package-manager-files-for-seed-comparison-check
Adding new failures to allowed list
Creating a path to input non-deterministic files to the cached-seed action file so that it can ignore those files for the validation step (typically for lock files)

## Changes Made
- ts-sdk seed.yml updates for new allowed failures
- php-model seed.yml updates for new allowed failures
- cached-seed.yml added new input for files to skip at validation step
- workflows/seed.yml and workflows/update-seed.yml updated to pass nondeterministic files

## Testing
Testing on CI here: https://github.com/fern-api/fern/pull/9280 then cleaned up and made this PR

### DIFF
--- a/.github/actions/cached-seed/action.yaml
+++ b/.github/actions/cached-seed/action.yaml
@@ -91,6 +91,7 @@ runs:
       id: validate
       shell: bash
       run: |
+        # Validate results
         # Split the input string into an array
         IFS=',' read -ra FOLDERS <<< "${{ inputs.validation-exclude-paths }}"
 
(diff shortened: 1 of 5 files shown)


tjb9dc (approved):

.github/actions/cached-seed/action.yaml
```diff
@@ -24,6 +24,10 @@ inputs:
     description: "Whether to skip caching and always regenerate"
     required: false
     default: "true"
+  validation-exclude-paths:
+    description: "Comma separated list of paths to exclude from validation step, typically package manager/lock files"
```
tjb9dc, line 28: up to you if you like comma separated vs newline-separated, they both seem like fine options and I see both in github actions world!

Sad that GHA doesn't support a list input type :/

chrismallinson: spent a decent amount of time trying to hack a list together for this with no luck ☹️ If you don't mind it either way, I will keep it as is to avoid another test cycle and falling out of sync again, but if the devs find it easier to use new lines I can change it down the line once this is merged and enabled in our CI

tjb9dc: oh yeah the comma separated is totally fine!

PR #9298, opened 2025-09-09
chore(internal): Move large files to LFS

tjb9dc:
move 10MB+ files to LFS


musicpulpite (approved):
Approving since you already demonstrated that it worked in CI and if people who use husky complain about the new errors popping up we can deal with it as it comes

PR #9314, opened 2025-09-10
feat(go): generated wiremock tests

musicpulpite:
## Description
Linear ticket: [FER-6290](https://linear.app/buildwithfern/issue/FER-6290/go-implement-wire-tests-using-go-wiremock)

Following the basic structure laid out for Java [here](https://github.com/fern-api/fern/blob/main/generators/java-v2/sdk/src/wire-tests/WireTestGenerator.ts) create a `WireTestGenerator` that creates basic wire tests using [go-wiremock](https://pkg.go.dev/github.com/wiremock/go-wiremock). This feature will be enabled using the `customConfig.enableWireTests` flag.

## TO-DO Items
See pending items in the [[SDKs] Go Wire tests](https://linear.app/buildwithfern/project/sdks-go-wire-tests-3609ed54df72/overview) Linear project. Namely:
1. Fix existing dynamic snippets
2. ~~Refactor to follow pattern established for Java wire tests~~
3. Add request body validation and accurate mock response bodies
4. Refactor to use homebrewed mock server
5. Potentially refactor to co-locate tests with packages

## Changes Made

- Add methods to EndpointSnippetGenerator
- Create WireTestGenerator class to effectively wrap endpoint snippet generation
- Fix golang `AsIs` file functionality

## Testing

- I've generated auth0's go-sdk using FROND and played around with it


generators/go-v2/ast/src/ast/core/Writer.ts
```diff
@@ -94,6 +94,6 @@ export class Writer extends AbstractWriter {
         if (split[0] == null) {
             return s;
         }
-        return split[0].replace(INVALID_GO_IDENTIFIER_TOKEN, "");
+        return split.map((part) => part.replace(INVALID_GO_IDENTIFIER_TOKEN, "")).join("");
```
musicpulpite, line 97: Can't have `go-wiremock` alias as `go`

generators/go-v2/base/src/asIs/test/main_test.go_
```diff
@@ -0,0 +1,57 @@
+package wiremock
```
musicpulpite, line 1: This file is useful to run once for the entire test suite since initializing the wiremock test container does incur overhead. It can then be used in all individual test cases in the package

generators/go-v2/dynamic-snippets/src/DynamicSnippetsGenerator.ts
```diff
@@ -27,13 +28,17 @@ export class DynamicSnippetsGenerator extends AbstractDynamicSnippetsGenerator<
     }
 
     public async generate(
-        request: FernIr.dynamic.EndpointSnippetRequest
+        request: FernIr.dynamic.EndpointSnippetRequest,
+        options?: Options
     ): Promise<FernIr.dynamic.EndpointSnippetResponse> {
-        return super.generate(request);
+        return super.generate(request, options);
```
musicpulpite, line 34: Since options is untyped and optional its the best way for me to pass in new options to the dynamic snippets endpoint generator to create "unit test case" snippets without breaking the abstract interface for all other users of `DynamicSnippetsGenerator`

generators/go-v2/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
```diff
@@ -137,4 +144,24 @@ export class DynamicSnippetsGeneratorContext extends AbstractDynamicSnippetsGene
             importPath: this.rootImportPath
         });
     }
+
+    public static chainMethods(
```
musicpulpite, line 155: I could be convinced that a utility like this should be added onto the `go.FuncInvocation` type itself... open to other ideas here but the usefulness should be clear

generators/go-v2/sdk/src/reference/EndpointSnippetsGenerator.ts
```diff
@@ -21,12 +21,18 @@ export class EndpointSnippetsGenerator {
...
     private readonly snippetsCache: Map<string, EndpointSnippets> = new Map();
+    private snippetsCacheInitialized: boolean = false;
 
     constructor({ context }: { context: SdkGeneratorContext }) {
         this.context = context;
+        this.snippetsCacheInitialized = false;
     }
 
     public async populateSnippetsCache(): Promise<void> {
+        if (this.snippetsCacheInitialized) {
```
musicpulpite: It seems to me that for a really large dynamic ir that this could be an expensive operation that should not be recomputed so having this early exit is a good pattern. But haven't actually done profiling to prove that fact. There are a couple of different places that might call `context.snippetGenerator.populateSnippetsCache()` so didn't want to duplicate the work

generators/go-v2/sdk/src/SdkGeneratorCli.ts
```diff
@@ -59,6 +59,12 @@ export class SdkGeneratorCLI extends AbstractGoGeneratorCli<SdkCustomConfigSchem
 
         await context.snippetGenerator.populateSnippetsCache();
 
+        if (context.config.writeUnitTests) {
```
musicpulpite: Is it valid here to use the base config option `writeUnitTest` (currently unused) or should I create a new more specific `enableWireTests` etc flag? Will we ever want this specificity in the future?

generators/go-v2/sdk/src/reference/EndpointSnippetsGenerator.ts
musicpulpite: NVM this is not necessary

generators/go-v2/sdk/src/SdkGeneratorCli.ts
musicpulpite: The realization is that `writeUnitTests` is enabled by default across all generators (but check me on that) so in order to make this feature "opt-in" to begin with I will need to create a new `customConfig` flag for the go generator probably called `enableWireTests` as well!

generators/go-v2/sdk/build.mjs
```diff
@@ -0,0 +1,29 @@
...
+async function main() {
+  await tsup.build({
+    entry: ["src/cli.ts"],
+    format: ["cjs"],
+    minify: false,
+    outDir: "dist",
+    sourcemap: true,
+    clean: true,
+  });
+  const filesFoldersToCopy = [
```
musicpulpite, line 20: Useful for fixing the AsIs files behavior and matches the pattern used in Typescript [here](https://github.com/fern-api/fern/blob/main/generators/typescript/sdk/cli/build.mjs)

generators/go-v2/sdk/src/wire-tests/WireTestGenerator.ts
```diff
@@ -0,0 +1,291 @@
...
+        const response = await this.dynamicSnippetsGenerator.generate(snippetRequest, {
+            config: { outputWiremockTests: true }
+        });
+        if (!response.snippet) {
+            throw new Error("No snippet generated for example");
+        }
+        return response.snippet;
+    }
+
+    private parseEndpointTestCaseSnippet(fileString: string): [string, Map<string, string>] {
```
musicpulpite, line 118: The endpoint snippet (with `config.outputWiremockTests`) writes a complete test file for the single endpoint test method. I think its more meaningful to keep that behavior and then parse out the function body and imports in the consuming code. We can then collate the test methods by service

generators/go/sdk/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM node:22.12-alpine3.20 AS node
 RUN apk --no-cache add git zip
 RUN git config --global user.name "fern" && git config --global user.email "hey@buildwithfern.com"
 
-COPY generators/go-v2/sdk/dist/cli.cjs /dist/cli.cjs
+COPY generators/go-v2/sdk/dist/ /dist/
```
musicpulpite, line 7: This is necessary for getting the `AsIs` file behavior to work for the v2 go generator. The `AsIs` files get copied into /dist/ along with the javascript bundle during `pnpm dist:cli`.

seed/go-sdk/imdb/with-wiremock-tests/test/imdb_test.go
```diff
@@ -0,0 +1,72 @@
+package wiremock
```
musicpulpite, line 1: Here's the generated tests... which do pass!

generators/go-v2/ast/src/ast/core/Writer.ts
musicpulpite: I guess we do need to be careful that this is not breaking behavior for package consumers.

But I don't think it is since it's just changing the alias we use in generated files, it isn't actual changing package names that the consumer might try to import and use without aliasing. But double check me on that!

tstanmay13: Yea as this is a writer change can this impact any other code generation parts?

musicpulpite:
I'm still looking into why this new seed test fails and why this new dependency adds a "toolchain" line:
```
[go-sdk:imdb:with-wiremock-tests]:  Running script CGO_ENABLED=0 go test ./... on imdb:with-wiremock-tests
go: errors parsing go.mod:
/imdb_with-wiremock-tests/generated/go.mod:5: unknown directive: toolchain
```

tstanmay13 (approved):

PR #9324, opened 2025-09-11
feat(java): add Maven publishTarget support for local filesystem generation

tstanmay13:
## Description
Linear Ticket: https://linear.app/buildwithfern/issue/FER-6325/java-add-publishtarget-to-java-similar-to-how-other-generators-do-it 

Enable Java SDK generators to include Maven publishing configuration in generated build.gradle files for local filesystem generation, similar to Python's PyPI implementation.

## Changes Made

CLI
- Added Maven publishTarget creation for java
- Extract Maven coordinates from generator config supporting two styles -- group/artifact or package-prefix w/ packageName

Java
- extract Maven coordinates from IR's publishTarget using visitor pattern
- Pass coordinates to build.gradle generation
- some input validation in Maven coordinate parser and error logging

## Result
Generated `build.gradle` now includes Maven publishing configuration:
```gradle
publishing {
    publications {
        maven(MavenPublication) {
            groupId = 'com.bloomberg'
            artifactId = 'enterprise-orders-sdk'
            version = '0.0.0' // defaults when there is no version provided 
            from components.java
            // ... POM configuration
        }
    }
    repositories {
        maven {
            url "$System.env.MAVEN_PUBLISH_REGISTRY_URL"
            credentials {
                username "$System.env.MAVEN_USERNAME"
                password "$System.env.MAVEN_PASSWORD"
            }
        }
    }
}
```

## Testing 
- against bloomberg fern def provided in slack thread


generators/java/generator-utils/src/main/java/com/fern/java/AbstractGeneratorCli.java
```diff
@@ -314,6 +315,11 @@ public final void runInDownloadFilesMode(
             IntermediateRepresentation ir,
             K customConfig) {
         runInDownloadFilesModeHook(generatorExecClient, generatorConfig, ir, customConfig);
+
+        // Note: AtomicReference is used here to capture the value from within the visitor pattern
+        // Java requires effectively final variables in anonymous inner classes.
+        final AtomicReference<Optional<MavenCoordinate>> maybeMavenCoordinate = new AtomicReference<>(Optional.empty());
```
tjb9dc: Instead of this pattern, what are your thoughts on returning an object from the map function:
```
generateFullProject: Boolean;
mavenCoordinate: Optional<MavenCoordinate>
```

It should result in the same access pattern and gets rid of the need for the forward declaration (which is mutable) and atomic reference

tjb9dc: I realize that in Java it's a little tedious to declare these sorts of objects but IMO it's better practice

packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
```diff
@@ -224,6 +224,59 @@ function getPublishConfig({
...
+                    context.logger.warn("Java generator has package-prefix configured but packageName is missing");
+                }
+
+                return undefined;
+            })();
+
+            const coordinate = mavenCoordinate ? `${mavenCoordinate.groupId}:${mavenCoordinate.artifactId}` : undefined;
+
+            if (coordinate) {
+                const mavenVersion = version ?? "0.0.0";
```
tjb9dc, line 265: where does this version come from?

tjb9dc (approved):

PR #9334, opened 2025-09-11
feat(cli): Add type to property path in IR

Swimburger:
## Description
Add type to property path in IR

## Testing


packages/cli/generation/ir-migrations/src/migrations/v60-to-v59/migrateFromV60ToV59.ts
```diff
@@ -0,0 +1,266 @@
...
+                                            )
+                                    }
+                                }
+                            })
+                        });
+                    }
+                    default:
+                        assertNever(scheme.configuration.type);
+                }
+                break;
```
tjb9dc, line 149: nit: Why do we need this break statement? I wonder if you did `return assertNever(scheme.configuration.type);` that the type checker would know it's unnecessary

tjb9dc (approved):

packages/cli/generation/ir-migrations/src/migrations/v60-to-v59/migrateFromV60ToV59.ts
Swimburger: Biome complains about it falling through. Biome isn't smart enough here

PR #9335, opened 2025-09-11
feat(java): add comprehensive request/response validation to wire tests

tstanmay13:
## Description
Linear Ticket: https://linear.app/buildwithfern/issue/FER-6516/request-and-response-body-validation

Implement full request and response body validation for Java SDK wire tests to achieve parity with TypeScript's comprehensive testing approach. Previously, Java wire tests only validated HTTP methods without checking actual request/response content.

## Changes Made

- Add WireTestDataExtractor class for clean separation of test data extraction from IR
- Enhance WireTestGenerator to validate both request and response JSON bodies
- Extract test data directly from static IR examples (user-specified and auto-generated)
- Generate tests with `MockWebServer` that validate full HTTP request/response cycle

## Testing
- Successfully tested with examples, wire-tests, and client-side-params fixtures that have the `enable-wire-tests` flag enabled
- All generated tests compile and execute correctly


seed/java-sdk/client-side-params/default/src/test/java/com/seed/clientSideParams/ServiceWireTest.java
```diff
@@ -87,6 +88,13 @@ public void testSearchResources() throws Exception {
         RecordedRequest request = server.takeRequest();
         Assertions.assertNotNull(request);
         Assertions.assertEquals("POST", request.getMethod());
+        
+        // Validate request body
+        String actualRequestBody = request.getBody().readUtf8();
+        String expectedRequestBody = "{\"query\":\"query\",\"filters\":{\"filters\":{\"key\":\"value\"}}}";
```
tjb9dc: It would maybe be nice to print this out as a multiline JSON formatted thing for readability

seed/java-sdk/examples/wire-tests/src/test/java/com/seed/examples/FileNotificationServiceWireTest.java
```diff
@@ -30,10 +31,18 @@ public void teardown() throws Exception {
     public void testGetException() throws Exception {
         server.enqueue(new MockResponse()
             .setResponseCode(200)
-            .setBody("{}"));
-        client.file().notification().service().getException("notification-hsy129x");
+            .setBody("{\"type\":\"generic\",\"exceptionType\":\"Unavailable\",\"exceptionMessage\":\"This component is unavailable!\",\"exceptionStacktrace\":\"<logs>\"}"));
+        var response = client.file().notification().service().getException("notification-hsy129x");
```
tjb9dc: what java version do we target? `var` is in 10+ right?

generators/java-v2/sdk/src/wire-tests/WireTestDataExtractor.ts
```diff
@@ -0,0 +1,253 @@
...
+                    object: (v) => {
+                        const result: Record<string, unknown> = {};
+                        v.properties.forEach((property) => {
+                            result[property.name.wireValue] = this.createRawJsonExample(property.value);
+                        });
+                        return result;
+                    },
+                    union: () => jsonExample,
+                    undiscriminatedUnion: (v) => this.createRawJsonExample(v.singleUnionType),
+                    _other: () => jsonExample
```
tjb9dc: I don't really understand the `WithJsonExample` stuff on type reference.

Could we just use the json from that? How come it's okay to use it verbatim for unions here but not for the other stuff? Since unions can arbitrarily nest the other types

tjb9dc: also do we have test coverage of unions

seed/java-sdk/examples/wire-tests/src/test/java/com/seed/examples/FileNotificationServiceWireTest.java
tstanmay13: it is! We should be compitable with everything 8+. Will fix

seed/java-sdk/client-side-params/default/src/test/java/com/seed/clientSideParams/ServiceWireTest.java
tstanmay13: Agreed

generators/java-v2/sdk/src/wire-tests/WireTestDataExtractor.ts
tstanmay13: Great question! The reason for the different handling is due to how the IR structures these values:
Why we process types differently:
1. Primitives: The IR stores strings as {original: "value"} but we need just "value" for the test validation imo
2. Objects: We recursively process properties to extract their wire format from the IR structure
3. Unions: The jsonExample already contains the complete discriminated union structure exactly as it would be sent over the wire (e.g., {type: "actor", name: "John", id: "123"})

the jsonExample is already in the correct format I think because it uses the discrminator fiedl, has the completed nested strucutre, and nested values. For other types we need to extract the actual values from IR representation, we just want the value.

tstanmay13: The `examples` fixture I tested in has this:

- CastMember (undiscriminated union)
- EventInfo (discriminated union)
- Exception (discriminated union with nested types)

These are all tested in the generated ServiceWireTest.java with the testCreateBigEntity() method that tests thewse types

tjb9dc (approved):

PR #9358, opened 2025-09-12
fix(cli): Add missing reserved keywords for Ruby language.

jsklan:
## Description
Ruby reserved words list was missing important keywords that resulted in base object methods being overridden for on some defined types. e.g. if a type had a property called `obejct_id`

## Changes Made
- updated ruby reserved word list
- added ruby specific seed fixture to demonstrate functionality

## Testing


dsinghvi (approved):

seed/ruby-sdk-v2/client-side-params/lib/seed/service/client.rb
```diff
@@ -33,7 +33,7 @@ def list_resources(request_options: {}, **params)
       #
       # @return [Seed::Types::Types::Resource]
       def get_resource(request_options: {}, **params)
-        _query_param_names = ["include_metadata", "format"]
+        _query_param_names = ["include_metadata", "format"] +%i[include_metadata format]
```
chrismallinson: I think you also need the `include_metadata format` in quotes for this to work in Ruby. Mentioning it here in case I can't find it in the generator code

chrismallinson (approved):
I don't see where your code would have generated the one concern I had so it may be independent of your changes (something already in the code) but otherwise this LGTM. If you agree that comment is an issue and wasn't generated by your new changes then I think a separate linear ticket for that and a merge of this is fine

seed/ruby-sdk-v2/client-side-params/lib/seed/service/client.rb
jsklan: This is weird, since this wasn't part of my change in this PR. the other PR has the change for this code.

jsklan: yeah, i think something got messed up in my local docker state. I'm guessing this isn't getting flagged since failures are allowed. fixing locally now to revert these changes.

jsklan: ok these are reverted now

chrismallinson: LGTM 👍

PR #9361, opened 2025-09-12
fix(go): Disable reserved-keywords fixture for Go and re-enable checks in CI

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6487/go-make-packagego-file-deterministic-and-re-enable-checks-in-ci
Makes Go deterministic and re-enables a handful of CI steps we skipped while it was non-deterministic

## Changes Made
- Update to test fixture so the property starts with a letter, not a number
- Update to seed files
- Re-enabling go in update seed workflow
- Re-enabling reserved-keywords for go in CI seed verification step

## Testing
- Verified go seed generation locally on multiple runs
- CI changes will be verified by CI


seed/go-sdk/seed.yml
```diff
@@ -174,6 +174,7 @@ fixtures:
     - outputFolder: default-values
       customConfig:
         useDefaultRequestParameterValues: true
+  reserved-keywords: []
```
musicpulpite, line 177: Gotcha so this line here overwrites the shared test fixture to be an empty test just for golang?

musicpulpite (approved):

musicpulpite:
Thanks for fixing this!

seed/go-sdk/seed.yml
chrismallinson: yes, it skips the generation for this fixture and doesn't report it in the test outputs

PR #9380, opened 2025-09-13
feat(go): `gettersPassByValue` makes getters pass by value

patrickthornton:
## Description
Linear ticket:
https://linear.app/buildwithfern/issue/FER-6491/auth0-getters-for-client-types-in-go-return-pointers-rather-than-the

Setting `gettersPassByValue: true` makes it so getters generally try and return the dereferenced versions of optional arguments. So, instead of, say:
```go
func (b *BigEntity) GetCastMember() *CastMember {
	if b == nil {
		return nil
	}
	return b.CastMember
}
```
You would have:
```go
func (b *BigEntity) GetCastMember() CastMember {
	if b == nil || b.CastMember == nil {
		return CastMember{}
	}
	return *b.CastMember
}
```
where the best-fitting zero-value for the relevant unwrapped type is returned; this allows users to safely use the return value from a getter without worrying about it being `nil`.

## Changes Made
Mostly changes to the "v1" side of the generator.

## Testing
Tested thoroughly against Auth0's spec, as well as an added fixture on `examples`.


musicpulpite (approved):
Great stuff!

PR #9385, opened 2025-09-15
fix(java): handle nullable request bodies with Optional fields correctly

tstanmay13:
## Description
Linear Ticket: Fixes FER-6479

Fixes Java SDK generation for nullable request bodies wrapped in Optional. The Java generator was incorrectly calling `.isPresent()` directly on request objects that had nullable body fields wrapped in `Optional<>`. This caused compilation errors. 

## Changes Made
- Fixed `OnlyRequestEndpointWriter` to use `request.getBody().isPresent()` instead of `request.isPresent()` for wrapped request types with Optional body fields
- Properly distinguish between 
    - Nullable types: which can be null but aren't wrapped in Optional
    - Optional types: which are wrapped in Java's `Optional<>`

## Testing

- [X] Added fixture `java-nullable-named-request-types` 
- 
You can view the fix in `RawSeedApiClient.java`


generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/OnlyRequestEndpointWriter.java
```diff
@@ -215,39 +216,55 @@ private RequestBodyInitializer(
...
+        /**
+         * Extracts the request body getter name from a wrapped request if it has a ReferencedRequestBodyGetter.
+         *
+         * @return An Optional containing the getter name if a ReferencedRequestBodyGetter is present, empty otherwise
+         */
+        private Optional<String> getRequestBodyGetterName() {
+            if (generatedWrappedRequest != null
+                    && generatedWrappedRequest.requestBodyGetter().isPresent()
+                    && generatedWrappedRequest.requestBodyGetter().get()
+                            instanceof GeneratedWrappedRequest.ReferencedRequestBodyGetter) {
```
tjb9dc, line 228: did you do this instead of a visitor? I get it

tjb9dc (approved):

PR #9392, opened 2025-09-15
chore(cli): upload dynamic IR for php

chdeskur:
we construct a generatorConfig from the generatorInstance, since github publishing mode is not piped through


packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -428,11 +429,46 @@ async function generateLanguageSpecificDynamicIRs({
...
+
+                // construct a generatorConfig for php since it is not parsed by getDynamicGeneratorConfig
+                if (generatorInvocation.language === "php") {
+                    packageName = (generatorInvocation.config as { packageName?: string })["packageName"] ?? "";
+                    dynamicGeneratorConfig = {
+                        apiName: workspace.workspaceName,
+                        organization: organization,
+                        customConfig: generatorInvocation.config
+                    } as dynamic.GeneratorConfig;
+                }
```
dsinghvi, line 458: Why is `getDynamicGeneratorConfig` not handling php?

dsinghvi: ideally the code here should just be: 
```
if (generatorInvocation.language === "php") {
      packageName = (generatorInvocation.config as { packageName?: string })["packageName"] ?? "";
 }
```

chdeskur: because the outputMode is undefined

dsinghvi (approved):

PR #9393, opened 2025-09-15
fix(csharp): Fixing C# seed errors (examples fixture)

fearthecowboy:
## Description

## Changes Made

The C# generator can generate types and namespaces that are not always valid, as well as 
when emitting c# code, it may not correctly ascertian when to fully qualify a type.

This PR adds code to have the C# generator introduce disambiguation and resolution code when 
creating `ClassReference` objects.

- Added a 'precompute' step to the c# generator that walks thru the names of everything it intends 
to generate, which allows it to dynamically change names to avoid conflicts.

- Created a `NameRegistry` class that tracks the names of types and namespaces that are 
  created and used.

- to support the state of the name registry, the csharp.* global functions 
  had to be modified to ensure state is maintained while the generator is running.

  This was done by creating a `CSharp` class that exposed the same interface as the `csharp` global
  functions, but also maintained the state of the name registry. This necessitates that all creation
  of AST types should be going thru the `CSharp` class

  Due to the nature of using `csharp.*` global functions, AST types, Contexts, and other code had to be modified 
  to contain a reference to the `CSharp` instance.

  (Type references to the AST types have been moved to a `ast.` namespace)

- rewrote `ClassReference.writeInternal` to handle dismbiguation when writing out the 
  class reference into source code. 

- modified the `Writer.stringifyImports` be able to get the resolvedNamespace when writing
  out the `using` imports.

- A lot of the functionality the context classes (AbstractCsharpGeneratorContext, et al) had been moved 
  to the `csharp-codegen` package. 

- The context classes (model, SDK, and DSG) have been modified to support the `precompute` step.

- any classes that accessed the `csharp.*` global functions had to be modified to access the `CSharp` instance instead.

- Any references to non-generated types were consolidated into a series of objects in the `CSharp` class that 
  manages consistent definitions to those types and exposes constants/functions to access them.
  
- Removed a large number of hard-coded string-smashed type and namespaces from the generator and replaced them 
  with constants/functions in the `CSharp` class.

- ensured that all the `ClassReference` obejcts are created thru the `CSharp.classReference` factory function.

## Resolved bugs and build errors that are exposed by the `examples` generator:
- in the `EndpointSnippetGenerator` if a request body was marked optional and didn't have 
  a value in the example, the generator would not generate a `null` (even though the parameter
  itself was not optional).

- modified the `DynamicSnippetsGeneratorContext` to use the now-dynamically exapanding known
  identifier support.

- fixed the `DynamicTypeLiteralMapper` to correctly generate the initializer for unions with 
  discriminators. (it had been generating empty constructor calls)

- the `DynamicTypeLiteralMapper` now knows when to fully qualify a type when it is using
  a known identifier.

- When `UnionGenerator` and `RootClientGenerator` emits code to throw an `Exception` it 
  now uses the fully qualified name of the `Exception` class when there is a name collision 
  with a generated model.

- fixed a bug in the HttpEndpointGenerator that would generate faulty code if the a response 
  status code was duplicated

- the `XmlDocBlock` file was getting a circular dependency, changed `ClassReference` to
  be `type` imported.

## Testing

- [X] Added unit tests for the `NameRegistry` class
- [X] Updated test run data
- [X] Regenerated all the seed fixtures for the C# generator


generators/csharp/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
```diff
@@ -274,6 +243,64 @@ export class DynamicSnippetsGeneratorContext extends AbstractDynamicSnippetsGene
...
+                                });
+
+                                break;
+
+                            case "body":
+                                break;
+                        }
+                    }
+                } catch (error) {
+                    // skip endpoints that can't be resolved
```
tjb9dc: When would this happen? Will this have downstream consequences? Should we log a warning? Should we just throw an error?

tjb9dc: Also if we are needing to do this in multiple places after resolving, could be worth making a second function `resolveEndpointLocation()` that returns undefined in that case. That way we avoid the blanket catch

generators/csharp/codegen/src/utils/stacktrace.ts
tjb9dc, line 1: What do we use this file for? Is its output user-facing? If so, I think I'd want to keep output consistent in the CLI (so strive to use this everywhere or nowhere)

generators/csharp/sdk/versions.yml
```diff
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.1.16
```
tjb9dc: Two thoughts on version here:

1. Could this result in renames that are compile breaks? This should probably be at least a minor version bump and possibly a major bump
2. Is it worth doing an RC and testing this on some specs in the wild?

tjb9dc (approved):
Left a few thoughts, thank you for the big refactor!

generators/csharp/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
fearthecowboy: It did happen in one or more of the fixtures, but I can add one that doesn't throw easy enough.

generators/csharp/codegen/src/utils/stacktrace.ts
fearthecowboy: It's for identifying where in code a classRef is created after a precalculate step. 

The output is really only for internal developers - and it won't even output this information right now at all, I have to insert code to write it out.

I kinda wanted to leave this in for a bit until I was sure I got all the places that things were creating references. In the long run it's not likely to be necessary.

tjb9dc: do you have a plan to know about spots where this is hit in the wild?

generators/csharp/sdk/versions.yml
fearthecowboy: That is a really good question.

If something generates cleanly today - I would really not think it would end up doing any disambiguation/renames to avoid collisions. It 

The most likely thing that we'd see different is that it would do minor alterations to the `using` imports and/or change how fully-qualified the names are, but these should all be cosmetic .

I do really like the idea of doing an RC and doing some specs in the wild - I have high confidence, but ... I'd rather know that real specs are coming out as we'd like them to.

Should I just change it to `2.1.16-rc1` ? is that all I need to do?

generators/csharp/codegen/src/utils/stacktrace.ts
fearthecowboy: Well, so the good news is, even if it happens, it's not an actual problem - it will still work really hard to ensure that it's not colliding with things. 

The only potential issue is that if it generates some code with the SDK generator AND then it generates code with the Dynamic Snippet Generator, AND if there is something that gets named in a different order, then the DSG code might not mesh with the SDK generated code perfectly. 

If the DSG isn't used, then it really can't be a problem at all. 

If someone throws something new at it where that does happen, we can look into it pretty quickly.

generators/csharp/sdk/versions.yml
tjb9dc: yep! Let's keep it as a patch version change and 2.1.16-rc0 sounds perfect

Issue #9366, opened 2025-09-12
Update pydantic generated types to use Annotated instead of Field


aryehklein:
### Which Fern component?

SDK Generator

### How important is this?

P1 - High (Strongly needed)

### What's the feature?

From:
```
  AuthGroundTruthPage:
    properties:
      items: list<AuthGroundTruth>
      totalItems: integer
```
Fern generates:
```
class AuthGroundTruthPage(UniversalBaseModel):
    items: typing.List[AuthGroundTruth]
    total_items: int = pydantic.Field(alias="totalItems")
```
Using an alias in camelCase. When I then do:
```
return AuthGroundTruthPage(items=items, total_items=len(items))
```
This flags a pyright issue due to how pyright resolves types. You get :
```
 error: No parameter named "total_items" (reportCallIssue)
```
This is a known issue with type checkers: See https://github.com/pydantic/pydantic/issues/5893#issuecomment-2512807073. The recommended fix is to use Annotated instead which seems to align with Pydantics current recommendation specifically for this use case - https://docs.pydantic.dev/latest/concepts/fields/. I.e., the generated pydantic type would have:
```
    total_items: typing.Annotated[int, pydantic.Field(alias="totalItems")]
```

### Are you interested in contributing this feature?

No

PR #9445, opened 2025-09-16
[feature] Support generating pydantic types with Annotated attributes

aryehklein:
## Description
This PR comes to solve https://github.com/fern-api/fern/issues/9366.

## Changes Made

- Adds support for a new pydantic field use_annotated_field_aliases that generates the pydantic types needed

## Testing


generators/python/src/fern_python/generators/pydantic_model/custom_config.py
```diff
@@ -50,6 +50,13 @@ class BasePydanticModelCustomConfig(pydantic.BaseModel):
 
     use_pydantic_field_aliases: bool = True
 
+    use_annotated_field_aliases: bool = False
```
aryehklein, line 53: High level points:

1. If this looks good, I assume I also need a docs change - can you point me to where to make that? 
2. I added unit tests for this change, but didnt know how/where to do an end to end test with a full on generation?

tjb9dc: Great questions!

1. We document the pydantic config in a separate repo specifically for our docs. You can see an analogous feature flag docs PR here: https://github.com/fern-api/docs/pull/702. We're happy to make that PR though if you'd prefer! You've already been so generous with your time on this contribution.
2. Our strategy for most ETE testing these days uses our framework called seed, which is a regression testing framework that you can read more about in [CONTRIBUTING.md](https://github.com/fern-api/fern/blob/main/CONTRIBUTING.md#contributing).
    - For testing this, I'd probably recommend [adding a new alias-extends fixture](https://github.com/fern-api/fern/blob/169128515c36f4bfca717e6a88d8b587e2c2bfd3/seed/python-sdk/seed.yml#L48) (could also use exhaustive) with this custom config set
    - You'd then generate and check in the seed files `pnpm seed:local test --generator python-sdk --fixture alias-extends --outputFolder <whatever_you_name_the_output_folder>`. This will also run ruff and mypy against the files
    - If you want more test coverage against the SDK, you can check in additional test files to the generated directory and .fernignore them

generators/python/src/fern_python/pydantic_codegen/pydantic_model.py
```diff
@@ -125,7 +127,39 @@ def add_field(self, unsanitized_field: PydanticField) -> None:
...
+            del prev_fields["type_hint"]
+            field = PydanticField(
+                **(field.__dict__),
+                type_hint=annotated_type_hint,
+            )
+
+            # Use only the default value as initializer, not pydantic.Field
+            initializer = default_value
+
+        elif is_aliased and not self._use_pydantic_field_aliases:
```
aryehklein, line 161: There is a bit of complexity here with how this new feature interplays with use_pydantic_field_aliases. If use_pydantic_field_aliases is false, and _use_annotated_field_aliases is true, you get into double annotated types. The FastAPI generator defaults this to true, whereas the use_pydantic_field_aliases defaults to false. I think this if/elif block handles it, but lmk if there are other edge cases to consider!

seed/fastapi/exhaustive/use-annotated-field-aliases/resources/types/resources/object/types/object_with_optional_field.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    integer: typing.Optional[int] = None
+    long_: typing_extensions.Annotated[typing.Optional[int], pydantic.Field(alias="long")] = None
+    double: typing.Optional[float] = None
+    bool_: typing_extensions.Annotated[typing.Optional[bool], pydantic.Field(alias="bool")] = None
+    datetime: typing.Optional[dt.datetime] = None
+    date: typing.Optional[dt.date] = None
+    uuid_: typing_extensions.Annotated[typing.Optional[uuid.UUID], pydantic.Field(alias="uuid")] = None
+    base_64: typing_extensions.Annotated[typing.Optional[str], pydantic.Field(alias="base64")] = None
```
aryehklein, line 25: @tjb9dc This is a key line to look at to understand the diff. This line differs from the existing ones (e.g. https://github.com/fern-api/fern/blob/main/seed/fastapi/exhaustive/pydantic-v2/resources/types/resources/object/types/object_with_optional_field.py#L24) and is the core thing I want to solve.

generators/python/src/fern_python/pydantic_codegen/pydantic_model.py
tjb9dc: As far as I know this is reasonable! I'm good with the approach of just preferring it over the other flag (rather than e.g. breaking the config shape), as long as we document that that's the behavior.

seed/fastapi/exhaustive/use-annotated-field-aliases/core/serialization.py
```diff
@@ -0,0 +1,276 @@
...
+    maybe_annotated_type = _get_annotation(type_)
+
+    if maybe_annotated_type is not None:
+        # The actual annotations are 1 onward, the first is the annotated type
+        annotations = typing_extensions.get_args(maybe_annotated_type)[1:]
+
+        for annotation in annotations:
+            if isinstance(annotation, FieldMetadata) and annotation.alias is not None:
+                return annotation.alias
+    return None
```
tjb9dc, line 265: I do wonder how/if this code will interact with this change. It's called (transitively) from the pydantic_utilities files, and we use different ones depending on the aliasing mode we're in. That decision (also factors in v1 vs v2) lives in `src/fern_python/generators/core_utilities/core_utilities.py`.

I think it's worth testing serializing some of your test models and making sure things look expected. From reading the code I think it's possible that we'll need to use the aliased pydantic utilities, which would mean updating the conditional statement in the three `core_utilities.py` files to provide that version if using the new feature flag you're adding. But definitely easier to figure that out by testing :)

PR #9448, opened 2025-09-16
feat(ruby): ApiError class and hierarchy

chanks:
## Description

Introducing a hierarchy of error classes that SDKs built with the v2 Ruby generator can raise in response to failed SDK calls. Matching the approach laid out in https://github.com/square/square-ruby-sdk/pull/196

## Changes Made

- Add templates for the error class hierarchy.
- Modify the Ruby generator's `AsIs` mechanism so that it can generate files outside `/internal`.
- Raise the new error classes when API calls fail.

## Testing

- Added unit tests for new AST components.
- Manually tested to make sure the gem loads and that error class selection/instantiation works:
```
christopher@fern imdb % irb
irb(main):001:0> require_relative "lib/seed"
=> true
irb(main):002:0> Seed::Errors::ResponseError.subclass_for_code(503)
=> Seed::Errors::ServiceUnavailableError
irb(main):003:0> Seed::Errors::ResponseError.subclass_for_code(504)
=> Seed::Errors::ServerError
irb(main):004:0> Seed::Errors::ServerError::new("body", code: 404)
=> #<Seed::Errors::ServerError: body>
irb(main):005:0> e = _
=> #<Seed::Errors::ServerError: body>
irb(main):006:0> e.code
=> 404
```


generators/ruby-v2/model/src/ModelGeneratorContext.ts
```diff
@@ -45,7 +45,15 @@ export class ModelGeneratorContext extends AbstractRubyGeneratorContext<ModelCus
...
     public getCoreAsIsFiles(): string[] {
         const files = [
-            // Errors
+            // Public errors
+            AsIsFiles.ApiError,
+            AsIsFiles.ClientError,
+            AsIsFiles.RedirectError,
+            AsIsFiles.ResponseError,
+            AsIsFiles.ServerError,
+            AsIsFiles.TimeoutError,
```
chanks: @dsinghvi unsure if these changes are right since I don't know what model generation is for, please let me know if this is wrong.

dsinghvi: Good shout -- its just for generating the underlying types but you should be good to ignore that

dsinghvi (approved):

PR #9487, opened 2025-09-16
feat(go): setters, support for explicit null

patrickthornton:
## Description
Linear ticket: Fixes FER-3852, Fixes FER-6561
Adds setter methods for every object. These setters also mark a bitfield that keeps track of which fields have been set; just before serialization, these fields are made non-optional (i.e. `omitempty` is removed) which allows users to pass explicit null even when the argument is originally marked optional.

## Changes Made
Changes to the `v1` side of the generator; addition of two as-is files, `explicit_fields.go` and `explicit_fields_test.go`, that handle preprocessing of structs to un-optionalize any optional arguments that have been explicitly set to some value.

## Testing
Tested against Auth0's API and against all fixtures (hence this PR being completely massive; it adds about 400 lines of as-is Go to every fixture, for starters).

### DIFF
--- a/generators/go/internal/generator/sdk/internal/explicit_fields.go
+++ b/generators/go/internal/generator/sdk/internal/explicit_fields.go
@@ -1,0 +1,116 @@
+package internal
+
+import (
+	"math/big"
+	"reflect"
+	"strings"
+)
+
+// HandleExplicitFields processes a struct to remove `omitempty` from
+// fields that have been explicitly set (as indicated by their corresponding bit in explicitFields).
+// Note that `marshaler` should be an embedded struct to avoid infinite recursion.
+// Returns an interface{} that can be passed to json.Marshal.
+func HandleExplicitFields(marshaler interface{}, explicitFields *big.Int) interface{} {
+	val := reflect.ValueOf(marshaler)
+	typ := reflect.TypeOf(marshaler)
+
+	// Handle pointer types
+	if val.Kind() == reflect.Ptr {
+		if val.IsNil() {
+			return nil
+		}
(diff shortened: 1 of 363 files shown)


generators/go/internal/generator/sdk/internal/explicit_fields.go
```diff
@@ -0,0 +1,116 @@
+package internal
```
musicpulpite, line 1: I will mention that AsIs files are fixed in the v2 generator as of https://github.com/fern-api/fern/pull/9314 but on second thought I think its fine to keep it here for now and we can migrate all of the AsIs files in one go at a future time. Carry on

generators/go/internal/generator/sdk/internal/explicit_fields_test.go
```diff
@@ -0,0 +1,356 @@
...
+		{
+			desc: "struct with boolean field",
+			giveInput: func() *testExplicitFieldsStruct {
+				s := &testExplicitFieldsStruct{}
+				s.SetEnabled(boolPtr(false)) // explicit false
+				return s
+			}(),
+			wantBytes: []byte(`{"enabled":false}`),
+		},
+		{
```
musicpulpite, line 185: Can you add a test case here for deeply-nested types perhaps?

musicpulpite (approved):
👏 👏

generators/go/internal/generator/sdk/internal/explicit_fields_test.go
patrickthornton: fair!

PR #9510, opened 2025-09-17
feat(go): `exportAllRequestsAtRoot` exports all requests at the root-level namespace

patrickthornton:
## Description
Linear ticket: Fixes FER-6559
Equivalent of https://github.com/fern-api/fern/pull/9154 in Go.

## Changes Made
This handles the change a little differently than TypeScript, which is a bit more forgiving with its import system; this moves inline requests to a root-level file and has subpackages import from the root-level package if they need them. Consequently, most of the changes are in the v1 side of the generator, but there's also changes in v2 and the dynamic snippet generator.

## Testing
A new fixture on `examples`; testing against Auth0's spec.


seed/go-sdk/examples/export-all-requests-at-root/requests.go
```diff
@@ -0,0 +1,75 @@
+// Code generated by Fern. DO NOT EDIT.
```
musicpulpite, line 1: So this is the new root-level file in question yeah?

musicpulpite (approved):

seed/go-sdk/examples/export-all-requests-at-root/requests.go
patrickthornton: yep

PR #9512, opened 2025-09-17
fix(cli): readme generation prefers user defined example values

dsinghvi:
## Description
Customer reported that example values in OpenAPI spec were not showing up in READMEs for certain generators

## Changes Made
- Example snippets are ordered [...user defined, ...autogenerated] so instead of updating the snippet we prioritize the first one seen.

## Testing


packages/seed/src/commands/run/runWithCustomFixture.ts
```diff
@@ -107,6 +107,7 @@ export async function runWithCustomFixture({
                 ...(customFixtureConfig?.customConfig as Record<string, unknown>),
                 ...(generatorGroup.invocation.config as Record<string, unknown>)
             },
+            readmeConfig: apiWorkspace.generatorsConfiguration?.rawConfiguration.readme,
```
dsinghvi, line 110: seed fix to respect readme configs as well

generators/go-v2/sdk/src/readme/ReadmeSnippetBuilder.ts
```diff
@@ -255,6 +255,9 @@ export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
             if (endpointSnippet.snippet.type !== "go") {
                 throw new Error(`Internal error; expected go snippet but got: ${endpointSnippet.snippet.type}`);
             }
+            if (snippets[endpointSnippet.id.identifierOverride] != null) {
+                continue;
+            }
```
aditya-arolkar-swe, line 260: Looks like `snippets[endpointSnippet.id.identifierOverride] != null` means it's an auto-generated snippet. Won't this just completely skip auto-generated examples though? What if there are no user-specified examples?

aditya-arolkar-swe (approved):

PR #9520, opened 2025-09-18
chore(deps): update vitest

davidkonigsberg:
## Description
Older versions of vitest contain vulnerabilities, so we want to move to the newest stable build.

**NOTE**:
- Moving to the newer vitest appears to have introduced a little more overhead on larger tests making things much slower and causing tests to timeout. To avoid this I followed the suggestions on this ticket to disable some of the metrics. https://github.com/vitest-dev/vitest/issues/7285

## Changes Made
Updated Vitest

## Testing
Testing in CI


davidkonigsberg:
Looks like I'm gonna need to update some tests/configs. Will work on this when I have time, but in case I forget, here are the dos to look at: https://vitest.dev/guide/migration.html#vitest-3

PR #9540, opened 2025-09-18
fix(go): final fixes for go  dynamic snippets blocking wire test generation

musicpulpite:
## Description
Linear ticket: This is the first of several PR's to address "Fix existing dynamic snippets" milestone in the [[SDKs] Go Wire tests](https://linear.app/buildwithfern/project/sdks-go-wire-tests-3609ed54df72/overview) project

Final fixes before we can move on with wire test generation

## Changes Made
- Always include a `strings.NewReader` instance even when the file example is not present in the IR (will reassess if I can fix this from within IR parser)
- For enums (with a single possible value) that are being typed simply as aliases due to the `coerce-enums-to-literals` [config option](https://buildwithfern.com/learn/api-definitions/openapi/generators-yml-reference#coerce-enums-to-literals) correctly instantiate the literal alias
- use correct check to see if a query param type is optional (by essentially copying the logic from the `AbstractGoGeneratorContext` [here](https://github.com/fern-api/fern/blob/af482f14d0fe56bf2719cd4eabc1a5ce7fc46e3e/generators/go-v2/base/src/context/AbstractGoGeneratorContext.ts#L446))

## Testing
- [x] Manual testing using frond against the auth0 sdk

### DIFF
--- a/generators/go-v2/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
+++ b/generators/go-v2/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
@@ -2,6 +2,7 @@ import {
     AbstractDynamicSnippetsGeneratorContext,
     FernGeneratorExec
 } from "@fern-api/browser-compatible-base-generator";
+import { assertNever } from "@fern-api/core-utils";
 import { FernIr } from "@fern-api/dynamic-ir-sdk";
 import { BaseGoCustomConfigSchema, go, resolveRootImportPath } from "@fern-api/go-ast";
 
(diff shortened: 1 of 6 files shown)


generators/go-v2/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
```diff
@@ -40,6 +41,25 @@ export class DynamicSnippetsGeneratorContext extends AbstractDynamicSnippetsGene
         });
     }
 
+    public isOptional(typeReference: FernIr.dynamic.TypeReference): boolean {
```
musicpulpite, line 44: Check me on this

generators/go-v2/dynamic-snippets/src/context/FilePropertyMapper.ts
```diff
@@ -62,10 +62,8 @@ export class FilePropertyMapper {
         property: FernIr.dynamic.FileUploadRequestBodyProperty.File_;
         record: Record<string, unknown>;
     }): go.TypeInstantiation {
-        const fileValue = this.context.getSingleFileValue({ property, record });
-        if (fileValue == null) {
-            return go.TypeInstantiation.nop();
-        }
+        // fix(williammcadams): always return a string reader even if no example was provided
+        const fileValue = this.context.getSingleFileValue({ property, record }) || "";
```
musicpulpite, line 66: For some reason these file upload body properties did not have example values populated in the `snippet.requestBody` so just default to "".

The robust solution would be to fix the snippet population to include the correct example string but I could not find where that was happening. I think it happens in the IR parser when populating `userExamples`/`autoGeneratedExamples`so would appreciate if someone can point me in the right direction

generators/go-v2/dynamic-snippets/src/context/DynamicSnippetsGeneratorContext.ts
patrickthornton: think this is right

patrickthornton (approved):

PR #9543, opened 2025-09-18
fix(python): Lazy imports properly use aliases on collisions

tjb9dc:
## Description
Lazy imports for collisions were not getting written. This should fix the issue here: https://github.com/merge-api/merge-python-client/pull/140/files

## Changes Made
- Resolve the reference before writing the lazy import
  - (Hackily) make this only happen at write time and not metadata resolution time. This is okay because we're very conscious advanced users of this
- Plumb through the reference resolver

## Testing
We had an existing seed fixture that was broken.


generators/python/src/fern_python/generators/sdk/client_generator/base_client_generator.py
```diff
@@ -234,18 +237,25 @@ def _write_lazy_import_property(
...
+            lambda writer: writer.write_line(
+                self._imports_manager.get_resolved_import_as_string(
+                    import_=service_import,
+                    reference_resolver=self._reference_resolver,
+                    noqas=["E402"],
+                )
+            )
+        )
+        # HACK(tjb9dc): We need this to be truly lazy, since we are managing the references/imports manually
+        lazy_import_statement.get_metadata = lambda: AstNodeMetadata()  # type:ignore[method-assign]
```
tjb9dc, line 250: I also considered making a `LazyImport` ASTNode but decided this would be a more reasonably sized change...if we ever start using lazy imports a lot we should probably make them a node idk

generators/python/src/fern_python/generators/sdk/client_generator/base_client_generator.py
```diff
@@ -289,10 +283,10 @@ def _write_lazy_import_property(writer: AST.NodeWriter) -> None:
                                     AST.Reference(qualified_name_excluding_import=("property",), import_=None)
                                 )
                             ],
-                            body=AST.CodeWriter(
```
tjb9dc, line 292: This never should've been a code writer I was just being silly before

aditya-arolkar-swe (approved):

PR #9548, opened 2025-09-19
fix(cli): Properly look at subtask failures

tjb9dc:
## Description
Noticed that we weren't failing on generator failures in seed. While fixing, it turned out this was a more pervasive issue where we weren't looking at subtask failures in the seed CLI or main CLI.

## Changes Made
- Inspect the result of generation in seed
- `getResult()` properly looks at subtasks. It looks like this was an oversight where it was implemented for `InteractiveTaskContextImpl` but not `TaskContextImpl`
  - This change was made both in seed and the main CLI

## Testing
- Tested that this properly fails when a generator raises
  - I suspect we'll get more seed failures we'll have to ignore :)
- I haven't tested the CLI


packages/cli/cli/src/cli-context/TaskContextImpl.ts
tjb9dc, line 1: lmk if you'd prefer to make the CLI change separately, or if there was some reason we like this behavior in the CLI

dsinghvi (approved):

PR #9561, opened 2025-09-19
changes without snapshot/seed updates

Swimburger:
## Description
Linear ticket:

## Changes Made

- 
- 

## Testing


packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
```diff
@@ -29,8 +30,17 @@ types:
...
       inferred: InferredAuthScheme
 
-  InferredAuthScheme:
+  BaseAuthScheme:
     extends: commons.WithDocs
+    properties:
+      key:
+        type: AuthSchemeKey
+        docs: |
+          The unique key for this auth scheme. This will be used to reference the scheme in endpoint auth requirements.
```
tjb9dc, line 40: I couldn't see where we were plumbing this. Could you point me at that?

Wanted to make sure that we were validating that the key exists before falling back to the type (and failing if it's not present). I guess we already get validation that there are no duplicate keys for free because of it being a map in YAML

fern/apis/fern-definition/definition/service.yml
```diff
@@ -55,6 +55,12 @@ types:
       source: optional<source.SourceSchema>
       retries: optional<RetriesConfiguration>
 
+  HttpEndpointAuth:
+    discriminated: false
+    union:
+      - boolean
+      - list<map<string, list<string>>>
```
tjb9dc: do we validate that all of the `Scopes` here are also in the predefined OAuthSchemeSchema? Supposedly that is required in OpenAPI

tjb9dc: Also, when you make the change for `Scope` and `AuthSchemeKey` to be aliases, I think it's worth making that change in `auth.yml` as well

packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
```diff
@@ -62,6 +63,7 @@ types:
       v2Responses: optional<V2HttpResponses>
       errors: ResponseErrors
       auth: boolean
+      security: optional<list<map<auth.AuthSchemeKey, list<string>>>>
```
tjb9dc: and then `Scope` alias here as well would be 👍

fern/apis/fern-definition/definition/service.yml
Swimburger: We don't validate that. We're simply grabbing what's provided. We could add a validator if this becomes an issue

tjb9dc: I think that makes sense, ultimately I'm not sure we do a ton of OpenAPI spec validating yet haha and I would put this in that camp

packages/ir-sdk/fern/apis/ir-types-latest/definition/auth.yml
Swimburger: It's copied from the FD here: https://github.com/fern-api/fern/pull/9561/files#diff-62341060014e41c5f95248fbbbc302ff3f1df40534ebbc95b59e8cdcb0df1e21R116-R125

PR #9573, opened 2025-09-22
chore(typescript): gitignore pnpm lock files

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6660/ci-ignore-pnpm-lock-files-inside-generators-like-we-did-for-other
Adding more package management files to our gitignore

## Changes Made
- Delete all pnpm-lock files in the seed/ts-sdk folder
- Add those files to the gitignore
- Add to path exclusions for seed verification checks

## Testing
Verified locally, will also verify in CI


tjb9dc (approved):
woot

PR #9581, opened 2025-09-22
fix(internal): `$ref`ed schemas get `nullable<>`

patrickthornton:
## Description
Linear ticket: Helps with FER-6680.
Makes it so nullable schemas are referenced as such in the Fern Definition when referred to via `$ref` in the OpenAPI spec.

## Changes Made
Adds a quick check of the relevant schema to `convertReferenceObject()`. Also does away with an unnecessary recursive call.

## Testing
Updated parser tests.

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/intercom.json
+++ b/packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/intercom.json
@@ -4738,9 +4738,7 @@ A "cursor" or pointer is used to keep track of the current position in the resul
             },
             "union": [
               "string",
-              {
-                "type": "optional<customObjectInstances.CustomObjectInstance>",
-              },
+              "optional<customObjectInstances.CustomObjectInstance>",
             ],
           },
           "CustomerRequest": {
@@ -9942,7 +9940,7 @@ types:
     discriminated: false
     union:
       - string
-      - type: optional<customObjectInstances.CustomObjectInstance>
+      - optional<customObjectInstances.CustomObjectInstance>
     source:
       openapi: ../openapi.yml
     inline: true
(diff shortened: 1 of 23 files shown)


packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi-ir/belvo.json
```diff
@@ -7031,27 +7035,6 @@
...
-                                    "type": "float"
-                                  },
-                                  "type": "primitive"
-                                }
-                              },
-                              "type": "object"
-                            }
-                          ],
-                          "type": "array"
-                        },
```
patrickthornton, line 7054: this is in the following path of the IR:
`endpoints/examples/response/value/value/properties/loan_data/properties/interest_rates`
which i believe is alright, as this property has `nullable: true` which now surfaces through a reference

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/intercom.json
```diff
@@ -40047,8 +40034,7 @@ service:
         openapi: ../openapi.yml
       display-name: Create a phone Switch
       request:
-        body:
-          type: optional<root.CreatePhoneSwitchRequest>
```
patrickthornton, line 40051: would a change like this matter?

dsinghvi (approved):

packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi/intercom.json
dsinghvi: nope

patrickthornton: kool

PR #9584, opened 2025-09-22
feat(ci): Parallelize update-seed workflow

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6694/ci-update-seed-update-workflow-like-seed-snapshot-tests
Closes FER-6694
Updating update-seed workflow to parallelize work like we did for seed tests

## Changes Made
- Move all update-seed jobs to ubuntu-latest (cheapest runner)
- Parallelize work for update-seed
- Make PRs per generator at workflow level in matrix instead of inside of action file, condenses down code
- Remove get-generator-changes now that we are doing it as a matrix for seed and update-seed workflows
- Add get-test-matrix as shared action file between update-seed and seed workflows

## Testing
Tested in CI


.github/workflows/seed.yml
```diff
@@ -125,7 +125,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        generator-name:
+        sdk-name:
```
Swimburger, line 128: nit: generator-name is more accurate, since not all generaters generate SDKs

chrismallinson: @Swimburger in my head, everything in the generators folder is a generator and everything created in the seed output folder is an SDK. This matrix is to loop over what will be generated outputs of the seed folder. If my understanding is not correct though then happy to change the name

dsinghvi (approved):

PR #9585, opened 2025-09-22
fix(postman): fix bundler transformation issue causing runtime error

kafkas:
## Description
Linear ticket: [FER-6385](https://linear.app/buildwithfern/issue/FER-6385/flagright-postman-issue)

Fixes a bundler issue causing a Postman generation error when writing Github workflows. `endent` is a cjs package that signals `__esModule: true` which is causing tsup's `__toESM` wrapper to create a `default` property that already exists. Using a dynamic require allows us to avoid the `__toESM` wrapper.

## Changes Made

- Fixes `Dockerfile` and `Dockerfile.dockerignore` to run the generator
- Updates `writePostmanGithubWorkflows` to use a dynamic require with a type cast
- Inits tsup config file

## Testing

### DIFF
--- a/generators/postman/Dockerfile
+++ b/generators/postman/Dockerfile
@@ -1,3 +1,5 @@
 FROM node:22.12-alpine3.20
-COPY dist /dist
+
+COPY generators/postman/dist /dist
+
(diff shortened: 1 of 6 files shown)


dsinghvi (approved):
🚀

PR #9593, opened 2025-09-23
feat(go, java, python, typescript): Add support for PR mode for self-hosted/local sdk generation

jsklan:
## Description
Linear ticket: [FER-6679](https://linear.app/buildwithfern/issue/FER-6679/)
Add support for PR mode for self-hosted/local sdk generation
NOTE:
- There are some issues this process has exposed where differences exist between the outputs of the local vs remote generation. This includes one that causes the typescript sdk to fail pnpm test. Given the already broad scope of this change set, I'm considering those bugs to be out of scope, and to be fixed in later change sets.

## Changes Made
- Base Generator
  - Pulled shared logic up out of abstract methods to minimize code duplication
- Java: Upgraded to IRv60
- Java-v2: Added mode option to github config extraction
- Python: Upgraded to IRv60
- Python-v2: Added mode option to github config extraction
- Go: Refactored ordering of v2 generator invocation.
  - Before, this happened at the end of the generator.go:generate function, which was before the project's files were persisted
  - Now this happens in the main.go:run function after the project's files are persisted
- Typescript: Added support for local github publishing
  - Different pattern here than the others because there is no v2 sdk generator for typescript
  - Instead, all logic was added to typescript's AbstractGeneratorCli class
- Cli: Updated IR migration logic to include new versions as IRv60 compatible

## Testing
- [x] Manually tested by creating a simple fern project, building generator images locally, and running local generation in PR mode to a sample repo. Results PRs are:
  - [x] [Java](https://github.com/fern-api/jsklan-empty/pull/14)
  - [x] [Go](https://github.com/fern-api/jsklan-empty/pull/13)
  - [x] [Python](https://github.com/fern-api/jsklan-empty/pull/15)
  - [x] [Typescript](https://github.com/fern-api/jsklan-empty/pull/17)


generators/python/src/fern_python/generators/sdk/core_utilities/client_wrapper_generator.py
```diff
@@ -784,6 +784,7 @@ def _get_bearer_auth_scheme(self) -> Optional[ir_types.BearerAuthScheme]:
                 # TODO: We need to support the token prefix. This will actually need to be handled as a
                 #       custom header auth scheme.
                 return ir_types.BearerAuthScheme(
+                    key="key",
```
jsklan: I'm not sure what the correct implementation is here. This was just something I put to get it to compile.

generators/java/generator-utils/src/main/java/com/fern/java/FeatureResolver.java
```diff
@@ -17,6 +18,7 @@
 public class FeatureResolver {
 
     public static final AuthScheme DEFAULT_BEARER_AUTH = AuthScheme.bearer(BearerAuthScheme.builder()
+            .key(AuthSchemeKey.of("key"))
```
jsklan: I'm not sure what the correct implementation is here. This was just something I put to get it to compile.

fern-support: The way it is in the IR generator itself is to grab explicitly given name, or fallback to the `type` property. I believe this would be `"bearer"`.

generators/python/src/fern_python/generators/sdk/core_utilities/client_wrapper_generator.py
fern-support: Same, but since `key` isn't used in Java, it doesn't really matter at this point.

generators/java/generator-utils/src/main/java/com/fern/java/FeatureResolver.java
tjb9dc: yeah the true default case here would be key `bearer`

tjb9dc (approved):
looks reasonable to me, can you go ahead and update `key` to be `bearer` for the default bearer auth scheme

tstanmay13 (approved):

PR #9594, opened 2025-09-23
fix(go): introduce error decoder to cut down on duplicated error handling code

patrickthornton:
## Description
Linear ticket: Fixes FER-6493
Integrates the error decoder that abstracts away the shared error-handling code between endpoints; see the `errors` fixture for an example of this working. (It's more pronounced in bigger customer SDKs.)

## Changes Made
~~This was achieved by moving `caller.go`, `caller_test.go`, `retrier.go`, `retrier_test.go`, `pager.go`, `pager_test.go`, and `streamer.go` up into the root-level directory (ew), since they need to be a part of the package that can access the types in `errors.go` (and `errors.go` can't move, as it needs the types available at root-level); these files are now also being generated by the v2 generator instead of the v1 generator, and the relevant duplicate code's been cut.~~

Actually, now only `error_codes.go` is at root level, phew! The error decoder initializer takes the error codes as an argument instead of needing to import them around everywhere.

Piggybacks off of @musicpulpite's work to get the actual `error_codes.go` and `error_decoder.go` files up to snuff.

## Testing


generators/go-v2/base/src/project/GoProject.ts
```diff
@@ -75,52 +83,138 @@ export class GoProject extends AbstractProject<AbstractGoGeneratorContext<BaseGo
...
-    }
+        filename: string;
+        templateVariables?: Record<string, string>;
+    }): Promise<File> {
+        let contents = (await readFile(this.getAsIsFilepath(filename))).toString();
+
+        // Process template variables
+        for (const [key, value] of Object.entries(templateVariables)) {
+            // Handle both {{{{ variable }}}} and {{{ variable }}} syntax
+            const tripleRegex = new RegExp(`\\{\\{\\{ *\\.${key} *\\}\\}\\}`, 'g');
```
musicpulpite: Great stuff!

musicpulpite (approved):
This is the way!

PR #9600, opened 2025-09-24
feat(rust): Add documentation comments and remove warning for unused imports/exports

iamnamananand996:
## Description

This PR add a Rust code generator with `documentation support` and improves overall code generation quality by fixing imports, `warnings`, and adding proper description handling.

Linear ticket: 
- https://linear.app/buildwithfern/issue/FER-6669/rust-sdk-eliminate-warning-from-the-sdk
- https://linear.app/buildwithfern/issue/FER-6671/rust-sdk-add-description-before-the-the-method-function-if-available

## Changes Made

- Added support for generating Rust documentation comments (`DocComment` AST node) with summaries, descriptions, parameters, and examples
- Implemented description support for `structs`, `enums`, `enum variants`, `fields`, and `client sub-modules`
- Fixed Rust codegen imports to reduce unnecessary dependencies and warnings
- Improved `chrono` type handling to only import `date/datetime` types when actually used
- Removed redundant imports (`serde_json::Value`, unused `QueryBuilder` imports)
- Enhanced `Default` derive logic for request bodies and improved module import patterns

## Extra's

- Updated utility functions and HTTP client implementations for better code quality
- Cleaned up request options and fixed module re-export patterns

## Testing
 
- [x] Units test added/updated 
- [x] Updated Rust seed test snapshots to validate new documentation generation
- [x] Verified improved import handling reduces compilation warnings

### DIFF
--- a/generators/rust/base/src/asIs/http_client.rs
+++ b/generators/rust/base/src/asIs/http_client.rs
@@ -1,4 +1,4 @@
-use crate::{Utils::join_url, ApiError, ClientConfig, RequestOptions};
+use crate::{join_url, ApiError, ClientConfig, RequestOptions};
 use reqwest::{
     header::{HeaderName, HeaderValue},
     Client, Method, Request, Response,
--- a/generators/rust/base/src/asIs/mod.rs
+++ b/generators/rust/base/src/asIs/mod.rs
@@ -8,4 +8,4 @@ mod utils;
 pub use http_client::HttpClient;
 pub use request_options::RequestOptions;
 pub use query_parameter_builder::{QueryBuilder, QueryBuilderError, parse_structured_query};
-pub use utils::Utils;
+pub use utils::*;
--- a/generators/rust/base/src/asIs/mod.rs
+++ b/generators/rust/base/src/asIs/mod.rs
@@ -8,4 +8,4 @@ mod utils;
 pub use http_client::HttpClient;
 pub use request_options::RequestOptions;
 pub use query_parameter_builder::{QueryBuilder, QueryBuilderError, parse_structured_query};
-pub use utils::*;
+pub use utils::join_url;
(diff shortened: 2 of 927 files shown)


generators/rust/base/src/asIs/http_client.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{Utils::join_url, ApiError, ClientConfig, RequestOptions};
+use crate::{join_url, ApiError, ClientConfig, RequestOptions};
 use reqwest::{
```
tjb9dc, line 2: not changed in this PR but is this a typo?

generators/rust/base/src/asIs/mod.rs
```diff
@@ -8,4 +8,4 @@ mod utils;
 pub use http_client::HttpClient;
 pub use request_options::RequestOptions;
 pub use query_parameter_builder::{QueryBuilder, QueryBuilderError, parse_structured_query};
-pub use utils::Utils;
+pub use utils::*;
```
tjb9dc: was this to get around unused imports/exports? I feel like it's more idiomatic to avoid the star imports. What do you think?

tjb9dc (approved):

generators/rust/base/src/asIs/http_client.rs
iamnamananand996: earlier, it was added inside a module like this

```rust
mod Utils {  ---> warning capital letters are not allowed for modules
  fn join_url() {

  }
}
```

so changes to direct funtions

```rust
fn join_url() {
 
}
```

generators/rust/base/src/asIs/mod.rs
iamnamananand996: Haven't added much work for wild card exports `*`; however, here we can fix this, as this file is a static template.

`pub use utils::join_url`

Let me update this.

------------------------

The wild card export will address this issue.

- https://linear.app/buildwithfern/issue/FER-6672/rust-sdk-have-mechanism-to-remove-unused-wildcard-export-whereever

cc - @tjb9dc

PR #9605, opened 2025-09-24
chore(internal): Remove dynamic IR dependency on IR types

tjb9dc:
## Description
The dynamic IR was referencing some IR types in a handful of places. This was making a breaking change to a common IR type prohibitively hard, as it resulted in a dynamic IR break. From talking to Deep, we'd like to keep these completely separate.

This will probably cause some minor import renaming pain during upgrade.

## Changes Made
- Remove references to regular IR from dynamic IR
  - For consistency, I removed all the type sharing
- Duplicate types as necessary (fortunately was quite minimal)

## Testing
Compile works


packages/ir-sdk/fern/apis/ir-types-latest/definition/dynamic/types.yml
```diff
@@ -107,3 +106,25 @@ types:
     union:
       boolean: boolean
       string: string
+
+  ObjectPropertyAccess:
+    enum:
+      - READ_ONLY
+      - WRITE_ONLY
+
+  PrimitiveTypeV1:
```
tjb9dc, line 115: Wondering if it will be particularly annoying or dangerous to sever the enum sharing

dsinghvi (approved):
👏

PR #9607, opened 2025-09-24
chore(go, cli): Upgrade go generators to IRv60

jsklan:
## Description
Upgrade go generators to IRv60

## Changes Made
- Added script to top level package.json to regenerate go ir-sdk to local output
- Regenerated local go ir-sdk to generators/go/internal/fern/ir
- Manually fixed compile issues in generated output
- Updated imports in go generator that were broken
- Updated usages in go generator that were broken. Main one was type change for PropertyPath
- Updated ir-sdk dependency for go-v2 to 60.3.1
- Added new versions for go-model, go-fiber, and go-sdk
- Updated migration logic in fern cli to use IRv60 for new go generator versions
- Added new version to cli to include migration logic

## Testing
- [ ] Ran seed test for all go-sdk fixtures to verify no changes in outputs


generators/go-v2/sdk/src/endpoint/utils/getPaginationInfo.ts
```diff
@@ -432,7 +432,7 @@ function getResponsePropertySetter({
             writer.writeNode(
                 getPropertyNilCheckCondition({
                     variableName: "response",
-                    propertyPath: responsePropertyPath
+                    propertyPath: responsePropertyPath.map((item) => item.name)
```
tjb9dc, line 435: hrm is this sufficient? Or do we need to unpack the name and then use the originalName or safeName or whatever

jsklan: the type for responsePropertyPath changed from Name[] to PropertyPathItem[], so the function this is being passed to expects Name[]. And it does compile, so I assume this works.

musicpulpite (approved):

PR #9610, opened 2025-09-24
break(cli): Compress Name in the IR whenever possible

tjb9dc:
## Description
Our IR is quite sizable, but if you look closely, up to 75% of it is the `Name` type which includes all of the casing variations (camelCase, PascalCase, snake_case, and SCREAMING_SNAKE_CASE). In _most_ cases, these casings aren't particularly interesting, and could be simply computed later without any context.

The producer (`CasingsGenerator.ts`) will now only emit casings that are non-trivially derivable (read: !== a naive call to lodash), with the onus on the consumer to fill them in. We have a typescript-based utility for this, but we'll also need to run this in other generators. I'm not too worried about this.

## Changes Made
- Checkpoint IR v60, stage IR v61
  - The `Name` type has been broken to be an undiscriminated union. In trivial cases, only a string is provided. In non-trivial cases, all non-trivial casings will be written, and trivials will be omitted (they are optional)
  - NOTE: This change was NOT made to dynamic IR. We'll need to come up with a migrations framework to pull that off
- IR migration which leverages a monkey patched serialization hook to expand new `Name` to old shape. Trying to write this explicitly would be hell on earth (and honestly more likely to result in a mistake)
- Some utility functions for extracting `originalName` and expanding back to the legacy `Name` type
  - These went in core-utils so that we can reference them in typescript generators without incurring an IR dependency
  - NOTE: We'll have to recreate these utilities in each generator language in order to upgrade generators. They use lodash behind the scenes. Lots of ways to do this of varying degrees of fun
- `CasingsGenerator` only emits cases for non-trivially derived casings, and simply emits `string` if all casings are trivially derived

## Testing
- Existing tests
- Migrator test


generators/typescript/model/union-generator/src/GeneratedUnionImpl.ts
```diff
@@ -23,7 +24,6 @@ import {
     VariableStatementStructure,
     WriterFunction
 } from "ts-morph";
-import { FernIr } from "../../../../../packages/ir-sdk/src";
```
tjb9dc, line 26: not sure what was up with this, kinda weird

packages/cli/generation/ir-migrations/src/migrations/v61-to-v60/migrateFromV61ToV60.ts
```diff
@@ -0,0 +1,87 @@
...
+            const raw = IrSerialization.V61.IntermediateRepresentation.jsonOrThrow(v61, {
+                unrecognizedObjectKeys: "strip",
+                skipValidation: true
+            });
+
+            // Deserialize as V60
+            return IrSerialization.V60.IntermediateRepresentation.parseOrThrow(raw, {
+                unrecognizedObjectKeys: "strip",
+                skipValidation: true
+            });
```
tjb9dc: TODO: We may be able to skip this..hard to say

PR #9621, opened 2025-09-25
fix(internal): path parameters respect URL order

patrickthornton:
## Description
Linear ticket: Fixes FER-6801
If a user specifies path parameters in a different order than they appear in the URL, as in:
```json
"here/{param1}/there/{param2}":
  "parameters": [
    {
      "name": "param2",
      "in": "path"
    },
    {
      "name": "param1",
      "in": "path"
    }
  ],
```
we now adhere to the URL order (param1, then param2) instead of the listed `parameters` order in all cases.

## Changes Made
Changes to the `openapi-ir-parser`.

## Testing
Tested against `openapi-ir-to-fern-tests` and Auth0.

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi-ir/deepgram.json
+++ b/packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi-ir/deepgram.json
@@ -4460,23 +4460,6 @@
             "file": "../openapi.yml",
             "type": "openapi"
           }
-        },
-        {
-          "description": "The unique identifier of the project",
-          "name": "project_id",
-          "schema": {
-            "schema": {
-              "example": "123456-7890-1234-5678-901234",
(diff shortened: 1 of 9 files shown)


packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/__snapshots__/openapi-ir/deepgram.json
```diff
@@ -4444,23 +4444,6 @@
...
-            },
-            "generatedName": "DeleteProjectRequestProjectId",
-            "groupName": [],
-            "type": "primitive"
-          },
-          "source": {
-            "file": "../openapi.yml",
-            "type": "openapi"
-          }
-        },
```
patrickthornton, line 4463: these deepgram deletions are interesting; as far as i can tell, these are all exact duplicates of the things just below them? i figure it's fine to shave off these

dsinghvi (approved):

PR #9632, opened 2025-09-26
feat(cli): Allow output folders per fixture when multiple are passed in to test

chrismallinson:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6805/cli-allow-each-fixture-to-have-its-own-output-folder-for-seed-test
Closes FER-6805
Allow output folders per fixture when passing multiple fixtures in at once to seed test

## Changes Made
- Remove `choices` for input fixtures. Parsing happens manually a bit later now so we can extract all available fixtures with output folders
- Move code for `getAvailableFixtures` to reusable function so that `seed get-available-fixtures` and `seed test` can both use it
- Update `testWorkspaceFixtures.ts` to handle new parsing

## Testing
- Verified on command line that you cannot have --outputFolder flag and output folder with `:` simultaneously
- Verified on command line that you cannot pass invalid fixtures (test inclusive of output folders)
- Verified on command line that the `get-available-fixtures` command still works

### DIFF
--- a/packages/seed/src/cli.ts
+++ b/packages/seed/src/cli.ts
@@ -140,14 +140,21 @@ function addTestCommand(cli: Argv) {
 
                 for (const fixture of argv.fixture) {
                     if (!availableFixtures.includes(fixture) && !availableFixturesWithOutputFolders.includes(fixture)) {
-                        throw new Error(`Fixture ${fixture} not found. Please make sure that it is a valid fixture for the generator ${generator.workspaceName}.`);
+                        throw new Error(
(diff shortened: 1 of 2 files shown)


musicpulpite (approved):

chrismallinson:
Need to fix one item - case where no fixture is passed and all fixtures are meant to run

chrismallinson:
> Need to fix one item - case where no fixture is passed and all fixtures are meant to run

Fixed in latest push, and fixed linting

PR #9633, opened 2025-09-26
fix(typescript): Fix output modes

Swimburger:
## Description
Fix output modes

## Changes Made
- Format GitHub Actions workflow files using prettier.
- Add .prettierignore to ensure dist and temporary files are not formatted.
- Fix non-GitHub output modes.

prettierignore will prevent `prettier` from trying to format temporary files from the parallel `pnpm install`, which was causing occasional errors in seed.

## Testing


musicpulpite (approved):
Seems right to me!

PR #9634, opened 2025-09-26
fix(cli): suppress stdout piping of entire generator-cli output

musicpulpite:
## Description
When generators fail they logging is polluted with the entire contents of the `readme.md` and `reference.md`. This behavior is a byproduct  of these files being created by the generator-cli being run as a subprocess with its stdout being piped to the parent process stdout (when all we really need to do is capture the stdout as a string) which bubbles up to the top-level generator during failure.

The simple solution is to pass `doNotPipeOutput: true` to all of the methods since the cli is memoized with the options of the first method invocation. **Could refactor it to not reuse the options if y'all think it matters**

### DIFF
--- a/generators/base/src/GeneratorAgentClient.ts
+++ b/generators/base/src/GeneratorAgentClient.ts
@@ -9,12 +9,13 @@ const GENERATOR_AGENT_NPM_PACKAGE = "@fern-api/generator-cli";
 export class GeneratorAgentClient {
     private logger: Logger;
     private skipInstall: boolean;
-    private cli: LoggingExecutable | undefined;
+    private installAttempted: boolean;
     private selfHosted: boolean;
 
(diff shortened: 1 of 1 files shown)


generators/base/src/GeneratorAgentClient.ts
```diff
@@ -97,7 +99,8 @@ export class GeneratorAgentClient {
 
         const cli = createLoggingExecutable("generator-cli", {
```
Swimburger, line 111: Can we create a `cli` instance just for printing the version, but then return a function that lets you create the `cli` with your own options, so you could specify the no output redirect on the commands you want specifically?

musicpulpite: Good idea

Swimburger (approved):

PR #9670, opened 2025-09-29
fix(cli): speed up how we deal with example generation on discriminated unions

kennyderek:
## Description
Make cli faster
Took full process down from 330 seconds to 110, of which I believe ~80sec is uploading to S3.

My hypothesis of why this works is the following:
In Elevenlabs recent PRs, they added nodes where within their `oneOf` operators, they reference themselves. See below for example
```json
     "ASTOrOperatorNode": {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "default": "or_operator",
            "const": "or_operator",
            "title": "Type"
          },
          "children": {
            "type": "array",
            "items": {
              "oneOf": [
                {
                  "$ref": "#/components/schemas/ASTStringNode"
                },
                {
                  "$ref": "#/components/schemas/ASTNumberNode"
                },
                {
                  "$ref": "#/components/schemas/ASTBooleanNode"
                },
                {
                  "$ref": "#/components/schemas/ASTLLMNode"
                },
                {
                  "$ref": "#/components/schemas/ASTDynamicVariableNode"
                },
                {
                  "$ref": "#/components/schemas/ASTOrOperatorNode"
                },
                {
                  "$ref": "#/components/schemas/ASTAndOperatorNode"
                },
                {
                  "$ref": "#/components/schemas/ASTEqualsOperatorNode"
                },
                {
                  "$ref": "#/components/schemas/ASTNotEqualsOperatorNode"
                },
                {
                  "$ref": "#/components/schemas/ASTGreaterThanOperatorNode"
                },
                {
                  "$ref": "#/components/schemas/ASTLessThanOperatorNode"
                },
                {
                  "$ref": "#/components/schemas/ASTGreaterThanOrEqualsOperatorNode"
                },
                {
                  "$ref": "#/components/schemas/ASTLessThanOrEqualsOperatorNode"
                }
              ],
              "discriminator": {
                "propertyName": "type",
                "mapping": {
                  "and_operator": "#/components/schemas/ASTAndOperatorNode",
                  "boolean_literal": "#/components/schemas/ASTBooleanNode",
                  "dynamic_variable": "#/components/schemas/ASTDynamicVariableNode",
                  "eq_operator": "#/components/schemas/ASTEqualsOperatorNode",
                  "gt_operator": "#/components/schemas/ASTGreaterThanOperatorNode",
                  "gte_operator": "#/components/schemas/ASTGreaterThanOrEqualsOperatorNode",
                  "llm": "#/components/schemas/ASTLLMNode",
                  "lt_operator": "#/components/schemas/ASTLessThanOperatorNode",
                  "lte_operator": "#/components/schemas/ASTLessThanOrEqualsOperatorNode",
                  "neq_operator": "#/components/schemas/ASTNotEqualsOperatorNode",
                  "number_literal": "#/components/schemas/ASTNumberNode",
                  "or_operator": "#/components/schemas/ASTOrOperatorNode",
                  "string_literal": "#/components/schemas/ASTStringNode"
                }
              }
            },
            "minItems": 2,
            "title": "Children"
          }
        },
        "required": [
          "children"
        ],
        "title": "ASTOrOperatorNode"
      },
```

I think there might be two issues going on here:
1. [TODO] our [isReferenceObject](https://github.com/fern-api/fern/blob/9d3af40bca0a70e3fc905d1437a7377455d2e136/packages/cli/api-importers/v3-importer-commons/src/converters/ExampleConverter.ts#L107) check does not handle this case above properly, since the type is `array` and the references exist in a property `items`. [See code for the check here](https://github.com/fern-api/fern/blob/92d4142ec95a8e9c640568b3e7ba814e89c6fd1e/packages/cli/api-importers/v3-importer-commons/src/AbstractConverterContext.ts#L701). I did not mess with this code path yet, but it means that we don't add the reference objects to our `seen` set.
2. [THIS CHANGE] Even given the above, the code still shouldn't be so slow. Until you notice that the AST has 13 variants. If we evaluate each to a depth of 12, thats quite a bit of branching logic going on... 13^12, potentially? @deep mentioned we only have to evaluate one example variant, so rather than evaluating all of them, and then picking a correct one, let's break out of the evaluation loop once we have 1 correct one. That means that in an ideal case of valid non-coerced examples, we'd only have 12 evaluations.

## Changes Made
- change how we process examples
- TODO: undiscriminated unions too

## Testing

before:
image

after:
image


packages/cli/cli/src/cli.ts
```diff
@@ -607,6 +609,10 @@ function addGenerateCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext)
                     strictBrokenLinks: argv.strictBrokenLinks,
                     disableTemplates: argv.disableSnippets
                 });
+                const endTime = Date.now();
+                const totalTime = endTime - startTime;
+                cliContext.logger.info(`Docs generation completed in ${totalTime}ms (${(totalTime / 1000).toFixed(2)}s)`);
```
dsinghvi: can you confirm that these logs are not duplicate of others?

dsinghvi (approved):

packages/cli/cli/src/cli.ts
kennyderek: removed - I don't think we need them since the github actions tells you how long it takes anyway, but it was helpful for local dev testing

PR #9695, opened 2025-09-29
chore(internal): update seed test packages

fern-support:
Auto-generated PR, triggered by nightly-seed-packaging workflow with: push from branch: mallinson/nightly-seed-packaging-testing-branch.
GitHub workflow run: https://github.com/fern-api/fern/actions/runs/18141813310

### DIFF
--- a/.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
+++ b/.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
@@ -1,0 +1,155 @@
+{
+  "total-test-time": 14420,
+  "split-cutoff-time": 30000,
+  "split-tests": false,
+  "packages": [
+    {
+      "fixtures": [
+        "auth-environment-variables",
+        "response-property",
+        "multiple-request-bodies",
+        "idempotency-headers",
+        "unions",
+        "single-url-environment-default",
+        "custom-auth",
+        "single-url-environment-no-default",
+        "multi-url-environment",
+        "pagination"
+      ],
+      "packageTotalTime": 1480
+    },
+    {
+      "fixtures": [
+        "alias-extends",
+        "nullable",
+        "nullable-optional",
+        "request-parameters",
+        "license",
+        "oauth-client-credentials-nested-root",
+        "objects-with-imports",
+        "oauth-client-credentials-with-variables",
+        "circular-references",
+        "trace"
+      ],
+      "packageTotalTime": 1482
+    },
+    {
+      "fixtures": [
+        "api-wide-base-path",
+        "version-no-default",
+        "server-sent-event-examples",
+        "undiscriminated-unions",
+        "client-side-params",
+        "error-property",
+        "streaming-parameter",
+        "websocket",
+        "circular-references-advanced"
+      ],
+      "packageTotalTime": 1415
+    },
+    {
+      "fixtures": [
+        "bytes-upload",
+        "csharp-system-collision",
+        "query-parameters",
+        "file-download",
+        "simple-fhir",
+        "variables",
+        "oauth-client-credentials-environment-variables",
+        "websocket-bearer-auth",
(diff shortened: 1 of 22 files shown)


.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
```diff
@@ -0,0 +1,155 @@
+{
+  "total-test-time": 14420,
```
tjb9dc, line 2: whenever providing time as a number, I think always prudent to include the unit. E.g. is this time in ms or s? And typically I just see that appended e.g. `total-test-time-ms`

.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
```diff
@@ -0,0 +1,155 @@
...
+  "split-cutoff-time": 9600,
```
tjb9dc: what does this track?

.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
```diff
@@ -0,0 +1,155 @@
...
+  "split-tests": true,
+  "packages": [
```
tjb9dc, line 5: nit: I feel like I don't complete love "package" here since it's so overloaded in build world. Not a strong conviction though.

Maybe `cohort` or the 5th grade version would be `group` ?

.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
```diff
@@ -0,0 +1,155 @@
...
+        "any-auth",
+        "path-parameters",
+        "inferred-auth-explicit",
+        "mixed-file-directory",
+        "cross-package-type-names"
+      ],
+      "packageTotalTime": 1415
+    },
+    {
+      "fixtures": ["leftovers"],
```
tjb9dc, line 151: I take it this is a reserved fixture name?

A couple of solutions I think are cleaner:

1. Make `fixtures` key an undiscriminated union of the literal `"leftovers" | String[]`. That way we don't have to mess with reserving a magic fixture name, and we get strong validation that "leftovers" is special
2. Make this implicit (ie remove it) and always allocate it. Would we ever want a case where we _don't_ run the leftovers? I feel like that would result in dropped tests, which we would never want

.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
chrismallinson: This is my cutoff time for when the file reports back that we should split tests across runners vs running it all in one runner. Per our conversation though, I will move this to the bash side of it, we don't need this answer here.

.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
chrismallinson: I did in parts of the code and didn't pull it up to this output file, will add in the next commit 👍

.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
chrismallinson: from in person conversation, going with option 2, will add the leftovers group in bash. Removing it from here

.github/workflow-resource-files/seed-packages/csharp-model-seed-packages.json
chrismallinson: Changing mentions of packages to groups, I like the 5th grade version 👶

chrismallinson:
@tjb9dc as part of the name change from package->group, also updated the branch name that gets created by CI so the changes are all reflected here: https://github.com/fern-api/fern/pull/9702. If you are happy with the updates I will retarget the change to main and have it the automation set to merge

PR #9697, opened 2025-09-30
fix(ruby): Fix up Ruby v2's reference.md and dynamic snippets

chanks:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6309/ruby-add-referencemd-generation
Some general tweaks to make dynamic snippets and `reference.md` more idiomatic Ruby.

## Changes Made
- Use snake_case and PascalCase more appropriately: the former for method names, the latter for class names.
- When writing hash literals, always include a space between keys and values.
- Updated `EndpointSnippetGenerator` to use keyword args in method calls instead of hash literals.
  - While doing this I realized that back in #9531 when I was adding support for keyword arguments to the MethodInvocation AST I didn't use the `KeywordArgument` nodes that already existed, oops. Updated MethodInvocation appropriately.

## Testing
- Inspected generated reference.md files.


seed/ruby-sdk-v2/bytes-upload/reference.md
```diff
@@ -1,6 +1,6 @@
 # Reference
-## Service
-<details><summary><code>client.Service.Upload(request) -> </code></summary>
+## service
```
dsinghvi: can we make these capitalized?

chanks: No problem, but feel free to ignore my draft PRs, I open them only to offload seed generation to github.

dsinghvi (approved):

PR #9705, opened 2025-10-01
fix(python): Added support for poetry_v1, poetry_v2, uv. Added feature Flags to Indicate which tool (uv, poetry_v1, poetry_v2...)

sangjoon245:
## Description
Linear ticket: https://linear.app/buildwithfern/issue/FER-6797/generated-pyprojecttoml-files-are-technically-invalid
In this issue, we had compability issues with UV and the pyproject.toml fern generates.

The main issue is that uv does not understand poetry_v1 (which we currently support) syntax.

With these changes, we generate a giant concatenated toml file that supports all of poetry v1, poetry v2, and uv.

**Note: Poetry v2.x+ may complain and give warnings as it supports both syntaxes of poetry v1 and uv, but should not matter.**

## Changes Made
1. pyproject_toml now has different blocks that generates two toml file, legacy (generates toml file for poetry v1) and modern (generates toml file that is up to PEP621 standards, for uv and poetry v2).
2. It will now concat both of them, supporting all of our current python package managing dependency support

### DIFF
--- a/generators/python/src/fern_python/codegen/pyproject_toml.py
+++ b/generators/python/src/fern_python/codegen/pyproject_toml.py
@@ -290,7 +290,7 @@ httpx = "^0.24.0"
                     deps.append(f'{name} = ">={dep.version}"')
                 else:  # EXACT
                     # Check if version already contains a specifier
-                    if dep.version.startswith(("==", ">=", "<=", "~=", "!=")):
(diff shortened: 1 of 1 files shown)


sangjoon245:
Closed due to failing / hanging seed test.

PR #9706, opened 2025-10-01
chore(cli): add preview check to CLI

kennyderek:
## Description
 - We should warn people running CLI in terminal that they might affect production builds

## Changes Made
 - preview check and `--no-prompt` skip

## Testing

image


packages/cli/cli/versions.yml
```diff
@@ -1,4 +1,13 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- changelogEntry:
+    - summary: |
+        Add a confirmation prompt when generating docs for a production instance outside of GitHub Actions.
+        Add version logging.
```
kennyderek: i'll remove this one and add the explanation of `--no-prompt`

dsinghvi (approved):

PR #9713, opened 2025-10-01
fix(ruby): Fix Ruby v2 SDK construction to handle missing version configurations

chanks:
## Description
When SDK construction can't figure out an appropriate gem version, default to 0.0.0 and emit a warning. I'm not sure what we consider to be best practices here, so please feel free to tell me this is wrongheaded, but I do think it's an improvement over the current behavior of emitting a gem with the version string `"undefined"` and causing `bundle install` to error.

A better implementation is to accept a version being manually set as part of some kind of `downloadFiles` config, I'm just trying to unblock a build in the short term.

## Changes Made

- Updated `getVersionFromConfig` to never return `undefined` and instead emit a warning if it can't determine an appropriate version.

## Testing

- Relying on existing tests and type system.

### DIFF
--- a/generators/ruby-v2/base/src/project/RubyProject.ts
+++ b/generators/ruby-v2/base/src/project/RubyProject.ts
@@ -443,7 +443,7 @@ class VersionFile {
 
     public toString(): string {
         const seedName = this.context.getRootModuleName();
-        const version = this.context.getVersionFromConfig();
+        const version = this.context.getVersionFromConfig(this.context.logger);
 
         return dedent`
             # frozen_string_literal: true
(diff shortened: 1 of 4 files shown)


generators/ruby-v2/base/src/project/RubyProject.ts
```diff
@@ -443,7 +443,7 @@ class VersionFile {
 
     public toString(): string {
         const seedName = this.context.getRootModuleName();
-        const version = this.context.getVersionFromConfig();
+        const version = this.context.getVersionFromConfig(this.context.logger);
```
tjb9dc: it's weird to me that we're calling this.context.<function> and needing to pass in this.context.<field>, can it not just get the logger from the context in the function?

tjb9dc (approved):

generators/ruby-v2/base/src/project/RubyProject.ts
chanks: Ah, you're of course right, I'll tweak

PR #9715, opened 2025-10-01
fix(internal): Update dynamic IR to support the response types other than json

fearthecowboy:
## Description
Support examples with other than 'json' return types (streaming, fileDownload, etc) in dynamic IR

## Changes Made
- Changed dynamic IR to allow for more types
- Updated `local-workspace-runner` :
> the `local-workspace-runner` uses `convertIr` which always returns a `DynamicIntermediateRepresentation`
> that is actually of the latest version in the workspace.
> (regardless of version that the dynamic IR is being asked for in the language-specific generator)
> This appears to have been always an additive change, so this hasn't broken anything
> We're adding support for more types for the response, which is a mutation of the interface.
> This really shouldn't break the language-specific dynamic code generator, because there was never
> a need to check the `type` of the response.
> In order to not force a version bump of the language-specific dynamic code generator,
> we're casting the IR to `as unknown as any` until the individual generators have updated
- 

## Testing

- [x] Unit tests updated


packages/cli/generation/ir-generator/src/dynamic-snippets/DynamicSnippetsConverter.ts
```diff
@@ -140,7 +140,17 @@ export class DynamicSnippetsConverter {
...
-            response: DynamicSnippets.Response.json(),
+            response: (endpoint.response?.body?._visit<DynamicSnippets.Response>({
+                json: () => DynamicSnippets.Response.json(),
+                streaming: () => DynamicSnippets.Response.streaming(),
+                streamParameter: () => DynamicSnippets.Response.streamParameter(),
+                fileDownload: () => DynamicSnippets.Response.fileDownload(),
+                text: () => DynamicSnippets.Response.text(),
+                bytes: () => DynamicSnippets.Response.bytes(),
+                _other: () => DynamicSnippets.Response.json()
+            }) ?? DynamicSnippets.Response.json()) as DynamicSnippets.Response.Json,
```
tjb9dc: why do we have the cast here? I get why we have the default but not the cast :)

packages/cli/generation/ir-generator/src/dynamic-snippets/DynamicSnippetsConverter.ts
```diff
@@ -140,7 +140,17 @@ export class DynamicSnippetsConverter {
             declaration: this.convertDeclaration({ name: endpoint.name, fernFilepath: endpoint.fernFilepath }),
             location,
             request: this.convertRequest({ endpoint }),
-            response: DynamicSnippets.Response.json(),
+            response: (endpoint.response?.body?._visit<DynamicSnippets.Response>({
```
tjb9dc: is it safe to go ahead and produce this at this point? Do our tests ensure that the consumers are graceful?

packages/cli/generation/ir-generator/src/dynamic-snippets/DynamicSnippetsConverter.ts
fearthecowboy: The cast isn't necessary, dropping

packages/cli/generation/ir-generator/src/dynamic-snippets/DynamicSnippetsConverter.ts
fearthecowboy: It should be fine.

packages/cli/generation/ir-migrations/src/migrations/v60-to-v59/migrateFromV60ToV59.ts
tjb9dc, line 1: We need to add to the functionality here to downgrade the union shape to .json always in the dynamic IR.

After that, I would not expect to see a diff in any of the older IR snapshot tests (like packages/cli/generation/ir-migrations/src/migrations/v54-to-v53/__test__/__snapshots__/migrateFromV53ToV52.test.ts.snap)

PR #9723, opened 2025-10-02
fix(python): clear vulnerabilities of generator image for integration in enterprise automation pipelines

aditya-arolkar-swe:
## Description
Fixes: [FER-6650](https://linear.app/buildwithfern/issue/FER-6650/bloomberg-fix-vulnerabilities-in-docker-generator-image)

Fixes 6 unique vulnerabilities in the docker image for our python SDK generator. These need to be cleared before containers built off this image can be integrated into enterprise automation pipelines.

1. CVE-2024-27980 - Node.js 18.20.1: Command injection via batch files → upgrade to 18.20.2+
2. CVE-2024-36138 - Node.js 18.20.1: Bypass of previous command injection fix → upgrade to 18.20.4+
3. CVE-2025-23166 - Node.js 18.20.1: Crypto operations crash causing DoS → upgrade to 20.19.2+
4. CVE-2025-8194 - Python 3.11.13: Tarfile infinite loop on malicious archives → upgrade to 3.13.6+
5. GHSA-f5x3-32g6-xq36 - tar 6.2.0: DoS from malicious tar files → upgrade to 6.2.1+ 
6. GHSA-3xgq-45jj-v275 - cross-spawn 7.0.3: ReDoS vulnerability → upgrade to 7.0.5+ 

## Changes Made

- Updated python version: 3.11.13 -> 3.13.7
- Used base image `python:3.13.7-slim` instead of `python:3.13.7` for quicker build times (full image wasn't needed) 
- Updated node version: 18.20.1 -> 20.19.2
- Use latest npm to fix cross-spawn vulnerability
- Removed unnecessary layers 

## Testing


generators/python/sdk/Dockerfile
```diff
@@ -1,18 +1,9 @@
-# Stage 1: Build Node CLI
-FROM node:22.12-alpine3.20 AS node
```
tjb9dc, line 2: I'm pretty sure this will be safe :) it bit us on FDR and broke Prisma lol but this should be fine here

generators/python/sdk/Dockerfile
```diff
@@ -1,18 +1,9 @@
...
-
-RUN apk --no-cache add git zip
-RUN git config --global user.email "115122769+fern-api[bot]@users.noreply.github.com" && \
-    git config --global user.name "fern-api"
```
tjb9dc, line 6: Do you know why this was here? Seems like we would have had some git automation. A bit surprised that this would live in the python generator instead of base

tjb9dc: ah I see we have it at the bottom still

generators/python/sdk/Dockerfile
```diff
@@ -1,18 +1,9 @@
...
-# Stage 2: Base Python image with dependencies
-FROM python:3.11.13 AS python-base
+# Stage 1: Base Python image with dependencies
+FROM python:3.13.7-slim AS python-base
 
 # Install node and npm.
-ENV NODE_VERSION=18.20.1
-RUN apt install -y curl
+ENV NODE_VERSION=20.19.2
+RUN apt update && apt install -y curl git
```
tjb9dc, line 6: do we no longer need `zip`?

tjb9dc (approved):

generators/python/sdk/Dockerfile
aditya-arolkar-swe: This seems to have had a complicated history where we used to zip stuff inside it but no longer do. Worked to manually generate SDK with the current version, so leaving as-is unless issues arise in the future.

PR #9730, opened 2025-10-02
fix(internal): Bump IR to v61 - update dynamic IR to support the response types other than json 

fearthecowboy:
## Description
Bumps the IR version to v61 (breaking change)
Support examples with other than 'json' return types (streaming, fileDownload, etc) in dynamic IR

## Changes Made
- Changed dynamic IR to allow for more types

- Updated `local-workspace-runner` :
> the `local-workspace-runner` uses `convertIr` which always returns a `DynamicIntermediateRepresentation`
> that is actually of the latest version in the workspace.
> (regardless of version that the dynamic IR is being asked for in the language-specific generator)
> This appears to have been always an additive change, so this hasn't broken anything
> We're adding support for more types for the response, which is a mutation of the interface.
> This really shouldn't break the language-specific dynamic code generator, because there was never
> a need to check the `type` of the response.
> In order to not force a version bump of the language-specific dynamic code generator,
> we're casting the IR to `as unknown as any` until the individual generators have updated
- 

## Testing
- [x] Unit tests updated/regenerated output


packages/cli/generation/ir-migrations/src/migrations/v61-to-v60/migrateFromV61ToV60.ts
```diff
@@ -0,0 +1,96 @@
...
+            unrecognizedObjectKeys: "strip",
+            skipValidation: true
+        }),
+    migrateBackwards: (
+        v61: IrVersions.V61.IntermediateRepresentation
+    ): IrVersions.V60.ir.IntermediateRepresentation => {
+        return {
+            ...v61,
+            dynamic: v61.dynamic != null ? convertDynamic(v61.dynamic) : undefined
+        } as IrVersions.V60.ir.IntermediateRepresentation;
```
tjb9dc: we should never cast in these, I have PTSD from the issue I just fixed in https://github.com/fern-api/fern/pull/9734
```suggestion
        };
```

packages/cli/generation/ir-migrations/src/migrations/v61-to-v60/migrateFromV61ToV60.ts
```diff
@@ -0,0 +1,96 @@
...
+    endpoint: IrVersions.V61.dynamic.Endpoint
+): IrVersions.V60.dynamic.Endpoint | undefined {
+    return {
+        ...endpoint,
+        response: convertDynamicEndpointResponse(endpoint.response)
+    };
+}
+
+function convertDynamicEndpointResponse(response: IrVersions.V61.dynamic.Response): IrVersions.V60.dynamic.Response {
+    return IrVersions.V61.dynamic.Response._visit<IrVersions.V60.dynamic.Response>(response, {
```
tjb9dc, line 87: This works! For simplicity, could also just do a single IrVersions.V60.dynamic.Response.json line :)

packages/ir-sdk/fern/apis/ir-types-v60/changelog/CHANGELOG.md
```diff
@@ -0,0 +1,404 @@
+# Changelog
```
tjb9dc: We omit the changelog directory when cataloging IR versions

tjb9dc (commented):
Couple of things, but largely looks there

packages/cli/generation/ir-migrations/src/migrations/v61-to-v60/migrateFromV61ToV60.ts
fearthecowboy: I looked at other code in previous revisions on how to handle that, and this seemed actually elegant :D

packages/ir-sdk/fern/apis/ir-types-v60/changelog/CHANGELOG.md
fearthecowboy: fixing

packages/cli/generation/ir-migrations/src/migrations/v61-to-v60/migrateFromV61ToV60.ts
fearthecowboy: That was in the 60-to-59 code that I copied across.

tjb9dc (approved):
LGTM

PR #9738, opened 2025-10-02
fix(cli): Improve OpenAPI to Fern Definition parser

Swimburger:
## Description
Linear ticket:

## Changes Made
* Fix an issue where some optional properties were converted to required in the OpenAPI parser to Fern Definition.
* Fix an issue where some property descriptions were not being carried over in the OpenAPI parser to Fern Definition.
* Generate more compact Fern Definition when only a type is provided for optional properties.
  ```yml
  # Before
  prop:
    type: optional<Foo>

  # After
  prop: optional<Foo>
  ```

## Testing


tjb9dc (approved):
Reviewed sync, very very mildly concerned about the diff to generatedName but we decided that shouldn't matter.

PR #9744, opened 2025-10-03
fix(typescript): Handle colons in basic auth passwords

tjb9dc:
We weren't handling basic auth passwords with colons before. This change handles that, while being sure to still support the existing behavior that allows empty usernames and/or passwords, as long as they are properly "delimited" with a colon.


sahil485 (approved):
lgtm!

PR #9751, opened 2025-10-03
fix(cli): Fix whiteLabeled readme for remote generation

jsklan:
## Description
Previously this only worked for local generation. Now it works for botjh

## Changes Made
- Updated runRemoteGenerationForGenerator.ts to match logic in runLocalGenerationForWorkspace.ts for fetching org details from venus and finalizing ir values with settings

## Testing
- [x] Manual testing against sample project


packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts
```diff
@@ -88,6 +89,18 @@ export async function runRemoteGenerationForGenerator({
         dynamicGeneratorConfig
     });
 
+    const venus = createVenusService({ token: token.value });
+    const orgResponse = await venus.organization.get(FernVenusApi.OrganizationId(projectConfig.organization));
```
tjb9dc, line 93: is this smart enough to match that the token we're running with belongs to the org? E.g. if I have orgA's token and put orgB's name in my project config, what does this API call return

jsklan: I'm not sure, this is just what we do in local generation, so I copied that.

tjb9dc (approved):

PR #9755, opened 2025-10-03
fix(go): readme generation for new section on wire tests

musicpulpite:
self-explanatory. Will cut a new generator version for this after few other changes are merged as well


musicpulpite:
Most likely closing this since it doesn't make sense to have this info in the customer-exposed docs

PR #9772, opened 2025-10-06
feat(python): switch to `uv`

patrickthornton:
## Description
Linear ticket: Fixes FER-6797, Fixes FER-6633
Switches to `uv` internally for Python SDK package management, building, and CI/CD, which gives a little speedup. The `pyproject.toml`s generated with Python SDKs are now PEP-621-compliant, which means they can be used by any package manager that supports it; this includes `uv`, `poetry>=2.0`, `pip`, etc. This releases `5.0.0-rc0` of the Python SDK generator.

## Changes Made
Lots of little changes to `pyproject.toml` generation (plus an added snapshot test file for it), seed workflows, dockerfiles, and so on.

## Testing
All tests pass as before.


generators/python/.pre-commit-config.yaml
```diff
@@ -34,8 +34,3 @@ repos:
         args: [--fix, --config=generators/python/pyproject.toml]
       - id: ruff-format
         args: [--config=generators/python/pyproject.toml]
-  - repo: https://github.com/python-poetry/poetry
-    rev: "1.2"
-    hooks:
-      - id: poetry-check
-      - id: poetry-export
```
tjb9dc, line 41: we probably want the equivalent check here. Basically the intent is to make sure that the lockfile is up to date and what not

generators/python/pydantic/Dockerfile
```diff
@@ -4,16 +4,15 @@ FROM python:3.11.13 AS base
 ENV PYTHONPATH=${PYTHONPATH}:${PWD}
 ENV _TYPER_STANDARD_TRACEBACK=1
 
-RUN pip3 install poetry==1.8.5
-RUN poetry config virtualenvs.create false
+RUN pip3 install uv
```
tjb9dc: hm can we pin this to some reasonable version? Or at least pin to major version

generators/python/pyproject.toml
```diff
@@ -1,36 +1,59 @@
...
+dev = [
+    "mypy==1.13.0",
+    "pre-commit>=2.20.0,<3.0.0",
+    "pytest>=7.4.2,<8.0.0",
+    "pytest-only>=2.0.0,<3.0.0",
+    "ruff==0.11.5",
+    "setuptools>=80.9.0,<81.0.0",
+]
+
+[tool.poetry]
```
tjb9dc: do we care about this file being backcompat with poetry? Couldn't we just wholesale migrate to uv internally?

generators/python/setup-python.sh
```diff
@@ -4,5 +4,4 @@ export PYENV_ROOT="$HOME/.pyenv"
 [[ -d $PYENV_ROOT/bin ]] && export PATH="$PYENV_ROOT/bin:$PATH"
 eval "$(pyenv init -)"
 pyenv shell 3.9
-pip install poetry
-poetry env use 3.9
+pip install uv
```
tjb9dc: also would like to pin here

generators/python/src/fern_python/codegen/ast/dependency/dependency.py
```diff
@@ -1,22 +1,15 @@
...
-
 @dataclass(frozen=True)
 class Dependency:
     name: DependencyName
-    version: DependencyVersion
+    version: DependencyVersion  # Should include operator like "==1.0.0" or ">=1.0.0"
     # Using a tuple here to ensure immutability and make it hashable
     python: Optional[str] = None
     extras: Optional[Tuple[str, ...]] = None
-    compatibility: DependencyCompatibility = DependencyCompatibility.EXACT
```
tjb9dc, line 21: what is this change? We're just offloading the handling to uv?

generators/python/src/fern_python/cli/abstract_generator.py
```diff
@@ -288,15 +286,14 @@ def _get_github_workflow(self, output_mode: GithubOutputMode, write_unit_tests:
         uses: actions/setup-python@v4
         with:
           python-version: 3.8
-      - name: Bootstrap poetry
-        run: |
-          curl -sSL https://install.python-poetry.org | python - -y --version 1.5.1
+      - name: Install uv
+        run: pip install uv
```
tjb9dc: another spot we might want to pin

generators/python/pyproject.toml
patrickthornton: internally dropping the poetry stuff is probably fine, yeah

generators/python/pydantic/Dockerfile
patrickthornton: will do the pins

generators/python/src/fern_python/codegen/ast/dependency/dependency.py
patrickthornton: this is a bit of cruft around versions; this was like an enum that signified what the sign before a version number would be, but it was incomplete and would be really cumbersome if used literally (like what should it be if the version is `">=3.9,<4.0"`?) so now the version string is expected to contain the operator(s) and DependencyCompatibility is cut. if it lacks an operator it defaults to ==

generators/python/.pre-commit-config.yaml
patrickthornton: i think `uv sync` will just fail if the lock isn't up to date? and export isn't needed since there's no `requirements.txt` needed here when we just use `uv.lock`?

generators/python/core_utilities/shared/pydantic_utilities.py
```diff
@@ -8,14 +8,14 @@
...
-    from pydantic.v1.typing import is_literal_type as is_literal_type
-    from pydantic.v1.typing import is_union as is_union
+    from pydantic.v1.datetime_parse import parse_date as parse_date  # type: ignore[import]
+    from pydantic.v1.datetime_parse import parse_datetime as parse_datetime  # type: ignore[import]
+    from pydantic.v1.fields import ModelField as ModelField  # type: ignore[import]
+    from pydantic.v1.json import ENCODERS_BY_TYPE as encoders_by_type  # type: ignore[import, attr-defined]
+    from pydantic.v1.typing import get_args as get_args  # type: ignore[import]
+    from pydantic.v1.typing import get_origin as get_origin  # type: ignore[import]
+    from pydantic.v1.typing import is_literal_type as is_literal_type  # type: ignore[import]
+    from pydantic.v1.typing import is_union as is_union  # type: ignore[import]
```
tjb9dc: what's with this

patrickthornton: ah, this can just be conditional on v1 like the other overrides

PR #9784, opened 2025-10-06
feat(python): eliminate SSE dependency 'httpx_sse' by hard-forking into core_utilities

aditya-arolkar-swe:
## Description
Generator change to yield this output: https://github.com/cohere-ai/cohere-python/pull/700

Fixes FER-6826

## Changes Made

- **Hard forked `httpx_sse`, moved into `core_utilities`, made fixes to line parsing bugs and improved idiomatic class definition.**
- **updated parser to: copy `http_sse` code into output, use new in-house SSE `EventStream` class and better handle failed SSE parsing**

## Testing


generators/python/core_utilities/shared/http_sse/_api.py
```diff
@@ -0,0 +1,89 @@
...
+        decoder = SSEDecoder()
+
+        buffer = ""
+        for chunk in self._response.iter_bytes():
+            # Decode chunk and add to buffer
+            text_chunk = chunk.decode("utf-8", errors="replace")
+            buffer += text_chunk
+
+            # Process complete lines
+            while "\n" in buffer:
```
tjb9dc, line 58: did you test this with a multiline event (like a JSON blob containing a content field with extra lines)? This doesn't appear to handle that to me

generators/python/sdk/versions.yml
```diff
@@ -1,5 +1,13 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 # For unreleased changes, use unreleased.yml
+- version: 4.31.0
+  changelogEntry:
+    - summary: |
+        Eliminated SSE dependency "httpx-sse" by hard-forking the code into the generator
```
tjb9dc: Can we have it anchored in the user-facing outcome more? I also don't love "hard-forking" language in the changelog. I think the points we are trying to land are:

- **Fix** SSE handling of events longer than <blah> or containing escaped newlines
- **Remove** external dependency on httpx-sse

generators/python/core_utilities/shared/http_sse/_models.py
```diff
@@ -0,0 +1,15 @@
...
+@dataclass(frozen=True)
+class ServerSentEvent:
+    event: str = "message"
+    data: str = ""
+    id: str = ""
+    retry: Optional[int] = None
+
+    def json(self) -> Any:
+        """Parse the data field as JSON."""
+        return json.loads(self.data)
```
tjb9dc, line 15: Does this have all the built-in decoders that we use for other json deser? I wonder if we want to use the pydantic decoder/JSON loader for consistency

generators/python/core_utilities/shared/http_sse/_api.py
```diff
@@ -0,0 +1,93 @@
...
+                # => decoder will attempt to return an SSE Event
+                if sse is not None:
+                    yield sse
+        
+        # Process any remaining data in buffer
+        if buffer.strip():
+            line = buffer.rstrip('\r')
+            sse = decoder.decode(line)
+            if sse is not None:
+                yield sse
```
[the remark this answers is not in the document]
tjb9dc: I feel like we should catch this one with a test?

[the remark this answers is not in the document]
tjb9dc: I think this is wrong, and the SSE spec is very explicit about the double newline splitting events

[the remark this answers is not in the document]
aditya-arolkar-swe: This is actually intentional. If the SSE server doesn't send a final chunk that closes out an SSE event. we don't want to force the processing of an event, which is what the graphite suggested logic would do: 
```
            # Signal end of stream to finalize any pending event
            final_sse = decoder.decode('')
```
The current version would be a stricter way to handle SSE chunks, where the server must end on /n/n to close out events or our SDK won't process it.

generators/python/core_utilities/shared/http_sse/_api.py
aditya-arolkar-swe: I tested it out with Cohere's bug. The extra lines in the JSON blob must be escaped ("\\n\\n") and this works fine. If they are regular double new lines ("\n\n"), we will interpret it as a new event.

generators/python/core_utilities/shared/http_sse/_models.py
aditya-arolkar-swe: Our standard deserializer is the one I am using - also seen in `generators/python/src/fern_python/generators/sdk/client_generator/socket_client_generator.py`. Happy to migrate to something better.

tjb9dc: this is just the stdlib json parser right? I'm surprised we are using that in other places

aditya-arolkar-swe: Yeah - specifically https://docs.python.org/3/library/json.html

generators/python/tests/utils/test_http_sse.py
tjb9dc, line 1: So we _could_ include this file (or a version of it) like the other as-is unit test files and generate it for all the customers who have SSE included in their APIs.

I have a moderate preference for doing that, but regardless this looks good :)

tjb9dc (approved):

PR #9787, opened 2025-10-07
fix(ci): remove need for Auto Seed Update

musicpulpite:
## Summary
1. In order to require all seed tests we needed to pass the `git diff` check ensuring that we have committed all seed changes in our PR
2. The quick fix to address this was to run "Auto Seed Update" to commit the changes to the PR immediately. But this has the unintended consequence of rerunning the original seed test from scratch that it was intended to fix - essentially doubling our time-to-first-signal from seed tests and slowing our developer velocity.
3. Since we still have the jobs to update seed after a PR is merged we can rely on that to keep the repo in sync (for now but we'll look for better alternatives long-term) and don't need to require these changes to be included in PR's at all. All we need is that the seed tests are fast and accurate

### DIFF
--- a/.github/actions/cached-seed/action.yaml
+++ b/.github/actions/cached-seed/action.yaml
@@ -96,40 +96,6 @@ runs:
           ${{ inputs.skip-scripts == 'true' && '--skip-scripts' || '' }} \
           ${{ inputs.allow-unexpected-failures == 'true' && '--allow-unexpected-failures' || '' }}
 
-    - name: Validate results
-      if: inputs.validate-git-diff == 'true'
-      id: validate
-      shell: bash
-      run: |
-        # Validate results
-        # Split the input string into an array
-        IFS=',' read -ra FOLDERS <<< "${{ inputs.validation-exclude-paths }}"
-
-        # Build the exclude string
-        EXCLUDE_STRING=""
-        for folder in "${FOLDERS[@]}"; do
-          # Trim whitespace
-          folder=$(echo "$folder" | xargs)
-          if [ -n "$folder" ]; then
-            EXCLUDE_STRING="$EXCLUDE_STRING\":(exclude)$folder\" "
-          fi
-        done
-
-        # Remove trailing space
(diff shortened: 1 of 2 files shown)


.github/workflows/update-seed.yml
```diff
@@ -1,1449 +0,0 @@
...
-      - "packages/cli/generation/ir-generator/**"
-      - "test-definitions/**"
-      - "test-definitions-openapi/**"
-      - "generators/**"
-      - "seed/**"
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
-    branches:
-      - main
-  workflow_dispatch:
```
chrismallinson, line 20: To remove from the PR workflow, you want to remove the `pull_request` section from here. The workflow to update seed with merges to main uses the same code as the one that runs at a commit level and this is where that gets triggered. I would also like to keep the `workflow_dispatch` so we can use it for future triggers if we want to manually make an update

Please remove the pull_request section but add back the rest (you can also keep the part in `cached-seed` deleted if we aren't running that check anymore

chrismallinson (changes requested):

.github/workflows/update-seed.yml
musicpulpite: Oops you are so right! Man this is harder than I thought 😆

.github/actions/cached-seed/action.yaml
```diff
@@ -96,40 +96,6 @@ runs:
           ${{ inputs.skip-scripts == 'true' && '--skip-scripts' || '' }} \
           ${{ inputs.allow-unexpected-failures == 'true' && '--allow-unexpected-failures' || '' }}
 
-    - name: Validate results
-      if: inputs.validate-git-diff == 'true'
```
chrismallinson, line 100: not 100% necessary but would be nice to remove this from the inputs too now that it is dead code: `validate-git-diff`. Sorry I missed that on my first run through, will approve otherwise

chrismallinson (approved):

.github/actions/cached-seed/action.yaml
musicpulpite: You are 100% correct sir

PR #9794, opened 2025-10-07
feat(typescript): `timeoutMs` and `maxRetries` configurable at client-level

patrickthornton:
## Description
Linear ticket: Fixes FER-7110
Allows for setting `timeoutInSeconds` and `maxRetries` when instantiating the client, like so:
```
const client = new SeedExhaustiveClient({
      environment: "production",
      token: "my-token",
      timeoutInSeconds: 30,  // All requests default to 30s timeout
      maxRetries: 5           // All requests default to 5 retries
});

// This request uses the client defaults (30s timeout, 5 retries)
await client.endpoints.getPrimitive();

// This request overrides the timeout but uses the client default for retries (5)
await client.endpoints.getPrimitive({
    timeoutInSeconds: 60  // Override: use 60s instead of the client default 30s
});
```

## Changes Made
Updates mostly to the `requestOptionsParameter.ts`.

## Testing
Autogenerating seed tests now!


generators/typescript/sdk/versions.yml
```diff
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.5.0
+  changelogEntry:
+    - summary: |
+        Allow for configuring certain request options at the client-level as well as the request level.
```
Swimburger: Let's mention which options in the changelog

Swimburger (approved):

PR #9795, opened 2025-10-07
fix(csharp): Address nullability issues for seed failures

fearthecowboy:
## Description

This PR addresses issues with nullability related to seed fixture failures. 

## Changes Made
- when creating the headers block, parameters are guarded against null with `?? ""` 

- ensure no name collisions when throwing `System.Exception`.

- `ToString()` implementation guards against null with `?? ""`

- when deserializing discriminated unions, if the type is a class, struct, or record it always deserializes  via the optional type and then if it is not nullable, throws on null  This makes the code consistent between classes/structs/records  because the record struct is not nullable, and use of `??` is a compiler error.

- in generated examples (server tests and snippets), dictionaries with optional values were not being generated with the optional type on the dictionary, causing a mismatch between the dictionary and the type of the dictionary.

- if the inner type of a union was called "Value" or "Type" the code generator renames the type to "ValueInner" or "TypeInner". The dynamic snippet generator didn't follow that.

- When generating the class `NotFoundError` if the body type is still `object` the code generator does not need to override the body property at all. 

- reduced number of 'allowed failures' to 6 

- in #8992 - the concept of a nullable parameter that is not declared optional was introduced. This ensures that if the value is null, that the parameter is checked before pushing into the _query

- if a dynamic snippet was creating a `List<T>` and the element typewas not supposed to be nullable or optional,  the code was creating the element type in the `List<T?>` as optional regardless. 

- when using the `clientOptions` there is a if block that checks to see if a value is not null, but the compiler can't propagate the nullness to the member when calling `ToString()` so this adds a `!` to the `ToString()` call.

- idempotency headers should have been required if they are not nullable. 

- in the generated example tests, idempotency headers were not initialized to a value, the interface was incorrectly  implemented.

- [x] Updated versions.yml

## Testing

- [x] Seed fixtures regenerated


generators/csharp/model/src/union/UnionGenerator.ts
```diff
@@ -666,25 +665,24 @@ export class UnionGenerator extends FileGenerator<CSharpFile, ModelCustomConfigS
...
                                     assertNever(type.shape);
                             }
-                            writer.write(".Deserialize<");
-                            writer.writeNode(csharpType);
-                            writer.write(">(options)");
-                            if (csharpType.isOptional()) {
-                                writer.writeLine(",");
+                            if (csharpType.isReferenceType() === false) {
+                                // non-reference types can be always be deserialized directly as is
+                                writer.write(".Deserialize<", csharpType, ">(options)");
```
Swimburger, line 670: Was the `writer.write` updated to support multiple args, and to write a node instead of using `writeNode`?

fearthecowboy: Yes

generators/csharp/sdk/versions.yml
```diff
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.4.1
+  changelogEntry:
+    - type: fix
+      summary: |
+        Fix nullability issues/seed failures
```
Swimburger: I wouldn't mention "seed" since the changelog is for our users to read and understand. Let's focus on how it improves the SDK for users.

generators/csharp/base/src/asIs/RawClient.Template.cs
```diff
@@ -214,7 +214,7 @@ and not string
                 {
                     var items = collection
                         .Cast<object>()
-                        .Select(value => $"{Uri.EscapeDataString(queryItem.Key)}={Uri.EscapeDataString(value.ToString())}")
+                        .Select(value => $"{Uri.EscapeDataString(queryItem.Key)}={Uri.EscapeDataString(value.ToString() ?? "null")}")
```
Swimburger: I vaguely remember there are some classes that handle conversion to query string in the C# generator.

Swimburger: The `ValueConvert` class with `ToQueryStringValue`. It also converts `null` to `""`, not `"null"`. Is there a customer that is asking for it to convert to `"null"`?

generators/csharp/sdk/src/test-generation/mock-server/BaseMockServerTestGenerator.ts
```diff
@@ -80,14 +80,29 @@ export class BaseMockServerTestGenerator extends FileGenerator<CSharpFile, SdkCu
...
+            // create an initializer for the fields
+            const initializer = this.csharp.codeblock((writer) => {
+                writer.writeLine("new() {");
+                writer.indent();
+                fields.forEach((field) => {
+                    if (field.isRequired) {
+                        writer.write(field.name, " = ", this.csharp.getDefaultValue(field.type), ",");
+                        writer.writeLine();
+                    }
+                });
```
Swimburger, line 93: Yes 🔥

seed/csharp-model/examples/src/SeedExamples/Types/Metadata.cs
```diff
@@ -189,9 +189,9 @@ JsonSerializerOptions options
 
             var value = discriminator switch
             {
-                "html" => json.GetProperty("value").Deserialize<string>(options)
+                "html" => json.GetProperty("value").Deserialize<string?>(options)
                 ?? throw new JsonException("Failed to deserialize string"),
-                "markdown" => json.GetProperty("value").Deserialize<string>(options)
+                "markdown" => json.GetProperty("value").Deserialize<string?>(options)
```
Swimburger, line 194: Is `string?` here a string because `value` is a nullable property? If so, we should remove the `?? throw ...`.
Otherwise, good as is

Swimburger:
I would clarify whether `null` should become `"null"` in the querystring, other than that, LGTM!

generators/csharp/base/src/asIs/RawClient.Template.cs
fearthecowboy: No. And now that I'm looking at this, I have no idea why I put `"null"` in there.  Fixing.

seed/csharp-model/examples/src/SeedExamples/Types/Metadata.cs
fearthecowboy: It's not actually a nullable property -- the generator won't put a throw on a nullable property deserializing as null. This is intentional

generators/csharp/sdk/src/test-generation/mock-server/BaseMockServerTestGenerator.ts
fearthecowboy: Yeah, it was a tiny change, but the usability is so much better.

generators/csharp/sdk/versions.yml
fearthecowboy: fixed

Swimburger (approved):

PR #9800, opened 2025-10-08
chore(internal): Add .cursorindexingignore

jsklan:
https://cursor.com/docs/context/ignore-files#limit-indexing-with-cursorindexingignore


WARP.md
tjb9dc, line 1: what's this

tjb9dc (approved):

.cursorindexingignore
```diff
@@ -0,0 +1,61 @@
...
+tests
+
+# Seed files
+seed/csharp-model/*
+!seed/csharp-model/seed.yml
+seed/csharp-sdk/*
+!seed/csharp-sdk/seed.yml
+seed/fastapi/*
+!seed/fastapi/seed.yml
+seed/fern-cli/*
```
chrismallinson, line 19: nit: would it be simpler to have seed/* and then !**/seed.yml ?

chrismallinson (approved):
nit, otherwise approved

.cursorindexingignore
kafkas: Maybe this instead ?

```
seed/**/*
!seed/*/seed.yml
```

jsklan: Yes, but cursorignore doesn't support this format. 
https://cursor.com/docs/context/ignore-files#negation-pattern-limitations

kafkas: Interesting, you're right. Neither does git apparently

WARP.md
jsklan: It's just a file that points warp to claude for context. Warp is compatible with most claude and cursor settings files, so I may be adding more of these small things to make sure it points to those if folks are using it.
https://www.warp.dev/

PR #9815, opened 2025-10-09
fix(cli): Use FDR instead of hardcoded values to determine which IR version to supply to the generator

jsklan:
## Description

Improve IR migration by dynamically fetching IR version information from FDR instead of relying on hardcoded values. This change enhances the reliability and maintainability of the generator version compatibility system by querying the Fern registry for the exact IR version required by each generator.

  ## Changes Made
  - Added `getIrVersionForGenerator` function to
  `packages/core/src/services/fdrGeneratorsSdk.ts` to fetch IR versions from FDR
  - Updated local generation (`getIntermediateRepresentation.ts`) to use FDR-sourced IR
   versions as fallback when no override is specified
  - Updated remote generation (`createAndStartJob.ts`) to use FDR-sourced IR versions
  as fallback for Fiddle service
  - Refactored `compatible-ir-versions` validation rule to use the centralized
  `getIrVersionForGenerator` function
  - Added version `0.86.1` to CLI versions with appropriate changelog entry
  - Exported new `getIrVersionForGenerator` function from core services module

  ## Testing


packages/cli/generation/local-generation/local-workspace-runner/src/getIntermediateRepresentation.ts
```diff
@@ -61,12 +62,16 @@ export async function getIntermediateRepresentation({
         intermediateRepresentation.sourceConfig = sourceConfig;
     }
     context.logger.debug("Generated IR");
+    const irVersionFromFdr = await getIrVersionForGenerator(generatorInvocation).then((version) =>
+        version == null ? undefined : "v" + version.toString()
```
tjb9dc, line 66: bruhh are these not normalized? We have to add a v to it?

tjb9dc (approved):

PR #9828, opened 2025-10-10
fix(cli): Fix diff command to output no version change when IR is equal

jsklan:
## Description
This PR enhances the Fern CLI's diff command to properly handle identical API definitions by returning null when there are no changes. This provides more accurate versioning information when comparing APIs that haven't changed.

## Changes Made

- Updated diff.ts to handle null bump values for identical IRs, preventing unnecessary version increments
- Enhanced IntermediateRepresentationChangeDetector to return null bump when IRs are identical (using JSON hash comparison)
- Modified maxBump function logic to properly handle null values in bump priority calculations
- Updated all related test files to expect null bumps for identical comparisons instead of boolean breaking change flags
- Added comprehensive test case for identical IR comparison to ensure proper null bump detection
- Updated README.md generator (if applicable)

## Testing
- [x] Unit tests added/updated - Added test case for identical IRs in breakingChangeDetector.test.ts:71
- [x] Manual testing completed - Updated all existing test cases in diffSampleIr.test.ts to use new bump-based assertions
- [x] Tests verify proper handling of identical IRs returning null bump
- [x] Tests verify backward compatibility with existing breaking/non-breaking change detection


packages/cli/cli/src/commands/diff/diff.ts
```diff
@@ -8,15 +8,15 @@ import semver from "semver";
 
 import { CliContext } from "../../cli-context/CliContext";
 
-export type Bump = "major" | "minor" | "patch";
+export type Bump = "major" | "minor" | "patch" | null;
```
tjb9dc: I would prefer a sentinel value for null. Like you could just use `"null"` or `"no-change"`.

The reason I don't love this is that this type is `Bump | null | undefined`, and devs like to write code like `bump != null` which will match on two of the types, or `bump ?? foo` which will fall back to foo in both cases when maybe they need to handle `no-change` clearly

For consumers of the API, I think it's also more obvious to them that getting back `no-change` is different from getting back `undefined` or omitting the argument

packages/cli/cli/src/commands/diff/diff.ts
```diff
@@ -8,15 +8,15 @@ import semver from "semver";
...
 export interface Result {
     bump: Bump;
     nextVersion?: string;
     errors: string[];
 }
 
 interface InternalResult {
-    bump?: Bump;
+    bump?: Bump | null;
```
tjb9dc: I think these types in these functions are redundantly including `| null` because you changed the `Bump` type already

tjb9dc (approved):

PR #9851, opened 2025-10-11
chore(cli): update configs

chdeskur:
adds docs.yml configs for: 
- rust sdk snippets
- canonical host
- prefer javascript over typescript for snippets

removes duplicate config for disabling on-page feedback


dsinghvi (approved):

fern/apis/docs-yml/definition/docs.yml
```diff
@@ -1094,6 +1096,7 @@ types:
       "twitter:site": optional<string>
       "twitter:url": optional<string>
       "twitter:card": optional<TwitterCardSetting>
+      canonical-host: optional<string>
```
dannysheridan, line 1099: Why no quotes? Can we make consistent?

packages/cli/cli/versions.yml
```diff
@@ -1,4 +1,15 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- changelogEntry:
+    - summary: |
+        Adds support for the following docs.yml features:
+        - `settings.use-javascript-as-typescript`:
```
dannysheridan: Can you define the settings here? Helpful for Ask Fern.

dannysheridan (approved):

fern/apis/docs-yml/definition/docs.yml
chdeskur: we only need quotes if there is a non-standard character in the key itself. the ones above have a `:`, so they need quotes to be parsed properly

packages/cli/cli/versions.yml
chdeskur: i'll be updating the docs when these actually get added to the frontend. for now, technically these will not work

PR #9872, opened 2025-10-15
chore(rust): bump `ir-sdk` version and migration to `v61` support `--version` params

iamnamananand996:
## Description
Linear ticket:
- https://linear.app/buildwithfern/issue/FER-7202/rust-sdk-add-support-for-fern-generate-api-v1-group-rust-sdk-version

## Changes Made

- Upgraded Rust generators (base, model, SDK) to use IR SDK v61.4.0 (from v59.7.0)
- Added support for `--version` parameter in Rust SDK generation with proper priority order:
  -  `customConfig.crateVersion` (explicitly set in `generators.yml` or for `--local` generation)
  - `publishConfig.publishTarget.version` (set via `output.location = crates` for remote generation)
  - Default: `"0.1.0"`
- Updated publish target type check from `pypi` to `crates` for proper version handling
- Added IR migration entries for Rust SDK versions 0.7.2 (IR v60) and 0.7.3 (IR v61)

## Testing

### DIFF
--- a/generators/rust/sdk/versions.yml
+++ b/generators/rust/sdk/versions.yml
@@ -1,5 +1,19 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 
+- version: 0.7.3
+  createdAt: "2025-10-15"
+  changelogEntry:
+    - type: chore
+      summary: Upgraded to IR v61 and --version params support
+  irVersion: 61
+
+- version: 0.7.2
+  createdAt: "2025-10-15"
+  changelogEntry:
+    - type: chore
+      summary: Upgraded to IR v60
+  irVersion: 60
+
 - version: 0.7.1
   createdAt: "2025-10-15"
   changelogEntry:
--- a/generators/rust/sdk/versions.yml
+++ b/generators/rust/sdk/versions.yml
@@ -1,18 +1,11 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 
-- version: 0.7.3
+- version: 0.7.2
   createdAt: "2025-10-15"
   changelogEntry:
     - type: chore
       summary: Upgraded to IR v61 and --version params support
   irVersion: 61
-
-- version: 0.7.2
-  createdAt: "2025-10-15"
-  changelogEntry:
-    - type: chore
-      summary: Upgraded to IR v60
-  irVersion: 60
 
 - version: 0.7.1
   createdAt: "2025-10-15"
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,6 +1,15 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 - changelogEntry:
     - summary: |
+        Prevent accidental changes to doc sites by adding a confirmation prompt when generating docs for a production instance outside of common CI/CDs.
+        Bypass this prompt by adding a `--no-prompt` flag to the `fern generate` command.
+      type: feat
+  irVersion: 61
+  createdAt: "2025-10-14"
+  version: 0.90.1
+
+- changelogEntry:
+    - summary: |
         Add Rust crates.io publishing support. Rust SDK now uses --version flag for Cargo.toml version.
       type: feat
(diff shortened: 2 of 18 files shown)


generators/rust/sdk/versions.yml
```diff
@@ -1,5 +1,19 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 
+- version: 0.7.3
```
iamnamananand996: hey @tjb9dc, 

Just a mention, if we observe here on the version bumps, have need to add 2 version bumps up, because of the upgrade of `ir-sdk` from `v59` --> `v61`

Want to double-check, this kind of scenario is fine?

tjb9dc: yep! Totally fine to bump an arbitrary number of versions at once

packages/cli/cli/versions.yml
```diff
@@ -3,7 +3,7 @@
     - summary: |
         Prevent accidental changes to doc sites by adding a confirmation prompt when generating docs for a production instance outside of common CI/CDs.
         Bypass this prompt by adding a `--no-prompt` flag to the `fern generate` command.
-      type: feat
+      type: chore
```
tjb9dc, line 6: why did this change?

generators/rust/sdk/versions.yml
tjb9dc: I think we only want one of these new versions right? So just 0.7.2

packages/cli/generation/ir-migrations/src/migrations/v60-to-v59/migrateFromV60ToV59.ts
```diff
@@ -39,7 +39,7 @@ export const V60_TO_V59_MIGRATION: IrMigration<
         [GeneratorName.PHP_MODEL]: GeneratorWasNeverUpdatedToConsumeNewIR,
         [GeneratorName.PHP_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR,
         [GeneratorName.RUST_MODEL]: GeneratorWasNeverUpdatedToConsumeNewIR,
-        [GeneratorName.RUST_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR
+        [GeneratorName.RUST_SDK]: "0.7.2"
```
tjb9dc: In which case this should go back to `GeneratorWasNeverUpdatedToConsumeNewIR`

packages/cli/generation/ir-migrations/src/migrations/v61-to-v60/migrateFromV61ToV60.ts
```diff
@@ -38,7 +38,7 @@ export const V61_TO_V60_MIGRATION: IrMigration<
         [GeneratorName.PHP_MODEL]: GeneratorWasNeverUpdatedToConsumeNewIR,
         [GeneratorName.PHP_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR,
         [GeneratorName.RUST_MODEL]: GeneratorWasNeverUpdatedToConsumeNewIR,
-        [GeneratorName.RUST_SDK]: GeneratorWasNeverUpdatedToConsumeNewIR
+        [GeneratorName.RUST_SDK]: "0.7.3"
```
tjb9dc: And then this would be 0.7.2

packages/cli/cli/versions.yml
iamnamananand996: Looks like done accidentally by @kennyderek 

the CLI bump was `0.90.0` --> `0.90.1` 

But the type was written `feat`, which was failing in the change log check.

So either we should update the version number to `0.91.0` or have type - `chore`

tjb9dc: ohh sounds good

tjb9dc: chore is good with me, could see either way though

generators/rust/sdk/versions.yml
iamnamananand996: updated

tjb9dc (approved):

tjb9dc (approved):

PR #9908, opened 2025-10-17
chore(rust): auto-generate publish workflow based on `publishConfig`

iamnamananand996:
## Description
Linear ticket:
- https://linear.app/buildwithfern/issue/FER-7214/rust-sdk-have-publish-mechanism-in-ci

## Changes Made
- **Publish Workflow Automation**: Changes for automatic generation based on `publishConfig`
  - Stores `publishConfig` in `AbstractRustGeneratorContext` for easy access
  - add `RustProject.generatePublishWorkflow()` to add CI workflow


generators/rust/codegen/src/custom-config/BaseRustConfigSchema.ts
```diff
@@ -13,7 +13,9 @@ export const BaseRustCustomConfigSchema = z.object({
     packageDescription: z.string().optional(),
     packageLicense: z.string().optional(),
     packageRepository: z.string().optional(),
-    packageDocumentation: z.string().optional()
+    packageDocumentation: z.string().optional(),
+    // CI/CD configuration
+    enablePublishWorkflow: z.boolean().optional().default(false)
```
tjb9dc: Why not default to true? I think we do that in other languages.

also can you make sure this aligns with whatever flags we use in other languages

tjb9dc (approved):

generators/rust/codegen/src/custom-config/BaseRustConfigSchema.ts
iamnamananand996: Lemme check, how it's done in other languages, will update it accordingly.

PR #9982, opened 2025-10-20
fix(cli): handle breaking errors at top level

kennyderek:
handle some errors that cause the entire flow to break, at a higher level

With `fern check` changes upcoming, these should be caught at the `fern check` level. `fern generate` should not fail


dsinghvi (approved):
Can you create a `versions.yml` so that we can release the change?

PR #9990, opened 2025-10-21
fix(python): resolve circular import issues with self-referencing union types

tjb9dc:
Fixes circular import errors that occurred when generating Python SDKs with complex self-referencing union types (e.g., ElevenLabs SDK with AST node types).

The issue was that union variant classes were adding ghost references that caused imports to be written in the middle of files, breaking Python's import resolution for circular dependencies.

Changes:
- Use FernAwarePydanticModel instead of PydanticModel for single-property union variants to properly handle circular references
- Skip adding ghost references for union variants to defer all imports until after the union type definition
- Skip calling update_forward_refs() on union variants to avoid infinite recursion in Pydantic's schema resolution

This ensures all imports are written after class definitions, allowing Python's forward reference mechanism to resolve circular dependencies correctly.


generators/python/src/fern_python/generators/pydantic_model/type_declaration_handler/pydantic_models/pydantic_model_simple_discriminated_union_generator.py
```diff
@@ -108,32 +108,19 @@ def _generate_single_property_member(
         properties: List[PydanticField],
         fern_aware_properties: List[FernAwarePydanticField],
     ) -> LocalClassReference:
-        # TODO: turn this into a FernAwarePydanticModel to be able to handle circular references
```
tjb9dc, line 111: 😅

tjb9dc:
subsumed by https://github.com/fern-api/fern/pull/10071

PR #10038, opened 2025-10-23
fix(python): Fix union type deserialization with literal discriminators

tjb9dc:
## Problem

In python SDKs with "undiscriminated" unions and validations disabled, our union parsing code was still leveraging validation under the hood. For slightly invalid payloads with the correct discriminator (as a literal), this would cause them to wholesale fail validation and fall through to the default case, which would return the first member.

TL;DR: Even with validation disabled, an invalid union member with a valid discriminator would get returned as the first defined member.

## Changes

Changes the union parsing algorithm accordingly:

1. Try to parse into a valid member via pydantic
2. [New] Check if all literals match the payload, and return that member if so
3. Return the first member

Relevant files to review:

- generators/python/core_utilities/shared/unchecked_base_model.py
- seed/python-sdk/unions/no-custom-config/tests/custom/test_deser_fallbacks.py


generators/python/core_utilities/shared/unchecked_base_model.py
```diff
@@ -173,7 +173,42 @@ def _convert_undiscriminated_union_type(union_type: typing.Type[typing.Any], obj
...
+            for field_name, field in fields.items():
+                # Check if this field has a Literal type
+                if IS_PYDANTIC_V2:
+                    field_type = field.annotation
+                else:
+                    field_type = field.outer_type_
+
+                if is_literal_type(field_type):
+                    field_default = _get_field_default(field)
+                    name_or_alias = get_field_to_alias_mapping(inner_type).get(field_name, field_name)
```
tjb9dc, line 192: it could be worth caching this call, and possibly the one above it

tjb9dc: it's probably fine though? I don't expect this to get hit a ton, and especially not for deeply nested unions

aditya-arolkar-swe (approved):

PR #10047, opened 2025-10-23
fix(cli): add x-fern-example requests

kennyderek:
https://linear.app/buildwithfern/issue/FER-7157/x-fern-examples-path-parameter-not-rendering-for-openapi


dsinghvi (approved):
@kennyderek do you ind adding a versions.yml so that a new version of the CLI can be released?

PR #10102, opened 2025-10-27
feat(cli): silence noise + add helpful error for missing references

kennyderek:
tested with Square and SPS.

We don't have a 1000+ line laundry list of warnings now. More helpful errors are surfaced via fern check

image


dsinghvi (approved):
@kennyderek can you include an entry in versions.yml as well so we can release for the CLI

PR #10103, opened 2025-10-27
feat(cli): Add support for ignoring properties in OpenAPI schemas using the `x-fern-ignore` extension.

Swimburger:
## Description
Add support for ignoring properties in OpenAPI schemas using the `x-fern-ignore` extension.

## Testing

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertObject.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertObject.ts
@@ -71,7 +71,7 @@ export function convertObject({
     source: Source;
 }): SchemaWithExample {
     const allRequired = [...(required ?? [])];
-    const propertiesToConvert = { ...getNonIgnoredProperties(properties, context) };
+    const propertiesToConvert = { ...getNonIgnoredProperties({ properties, breadcrumbs, context }) };
     let inlinedParentProperties: ObjectPropertyWithExample[] = [];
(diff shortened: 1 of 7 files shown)


packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertObject.ts
```diff
@@ -388,6 +388,21 @@ export function wrapObject({
...
 
+function getNonIgnoredProperties(
+    properties: Record<string, OpenAPIV3.ReferenceObject | OpenAPIV3.SchemaObject>,
+    context: SchemaParserContext
+): Record<string, OpenAPIV3.ReferenceObject | OpenAPIV3.SchemaObject> {
+    return Object.fromEntries(
+        Object.entries(properties).filter(([_, propertySchema]) => {
+            const shouldIgnore = getExtension<boolean>(propertySchema, FernOpenAPIExtension.IGNORE);
+            if (shouldIgnore) {
+                context.logger.debug(`Ignoring property due to the ${FernOpenAPIExtension.IGNORE} extension.`);
```
tjb9dc: can we log something about the property or the path we're in? Or is that already handled by the contextual logger

tjb9dc (approved):

packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertObject.ts
fern-support: Good catch. fixed.

PR #10132, opened 2025-10-28
chore(deps): upgrade packages, pnpm, move to catalog dep

davidkonigsberg:
## Description
Update dependencies to move away from critical vulnerabilities 

## Changes Made
- Updated internal packages to newer versions
- Moved these into catalog dependencies

## Testing
Tested in CI


davidkonigsberg:
going to go with an override instead.

PR #10148, opened 2025-10-29
fix(java): correctly  sync client delegation causing array query param compilation errors

tstanmay13:
## Description
Linear ticket: Closes [FER-7490: Schematic fix for Java](https://linear.app/buildwithfern/issue/FER-7490/schematic-fix-for-java)

Java SDKs failed to compile when using array query parameters:
```
error: incompatible types: List<String> cannot be converted to String
httpUrl.addQueryParameter("ids", request.getIds().get());
```

The sync client generator was returning the base name directly instead of using it to get the correct client names. This broke the delegation pattern and caused inline HTTP generation to fails with array query parameters. This was seen in schematic here: https://github.com/SchematicHQ/schematic-java/actions/runs/18884400968/job/53895910169?pr=30

The `clientImplName()` parameter was wrongly named rawClientImplName but actually received the base name ( "AccountsClient"). The sync implementation returned it directly which broke 

## Changes Made

- Renamed parameter from `rawClientImplName` to `baseClientName` for clarity
- Sync delegating clients by return base name unchanged ("AccountsClient")
- Sync raw clients by adding "Raw" prefix via rawClientImplName() ("RawAccountsClient")

## Testing

- [x] Manual testing completed - against Schematic def directly


generators/java/sdk/src/main/java/com/fern/java/client/generators/AbstractClientGeneratorUtils.java
```diff
@@ -196,6 +197,9 @@ public Result buildClients() {
 
                 generatedWrappedRequests.addAll(httpEndpointMethodSpecFactory.getGeneratedWrappedRequests());
             }
+        } else {
+            System.out.println("[FERN-DEBUG]   -> Skipping Raw client generation (maybeHttpService is empty)");
+            System.out.println("[FERN-DEBUG]   -> This will likely cause inline HTTP generation or delegation issues!");
```
tjb9dc: what's this? Is this the pattern we are using to debug log in the Java generator? Or leftover code.

If it's intended to stay I think it is kind of unintelligible to the end user

tjb9dc (approved):

generators/java/sdk/src/main/java/com/fern/java/client/generators/AbstractClientGeneratorUtils.java
tstanmay13: Ah forgot to get rid of it haha

PR #10149, opened 2025-10-29
feat(typescript): Expose response on Page

tjb9dc:
This change makes the response field on the Page class public and properly typed, allowing SDK users to access the full API response object while maintaining auto-pagination functionality.

Changes:
- Made Page.response field public (was private)
- Added generic type parameter R to Page class for response typing with `unknown` default for backcompat
- Removed Pageable class (was just a thin wrapper with no added value)
- Updated endpoint generator to construct Page with both type parameters

Generated SDKs will now expose:
- page.data - Array of items (existing)
- page.response - Full typed API response (newly exposed)
- page.rawResponse - Raw HTTP response (existing)


dsinghvi:
@tjb9dc can you make sure that this is documented in the README in the `pagination` section?

generators/typescript/utils/core-utilities/src/core/pagination/Pageable.ts
```diff
@@ -7,12 +7,8 @@ export declare namespace Pageable {
...
+        loadPage: (response: Response) => HttpResponsePromise<Response>;
     }
 }
 
-export class Pageable<R, T> extends Page<T> {
-    constructor(args: Pageable.Args<R, T>) {
-        super(args as any);
-    }
-}
+export class Pageable<R, T> extends Page<T, R> {}
```
tjb9dc, line 14: This was getting linted away in the generated code so I removed it

Swimburger (approved):

PR #10155, opened 2025-10-29
fix(go): expose raw paginated response on core.Page

musicpulpite:
## Summary
Customers want to be able to access the raw cursor property from paginated response, not just use the `GetNextPage` helper function.

## Testing
I ran the `pagination` seed case, are there any other seed cases that might be useful to prove this functionality out?

### DIFF
--- a/generators/go-v2/sdk/src/endpoint/utils/getPaginationInfo.ts
+++ b/generators/go-v2/sdk/src/endpoint/utils/getPaginationInfo.ts
@@ -530,23 +530,13 @@ function getPagePropertyInitializer({
                     );
                     writer.writeLine(" {");
                     writer.indent();
-                    writer.writeLine("var err error");
-                    writer.write("if next, err = ");
                     writer.writeNode(
-                        go.invokeFunc({
-                            func: go.typeReference({
-                                name: "Atoi",
-                                importPath: "strconv"
-                            }),
-                            arguments_: [getQueryParameter({ key: pagination.page.property.name.wireValue })],
-                            multiline: false
+                        getPageValueParser({
+                            pageType,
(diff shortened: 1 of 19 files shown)


generators/go-v2/sdk/src/endpoint/utils/getPaginationInfo.ts
```diff
@@ -530,23 +530,13 @@ function getPagePropertyInitializer({
...
-                    writer.write("if next, err = ");
                     writer.writeNode(
-                        go.invokeFunc({
-                            func: go.typeReference({
-                                name: "Atoi",
-                                importPath: "strconv"
-                            }),
-                            arguments_: [getQueryParameter({ key: pagination.page.property.name.wireValue })],
-                            multiline: false
+                        getPageValueParser({
```
musicpulpite, line 534: This change is not directly related to the purpose of this ticket (to expose the raw cursor) I just noticed while testing seed that the functionality to use the `page` query param as the first cursor did not work for the cursor type of this example... so added more cases

seed/go-sdk/pagination/users/client.go
```diff
@@ -271,15 +271,15 @@ func (c *Client) ListWithDoubleOffsetPagination(
 	var next float64 = 1
 	if queryParams.Has("page") {
 		var err error
-		if next, err = strconv.Atoi(queryParams.Get("page")); err != nil {
+		if next, err = strconv.ParseFloat(queryParams.Get("page"), 64); err != nil {
```
musicpulpite, line 274: The unrelated changes were to ensure this conversion respects all possible cursor types

dsinghvi (approved):

PR #10223, opened 2025-10-31
fix(cli): fix fern definition creating duplicate types in 2 files

aditya-arolkar-swe:
## Description
Linear ticket: Closes 
If we had a nested referenced schema in the OpenAPI spec, we would try to write it in both the __package__.yml and the namespace specific .yml file. 
## Changes Made

- Ensures we don't write it twice

## Testing


dsinghvi (approved):

packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildTypeReference.ts
```diff
@@ -135,6 +139,7 @@ export function buildTypeReference({
         case "oneOf":
             return buildOneOfTypeReference({
                 schema: schema.value,
+                fullSchema: schema,
```
tjb9dc, line 140: for refactors like this, I think we should instead make the param named `schema` and then get rid of the redundant param, just having the function access `schema.value`

Will also let you achieve the TypeScript nirvana of just providing `{schema, ...}`

tjb9dc (approved):

packages/cli/api-importers/openapi/openapi-ir-to-fern/src/buildTypeReference.ts
aditya-arolkar-swe: There was a bit of type checking annoyance with moving it into the function. Feel like i've been spending too much time on this fix already, think it's ok to leave as is and move onto other higher prio tickets - unless you feel super strongly otherwise.

PR #10234, opened 2025-10-31
fix(csharp): Significant name/namespace fixes for c# generation (targeting allowed failures)

fearthecowboy:
_NOTE: There are 11 commits categorized to make reviewing easier. Well, easier than it would be as 2000 files changed. Sorry._

# Issues Fixed correcting `allowedFailures` error resolution:
- FormRequest.template.cs - removed useless check on AdditionalBodyProperies
- Pagination fixture now generates code that can actually compile
  - generates many more files now because some files were getting clobbered by others with the same name 
    (name registry now moves colliding classes when they are not the same)
- Dynamic Snippets were previously generating incorrect examples for endpoints where there were examples 
  with parameters with the same names (or subsets of them). This resulted in a number of changed dynamic snippet
  files that are now generating the appropriate examples.
- added .csproj implicit namespace declarations for .NET 462/netstandard2.0 to match what the compiler automatically 
  uses for newer .NET frameworks. 
- reduced the namespace clutter in snippets (json/markdown files)  
- Now that implicit namespaces are accurately tracked, implicit namespaces are automatically trimmed from .cs files when
  possible
- Namespace collision detection (vs class names and ambient namespaces) is vastly improved and permits accurate scope reductions 
  in snippets and dynamic examples

  > Before
  ```csharp
    Test = new SeedExamples.Test(
        new SeedExamples.Test.And()
    ),
  ```
  > After
  ```csharp
    Test = new Test(
        new Test.And()
    ),
  ```

- Member Names that collide with any keywords, built-in c# members, or other properties are automatically 
  renamed to avoid collsions. This is universal for properties now. And this removes the one-off
  mechanisms that were used in the past.

  > Before
  ```csharp
    public static readonly SingleFilterSearchRequestOperator Equals = new(Values.Equals);
  ```
  > After
  ```csharp
    public static readonly SingleFilterSearchRequestOperator Equals_ = new(Values.Equals_);
  ```

- Slight order change in generating hand-named members - the names of generated members must be done first
  so that hand crafted names are not clobbering the generated names

- handling of `Task<T>` vs client type `Task` declarations are consistent with implicit namespaces and require no special handling

  > Before
  ```csharp
    public async global::System.Threading.Tasks.Task<Task> CreateTaskAsync(
        Task request,
        RequestOptions? options = null,
    )
  ```
  > After - (and to be honest, I'm a little amazed that this works and compiles perfectly.)
  ```csharp
    public async Task<Task> CreateTaskAsync(
        Task request,
        RequestOptions? options = null,
    )
  ```

# Changes:

## "Origin" tracking for `AstNode`s
The ultimate goal with Origin tracking is to ensure that when a node is constructed and decisions are made about the 
name/namespace/etc that 
 - These decisions are made only once *EVER*
 - Other generation code can ask for the name of soemthing by referencing it's origin. This ensures that the same name is used everywhere. 
 - The `NameRegistry` can ensure that name collisions are resolved and renamed appropriately. Cases where it can identify these:
   - keyword avoidance
   - C# class member avoidance (`Equals`)
   - C# Type name avoidance (types can not have members with the same name as the type)
   - duplicate identifer avoidance (can't generate a method and field with the same name)
   - explicit member names that collide with IR-derrived names (ie `Value`, `Index`)

### `Provenance` 
In order to uniquely identify the "source" (what implies it's existence) of a node, we now create a `Provenance` for nodes during creation.
The `Provenance` is tied to either a node in the IR/DynamicIR or an explicitly named element relative to another `Provenance` (ultimately - EVERYTHING ties to some IrNode somewhere)
After that, when any other code attempts to resolve the name of something with `Provenance` we can identify the actual resolved name.
Ultimately, Provenance boils down to a JsonPath to the originating node in the IR/DynamicIR.

The `Args` for `AstNode`s that are named now support an `origin` property which is used to associate the orignating node with the `AstNode` 
And if the name is not present, will derive the desired name from the originating node. 

`AstNode`s that are created with explicitly provided names (as opposed to the name being based on the originating node's name) are still 
associated with an originating node by using the `model.explict(...)` and `model.explicitStatic(...)` methods to provide a `Provenance` 
that notes that the name was explicitly provided. ('static' really means project-wide and is essentially associated to the root of the IR).

### `ModelNavigator` 
The `ModelNavigator` indexes the IR at initialization to enable Node to Provenance tracking
- there are convenience methods to normalize access to `Provenance` information so that it is easier to navigate the IR Model 
- The `ModelNavigator` transparently handles both IR and DynamicIR variants

## `AbstractWriter` and `StackTrace` support.
Improved support for `FERN_STACK_TRACK` and provided a facility to allow language-specific handling for crafting the generated stack comments
- `FERN_STACK_TRACK`  can be declared as `single` `multiline` or `box`  (any non-false value will assume `single`) to change the style of comments generated
- the stack now contains all the stack locations where AstNode object were created since the last comment emitted.
- the non-relevant stack frames are filtered out by default
- `AbstractWriter` stores the buffer internally as an array of strings 

## Generation Info
Created `Generation` information class that provides consistent across-the-board access to csharp,types,externs,model,namespaces,names, constants and settings 
  - removes duplicate interpretations of settings 
  - consolidates declarations of all external types into a singular source of truth
  - a singular reference point for all namespaces,names,paths,etc
  - deduplicated all the constants sprinkled around the project
  - convenience members for generation-info added in strategic locations  
  - concentrated all aribtrary constants into a single location 

### `Lazy` initializer  
The `Lazy` initializer takes a set of factory functions and returns an object that provides lazily initialized properties
Transforms an object of factory functions into an object where:
  - Zero-parameter functions become lazy-evaluated, cached properties (computed once on first access)
  - Functions with parameters remain as regular methods (called every time)

This ensures that single-result computational values are only ever evaluated once.  

## dead code removal / code quality improvements
- removed `getPublicCoreTestAsIsFiles()` - was completely empty set of files in all cases.
- removed superflous/duplicate parmeters that are already being passed around (`rootNamespace` - access thru `<this>.namespaces.root` )
- removed unused class `PrebuiltUtilities` - zero usage 
- removed usage of `lodash-es` (was only using 2 very trivial functions)
- `tsconfig.json` add `noImplicitOverride` to ensure that overrides are explicitly managed. 
  - ensure that overriden members are consistent (many child classes were creating/setting additional properties for properties declared in the parent)
  - ensures that overriden methods are easy to identify
- heavily enforce access to AstNode construction via the `CSharp` factory methods.
- project now uses template literal strings everywhere for consistency.

# Testing
- [X] Seed fixtures generated (alot)


generators/csharp/codegen/src/ast/types/Field.ts
```diff
@@ -15,9 +17,9 @@ export declare namespace Field {
         init?: (writer: Writer) => void;
     };
 
-    interface Args {
+    interface Args extends MemberNode.Args {
         /* The name of the field */
-        name: string;
+        name?: string;
```
Swimburger, line 22: Is it necessary for this to be optional? Why would a field not have a name?

generators/csharp/codegen/src/context/generation-info.ts
```diff
@@ -0,0 +1,876 @@
...
+            | IntermediateRepresentation
+            | FernIr.dynamic.DynamicIntermediateRepresentation,
+        private readonly apiName: string,
+        private readonly customConfig: BaseCsharpCustomConfigSchema,
+        private readonly generatorConfig: FernGeneratorExec.config.GeneratorConfig,
+        private readonly makeRelativeFilePath: (path: string) => TRelativeFilePath = (path) =>
+            path as TRelativeFilePath,
+        private readonly makeAbsoluteFilePath: (path: string) => TAbsoluteFilePath = (path) =>
+            path as TAbsoluteFilePath,
+        public readonly logger: {
```
Swimburger, line 72: There should be a common type for this somewhere that you can reuse.
It looks a little wild to have this type inline here.

generators/csharp/codegen/src/utils/ir-type-guards.ts
Swimburger, line 1: Is this generated somehow?

generators/csharp/codegen/src/utils/dynamic-ir-type-guards.ts
Swimburger, line 1: Is this generated somehow?

generators/csharp/model/src/enum/StringEnumGenerator.ts
```diff
@@ -1,29 +1,22 @@
...
-
+import { EnumTypeDeclaration, TypeDeclaration } from "@fern-fern/ir-sdk/api";
 import { ModelCustomConfigSchema } from "../ModelCustomConfig";
 import { ModelGeneratorContext } from "../ModelGeneratorContext";
 
 export class StringEnumGenerator extends FileGenerator<CSharpFile, ModelCustomConfigSchema, ModelGeneratorContext> {
     private readonly classReference: ast.ClassReference;
     private readonly customMethodName: string;
-    private valuePropertyName: string = "Value";
+    // private valuePropertyName: string = "Value";
```
Swimburger, line 11: Let's delete this

generators/csharp/codegen/src/proto/ProtobufResolver.ts
```diff
@@ -48,24 +44,23 @@ export class ProtobufResolver {
...
                         .split(".")
-                        .filter((segment) => !rootNamespace.split(".").includes(segment))
+                        .filter((segment) => !this.namespaces.root.split(".").includes(segment))
                         .join("_")
                 );
                 return this.csharp.classReference({
                     name: protobufType.name.originalName,
                     namespace: this.context.protobufResolver.getNamespaceFromProtobufFileOrThrow(protobufType.file),
-                    namespaceAlias: `Proto${aliasSuffix.charAt(0).toUpperCase() + aliasSuffix.slice(1)}`
+                    namespaceAlias: `Proto${aliasSuffix.charAt(0).toUpperCase()}${aliasSuffix.slice(1)}`
```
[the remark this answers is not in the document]
fearthecowboy: It is coming in the future

generators/csharp/codegen/src/csharp.ts
```diff
@@ -117,41 +125,48 @@ export class CSharp {
...
+            | ClassRefWithOriginAndNamespace
+            | ClassRefWithOriginAndEnclosingType
+    ): ClassReference {
+        return this.classReferenceInternal(argz as ClassReference.Args);
+    }
+
+    public classReferenceInternal(args: ClassReference.Args): ClassReference {
+        // if we're given an origin, we need to turn that into a name
+        if (!args.name && "origin" in args && args.origin) {
+            args.name = this.model.getClassNameFor(args.origin);
```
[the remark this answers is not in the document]
fearthecowboy: No.

Swimburger (approved):

generators/csharp/codegen/src/ast/types/Field.ts
fearthecowboy: When `name` is not present, then the system derives the name from the `origin` node - which is what we want all the time.

In the not too distant future this might actually get removed from the exposed interface completely  - hand-named things require the use of `origin: class.explicit(` anyways.

generators/csharp/codegen/src/context/generation-info.ts
fearthecowboy: Yeaaaah. This was an odd edge case - because the DSG doesn't have any actual filesystem dependency, code shared between the two can't actually depend on the types that we have for that (because they're tied to nodejs' `fs` and `path`  module.

I think  I can refactor this out a bit more in the not-too-distant future, and/or push an interface up into the base project so that at least there isn't parallel types for this.

generators/csharp/codegen/src/utils/ir-type-guards.ts
fearthecowboy: Yeees. I nudged the AI (Claude, or Cursor) enough to have it generate this cleanly, and then told it to put in sufficient comments in the code so that when we need to do it again, it will be able to not break anything.

The dynamic version of the type guards for this was the 'replicate this functionality on a different set of types)  and it did precisely what I was interested in.

generators/csharp/codegen/src/utils/dynamic-ir-type-guards.ts
fearthecowboy: yes (see above)

generators/csharp/model/src/enum/StringEnumGenerator.ts
fearthecowboy: I'm sad now. I was soooooo sure I got them all. Even the AI betrayed me (I asked it to find any commented out code that I left behind)  😭

fearthecowboy: I'll get it cleaned up on my next PR which isn't far off now.

PR #10306, opened 2025-11-04
fix(docs): do background validation for docs dev

coltondotio:
## Description

no longer block front-end updates on docs validation during `fern docs dev`

## Testing

tested locally


dsinghvi (approved):
lets add a `versions.yml` for releasing?

PR #10367, opened 2025-11-05
feat(python): hard-set retries disabled config at `raw_client.py` level

aditya-arolkar-swe:
## Description
Closes FER-6558

Integrates `x-fern-retries` extension into the generated Python SDK. 

I.e. specify in your OpenAPI spec:
```
paths:
  /imdb/movies/:
    get:
      x-fern-retries:
        disabled: true
        ...
```
 
 and the generated `raw_client.py` will enforce no retries:
 
```
class RawClient:
    ...

    def get_movie(
        ...
    ) -> HttpResponse[Movie]:
        _request_options_with_retries_disabled = (
            # even if end-user specifies max_retries, will override to 0
            {**request_options, "max_retries": 0} if request_options is not None else {"max_retries": 0}
        )
        _response = self._client_wrapper.httpx_client.request(
            ...
            json={
                ...
                request_options=_request_options_with_retries_disabled,
                ...
            }
            ...
        )
```

## Changes Made

- updated `endpoint_function_generator.py` to pass in specified request_options
- kept `http_client.py` implementation the same

## Testing


generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
```diff
@@ -608,15 +608,32 @@ def write(writer: AST.NodeWriter) -> None:
...
                         )
 
+            # Get the request_options variable name from the last parameter if it exists
+            request_options_variable_name = EndpointFunctionGenerator.REQUEST_OPTIONS_VARIABLE
+            if named_parameters and len(named_parameters) > 0:
+                last_param = named_parameters[-1]
+                request_options_variable_name = last_param.name
+
+            # If retries are disabled, enforce max_retries=0 by overriding request_options
+            if endpoint.retries is not None and hasattr(endpoint.retries, "disabled") and endpoint.retries.disabled:
```
tjb9dc: does the type narrowing work for `hasattr(endpoint.retries, "disabled")` ? I tend to prefer an `isinstance` check so that type checking works better afterwards. Would be more relevant if we were doing more interesting stuff than just the boolean disabled

generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
```diff
@@ -608,15 +608,32 @@ def write(writer: AST.NodeWriter) -> None:
...
+                overridden_request_options_var = "_request_options_with_retries_disabled"
+
+                def write_override_logic(writer: AST.NodeWriter) -> None:
+                    writer.write(f'({{**{request_options_variable_name}, "max_retries": 0}} ')
```
tjb9dc: wow the kwargs over a dict union, interesting choice 😅 

jk I'm not sure which is more idiomatic

tjb9dc: like `(request_options_variable_name or {}) | {"max_retries": 0}`

tjb9dc (approved):

generators/python/src/fern_python/generators/sdk/client_generator/endpoint_function_generator.py
aditya-arolkar-swe: spoke offline: keeping current change for easier readability

PR #10487, opened 2025-11-11
fix(internal): use numerical suffixes universally to avoid buggy self-references

patrickthornton:
## Description
Linear ticket: Closes FER-7672. [Fixes Vapi.]
When generating names for the subtypes of an undiscriminated `oneOf` in the spec-to-OpenAPI-IR parser, it adds numerical suffixes (One, Two, etc.; called prefixes in the code, but they're definitely suffixes [I renamed them]) to avoid name conflicts. However, there was logic to try and skip this step whenever the type was primitive and therefore would never be made into a standalone type, as it would be unnecessary in that case. The logic was incomplete, though; it deemed a two-valued enum to be a primitive string and therefore caused a name collision. 

I think the best approach here is just to ditch the extra logic and slap on the numerical suffixes in every case, as all it'll do is add a few characters to the IR in exchange for a virtual guarantee that we won't hit a name conflict like we did here in any possible case.

## Changes Made
Changes to `convertUndiscriminatedOneOf.ts` in the openapi-ir-parser.

## Testing
Added a test fixture called `vapi-enum` that minimally reproduces the error we saw with Vapi; in previous versions of Fern, a Fern Definition generated using that fixture will have a type that immediately references itself:
```
types:
  Thing:
    properties:
      problem: optional<ThingProblem>
    source:
      openapi: openapi.json
  ThingProblem:
    discriminated: false
    inline: true
    source:
      openapi: openapi.json
    union:
      - ThingProblem
      - string
```
Now, this will look like:
```
types:
  Thing:
    properties:
      problem: optional<ThingProblem>
    source:
      openapi: openapi.json
  ThingProblem:
    discriminated: false
    inline: true
    source:
      openapi: openapi.json
    union:
      - ThingProblemZero
      - string
  ThingProblemZero:
    enum:
      - foo
      - bar
    inline: true
    source:
      openapi: openapi.json
```
Updating seed!

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v2/parseAsyncAPIV2.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/asyncapi/v2/parseAsyncAPIV2.ts
@@ -17,7 +17,7 @@ import { FernOpenAPIExtension } from "../../openapi/v3/extensions/fernExtensions
 import { ParseOpenAPIOptions } from "../../options";
 import { convertAvailability } from "../../schema/convertAvailability";
 import { convertReferenceObject, convertSchema } from "../../schema/convertSchemas";
-import { convertUndiscriminatedOneOf, UndiscriminatedOneOfPrefix } from "../../schema/convertUndiscriminatedOneOf";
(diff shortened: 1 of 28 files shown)


tjb9dc (approved):
don't forget that versions.yml!

PR #10491, opened 2025-11-11
fix(java): handle nullable and literal container types in SDK generation

tstanmay13:
## Description
Linear ticket: Closes [FER-7674: \[Java, Speechify\] Unexpected container type](https://linear.app/buildwithfern/issue/FER-7674/java-speechify-unexpected-container-type)

Speechify was getting `Unexpected container type` errors for APIs with nullable types. added some literal container detection as well. 

## Changes Made

  - Add support for nullable containers in response parsing and pagination
  - Add literal container detection, while literals shouldn't appear in response parsing, this ensures they're handled gracefully rather than falling through to a generic error.
  - Refactored some collection type handling to be more type safe and readble.instead of using a string variable to build method names dynamically, each collection type now has its explicit handler
  - Add defensive null checking to prevent NPEs

## Testing

- [x] Manual testing completed - speechify can generate now

### DIFF
--- a/generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/AbstractHttpResponseParserGenerator.java
+++ b/generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/AbstractHttpResponseParserGenerator.java
@@ -47,6 +47,16 @@ public abstract class AbstractHttpResponseParserGenerator {
     private static final String INTEGER_ONE = "1";
     private static final String DECIMAL_ONE = "1.0";
 
+    /** Helper method to generate diagnostic string for container types. Useful for debugging and error messages. */
+    private static String getContainerDiagnosticString(com.fern.ir.model.types.ContainerType container) {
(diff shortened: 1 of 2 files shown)


generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/AbstractHttpResponseParserGenerator.java
```diff
@@ -47,6 +47,16 @@ public abstract class AbstractHttpResponseParserGenerator {
     private static final String INTEGER_ONE = "1";
     private static final String DECIMAL_ONE = "1.0";
 
+    /** Helper method to generate diagnostic string for container types. Useful for debugging and error messages. */
+    private static String getContainerDiagnosticString(com.fern.ir.model.types.ContainerType container) {
```
tstanmay13, line 51: Thought we could have a better error message but if you feel that we don't need to provide this to the customer I get that too.

tjb9dc (approved):

PR #10506, opened 2025-11-12
fix(cli, go): Fix go local github generation to match remote generation

jsklan:
## Description
This PR fixes a mismatch between local and remote Go SDK generation by ensuring the CLI can properly retrieve the package name from `module.path` configuration during local generation workflows. Previously, local generation was failing because the package name couldn't be extracted from the generator configuration, while remote generation (via Fiddle) worked correctly.

## Changes Made
- **CLI package name extraction**: Added support for reading `module.path` from Go generator config in local workspace runner
- **Version normalization**: Go generator now adds 'v' prefix to version strings following Go conventions
- **Custom license file support**: Go v2 generator now copies custom LICENSE files during local generation (remote generation handled by Fiddle)

## Testing
- [x] Manual testing [completed](https://github.com/fern-api/lattice-sdk-go/compare/fern-bot/2025-11-12T01-52Z..fern-bot/2025-11-12_04-23-12) with local generation
- [x] Verified version normalization with/without 'v' prefix
- [x] Tested custom license file copying in local generation environment


packages/cli/generation/local-generation/local-workspace-runner/src/runLocalGenerationForWorkspace.ts
```diff
@@ -206,8 +206,13 @@ function getPackageNameFromGeneratorConfig(generatorInvocation: GeneratorInvocat
         if (packageName != null) {
             return packageName;
         }
-    }
 
+        // go-sdk generator uses module.path to set the package name
```
musicpulpite, line 210: Great catch here

generators/go/internal/cmd/cmd.go
```diff
@@ -109,6 +109,12 @@ func run(fn GeneratorFunc) (retErr error) {
 	if err != nil {
 		return err
 	}
+
+	// Normalize the version by adding 'v' prefix if missing (Go convention)
```
musicpulpite, line 113: I'm a little concerned about this one here. We already have code in the v1 generator for adding v<major-version-number> to the module path here: https://github.com/fern-api/fern/blob/f26e1712b37802cf9587b8d6fcdab87857a3475f/generators/go/internal/cmd/cmd.go#L146 so just want to be sure that it doesn't interfere with existing logic? E.g. does this still behave well for sdk versions >= 2.0.0?? But overall I trust your judgement enough to approve now!

musicpulpite (approved):

generators/go/internal/cmd/cmd.go
jsklan: Honestly I wasn't sure about this. Spent a few hours trying to figure out why the v1 generator wasn't resolving the module paths with the v4 in them when I passed --version 4.0.0 when running generate. So if you know a better way to factor this, I would defer to you.

musicpulpite: Gotcchhhhaaa. So after you added this you saw the path correctly templated as `github.com/<org>/<repo>/v4`?? Then I guess this is the way, there may be another place in the logic where it falls through the cracks without it.

PR #10521, opened 2025-11-12
feat(internal): global `api.settings` defaults in `generators.yml`

patrickthornton:
## Description
Adds a global settings flag to `generators.yml`:
```
api:
  settings:
    coerce-enums-to-literals: false
  specs:
    - openapi: ./openapi.json
```
The precedence hierarchy for this is that generator-level settings (`groups.generators[].api.settings`) beats spec-level settings (`api.specs[].settings`), which in turn beats global settings (`api.settings`). The lattermost of these I feel is semantically necessary for config flags for the IR generator, which technically shouldn't be bound to any particular spec (as they can be run from OpenAPI specs or from native Fern Definitions) or generator (as they should be triggered when running `fern ir`, for instance).

## Testing
Added new tests to `convertGeneratorsConfiguration.test.ts` to test precedence.


fern/apis/generators-yml/definition/generators.yml
```diff
@@ -19,6 +19,13 @@ types:
         type: optional<OutputMetadataSchema>
       readme:
         type: optional<ReadmeSchema>
+      settings:
+        type: optional<BaseAPISettingsSchema>
```
Swimburger: Should this be `BaseAPISettingsSchema`?
I would expect this to be an `allOf` of the Async, OpenAPI, Conjure, and proto settings

patrickthornton: ah, yep

Swimburger (approved):

PR #10592, opened 2025-11-14
feat(cli): Add sdk-diff command

tjb9dc:
Adds `sdk-diff` command that will use the `ai` config block in `generators.yml` to analyze a diff.


fern/apis/generators-yml/definition/ai.yml
```diff
@@ -0,0 +1,22 @@
...
+  AIServiceBaseSchema:
+    properties:
+      provider: ModelProvider
+      model: ModelName
+
+  ModelProvider:
+    enum:
+      - openai
+      - anthropic
+      - bedrock
```
tjb9dc, line 11: we shooould support all of https://docs.boundaryml.com/ref/baml/client-llm#fields

But I think start with the enum and we can relax later

fern-support (approved):

PR #10595, opened 2025-11-16
fix(python): Discriminated unions use Python field names for Pydantic v2

tjb9dc:
Fix Field(discriminator=...) and UnionMetadata(discriminant=...) to use
Python field names instead of JSON aliases for Pydantic v2 compatibility.

In Pydantic v2, discriminated unions require the discriminator parameter
to reference the Python attribute name, not the serialization alias. This
was causing PydanticUserError when discriminator fields had aliases.

Changes:
- pydantic_model_simple_discriminated_union_generator.py: Use
  get_discriminant_parameter_name() for both Field(discriminator) and
  UnionMetadata(discriminant)
- discriminated_union_with_utils_generator.py: Use
  _get_discriminant_attr_name() for UnionMetadata(discriminant)


dsinghvi (approved):

generators/python/src/fern_python/generators/pydantic_model/type_declaration_handler/discriminated_union/discriminated_union_with_utils_generator.py
```diff
@@ -61,7 +61,7 @@ def _add_conditional_base_methods(
                         kwargs=[
                             (
                                 "discriminant",
-                                AST.Expression(f'"{self._union.discriminant.wire_value}"'),
+                                AST.Expression(f'"{self._get_discriminant_attr_name()}"'),
```
tjb9dc, line 64: hmm surprised this doesn't use `get_discriminant_parameter_name`

tjb9dc: or, rather, I think we probably need to update the impl:

```
    def _get_discriminant_attr_name(self) -> str:
        return self._union.discriminant.name.snake_case.safe_name
```

PR #10629, opened 2025-11-17
feat(go): get raw HTTP response from paginated responses

patrickthornton:
## Description
Adds `.HTTPRawResponse` to a page, which allows a user to access the raw headers and status code associated with a paginated endpoint's response. Like in TS, this is always included in the page; you don't need to opt into it via `withRawResponse()`.

## Changes Made
Small changes to as-is files in the go-v2 generator.

## Testing
Regenerating seed now!


fern-support (approved):
🥂 🥂

PR #10836, opened 2025-11-25
fix(java): preserve parameter order for staged builders in dynamic snippets

tstanmay13:
## Description

Refs: Requested by @tstanmay13

Link to Devin run: https://app.devin.ai/sessions/a5d200d8904e4958b37da0e3d584d474

Java v2 dynamic snippet generator was producing builder method calls in incorrect order for staged builders, causing compilation failures. The root cause was that `associateByWireValue()` in the shared base code iterates over example values (arbitrary order) instead of schema parameters (definition order). This breaks Java's staged builders which require compile-time type-state ordering.

## Changes Made

- Added `associateByWireValue` override in `DynamicSnippetsGeneratorContext.ts` that calls the base implementation to preserve all error handling semantics, then reorders results to match schema parameter order
- Added new test fixture `java-staged-builder-ordering` with 6 types and 11 examples demonstrating various wrong orderings (reverse, random, nested)
- Updated seed files for the new fixture and existing `examples` fixture
- Added snapshots for ir-to-jsonschema and Swift SDK tests to account for the new fixture
- Added `java-staged-builder-ordering` to allowed failures for PHP and Ruby generators (since this is a Java-specific test fixture)
- Added `versions.yml` entry for version 3.18.11

## Testing

- [x] Java SDK `examples` fixture passes (4/4 tests)
- [x] Java SDK `java-staged-builder-ordering` fixture passes (1/1 tests)
- [x] Python SDK `examples` fixture passes (confirms no cross-generator impact)
- [x] Lint checks pass
- [x] Verified generated snippets use correct schema order regardless of example JSON order
- [x] All 29 dynamic snippets unit tests pass locally

## Human Review Checklist

- [ ] Verify the super + reorder approach in `associateByWireValue` correctly preserves base error handling (e.g., `ignoreMissingParameters` behavior for read-only fields)
- [ ] Check if changes to `Metadata.html("metadata")`, `Data.string("data")`, `Test.and(true)` in `examples` fixture are expected (these appear to have gained arguments)
- [ ] Confirm ordering changes in snapshots (e.g., `name` before `id`, `metadata` before `cast`) match the schema definition order


tstanmay13:
add to versions.yml for Java as a fix keep it short and we should be good here.

tjb9dc (approved):

PR #10883, opened 2025-11-26
fix(cli): correct property ordering in dynamic snippets for Java staged builders

tstanmay13:
## Description
Linear ticket: Closes [FER-7835: \[Java, Payroc\] one last payroc fix](https://linear.app/buildwithfern/issue/FER-7835/java-payroc-one-last-payroc-fix)

Fixes Java SDK dynamic snippet generation for types that extend other types through interface inheritance. The issue was that java sdk gen generated staged builders where type implementing self-interfaces have their own properties ordered first and hten into inherited props. The SDG was producing builders in the wrong order because the dynamic IR flattenes inherited props. However, the types that are extended by others like self-interfaces, it eneds up being opposite of what is needed. 

### Changes
Added logic in `DynamicSnippetsConverter.ts` to:
  1. Detect which types are extended by other types (by checking if one type's properties are a prefix of another's)
  2. For extended types, reorder properties so own properties come before inherited properties

This aligns the dynamic snippet property order with Java v1's staged builder expectations.

## Testing

- [x] Unit tests added/updated -- no regression in snapshots
- [x] Manual testing completed -- testing against payroc no failing snippets


packages/cli/generation/ir-generator/src/dynamic-snippets/DynamicSnippetsConverter.ts
```diff
@@ -480,12 +504,20 @@ export class DynamicSnippetsConverter {
...
-        object
+        object,
+        typeId
     }: {
         declaration: DynamicSnippets.Declaration;
         object: ObjectTypeDeclaration;
+        typeId: TypeId;
     }): DynamicSnippets.NamedType {
-        const properties = [...(object.extendedProperties ?? []), ...object.properties];
+        // If this type is extended by other types, it has a self-interface in Java
```
tjb9dc: nit: This isn't Java-specific right? In this package at least

tjb9dc (approved):

packages/cli/generation/ir-generator/src/dynamic-snippets/DynamicSnippetsConverter.ts
tstanmay13: true updating the comment now

PR #10889, opened 2025-11-26
fix(python): match request retry strategy of TS, Go, Java, etc.

patrickthornton:
## Description
As described: changes constants and adds `X-RateLimit-Reset` check to match https://github.com/fern-api/fern/blob/main/generators/typescript/utils/core-utilities/src/core/fetcher/requestWithRetries.ts.

## Testing
Regenerating seed now!


seed/pydantic/simple-fhir/no-inheritance-for-extended-models/src/seed/api/account.py
```diff
@@ -31,4 +31,4 @@ class Config:
 from .script import Script  # noqa: E402, I001
 from .resource_list import ResourceList  # noqa: E402, I001
 
-update_forward_refs(Account, Patient=Patient, Practitioner=Practitioner, Script=Script)
```
tjb9dc, line 34: aside, looks like we need to sort these forward refs or store them in an OrderedSet as we accrue them

tjb9dc (approved):

seed/pydantic/simple-fhir/no-inheritance-for-extended-models/src/seed/api/account.py
tjb9dc: fixed in separate PR

PR #11064, opened 2025-12-05
fix(typescript): Update WebSocket allow-multiple query params to include undefined in type

tjb9dc:
## Description

Refs: Follow-up to #10966

Updates WebSocket `ConnectArgs` interface to conditionally include `undefined` in the union type for `allow-multiple: true` query parameters based on whether the parameter is optional:
- Non-optional, allow-multiple: `foo: T | T[]`
- Optional, allow-multiple: `foo?: T | T[] | undefined`

**Link to Devin run:** https://app.devin.ai/sessions/3df4d8e6eeef4b0bb25291da9228d746
**Requested by:** thomas@buildwithfern.com (@tjb9dc)

## Changes Made

- Updated `GeneratedDefaultWebsocketImplementation.ts` to conditionally add `undefined` to the union type for allow-multiple query parameters only when `context.type.isOptional(queryParameter.valueType)` is true
- Added version 3.35.10 changelog entry

Based on reviewer feedback, changed from always adding `undefined` to only adding it when the query parameter's value type is optional.

## Testing

## Human Review Checklist

- [ ] Verify the conditional behavior (only adding `undefined` when optional) is the intended behavior
- [ ] Note: The changelog entry in `versions.yml` says `T | T[] | undefined` unconditionally - may want to update it to reflect the conditional behavior
- [ ] Confirm this differs from `GeneratedRequestWrapperImpl.ts` (lines 222-227) which uses `T | T[]` without explicit `undefined` - is this intentional divergence?


generators/typescript/sdk/client-class-generator/src/websocket/GeneratedDefaultWebsocketImplementation.ts
```diff
@@ -176,7 +176,8 @@ export class GeneratedDefaultWebsocketImplementation implements GeneratedWebsock
                     const typeNode = queryParameter.allowMultiple
                         ? ts.factory.createUnionTypeNode([
                               type.typeNodeWithoutUndefined,
-                              ts.factory.createArrayTypeNode(type.typeNodeWithoutUndefined)
+                              ts.factory.createArrayTypeNode(type.typeNodeWithoutUndefined),
+                              ts.factory.createKeywordTypeNode(ts.SyntaxKind.UndefinedKeyword)
```
tjb9dc: um I guess this works? Should we instead be keying off of whether this allows undefined?

fern-support (approved):

PR #11072, opened 2025-12-06
fix(java): Fix duplicate case label errors when multiple errors map to same status code

tstanmay13:
## Description
Refs: Slack thread from @tstanmay13

Fixes Java SDK compilation errors caused by duplicate `case` labels in switch statements when multiple error types map to the same HTTP status code (e.g., both `FooTooLittle` and `InternalServerError` mapping to status code 500).

## Changes Made
- Added deduplication logic in `AbstractHttpResponseParserGenerator.addMappedFailuresCodeBlock()` to keep only the first error declaration for each status code
- Uses `LinkedHashMap` to preserve order while deduplicating
- Updated Java SDK versions.yml with version 3.23.1 changelog entry
- Updated seed files for `errors` fixture (removed duplicate `case 500:` labels)
- Removed `errors` from `allowedFailures` in `seed/java-sdk/seed.yml`

## Behavior
When multiple errors share the same status code, the first error in the list is used and subsequent ones are dropped. This matches the effective behavior in other language generators:
- **Python**: Uses `if` statements, so first match wins (subsequent checks are unreachable after throw)
- **TypeScript**: Also uses switch with duplicate cases (second case is unreachable)
- **Go**: Uses map with duplicate keys (last wins, but effectively similar)

## Testing
- [x] Local seed test passed: `pnpm seed:local test --generator java-sdk --fixture errors`
- [x] The `errors` fixture now compiles successfully
- [x] Removed `errors` from `allowedFailures` in seed.yml
- [x] CI passed

## Human Review Checklist
- [ ] Verify the "first wins" deduplication behavior is acceptable (endpoint-specific errors like `FooTooLittle` take precedence over global errors like `InternalServerError`)
- [ ] Note: No warning is currently emitted when duplicate status codes are detected - this can be addressed in a follow-up PR at the CLI level

---
Link to Devin run: https://app.devin.ai/sessions/1512057fccb54cc186e68f48ae9f367d
Requested by: tanmay.singh@buildwithfern.com (@tstanmay13)


fern-support (approved):

tstanmay13:
generators/java/generator-utils/src/main/java/com/fern/java/GeneratorLogging.java are the changes here neceessary I don't see it being used anywhere

PR #11076, opened 2025-12-06
fix(java): fix dynamic snippets to correctly reference inline types as nested classes

tstanmay13:
## Description
Refs: Slack thread from @tstanmay13

Fixes a bug where Java SDK dynamic snippets were referencing inline types from the wrong package location. When `enable-inline-types` is configured, types are generated as nested static classes within parent types (e.g., `GetDiscriminatedUnionRequest.Bar`), but the snippet generator was trying to import them from `com.seed.object.types.*` where they don't exist.

**Link to Devin run:** https://app.devin.ai/sessions/b8576a2505a54491902ab0e494874424
**Requested by:** tanmay.singh@buildwithfern.com (@tstanmay13)

## Changes Made
- Added `inline` flag to the dynamic IR `Declaration` type to track whether a type should be generated as a nested class
- Updated `DynamicSnippetsConverter` to populate the `inline` flag from the full IR
- Implemented inline type mapping in `DynamicSnippetsGeneratorContext` that:
  - Traverses all types to find inline children and records their owner class + nested path
  - Handles name collisions by appending underscores (matching Java generator behavior)
  - Supports deeply nested inline types
- Added `getJavaClassReferenceForNamedType()` method that returns nested class references for inline types
- Updated `DynamicTypeMapper` and `DynamicTypeLiteralMapper` to use the new inline-aware method

## Testing
- [x] Seed tests pass for all 4 java-inline-types configurations:
  - `java-inline-types:inline`
  - `java-inline-types:no-inline`
  - `java-inline-types:no-wrapped-aliases`
  - `java-inline-types:enable-forward-compatible-enums`

## Human Review Checklist
- [ ] Verify the inline type naming logic in `getUniqueNestedName()` matches the Java generator's collision handling
- [ ] Check that type assertions for `inline` property (lines 82-84, 320-322, 501) are safe workarounds until the SDK is published
- [ ] Confirm traversal logic in `traverseTypeForInlineChildren()` handles all type structures correctly


tstanmay13:
update the snapshots those are correct changes as we added a property

fern-support (approved):

PR #11084, opened 2025-12-08
fix(typescript): use bracket notation for non-identifier property names in file upload requests

tjb9dc:
## Description

Refs: https://github.com/fern-demo/uipath-typescript/commit/cde7c73511239a0bc22deb07d709f5b606379583

Fixes a bug where the TypeScript SDK generator produced invalid JavaScript when file upload request parameters had property names containing hyphens (e.g., header names like `X-Custom-Header`).

## Changes Made

- Modified `FileUploadRequestParameter.getReferenceToProperty()` to use bracket notation (`request["X-Custom-Header"]`) instead of dot notation when the property name is not a valid JavaScript identifier
- Added shared `isValidIdentifier()` utility to `@fern-typescript/commons` using `esutils.keyword.isIdentifierNameES6` for proper ES6 identifier validation
- Added changelog entry for version 3.38.4

**Before (invalid JS - interpreted as subtraction):**
```typescript
request.X - Custom - Header != null
    ? request.X - Custom - Header.toString()
    : undefined
```

**After (valid JS):**
```typescript
request["X-Custom-Header"] != null
    ? request["X-Custom-Header"].toString()
    : undefined
```

- Refactored to use shared `isValidIdentifier` utility from `@fern-typescript/commons` instead of a local regex, per reviewer feedback
- The utility uses `esutils.keyword.isIdentifierNameES6` which is already used elsewhere in the codebase (e.g., `getPropertyKey.ts`)
- Merged with main branch and updated version to 3.38.4

## Testing

- [x] Manual code review - verified the fix matches the pattern used in `GeneratedQueryParams.ts` which already uses `createElementAccessExpression` for similar cases

## Human Review Checklist

- [ ] Verify `esutils.keyword.isIdentifierNameES6("X-Custom-Header")` returns `false` (triggering bracket notation)
- [ ] Consider if a seed test fixture should be added for regression coverage
- [ ] Consider if the new shared `isValidIdentifier` utility needs unit tests

---

Link to Devin run: https://app.devin.ai/sessions/0522dde80b444962aea65de6933ee41a


generators/typescript/sdk/client-class-generator/src/request-parameter/FileUploadRequestParameter.ts
```diff
@@ -111,9 +111,17 @@ export class FileUploadRequestParameter extends AbstractRequestParameter {
...
+        if (!this.isValidIdentifier(propertyName)) {
+            return ts.factory.createElementAccessExpression(
+                requestIdentifier,
+                ts.factory.createStringLiteral(propertyName)
+            );
+        }
+        return ts.factory.createPropertyAccessExpression(requestIdentifier, propertyName);
+    }
+
+    private isValidIdentifier(name: string): boolean {
```
tjb9dc: do we have utilities to do this elsewhere?

PR #11105, opened 2025-12-08
feat(python): add Python 3.11+ StrEnum compatibility

tjb9dc:
## Thomas-authored context

According to https://github.com/python/cpython/issues/100458, `StrEnum` is a drop-in replacement for `(str, Enum)`

## Description

Refs: Slack request from @tjb9dc

Python 3.11 changed how enum mixins work, breaking the `(str, Enum)` pattern that was previously used. This PR adds a version-aware enum base class that uses `StrEnum` for Python >= 3.11 and falls back to the `(str, Enum)` mixin for older versions.

## Changes Made

- Add `enum.py` utility file to both pydantic and SDK core utilities with version-aware `StrEnum` base class
- Update pydantic CoreUtilities to copy `enum.py` and provide `get_fern_enum()` reference method
- Update SDK CoreUtilities to copy `enum.py` to generated SDKs (not re-exported from `__init__.py` - internal use only)
- Update enum generator to use `StrEnum` instead of hardcoded `(str, enum.Enum)` base classes
- Add changelog entries for all three Python generators:
  - SDK: v4.44.0 (feat - minor bump)
  - Pydantic: v1.10.2 (fix - patch bump)
  - FastAPI: v1.12.2 (fix - patch bump)
- [x] Updated README.md generator (if applicable) - N/A

## Testing

- [x] Seed tests run for `enum:real-enum` and `enum:real-enum-forward-compat` fixtures
- [x] Seed tests run for `pydantic:enum` fixture
- [x] Pre-commit hooks pass
- [ ] Unit tests added/updated - relies on existing seed tests

## Human Review Checklist

- [x] Verify `StrEnum` implementation correctly handles both Python < 3.11 and >= 3.11
- [x] Verify forward-compatible enums with `_missing_` method still work correctly with new base class
- [x] Verify `enum.py` filename doesn't cause import shadowing issues (should be fine since it's in a package)
- [x] Verify module import pattern (`from ..core import enum` then `enum.StrEnum`) works correctly
- [x] Note: `str(enum_member)` behavior differs on Python 3.11+ (returns value like `"foo"` instead of `"EnumClass.FOO"`) - this is expected stdlib StrEnum behavior

- Changed changelog type from `feat` to `fix` for fastapi (v1.12.2) and pydantic (v1.10.2) generators to satisfy semver validator
- Changed import pattern from `from ..core.enum import StrEnum` to `from ..core import enum` so generated enums use `enum.StrEnum` instead of `StrEnum` directly. This avoids potential naming conflicts with user-defined classes named `StrEnum`.
- Bumped SDK version from v4.43.0 to v4.44.0 after merging main (another PR took v4.43.0)

---

Link to Devin run: https://app.devin.ai/sessions/ad1b79e5c3524fa3b6d9b6dc70710b5b


generators/python/core_utilities/pydantic/fern_enum.py
```diff
@@ -0,0 +1,9 @@
+import enum
+import sys
+
+if sys.version_info >= (3, 11):
+    from enum import StrEnum as FernEnum
```
tjb9dc: let's just leave this named `StrEnum`, and rename the other one to `StrEnum`!

I think that will be the cleanest solution, and since the imports are all contained anyways should be safe

tjb9dc: let's also rename the file to `enum.py`.

And let's add comments to this file

generators/python/src/fern_python/generators/sdk/core_utilities/core_utilities.py
```diff
@@ -53,6 +53,15 @@ def copy_to_project(self, *, project: Project) -> None:
...
             exports={"serialize_datetime"} if not self._exclude_types_from_init_exports else set(),
         )
+        self._copy_file_to_project(
+            project=project,
+            relative_filepath_on_disk="fern_enum.py",
+            filepath_in_project=Filepath(
+                directories=self.filepath,
+                file=Filepath.FilepathPart(module_name="fern_enum"),
+            ),
+            exports={"FernEnum"} if not self._exclude_types_from_init_exports else set(),
```
tjb9dc: W should never reexport once we rename it to `StrEnum` (we will import it directly and it will only be used internally)

seed/python-sdk/enum/real-enum/src/seed/types/enum_with_special_characters.py
```diff
@@ -1,12 +1,13 @@
...
 
-import enum
 import typing
 
+from ..core.enum import StrEnum
+
 T_Result = typing.TypeVar("T_Result")
 
 
-class EnumWithSpecialCharacters(str, enum.Enum):
```
tjb9dc, line 9: Is there any breaking change to consider here? Off the top of my head I don't think so -- SDK user code like `isinstance(EnumWithSpecialCharacters, enum.Enum)` would continue to work the same I think

tjb9dc:
oh also, can we make sure this is added to the fastapi and pydantic generators?

And those versions.yml files should all get updates as well

generators/python/src/fern_python/generators/core_utilities/core_utilities.py
```diff
@@ -119,6 +129,12 @@ def get_serialize_datetime(self) -> AST.Reference:
             ),
         )
 
+    def get_fern_enum(self) -> AST.ClassReference:
+        return AST.ClassReference(
+            qualified_name_excluding_import=(),
+            import_=AST.ReferenceImport(module=AST.Module.local(*self._module_path, "enum"), named_import="StrEnum"),
```
tjb9dc: one more edge case:

Can we have this show up as:
```
import ..core.enum

class Foo(enum.StrEnum):
```

I want to avoid a conflict with a class named StrEnum

seed/python-sdk/enum/real-enum/src/seed/types/enum_with_special_characters.py
tjb9dc: Are you absolutely sure there is no change to serialization?

generators/python/src/fern_python/generators/sdk/core_utilities/core_utilities.py
```diff
@@ -53,6 +53,15 @@ def copy_to_project(self, *, project: Project) -> None:
             ),
             exports={"serialize_datetime"} if not self._exclude_types_from_init_exports else set(),
         )
+        self._copy_file_to_project(
```
tjb9dc: can we only copy this file when relevant to generation? I.e. when `_use_str_enums` is False? Don't want to pollute the core folder when not relevant

Check me on that but I believe that's the only case when this code path is relevant

seed/python-sdk/enum/real-enum/src/seed/types/enum_with_special_characters.py
tjb9dc: okay one more question: Does our implementation fully support mypy typing?

tjb9dc: can you also test the true positive case? (where mypy correctly catches a type error for `takes_operand`)

tjb9dc: err nevermind I don't think that's necessary

kafkas (approved):
I pulled this branch and ran a few small scripts against the generated enum SDKs under both Python 3.10 and 3.13 interpreters, and everything looks good with the new `StrEnum` helper and import pattern.

One question about `str(enum)`: in the `enum:real-enum` fixture’s `RawHeadersClient` we serialize header enums with `str(operand)`, which on older Python sends the enum name `Operand.GREATER_THAN` instead of the wire value `>`. Does this effectively mean header enums are currently wrong on Python < 3.11?

PR #11150, opened 2025-12-10
feat(go): Add OAuth token fetching support

tjb9dc:
## Description
Refs: Completes OAuth2 support started in #11142

Requested by: @tjb9dc
Link to Devin run: https://app.devin.ai/sessions/a9090504ff1c4a78a8de4a3e6e6ff31f

This PR completes the OAuth2 client credentials flow for the Go SDK by implementing automatic token fetching and caching. When OAuth is configured, the SDK now automatically fetches tokens from the OAuth endpoint and caches them, refreshing when expired by re-exchanging client credentials.

## Changes Made
- Added `TokenGetter` function type and `tokenGetter` field to `RequestOptions` in the Go generator
- Updated `ToHeader()` to call `tokenGetter` if no token is manually set
- Added `SetTokenGetter` method to `RequestOptions` for internal use
- Updated `ClientGenerator.ts` to generate OAuth token fetching setup in root client constructor
- Root client now creates an `OAuthTokenProvider`, a separate auth client (to avoid recursion), and sets up a token getter function
- Added version 1.19.0 entry to versions.yml

Fixed CI failures by extracting endpoint information from IR instead of hardcoding:
- Get method name from endpoint (e.g., `GetToken` vs `GetTokenWithClientCredentials`) instead of hardcoding
- Get request type reference using serviceId and wrapper name from IR
- Get field names from `requestProperties` (e.g., `Cid`/`Csr` for custom fixtures) instead of hardcoding `ClientId`/`ClientSecret`
- Use struct copy pattern (`authOptions := *options`) instead of manual field assignment for cleaner code

All 4 OAuth fixtures now pass: `oauth-client-credentials`, `oauth-client-credentials-default`, `oauth-client-credentials-custom`, `oauth-client-credentials-nested-root`

## Testing
- [x] Seed tests pass for all OAuth fixtures

## Human Review Checklist
- [ ] Verify the OAuth token fetching logic in `sdk.go` (lines 1159-1189) - this appears to be duplicate/unused code since `ClientGenerator.ts` generates the actual client code
- [ ] Note: This implementation re-exchanges client credentials on expiry. Refresh token support (if configured in IR) is not implemented - confirm if this is acceptable or needs follow-up

### DIFF
--- a/seed/go-sdk/oauth-client-credentials/client/client.go
+++ b/seed/go-sdk/oauth-client-credentials/client/client.go
@@ -3,6 +3,8 @@
 package client
 
 import (
+	context "context"
+	fern "github.com/oauth-client-credentials/fern"
 	auth "github.com/oauth-client-credentials/fern/auth"
 	core "github.com/oauth-client-credentials/fern/core"
 	internal "github.com/oauth-client-credentials/fern/internal"
@@ -25,6 +27,30 @@ type Client struct {
 
 func NewClient(opts ...option.RequestOption) *Client {
 	options := core.NewRequestOptions(opts...)
+	oauthTokenProvider := core.NewOAuthTokenProvider(
+		options.ClientID,
+		options.ClientSecret,
+	)
+	authOptions := core.NewRequestOptions()
+	authOptions.BaseURL = options.BaseURL
+	authOptions.HTTPClient = options.HTTPClient
+	authClient := auth.NewClient(
+		authOptions,
+	)
+	options.SetTokenGetter(func() (string, error) {
+		if token := oauthTokenProvider.GetToken(); token != "" {
+			return token, nil
+		}
+		response, err := authClient.GetTokenWithClientCredentials(context.Background(), &fern.GetTokenRequest{
+			ClientId:     options.ClientID,
+			ClientSecret: options.ClientSecret,
+		})
+		if err != nil {
+			return "", err
+		}
+		oauthTokenProvider.SetToken(response.AccessToken, response.ExpiresIn)
+		return response.AccessToken, nil
+	})
 	return &Client{
 		Auth:         auth.NewClient(options),
 		NestedNoAuth: client.NewClient(options),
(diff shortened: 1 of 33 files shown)


seed/go-sdk/oauth-client-credentials/client/client.go
```diff
@@ -25,6 +27,30 @@ type Client struct {
...
+		response, err := authClient.GetTokenWithClientCredentials(context.Background(), &fern.GetTokenRequest{
+			ClientId:     options.ClientID,
+			ClientSecret: options.ClientSecret,
+		})
+		if err != nil {
+			return "", err
+		}
+		oauthTokenProvider.SetToken(response.AccessToken, response.ExpiresIn)
+		return response.AccessToken, nil
+	})
```
tjb9dc, line 51: will this handle requesting a new token upon expiry?

seed/go-sdk/oauth-client-credentials/client/client.go
```diff
@@ -25,6 +27,30 @@ type Client struct {
...
+	oauthTokenProvider := core.NewOAuthTokenProvider(
+		options.ClientID,
+		options.ClientSecret,
+	)
+	authOptions := core.NewRequestOptions()
+	authOptions.BaseURL = options.BaseURL
+	authOptions.HTTPClient = options.HTTPClient
+	authClient := auth.NewClient(
+		authOptions,
+	)
```
tjb9dc: is there a simpler or more complete way to pass through the options? Or is this the best way

seed/go-sdk/oauth-client-credentials/client/client.go
tjb9dc: should we be using `nil` instead to be more explicit/obvious?

seed/go-sdk/oauth-client-credentials/client/client.go
tjb9dc: meh no let's do it this way. I just don't want to forget a field somehow

tjb9dc: something more like typescript destructuring would be better

fern-support (approved):

PR #11180, opened 2025-12-11
fix(java): add @JsonIgnore to prevent fluent setter conflicts for with* properties

tstanmay13:
## Description
Linear ticket: Closes 

## Changes Made
When a schema has both a property like `label` and `withLabel`, Jackson interprets the convenience method `withLabel(T)` as a fluent setter for `label` (stripping the `with` prefix), causing deserialization conflicts.

This fix detects when a property starts with `with` and another property matches the suffix, then adds @JsonIgnore to the convenience setter to prevent the conflict.

Fixes issue reported by Pipedream where their OpenAPI schema with both `label` and `withLabel` properties caused Jackson deserialization errors.

## Testing


seed/java-sdk/java-with-property-conflict/default/src/main/java/com/seed/javaWithPropertyConflict/types/ConfigurableProp.java
```diff
@@ -0,0 +1,248 @@
...
+            this.description = description;
+            return this;
+        }
+
+        /**
+         * <p>If true, requires label-value format</p>
+         * @return Reference to {@code this} so that method calls can be chained together.
+         */
+        @java.lang.Override
+        @JsonIgnore
```
tstanmay13, line 207: here!

tjb9dc (approved):

PR #11196, opened 2025-12-12
fix(python): fix missing root client headers param

kafkas:
## Description
Linear ticket: [FER-8029](https://linear.app/buildwithfern/issue/FER-8029/anduril-root-client-initializer-is-missing-headers-param) 

This fixes a bug where the generated root client constructor could omit the headers parameter even though the generated ClientWrapper supports custom headers. We now ensure the generic headers constructor parameter is always included in the shared constructor info so it’s available at the root client and passed through correctly.

## Changes Made

- Updated `_get_constructor_info` in`client_wrapper_generator.py` to prevent missing headers param
- Updated Seed snapshots

## Testing

### DIFF
--- a/seed/python-sdk/any-auth/src/seed/core/client_wrapper.py
+++ b/seed/python-sdk/any-auth/src/seed/core/client_wrapper.py
@@ -10,15 +10,15 @@ class BaseClientWrapper:
     def __init__(
         self,
         *,
-        headers: typing.Optional[typing.Dict[str, str]] = None,
         api_key: str,
         token: typing.Optional[typing.Union[str, typing.Callable[[], str]]] = None,
+        headers: typing.Optional[typing.Dict[str, str]] = None,
         base_url: str,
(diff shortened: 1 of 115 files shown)


seed/python-sdk/any-auth/src/seed/core/client_wrapper.py
```diff
@@ -56,14 +56,14 @@ class SyncClientWrapper(BaseClientWrapper):
     def __init__(
         self,
         *,
+        headers: typing.Optional[typing.Dict[str, str]] = None,
         api_key: str,
         token: typing.Optional[typing.Union[str, typing.Callable[[], str]]] = None,
-        headers: typing.Optional[typing.Dict[str, str]] = None,
```
dsinghvi, line 61: nit: headers should be the 3rd arg here (not first)

tjb9dc (approved):

PR #11197, opened 2025-12-12
feat(ruby): Add always-on OIDC publishing workflow and simplify wire tests

iamnamananand996:
Closes: [FER-7970](https://linear.app/buildwithfern/issue/FER-7970/generate-ciyml-file)

## Description

Adds an always-on GitHub Actions workflow for publishing gems to RubyGems.org using OIDC trusted publishing. The publish workflow is now generated for all Ruby v2 SDKs without requiring any configuration.

## Changes Made

- Updated `github-ci.yml` template to always include a `publish` job that:
  - Runs on tag pushes after lint and test jobs pass
  - Uses OIDC trusted publishing via `rubygems/release-gem@v1`
  - Requires `id-token: write` and `contents: write` permissions
- Simplified wire tests execution by running them inline with regular tests using `RUN_WIRE_TESTS=true` env var instead of a separate CI job
- Updated all seed test fixtures (100 files) to include the publish workflow
- Added `versions.yml` entry for version `1.0.0-rc68`

- Removed `enablePublishWorkflow` feature flag per review feedback - publish workflow is now always generated (no opt-in/opt-out)
- Regenerated all seed fixtures to include the publish job
- Updated changelog entry to reflect always-on behavior

## Testing

- [x] Seed fixtures updated and regenerated
- [x] Code compiles successfully

## Review Checklist

- [ ] Confirm always-on publish workflow is the desired behavior (no opt-out mechanism)
- [ ] Verify OIDC publish workflow configuration is correct for RubyGems.org
- [ ] Check that `id-token: write` and `contents: write` permissions are appropriate for all generated SDKs

---

Link to Devin run: https://app.devin.ai/sessions/072b7c866e94403e8cebd3081b5aa298
Requested by: naman.anand@buildwithfern.com (@iamnamananand996)


generators/ruby-v2/ast/src/custom-config/BaseRubyCustomConfigSchema.ts
```diff
@@ -8,6 +8,8 @@ export const BaseRubyCustomConfigSchema = z.object({
     customPagerName: z.optional(z.string()),
     // Generate wire tests for serialization/deserialization
     enableWireTests: z.boolean().optional(),
+    // Generate publish workflow for RubyGems.org with OIDC trusted publishing
+    enablePublishWorkflow: z.boolean().optional(),
```
tjb9dc: I'd like to avoid the feature flag and just make this always on. Thoughts?

tjb9dc (approved):

generators/ruby-v2/ast/src/custom-config/BaseRubyCustomConfigSchema.ts
iamnamananand996: Yes, thier is no harm in adding it by default.

Lemme see, how it done on other language will replicate the same

iamnamananand996: other generator `publishInfo` check, however, Ruby does not have this kind of ecosystem. So better to remove the `enablePublishWorkflow` and have also enabled.

PR #11199, opened 2025-12-12
fix(python): Fix empty query params stripping existing URL query strings

aditya-arolkar-swe:
## Description

Fixes a bug where empty query parameters would cause httpx to strip existing query parameters from URLs, breaking features like pagination cursors.

**Link to Devin run**: https://app.devin.ai/sessions/fd42729d974d4d8a9aaf037d47ec850a
**Requested by**: adi@buildwithfern.com (@aditya-arolkar-swe)

## Problem

The `HttpClient.request()` and `AsyncHttpClient.request()` methods (and their `stream()` variants) always passed `params=encode_query(...)` to httpx, even when the result was an empty list `[]`. 

httpx behavior: When `params={}` or `params=[]` is passed, httpx strips any existing query parameters from the URL. This breaks features that rely on URLs with pre-existing query strings (e.g., pagination cursors like `?after=123`).

## Changes Made

- Compute encoded params into `_encoded_params` variable first
- Pass `params=_encoded_params if _encoded_params else None` to httpx
- When params are empty, `None` is passed instead of `[]`, preserving the URL's existing query string
- Applied fix to all 4 affected methods: `HttpClient.request()`, `HttpClient.stream()`, `AsyncHttpClient.request()`, `AsyncHttpClient.stream()`
- Added version 4.45.7 changelog entry

## Testing

- [x] Pre-commit hooks passed
- [x] Seed tests passed for `pagination` fixture (2/2 test cases)
- [ ] No unit test added for this specific edge case

## Human Review Checklist

- [x] Verify the fix is consistently applied across all 4 methods in `http_client.py`
- [ ] Confirm `encode_query()` returns `[]` (falsy) for empty input, not a truthy empty value
- [x] Consider if a dedicated unit test should be added for this behavior

### DIFF
--- a/generators/python/tests/utils/test_http_client.py
+++ b/generators/python/tests/utils/test_http_client.py
@@ -152,12 +152,12 @@ def test_http_client_does_not_pass_empty_params_list() -> None:
         httpx_client=dummy_client,  # type: ignore[arg-type]
         base_timeout=lambda: None,
         base_headers=lambda: {},
-        base_url=lambda: "https://example.com/resource?after=123",
+        base_url=lambda: "https://example.com",
     )
 
-    # No params, no request_options.additional_query_parameters
+    # Use a path with query params (e.g., pagination cursor URL)
     http_client.request(
-        path="",
+        path="resource?after=123",
         method="GET",
         params=None,
         request_options=None,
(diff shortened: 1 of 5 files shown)


generators/python/tests/utils/test_http_client.py
```diff
@@ -105,3 +139,104 @@ def test_remove_none_from_dict_empty_dict() -> None:
...
+    http_client.request(
+        path="",
+        method="GET",
+        params=None,
+        request_options=None,
+    )
+
+    # We care that httpx receives params=None, not [] or {}
+    assert "params" in dummy_client.last_request_kwargs
+    assert dummy_client.last_request_kwargs["params"] is None
```
aditya-arolkar-swe, line 168: This unit test should also check that the query params in base URL wasn't erased. DevinAI please update this test accordingly.

aditya-arolkar-swe: and obviously apply the same fix to the async test below

Swimburger (approved):

PR #11226, opened 2025-12-15
fix(python): stringify non-string values in multipart form data requests

tjb9dc:
## Description

Fixes a critical runtime crash when making multipart form data requests with non-string values (int, bool, float). The httpx library requires all form field values to be strings or bytes for multipart/form-data encoding, but the `jsonable_encoder` preserves primitive types, causing a `TypeError` when httpx attempts to encode the request.

Link to Devin run: https://app.devin.ai/sessions/5ca6dc37ba074e2ea36f19543e88ff86

## Changes Made

- Added `_stringify_multipart_form_data()` helper function that converts all non-string/bytes form data values to strings
- Modified `_maybe_filter_none_from_multipart_data()` to use the new stringify function instead of just filtering None values
- Added version 4.46.1 changelog entry

- Added `json` import and updated `_stringify_multipart_form_data()` to use `json.dumps()` for dict/list values instead of `str()`, producing proper JSON wire format (e.g., `{"key": "value"}` instead of Python's `{'key': 'value'}`)

## Value Conversion Logic

| Input Type | Output |
|------------|--------|
| `None` | Skipped (not included) |
| `str`/`bytes` | Passed through unchanged |
| `dict`/`list` | JSON-encoded via `json.dumps()` |
| `int`/`float`/`bool`/etc. | Converted via `str()` |

## Testing

- [ ] Seed tests will run in CI to verify generated code

## Human Review Checklist

- [ ] Verify `json.dumps()` for nested structures produces the expected wire format for your API consumers
- [ ] Verify `str()` conversion for booleans produces acceptable output ("True"/"False" vs "true"/"false") - some servers may expect lowercase JSON-style booleans
- [ ] Confirm the fix handles edge cases: None values are skipped, bytes/strings pass through unchanged
- [ ] Consider if explicit unit tests should be added for this runtime behavior


generators/python/core_utilities/shared/http_client.py
```diff
@@ -121,18 +121,34 @@ def _should_retry(response: httpx.Response) -> bool:
...
+
+
 def _maybe_filter_none_from_multipart_data(
     data: typing.Optional[typing.Any],
     request_files: typing.Optional[RequestFiles],
     force_multipart: typing.Optional[bool],
 ) -> typing.Optional[typing.Any]:
     """
-    Filter None values from data body for multipart/form requests.
-    This prevents httpx from converting None to empty strings in multipart encoding.
```
tjb9dc, line 131: does this still convert none to empty string? That part is important

tjb9dc: oh okay great

generators/python/core_utilities/shared/http_client.py
```diff
@@ -121,18 +121,34 @@ def _should_retry(response: httpx.Response) -> bool:
...
+    httpx requires all form field values to be strings or bytes for multipart/form-data encoding.
+    """
+    result: typing.Dict[str, typing.Any] = {}
+    for key, value in data.items():
+        if value is None:
+            continue
+        if isinstance(value, (str, bytes)):
+            result[key] = value
+        else:
+            result[key] = str(value)
```
tjb9dc, line 144: can `value` here ever be a nested mapping? In which case I'm not sure that stringifying is sufficient

tjb9dc: is this really the correct behavior?

PR #11227, opened 2025-12-15
fix(python): fix retry logic off-by-two error in HTTP client

tjb9dc:
## Description

Refs: Issue #14 reported in Slack

Fixes the retry logic "off-by-two" error in the Python SDK's HTTP client. The `retries` parameter was incorrectly defaulting to 2 instead of 0, causing users to get `max_retries - 2` actual retries instead of the expected number.

**Example of the bug:**
| User Config | Expected Retries | Actual Retries (before fix) |
|-------------|------------------|----------------------------|
| `max_retries=3` | 3 | 1 |
| `max_retries=5` | 5 | 3 |

## Changes Made

- Changed `retries: int = 2` to `retries: int = 0` in all 4 HTTP client methods:
  - `HttpClient.request()`
  - `HttpClient.stream()` (for consistency, though it doesn't implement retry logic)
  - `AsyncHttpClient.request()`
  - `AsyncHttpClient.stream()` (for consistency, though it doesn't implement retry logic)
- Changed `max_retries` default from 0 to 2 (so users get 2 retries by default when not specified)
- Changed comparison from `max_retries > retries` to `retries < max_retries` (more readable, same logic)
- Added changelog entry in `versions.yml` for version 4.45.10

## Testing

- [x] Pre-commit hooks pass

## Human Review Checklist

- [ ] Verify the default behavior change is intentional: previously no retries by default, now 2 retries by default
- [ ] Verify `max_retries=0` still correctly disables retries
- [ ] Consider whether unit tests should be added for retry behavior
- [ ] Note: stream methods have the `retries` parameter but don't implement retry logic (no recursive call)

---
Link to Devin run: https://app.devin.ai/sessions/888c90b41dbb4e2f9c11a3d883d1998e

### DIFF
--- a/generators/python/core_utilities/shared/http_client.py
+++ b/generators/python/core_utilities/shared/http_client.py
@@ -235,7 +235,7 @@ class HttpClient:
         ] = None,
         headers: typing.Optional[typing.Dict[str, typing.Any]] = None,
         request_options: typing.Optional[RequestOptions] = None,
-        retries: int = 2,
+        retries: int = 0,
         omit: typing.Optional[typing.Any] = None,
         force_multipart: typing.Optional[bool] = None,
     ) -> httpx.Response:
@@ -338,7 +338,7 @@ class HttpClient:
         ] = None,
         headers: typing.Optional[typing.Dict[str, typing.Any]] = None,
         request_options: typing.Optional[RequestOptions] = None,
-        retries: int = 2,
+        retries: int = 0,
         omit: typing.Optional[typing.Any] = None,
         force_multipart: typing.Optional[bool] = None,
     ) -> typing.Iterator[httpx.Response]:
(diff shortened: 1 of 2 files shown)


generators/python/core_utilities/shared/http_client.py
```diff
@@ -235,7 +235,7 @@ def request(
         ] = None,
         headers: typing.Optional[typing.Dict[str, typing.Any]] = None,
         request_options: typing.Optional[RequestOptions] = None,
-        retries: int = 2,
+        retries: int = 0,
```
tjb9dc, line 238: hold on, we still want to default to 2 in the case when nothing is specified. At least the end outcome should be "the call is retried twice" from the user's POV

aditya-arolkar-swe (approved):

PR #11241, opened 2025-12-15
chore(cli): add test for webhook with dollar sign field value

sbawabe:
## Description

Refs: Requested by @sbawabe

Adds a test case to verify that webhooks with fields containing "$" characters (like currency values) are properly handled through the OpenAPI → IR → FDR conversion pipeline.

Link to Devin run: https://app.devin.ai/sessions/3b479a539bea4b9bbe77622186ef21b3

## Changes Made

- Added new OpenAPI fixture `webhook-dollar-field` with a webhook payload containing `Amount: "$99.99"`
- Added test case in `openapi-from-flag.test.ts` that validates the dollar sign value is preserved through the conversion pipeline
- Generated IR and FDR snapshots for regression testing

- Obfuscated test data to use generic placeholder values (e.g., "Acme Corp", "txn-abc123def456") while maintaining the same webhook structure and the key "$99.99" amount field

## Testing

- [x] Test passes locally with `pnpm test src/ir-to-fdr-converter/__test__/openapi-from-flag.test.ts -t "should handle OpenAPI with webhook containing dollar sign"`
- [x] Verified "$99.99" value is preserved in both IR and FDR snapshots

## Human Review Checklist

- [ ] Verify the OpenAPI fixture maintains the expected webhook structure (POST method, payload with string fields, nullable ExternalId)
- [ ] Confirm snapshot files correctly capture the "$99.99" value preservation
- [ ] Check that test data is sufficiently obfuscated from the original request


sbawabe:
proved this looks right, issue must be on FDR / docs def side

PR #11246, opened 2025-12-16
fix(go, java, python, typescript): Fix oauth edge cases for optional expires_in and default grant_type

fern-support:
## Description
Fixes OAuth token provider implementations across multiple language generators to handle cases where the `expires_in` property in OAuth token responses is optional. Some OAuth providers don't include `expires_in` in their token responses, which was causing failures. This change adds fallback behavior using a default expiration time of 1 hour (3600 seconds).

## Changes Made
- **Python SDK**: Handle optional `expires_in` with conditional fallback to default expiration
- **TypeScript SDK**: Check if `expires_in` is optional and use nullish coalescing to default value
- **Java SDK**: Handle optional `expires_in` using `Optional.orElse()` pattern; skip literal properties in OAuth request building
- **Go SDK**: Use pointer conversion for optional `client_id`/`client_secret` request fields; treat `expires_in` as pointer type
- **Seed tooling**: Transform `pip install` to `pipx install` to avoid PEP 668 errors on modern macOS

## Testing


packages/seed/src/commands/img/buildGeneratorImage.ts
tjb9dc, line 1: did you want this change in this PR?

generators/java/sdk/src/main/java/com/fern/java/client/generators/OAuthTokenSupplierGenerator.java
```diff
@@ -164,18 +172,31 @@ public GeneratedJavaFile generateFile() {
...
-                    EXPIRES_AT_FIELD_NAME,
-                    GET_EXPIRES_AT_METHOD_NAME,
-                    tokenPropertyName);
+            TypeReference expiresInType = expiresInProperty.getProperty().getValueType();
+            boolean isOptional = isOptionalType(expiresInType);
+            if (isOptional) {
+                // Handle optional expires_in with default fallback
+                // In Java, optional fields return Optional<T>, so use .orElse()
+                getMethodSpecBuilder.addStatement(
+                        "this.$L = $L(authResponse.get$L().orElse((int) $L))",
```
tjb9dc: what does this code look like? Is it `.orElse((int) "3600")`?

I think it should just be `.orElse($N)` which results in `.orElse(3600)`

packages/seed/src/commands/img/buildGeneratorImage.ts
jsklan: I'm fine with it here if you are.

generators/java/sdk/src/main/java/com/fern/java/client/generators/OAuthTokenSupplierGenerator.java
jsklan: good catch! fixed

tjb9dc (approved):

PR #11263, opened 2025-12-16
fix(go): reset request body before retry attempts

tjb9dc:
## Description

Refs: Customer issue reported in Slack - https://buildwithfern.slack.com/archives/C05KKD1P8BG/p1765560393682989

Fixes a bug where the Go SDK's retry logic would send empty request bodies on retry attempts. When the SDK retried a request, the `io.Reader` body was already consumed from the first attempt, causing servers to receive empty payloads.

Link to Devin run: https://app.devin.ai/sessions/f1cee93fd32f4cdb9c5937e6167646ac

## Changes Made

- Added body reset logic in `retrier.go_` that calls `request.GetBody()` before each retry attempt (when `retryAttempt > 0` and `GetBody` is available)
- Added `TestRetryWithRequestBody` test case that verifies POST requests with bodies are properly retried with the same body content
- Updated `versions.yml` with changelog entry for version 1.20.3
- [x] Updated seed output for exhaustive fixture

## Testing

- [x] Unit tests added - `TestRetryWithRequestBody` verifies the fix
- [x] Ran seed tests locally - internal tests pass (`ok github.com/exhaustive/fern/internal`)

## Human Review Checklist

- [ ] Verify the fix correctly handles non-replayable bodies (when `GetBody` is nil) - in this case, the body is not reset and the retry proceeds with whatever state the body is in
- [ ] Confirm the error handling when `GetBody()` fails is appropriate (returns error, doesn't retry with empty body)


generators/go-v2/base/src/asIs/internal/retrier.go_
```diff
@@ -98,6 +98,15 @@ func (r *Retrier) run(
 		return nil, err
 	}
 
+	// Reset the request body for retries since the body may have already been read.
+	if retryAttempt > 0 && request.GetBody != nil {
```
tjb9dc: would `GetBody` really ever be nil? Or is this the idiomatic way to do things

fern-support (approved):

PR #11266, opened 2025-12-16
fix(cli): treat dollar-prefixed strings as literals when they don't match example reference format

sbawabe:
Fixes FER-8063

## Description

Refs: Reported by Sarah Bawabe (@sbawabe)

Fixes an issue where strings starting with `$` in examples (like `"$3.00"` for currency values) were incorrectly interpreted as example references, causing "Failed to resolve examples" errors.

Link to Devin run: https://app.devin.ai/sessions/0a1b0fe1728741b48f96166212398916

## Changes Made

- **ExampleResolver.ts**: Added validation in `parseExampleReference` to ensure type/import names start with a letter (per Fern's naming rules). Updated `resolveExample` to fall back to treating the string as a literal when it doesn't match a valid example reference format.
- **visitAllReferencesInExample.ts**: Added the same validation for consistency during the validation phase, so `fern check` behaves consistently with `fern generate`.
- Added test fixture `dollar-string-examples` with webhook examples containing `"$3.00"` values.
- Updated `no-undefined-example-reference` test expectations: `$malformed-example` is now treated as a literal string (not flagged as malformed), which is intentional.
- Added `dollar-string-examples` to ruby-sdk `allowedFailures` list (fixture uses webhooks which ruby-sdk doesn't fully support).
- Updated snapshots for ir-generator-tests, ir-to-jsonschema, and swift-sdk.
- **versions.yml**: Added version 3.27.2 changelog entry for CLI release.

- Updated versions.yml from 3.24.1 to 3.27.2 after merging latest main (someone else had merged newer versions in the meantime).

## Human Review Checklist

- [ ] Verify the validation logic is consistent between `ExampleResolver.ts` and `visitAllReferencesInExample.ts` (they have similar but separate implementations)
- [ ] Check if there are other usages of `EXAMPLE_REFERENCE_PREFIX` that might need similar updates (I found `validateTypeReferenceExample.ts` also uses it but it already handles `parseExampleReference` returning null gracefully)
- [ ] Confirm the test fixture exercises the fix correctly
- [ ] Verify the change to `no-undefined-example-reference.test.ts` is correct: `$malformed-example` should now be treated as a literal string, not flagged as a malformed reference

## Testing

- [x] Local build succeeds (`pnpm run fern:build`)
- [x] ir-generator-tests pass locally (277 tests)
- [x] fern-definition-validator tests pass locally (69 tests)
- [x] CI checks passing (250/252 - remaining failures are unrelated php-sdk rate limiting issues)


packages/cli/fern-definition/validator/src/rules/no-undefined-example-reference/__test__/no-undefined-example-reference.test.ts
```diff
@@ -14,21 +14,10 @@ describe("no-undefined-example-reference", () => {
             )
         });
 
+        // Note: $malformed-example is now treated as a literal string (not an example reference)
+        // because it doesn't match the valid example reference format ($Type.Example or $import.Type.Example).
+        // This is intentional - strings like "$PATH", "$USD", "$3.00" should not require escaping.
         expect(violations).toEqual([
-            {
-                message:
-                    "Example $malformed-example is malformed. Examples should be formatted like $YourType.ExampleName",
```
dsinghvi, line 20: can you double check that `$malformed-example` is supposed to be a string, i.e. if a property is typed as an object we don't want to coerce the example into a string

dsinghvi (approved):

PR #11268, opened 2025-12-16
fix(cli): prevent optional wrapping of literal x-fern-type parameters

tjb9dc:
## Description
Refs: Slack thread from thomas@buildwithfern.com

Fixes a regression where `x-fern-type: literal<"...">` on parameters was being incorrectly parsed as `optional<literal<"...">>` instead of just `literal<"...">` when the parameter was not marked as `required: true`.

This was reported as a breaking change between CLI versions 0.84.1 and 2.6.2.

## Changes Made
- Modified `convertParameters.ts` to detect when a parameter's schema has `x-fern-type` that is a literal type
- When a literal `x-fern-type` is detected, the parameter is treated as required for optional/nullable wrapping purposes
- Added test fixture `x-fern-type-literal-header` to validate the fix

## Testing
- [x] All 241 existing tests pass
- [x] Lint checks pass

## Human Review Checklist
- [ ] Verify the string check `fernType.startsWith("literal<")` is robust enough for all valid literal type syntaxes
- [ ] Confirm the fix scope (parameters only, not object properties) is the correct behavior
- [ ] Check if the fallback `resolvedParameter.schema ?? resolvedParameter` correctly handles both schema-level and parameter-level `x-fern-type` extensions

---
Link to Devin run: https://app.devin.ai/sessions/bb19722f349c438cafa622e373eee1f7

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/endpoint/convertParameters.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/endpoint/convertParameters.ts
@@ -66,9 +66,18 @@ export function convertParameters({
         const parameterBreadcrumbs = [...requestBreadcrumbs, resolvedParameter.name];
         const generatedName = getGeneratedTypeName(parameterBreadcrumbs, context.options.preserveSchemaIds);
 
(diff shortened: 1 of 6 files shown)


tjb9dc:
found a workaround, also don't love how special-cased this is

PR #11270, opened 2025-12-16
fix(fastapi): update fastapi version to remove need for workaround, remove starlette direct dependency

sukeerthik-fern:
## Description
Linear ticket: Closes https://linear.app/buildwithfern/issue/FER-8014/candid-remove-fastapi-wrapper-patching
FastAPI couldn't find forward ref type hints (ex. "SomeModel") in a wrapped func (wrapped with `@functools.wraps`) without this workaround that copies wrapped func's `__globals__dict` into the wrapper's `__global__` dict (which fastapi inspects). 

FastAPI fixed the issue https://github.com/fastapi/fastapi/pull/5077 so we no longer need to do this workaround.

Bump the following deps:
```
fastapi = ">=0.123.5"
```
as well as `fernapi/fern-fastapi-server` to 2.0.0 to indicate major/breaking change.

## Changes Made
- [x] removed unneeded workaround lines that updated wrapper's `__globals__` dict
- [x] removed direct starlette dependency to let [FastAPI manage dependency transitively](https://fastapi.tiangolo.com/deployment/versions/#about-starlette)
- [x] bumped dependencies and fernapi/fern-fastapi-server version with new major version (2.0.0).
- [x] added custom config for `examples` test fixture. 

## Testing

- [x] added new test fixture config `generate-package-deps` to fastapi `examples` to output pyproject.toml that shows removed starlette direct dep and fastapi version locked
- [x] Unit tests pass


generators/python/src/fern_python/generators/fastapi/service_generator/endpoint_generator.py
```diff
@@ -183,15 +183,6 @@ def _write_init_body(writer: AST.NodeWriter) -> None:
             )
             writer.write_line()
 
-            writer.write_line("# this is necessary for FastAPI to find forward-ref'ed type hints.")
```
tjb9dc, line 186: Note that this was only fixed recently in 0.123.5 https://github.com/fastapi/fastapi/pull/5077

Although our existing code wasn't perfect and had some bugs. What do you think we should do to make sure that the break here is minimal?

seed/fastapi/seed.yml
```diff
@@ -28,7 +28,7 @@ test:
     command: docker build -f ./generators/python/fastapi/Dockerfile -t fernapi/fern-fastapi-server:latest ./generators/python
 language: python
 generatorType: Server
-defaultOutputMode: local_files
+defaultOutputMode: github
```
sukeerthik-fern: This is for testing with `seed` so that a `toml` gets generated by default and can be used for debugging changes to generator. I tested the output of `seed/fastapi/examples/pyproject.toml` to see if fastapi and starlette changed

```
[tool.poetry.dependencies]
python = "^3.8"
fastapi = ">=0.123.5"
pydantic = ">= 1.9.2"
pydantic-core = ">=2.18.2"
starlette = ">=0.40.0"
```

generators/python/fastapi/versions.yml
```diff
@@ -1,6 +1,17 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 # For unreleased changes, use unreleased.yml
 
+- version: 2.0.0
```
sukeerthik-fern, line 4: @tjb9dc @musicpulpite
So end users using the CLI can control whether a pyproject.toml is generated using their generators.yml file. They'll get a toml when they set `output.location` as github or npm / pypi / maven , but not when set to `local-file-system`. 

we need the following bumps as part of removing the lines for wrapper

fastapi = ">=0.123.5"
starlette = ">=0.40.0" (bc of fastapi bump)

If we want to "force" users to always get a toml along with their generated code, we'd need to consider updating logic of handling `output.location: local-file-system` to include a toml as part of generation. That might be more than we want to do but can be done in `generators/python/src/fern_python/cli/abstract_generator.py` specifically for the fastapi generator. 

In short term, we can bump fern (major ver) as part of these updates is my take and just emphasize to SDK users that they will need to update fastapi as well (which they should see if they chose `output.location` != `local-file-system`).

tjb9dc: Question:
Do you know why we are declaring both fastapi and starlette? Is starlette a direct dependency or a transitive (of FastAPI)?

Re: toml, we can't assume that because folks use `local-file-system` (specifically with FastAPI it's more common) to generate the files into a source directory rather than a complete package. Kind of weird, we should probably encourage standalone package generation. But FastAPI is kinda a legacy thing. We strive for zero dependencies in the other languages but really not possible for server stubs.

Anyways, that's all to say:

- I think we should make sure our declared version is bumped (for when a toml is generated)
- Bumping to 2.0.0 with a clear note about why sounds great since we know that the main customer this is for uses local output mode

seed/fastapi/seed.yml
tjb9dc: huh are there any consequences of doing this? Maybe we should just do this lol, or at least do it in some places

sukeerthik-fern: > huh are there any consequences of doing this? Maybe we should just do this lol, or at least do it in some places

We do it in most places, unsure if it was intentionally omitted but the effect is it just outputs `toml` files so we can inspect the deps. If we want to commit these `.toml` files as part of our `seed` tests, we could perhaps do that to make sure deps are being set as intended for our fixtures

Screenshot 2025-12-16 at 10 27 54 PM

generators/python/fastapi/versions.yml
sukeerthik-fern: startlette is a transitive dependency of fastapi but we [shouldn't need to pin the version](https://fastapi.tiangolo.com/deployment/versions/#about-starlette) ourselves. The reason we have it (from my knowledge) is because in various files like `generators/python/src/fern_python/generators/fastapi/service_generator/endpoint_generator.py`, `generators/python/src/fern_python/generators/fastapi/register/register_file_generator.py`, and a few other places, we explicitly return or type hint with Starlette HTTP exceptions (`Starlette.HTTPException`) when we should probably just use FastAPI HTTPexceptions which are based on Starlette httpexception classes. 

> * I think we should make sure our declared version is bumped (for when a toml is generated)

I added the`defaultOutputMode: github` change was so I could see the toml generated and the versions are bumped. assuming that our seed outputs are representative behavior of fern CLI as an end-user would use it, we should be getting outputs when a user generates a package as opposed to `local-file-system`. But let me know if there's a better way to test this output.

> * Bumping to 2.0.0 with a clear note about why sounds great since we know that the main customer this is for uses local output mode

Got it, so as-in update the current messaging in the 2.0.0 version bump to be very specific.

seed/fastapi/seed.yml
tjb9dc: sounds good, let's do it for some of the fixtures but not all maybe?

sukeerthik-fern:
@tjb9dc made some changes that should clean things up
- removed direct starlette dep (fastapi manages it transitively)
- updated fastapi version to no longer need workaround
- added new test fixture config to generate a pyproject.toml so we can validate fastapi version dependency generated
- bumped fastapi generator version to 2.0.0 with clear messaging on changes

seed/fastapi/examples/generate-package-deps/pyproject.toml
sukeerthik-fern, line 1: In this file, you'll see the removal of starlette and the updated fastapi version requirement.

generators/python/fastapi/versions.yml
```diff
@@ -1,6 +1,23 @@
...
+  changelogEntry:
+    - type: break
```
tjb9dc, line 6: Can we also mention the removal of the patching code? And link to the FastAPI issue

tjb9dc (approved):

PR #11273, opened 2025-12-16
fix(cli, typescript): Fix handling of empty allOf in additionalProperties

jsklan:
## Description
Fixes two TypeScript compilation errors when generating SDKs for OpenAPI specs that use hyphenated header names and empty `allOf` schemas in `additionalProperties`.

These issues were discovered when generating a custom fixture, where headers like `if-match` and `content-md5` caused syntax errors, and certain metadata fields produced references to non-existent types.

## Changes Made
- **FileUploadRequestParameter.ts**: Use bracket notation (`request["if-match"]`) instead of dot notation for header properties with hyphens or other invalid JS identifier characters
- **convertAdditionalProperties.ts**: Treat `additionalProperties.allOf` as "any" type when all elements are empty or metadata-only schemas (only `example`, `description`, etc.), generating `Record<string, unknown>` instead of a non-existent type reference

## Testing
- [x] Manual testing: Generated Box SDK compiles successfully with `pnpm build`
- [x] Verified hyphenated headers use bracket notation in generated code
- [x] Verified empty `allOf` produces `Record<string, unknown>` instead of broken type reference


packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertAdditionalProperties.ts
```diff
@@ -230,7 +230,29 @@ export function isAdditionalPropertiesAny(
         return false;
     }
     if ("allOf" in additionalProperties && additionalProperties.allOf != null) {
-        return false;
+        // Check if allOf contains only empty schemas or schemas with only metadata (example, description, etc.)
+        // If so, treat it as "any" type
+        const hasTypeDefiningSchema = additionalProperties.allOf.some((schema) => {
```
tjb9dc, line 235: Should we just scrub out the empty `allOf` in the CLI? I imagine this will result in weirdness in other languages.

What is the diff in the type this will generate for the Box spec?

generators/typescript/sdk/client-class-generator/src/request-parameter/FileUploadRequestParameter.ts
```diff
@@ -111,9 +111,21 @@ export class FileUploadRequestParameter extends AbstractRequestParameter {
     }
 
     private getReferenceToProperty(propertyName: string): ts.Expression {
-        return ts.factory.createPropertyAccessExpression(
-            ts.factory.createIdentifier(this.getRequestParameterName()),
-            propertyName
-        );
+        // Check if property name is a valid JavaScript identifier
+        // Valid identifiers: start with letter, _, or $, followed by letters, numbers, _, or $
+        const isValidIdentifier = /^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(propertyName);
```
tjb9dc, line 116: I think we have a util somewhere else for this

packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertAdditionalProperties.ts
jsklan: I'm not sure I know what you mean. This is in the cli and returns false which is already one of the expected values for this code path.

generators/typescript/sdk/client-class-generator/src/request-parameter/FileUploadRequestParameter.ts
jsklan: I'm searching, but not finding one...

eyw520 (approved):

PR #11326, opened 2025-12-18
feat(fastapi): use temp_pydantic_v1_params for pydantic v1/v1_on_v2 modes

tjb9dc:
## Description
Refs: Request from @tjb9dc via Slack

Adjusts the FastAPI generator to use FastAPI's `temp_pydantic_v1_params` module for parameter markers (Path, Query, Header, Body, File) when in pydantic v1 or v1_on_v2 mode. Additionally, uses `typing.get_type_hints()` to resolve forward references before applying parameter annotations for ALL pydantic versions.

Reference: https://github.com/fastapi/fastapi/blob/master/fastapi/temp_pydantic_v1_params.py
Related issue: https://github.com/fastapi/fastapi/issues/13056

## Changes Made
- Added new `FastAPIParams` class that generates version-aware imports based on pydantic compatibility mode
- Updated `FastApiGeneratorContext` to expose a `fastapi_params` instance
- Updated all endpoint parameter classes to use `self._context.fastapi_params.X()` instead of static `FastAPI.X()` methods
- Modified `endpoint_generator.py` to use `typing.get_type_hints()` for forward reference resolution (cleaner than per-parameter `evaluate_forwardref` calls)
- Set `response_model=None` for all pydantic model endpoints to disable response model validation/serialization (per https://fastapi.tiangolo.com/tutorial/response-model/#disable-response-model)
- Updated seed test outputs for all fixtures
- Added changelog entry for version 2.1.0

## Testing
- [x] Pre-commit hooks pass
- [x] Seed tests pass (7/7 for exhaustive fixture)
- [x] Verified pydantic-v1 fixture generates `fastapi.temp_pydantic_v1_params.Body()` etc.
- [x] All CI checks pass (97 passed)

## Human Review Checklist
- [ ] Verify `response_model=None` doesn't affect OpenAPI schema generation negatively (FastAPI should still infer from wrapper's return type annotation)
- [ ] Confirm `typing.get_type_hints()` correctly resolves forward references in all edge cases
- [ ] Verify seed test outputs are up to date with the latest generator changes

- Changed forward ref resolution from per-parameter `evaluate_forwardref()` calls to using `typing.get_type_hints()` once per endpoint (per reviewer feedback with diff from Candid API)
- Changed `response_model` handling: now explicitly sets `response_model=None` for all pydantic model endpoints instead of omitting it (per reviewer feedback citing https://fastapi.tiangolo.com/tutorial/response-model/#disable-response-model)

---
Link to Devin run: https://app.devin.ai/sessions/75b0c5ac25314f3d8bec4af9136950d4


generators/python/src/fern_python/generators/fastapi/service_generator/endpoint_generator.py
```diff
@@ -271,12 +272,26 @@ def _write_update_endpoint_signature(self, writer: AST.NodeWriter) -> None:
...
-                    writer.write(f"{NEW_PARAMETERS_VARIABLE_NAME}.append(")
-                    writer.write(f"{PARAMETER_VALUE_VARIABLE_NAME}.replace(")
-                    writer.write("annotation=typing.Annotated[")
-                    writer.write(f"{PARAMETER_VALUE_VARIABLE_NAME}.annotation, ")
-                    writer.write_node(parameter.get_fastapi_marker())
-                    writer.write("]")
+                    if uses_pydantic_v1_params:
+                        EVALUATED_VARIABLE_NAME = "evaluated"
+                        writer.write(f"{EVALUATED_VARIABLE_NAME} = ")
+                        writer.write_reference(FastAPI.evaluate_forwardref)
```
nicknicolini: this is correct, but you also need to evaluate forward refs in the else block below for pydantic v2. forward refs need to be evaluated for all fastapi params that use `Annotated`, regardless of pydantic version.

generators/python/src/fern_python/generators/fastapi/service_generator/endpoint_generator.py
```diff
@@ -271,12 +272,26 @@ def _write_update_endpoint_signature(self, writer: AST.NodeWriter) -> None:
...
                     python_default = parameter.get_python_default()
-                    writer.write(f"{NEW_PARAMETERS_VARIABLE_NAME}.append(")
-                    writer.write(f"{PARAMETER_VALUE_VARIABLE_NAME}.replace(")
-                    writer.write("annotation=typing.Annotated[")
-                    writer.write(f"{PARAMETER_VALUE_VARIABLE_NAME}.annotation, ")
-                    writer.write_node(parameter.get_fastapi_marker())
-                    writer.write("]")
+                    if uses_pydantic_v1_params:
+                        EVALUATED_VARIABLE_NAME = "evaluated"
+                        writer.write(f"{EVALUATED_VARIABLE_NAME} = ")
```
nicknicolini: can you also write a comment into the generated code that links to the fastapi issue that this works around? here is the link: https://github.com/fastapi/fastapi/issues/13056

tjb9dc (commented):
While we're here, let's remove `response_model` specification from our generated stubs. It is not necessary when the type is the same as the return type. See: https://docs.astral.sh/ruff/rules/fast-api-redundant-response-model/

tjb9dc:
> While we're here, let's remove `response_model` specification from our generated stubs. It is not necessary when the type is the same as the return type. See: https://docs.astral.sh/ruff/rules/fast-api-redundant-response-model/

Actually, the proper thing is to include them but type them as “None”

fern-support (approved):

PR #11334, opened 2025-12-18
feat(python): add positional_single_property_constructors config option

aditya-arolkar-swe:
## Description

Refs: Slack thread from customer requesting improved ergonomics for single-property wrapper types

Adds a new `pydantic_config.positional_single_property_constructors` option that generates custom `__init__` methods for models with a single required field, allowing positional argument construction.

**Before:**
```python
Isin(isin="US0378331005")
Quantity(quantity=10000.0)
```

**After (with config enabled):**
```python
Isin("US0378331005")
Quantity(10000.0)
```

Link to Devin run: https://app.devin.ai/sessions/146a1e0c5ea14526b52841f2db4f81c9
Requested by: adi@buildwithfern.com (@aditya-arolkar-swe)

## Changes Made

- Added `positional_single_property_constructors: bool = False` config option to `BasePydanticModelCustomConfig`
- Implemented `_maybe_add_positional_init()` in `PydanticModel` class that:
  - Detects models with exactly one required non-discriminator field
  - Generates `__init__(self, field: Type, **kwargs) -> None` that calls `super().__init__()`
  - Properly excludes optional fields and literal discriminator fields from the count
- Plumbed config through `FernAwarePydanticModel` and discriminated union generators
- Added generation-time warning in both Python generator (stderr) and CLI (terminal output)
- Added CLI-side warning in `runLocalGenerationForWorkspace.ts` that displays bold red warning right after "Wrote files to..." line
- Added version 4.47.0 to Python SDK versions.yml (irVersion: 62)
- Added version 3.31.0 to CLI versions.yml for the CLI-side warning feature
- Added JSON schema snapshots for the new test fixture

## Testing

- [x] Created new `python-positional-single-property` test fixture with test cases for:
  - Simple single-property types (`Isin`)
  - Single-property with discriminator literal (`IsinWithDiscriminator`, `Quantity`)
  - Multiple discriminator literals (`MultipleDiscriminators`)
  - Two required fields - should NOT get positional init (`TwoFields`)
  - One required + one optional - should get positional init (`OneRequiredOneOptional`)
  - Nested wrapper types (`Trader`, `TakerParty`, `BondSingleLeg`)
  - Union types (`Identifier`, `Cusip`)
- [x] Seed tests pass for both `no-custom-config` and `with-positional-constructors` configurations
- [x] JSON schema snapshot tests pass
- [x] Pre-commit checks pass

- **Fixed CI cache issue causing go-sdk/php-model/php-sdk failures**: Added a comment to `packages/ir-sdk/src/index.ts` to force Turbo cache invalidation. The CI was using stale build artifacts for `@fern-api/ir-sdk` that didn't include the `ExampleInlinedRequestBodyProperty` export, causing TypeScript compilation errors in `@fern-api/ir-utils`. This is a workaround for stale CI cache, not a code bug.
- Previous: Moved CLI warning from `GenerationRunner.ts` to `runLocalGenerationForWorkspace.ts` for local generation support

## Human Review Checklist

- [ ] Verify discriminator detection logic in `_maybe_add_positional_init()` handles edge cases correctly
- [ ] Confirm generated `__init__` with `**kwargs` is compatible with both Pydantic v1 and v2
- [ ] Check that config being `False` by default means no changes to existing SDK behavior
- [ ] Verify the CLI warning now appears in terminal output when running `fern generate --local` with the config enabled
- [ ] Confirm the config path traversal (`pydantic_config.positional_single_property_constructors`) matches the actual config structure in `generatorInvocation.config`
- [ ] Note: The IR SDK cache invalidation comment in `packages/ir-sdk/src/index.ts` is a workaround for stale Turbo cache in CI - can be removed once CI cache is refreshed


generators/python/sdk/versions.yml
```diff
@@ -1,5 +1,16 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
 # For unreleased changes, use unreleased.yml
+- version: 4.47.0
+  changelogEntry:
+    - summary: |
+        Add `positional_single_property_constructors` option to `pydantic_config`. When enabled, generates
+        a custom `__init__` method for models with a single required field (ignoring discriminator/literal
+        type fields with defaults). This allows positional argument construction like `Isin("US0378331005")`
+        instead of `Isin(isin="US0378331005")`.
```
tjb9dc: Can we make this example more generic? And maybe more terse

tjb9dc: would love to also include the nuance about how this wrecks future backcompat

tjb9dc (approved):

generators/python/sdk/versions.yml
aditya-arolkar-swe: Take this backwards compatibility a step further - add a loud warning about the compatibility issue in the fern generation itself when this flag is turned on. Add that warning to the end of the generation log, and highlight it red so it's clear to the generating user.

PR #11336, opened 2025-12-18
fix(typescript): Fix typescript wire tests for pagination

jsklan:
## Description
Fix wire test generation for paginated endpoints when the example response has a null or missing cursor field. Previously, the test generator incorrectly assumed pagination results were "missing" because it looked for camelCase property names (e.g., `transferOrders`) in JSON examples, but IR JSON examples are always in wire format (e.g., `transfer_orders`). This caused tests to incorrectly assert `expect(page.data).toEqual([])` instead of the actual expected data.

## Changes Made
- **TestGenerator.ts**: Fixed `isPaginationResultsPathMissingInExample` to always use wire values when walking JSON examples
- **TestGenerator.ts**: Added `isPaginationCursorMissingInExample` function to detect when cursor is missing and skip `hasNextPage()` assertions
- **users.yml**: Added test examples with and without cursor to validate the fix

## Testing
- [x] Manual testing with Square SDK generation - all 1067 tests pass
- [x] Seed pagination fixture tests pass with updated examples

### DIFF
--- a/generators/typescript/asIs/tests/mock-server/withJson.ts
+++ b/generators/typescript/asIs/tests/mock-server/withJson.ts
@@ -28,7 +28,9 @@ export function withJson(expectedBody: unknown, resolver: HttpResponseResolver):
         }
 
         const mismatches = findMismatches(actualBody, expectedBody);
-        if (Object.keys(mismatches).filter((key) => !key.startsWith("pagination.")).length > 0) {
+        // Filter out pagination-related fields (e.g., cursor) that change between paginated requests
+        const isPaginationField = (key: string) => key.startsWith("pagination.") || key === "cursor";
+        if (Object.keys(mismatches).filter((key) => !isPaginationField(key)).length > 0) {
             console.error("JSON body mismatch:", toJson(mismatches, undefined, 2));
             return passthrough();
         }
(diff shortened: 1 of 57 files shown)


generators/typescript/asIs/tests/mock-server/withJson.ts
```diff
@@ -28,7 +28,9 @@ export function withJson(expectedBody: unknown, resolver: HttpResponseResolver):
         }
 
         const mismatches = findMismatches(actualBody, expectedBody);
-        if (Object.keys(mismatches).filter((key) => !key.startsWith("pagination.")).length > 0) {
+        // Filter out pagination-related fields (e.g., cursor) that change between paginated requests
+        const isPaginationField = (key: string) => key.startsWith("pagination.") || key === "cursor";
```
Swimburger: how would it know what are pagination related fields? I don't think we can hardcode this

Swimburger (changes requested):
Not sure how to solve this, but we shouldn't hardcode fieldnames here

generators/typescript/asIs/tests/mock-server/withJson.ts
jsklan: oops, yeah I agree...

fern-support (approved):

Swimburger (approved):

PR #11406, opened 2026-01-05
fix(openapi): handle discriminator mapping without oneOf in both conversion paths

thesandlord:
## Description
Refs: Slack thread from Sandeep Dinesh

Link to Devin run: https://app.devin.ai/sessions/eddb3e06655c47ec89a88b5bced08a4e
Requested by: Sandeep Dinesh (@thesandlord)

Fixes handling of OpenAPI schemas that have `discriminator.mapping` but no explicit `oneOf`/`anyOf`. This pattern is used by some APIs (like Dropbox Sign) where a base schema defines a discriminator with mapping to variant schemas, but doesn't list them in a `oneOf`.

Previously, such schemas were incorrectly converted as plain objects. Now they are properly converted as discriminated unions.

## Changes Made
- Added `tryConvertDiscriminatorMappingSchema()` method in `SchemaConverter.ts` (for `--from-openapi` direct path) that:
  - Detects schemas with `discriminator.mapping` but no `oneOf`/`anyOf`
  - Synthesizes a `oneOf` from the mapping values
  - Delegates to `OneOfSchemaConverter` for proper discriminated union conversion
- Added handling in `convertSchemas.ts` (for OpenAPI→Fern conversion path) that:
  - Detects the same `discriminator.mapping` pattern
  - Calls `convertDiscriminatedOneOf` which uses the mapping internally to build union subtypes
- Added test fixture `discriminator-base-class` demonstrating this pattern
- Bumped CLI version to 3.33.0

- Rebased on latest main
- Removed smoke test API changes (per user request)
- Added version bump to 3.33.0 in `versions.yml`

## Testing
- [x] Unit tests added/updated (new test fixture with snapshots)
- [x] Manual testing completed (lint checks pass)
- [x] Verified fix works on Dropbox Sign docs preview (discriminated union now renders with "Hide 10 variants" UI)

## Human Review Checklist
- [ ] Verify the fix handles edge cases where `discriminator.mapping` exists but the schema should remain an object
- [ ] Confirm both code paths (SchemaConverter.ts and convertSchemas.ts) produce consistent results
- [ ] Note: The openapi-ir-parser path calls `convertDiscriminatedOneOf` directly without synthesizing oneOf first - verify this is correct since `convertDiscriminatedOneOf` uses `discriminator.mapping` internally (line 51-74 of convertDiscriminatedOneOf.ts)

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertSchemas.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/convertSchemas.ts
@@ -1123,6 +1123,36 @@ export function convertSchemaObject(
             });
         }
 
+        // handle discriminator.mapping without oneOf/anyOf
+        // Some OpenAPI specs define discriminated unions using discriminator.mapping
+        // without explicit oneOf/anyOf. The convertDiscriminatedOneOf function will
+        // use the discriminator.mapping to build the union subtypes.
+        if (
+            schema.discriminator?.mapping != null &&
+            Object.keys(schema.discriminator.mapping).length > 0 &&
+            schema.oneOf == null &&
+            schema.anyOf == null
+        ) {
+            return convertDiscriminatedOneOf({
+                nameOverride,
+                generatedName,
+                title,
+                breadcrumbs,
+                description,
+                availability,
+                discriminator: schema.discriminator,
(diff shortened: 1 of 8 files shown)


thesandlord:
image

chdeskur (approved):

PR #11420, opened 2026-01-06
feat(java): generate overloads without request wrapper when all query params are optional

tjb9dc:
## Description

Refs: Slack thread from @tjb9dc

When all query parameters in an endpoint are optional, adding new optional query parameters would previously cause compile breaks for SDK users because the method signatures changed. This PR generates additional method overloads that don't require the request wrapper, preserving backward compatibility.

**Before:** Adding an optional query parameter would change:
```java
getLatestInsurance(String userId)
getLatestInsurance(String userId, RequestOptions requestOptions)
```
to:
```java
getLatestInsurance(String userId, UserGetLatestInsuranceRequest request)
getLatestInsurance(String userId, UserGetLatestInsuranceRequest request, RequestOptions requestOptions)
```

**After:** All four overloads are generated when all query params are optional:
```java
getLatestInsurance(String userId)
getLatestInsurance(String userId, RequestOptions requestOptions)  // NEW
getLatestInsurance(String userId, UserGetLatestInsuranceRequest request)
getLatestInsurance(String userId, UserGetLatestInsuranceRequest request, RequestOptions requestOptions)
```

## Changes Made

- Added `getNoRequestBodyWithRequestOptionsMethodSpec()` to the `HttpEndpointMethodSpecs` interface hierarchy
- Updated `AbstractEndpointWriter` to generate the new overload when `SdkRequestIsOptional` returns true
- Added `addEndpointWithoutRequestWithRequestOptionsReturnStatement()` to sync/async response parser generators
- Updated `AbstractClientGeneratorUtils` to include the new method spec in generated clients
- Added `java-optional-query-params-overloads` test fixture to validate the fix
- Added JSON schema snapshots for the new test fixture types
- Added Swift SDK endpoint path snapshot for the new test fixture

## Testing

- [x] Seed test added (`java-optional-query-params-overloads`)
- [x] Seed test passes - generated SDK compiles successfully
- [x] Verified generated code contains expected 4 overloads for all-optional endpoint
- [x] Verified endpoints with required params still generate only 2 overloads
- [x] JSON schema snapshot tests pass
- [x] Swift SDK endpoint path snapshot tests pass

## Human Review Checklist

- [ ] Verify `SdkRequestIsOptional` visitor correctly identifies when ALL properties are optional (not just some)
- [ ] Check the delegation logic in the new overload passes correct default values (`builder().build()`, `Optional.empty()`, or `OptionalNullable.absent()`)
- [ ] Confirm this works for request bodies with optional properties, not just query params
- [ ] Review parameterized type handling for `OptionalNullable<T>` in response parser generators

---

Link to Devin run: https://app.devin.ai/sessions/c1a111911cc745adbd6d17abedcfaadd
Requested by: @tjb9dc


generators/java/sdk/src/main/java/com/fern/java/client/generators/endpoint/AbstractEndpointWriter.java
```diff
@@ -302,6 +302,65 @@ public Optional<String> _visitUnknown(Object o) {
                     endpointWithoutRequestBuilder, endpointWithRequestOptions, paramNamesWoBody, bodyParameterSpec);
             endpointWithoutRequest = endpointWithoutRequestBuilder.build();
         }
+
+        // Generate overload without request body but WITH RequestOptions when all properties are optional
+        MethodSpec endpointWithoutRequestWithRequestOptions = null;
```
tjb9dc: Couldn't we consolidate this with the if block above? Starting on line 262

Since we want to always generate the pair of overloads (one with and without RequestOptions), and that if block is already doing all the heavy lifting

fern-support (approved):

PR #11447, opened 2026-01-07
fix(python): support `bytes` for websocket methods, various `methodName` fixes

patrickthornton:
## Description
Preserves full message schema for referenced payloads and not just `type`; adds support for `bytes`-type arguments in websocket methods, which in turn use `send` instead of `send_model`; fixed potential mislabeling of websocket methods with the incorrect method name.

## Changes Made
Small change to IR generator; changes to Pydantic typing in Python generator to extract `bytes` from `validation: "binary"`.

## Testing
[The large addition in the diff is because of tables of contents in README.mds showing up, which is somebody else's change they just forgot to update seed for, I'm presuming.]

### DIFF
--- a/generators/python/src/fern_python/generators/context/type_reference_to_type_hint_converter.py
+++ b/generators/python/src/fern_python/generators/context/type_reference_to_type_hint_converter.py
@@ -225,6 +225,7 @@ class TypeReferenceToTypeHintConverter:
                 return False
 
             return string_type.validation.format == "binary"
+
         str_type_hint = AST.TypeHint.bytes if is_bytes() else AST.TypeHint.str_
 
         print(str_type_hint)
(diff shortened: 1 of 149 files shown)


generators/python/src/fern_python/generators/context/type_reference_to_type_hint_converter.py
```diff
@@ -212,10 +212,30 @@ def _get_type_hint_for_named(
...
+
+            string_type: ir_types.StringType = v2_union
+            if string_type.validation is None:
+                return False
+
+            return string_type.validation.format == "binary"
+
+        str_type_hint = AST.TypeHint.bytes if is_bytes() else AST.TypeHint.str_
+
+        print(str_type_hint)
```
tjb9dc, line 231: Removeme

tjb9dc (approved):

generators/python/src/fern_python/generators/context/type_reference_to_type_hint_converter.py
patrickthornton: shameful of me

PR #11465, opened 2026-01-08
fix(python-sdk): use endpoint ID for wire test snippet generation

tjb9dc:
## Description

Fixes a bug where Python SDK wire tests would call the wrong client methods when multiple namespaces have endpoints with the same HTTP method and path pattern.

**Root cause**: The wire test generator was using path-based endpoint lookup (`resolveEndpointLocationOrThrow`), which returns all endpoints matching a given HTTP method and path. When multiple namespaces have endpoints with the same path (e.g., `POST /v2/Services`), the first matching endpoint was selected instead of the correct one for the namespace being tested.

**Example failure**:
```
def test_beta_create_service() -> None:
    client = get_client(test_id)
    client.alpha.create_service()  # BUG: should be client.beta.create_service()
```

Link to Devin run: https://app.devin.ai/sessions/44c8492bf6774b9084046925596cc5c8
Requested by: @tjb9dc

## Changes Made

- Added `resolveEndpointById` and `resolveEndpointByIdOrThrow` methods to `AbstractDynamicSnippetsGeneratorContext` for direct endpoint lookup by ID
- Added `generateMethodCallSnippetAstById` method to Python v2 `DynamicSnippetsGenerator` that uses endpoint ID instead of path matching
- Updated `WireTestGenerator.generateApiCallAst` to use the new ID-based lookup
- Added test fixture `python-wire-test-path-collision` demonstrating the bug with two namespaces having the same path pattern

## Testing

- [x] Added seed test fixture that reproduces the path collision scenario
- [x] Verified fixture fails before fix (calls `client.alpha.*` instead of `client.beta.*`)
- [x] Verified fixture passes after fix (correctly calls `client.beta.*`)
- [x] Ran `pnpm check:fix` - passed

## Human Review Checklist

- [ ] Verify the new methods in `AbstractDynamicSnippetsGeneratorContext` are appropriate for the shared base class
- [ ] Consider if other generators (TypeScript, Java, etc.) might have similar wire test issues


fern-support:
succeeded by https://github.com/fern-api/fern/pull/11470

PR #11467, opened 2026-01-08
feat(openapi): Add generator support for OpenAPI overlays specs

schottra:
## Description
Linear ticket: Refs https://linear.app/buildwithfern/issue/FER-8144/add-openapi-overlays-support

## Changes Made
(note: merging to a feature branch, needs more work to be feature complete)

See the [announcement](https://www.openapis.org/blog/2024/10/22/announcing-overlay-specification) and [spec](https://github.com/OAI/Overlay-Specification/blob/main/versions/1.0.0.md) for context.

* Added core util to apply an overlay to a spec object, including a smattering of tests to try and exercise all the edge cases in the spec
* Added test fixture with a sample overlay and snapshots to exercise the generator path
* Updated generation code paths to support overlay field in YAML, read the file, and use the helper to apply it after overrides

I copied the test fixture for [applying overrides](https://github.com/fern-api/fern/blob/de7af91cda40486574c5402fd5d8844fc13ba400/packages/cli/api-importers/openapi/openapi-ir-to-fern-tests/src/__test__/fixtures/overrides-resolution), so it's not exhaustive.

## Open Questions / Issues
The test fixture for overlays was based on the overrides fixture next to it. However, when generated, the snapshot has one reference to `UserSettings` which ends up typed as `unknown` in the overlays code path. At first glance I'm not sure what's causing this, but wonder if it's something to with refs not getting copied over correctly.
image

Do we need to add this support to  https://github.com/fern-api/fern/blob/066edd00e9d74464fd150839fb539c716efd6d66/packages/cli/workspace/browser-compatible-fern-workspace/src/InMemoryOpenAPILoader.ts as well?

## Testing
Unit and snapshot tests cover this change for now. I haven't manually tested it, would do that as part of merging the feature branch.

## TODO / Follow-ons
- Add instructions for using overlays to public docs


packages/cli/workspace/lazy-fern-workspace/src/loaders/applyOverlays.ts
```diff
@@ -0,0 +1,140 @@
...
+        return false;
+    }
+
+    if (!overlay.info?.title || !overlay.info?.version) {
+        context.logger.error("Overlay file missing required 'info.title' or 'info.version' field");
+        return false;
+    }
+
+    if (!Array.isArray(overlay.actions) || overlay.actions.length === 0) {
+        context.logger.error("Overlay file must have at least one action");
```
fern-support, line 116: I think this should be a warning, it shouldn't fail anything.

schottra: Practically speaking I agree. This was meant to enforce a rule in the Overlays spec:
```REQUIRED An ordered list of actions to be applied to the target document. The array MUST contain at least one value.```

fern-support:
A couple of comments:
- The problem with $ref resolving can be solved, I think, not worried about it
- William pointed me to the `immer` module that helps with mutating data structures. I think this may be relevant here, too.

I asked Claude to review the PR, and it had some helpful feedback, but nothing worth holding the PR back on. I recommend asking it to review and doing some polishing based on that.

Swimburger (approved):

fern-support (approved):

PR #11484, opened 2026-01-09
fix(typescript): ignore cursor mismatches in wire test mock server

jsklan:
## Description

Fixes wire test failures for pagination endpoints where `getNextPage()` sends a different cursor value than the original request.

Link to Devin run: https://app.devin.ai/sessions/569f5162819c47c4ba95afbfde8313de
Requested by: @jsklan

## Root Cause

When pagination tests call `getNextPage()`, the SDK correctly sends the cursor from the response (e.g., `"next_cursor_value"`), but the mock server was rejecting the request because it expected the original request's cursor value (e.g., `"cursor_value"`). The mock server already filtered out `pagination.*` mismatches but not top-level cursor mismatches.

## Changes Made

- Added `WithJsonOptions` interface to `withJson.ts` with optional `ignoredFields` parameter
- Updated `mockEndpointBuilder.ts` to pass `ignoredFields` through the `jsonBody` method
- Modified `TestGenerator.ts` to extract pagination cursor field names from the IR and pass them to the mock server at code generation time
- Added cursor pagination example to the existing pagination test fixture
- Updated versions.yml with changelog entry for version 3.43.7
- Updated IR generator test snapshot for pagination fixture

## Approach

Instead of hardcoding field names to ignore (like `key !== "cursor"`), the solution uses IR metadata to determine which fields are pagination cursors. The `TestGenerator` extracts the cursor field name from `endpoint.pagination.page.property.name.wireValue` for cursor pagination endpoints and passes it to the mock server's `jsonBody` method. This approach:
- Uses IR metadata to determine which fields are pagination cursors
- Avoids false positives from legitimate fields named "cursor"
- Makes the ignored fields explicit in the generated test code

Note: Nested cursor fields (e.g., `pagination.cursor`) are already handled by the existing `!key.startsWith("pagination.")` filter.

## Testing

- [x] Verified fix against customer's generated SDK (Square) - all 659 tests passed
- [x] Added cursor pagination example to pagination fixture
- [x] Pagination seed tests pass (2/2)
- [x] Lint checks pass

## Review Checklist

- [ ] Verify the `WithJsonOptions` interface and `ignoredFields` parameter in `withJson.ts`
- [ ] Confirm `TestGenerator.ts` correctly extracts cursor field names from IR for top-level cursor pagination
- [ ] Verify the pagination example in `users.yml` is valid


generators/typescript/asIs/tests/mock-server/withJson.ts
```diff
@@ -28,7 +28,7 @@ export function withJson(expectedBody: unknown, resolver: HttpResponseResolver):
         }
 
         const mismatches = findMismatches(actualBody, expectedBody);
-        if (Object.keys(mismatches).filter((key) => !key.startsWith("pagination.")).length > 0) {
+        if (Object.keys(mismatches).filter((key) => !key.startsWith("pagination.") && key !== "cursor").length > 0) {
```
fern-support: Using the key seems like it could lead to issues. Instead, is there a way for us to determine whether a field corresponds to a cursor used by the pagination feature? Isn't there metadata in the IR that specifies how the request/response fields correspond to pagination metadata that we could use to determine this?

fern-support: yeah, if possible it would make more sense to have the invoker of withJson tell it which fields to ignore rather than have withJson assume it can ignore any field called "cursor". Go ahead and implement this more robust approach.

PR #11541, opened 2026-01-14
feat(cli): add placeholder page for python-docs in fern docs dev

paarthfern:
When running `fern docs dev`, Python library documentation cannot be generated since it requires server-side processing via FDR. This change adds a helpful placeholder page that:

- Explains why Python docs are not available in dev mode
- Shows the GitHub URL being used for the library
- Provides instructions on how to generate the full docs
- Lists what content will be generated

The placeholder page is visible in dev mode but is marked as noindex to prevent search engine indexing. In production mode, FDR replaces this placeholder with the actual generated documentation.

## Testing
Tested locally: 
Screenshot 2026-01-14 at 11 01 48 AM

### DIFF
--- a/packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
+++ b/packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
@@ -1373,9 +1373,9 @@ export class DocsDefinitionResolver {
 title: ${title}
 ---
 
-<Note>
-This is a placeholder page. Python library documentation is automatically generated when you run \`fern generate --docs\` or \`fern generate --docs --preview\`.
-</Note>
+<Warning>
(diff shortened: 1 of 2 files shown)


packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
```diff
@@ -1351,28 +1351,84 @@ export class DocsDefinitionResolver {
...
+
+        // Add placeholder markdown content to parsedDocsConfig.pages
+        const placeholderMarkdown = `---
+title: ${title}
+---
+
+<Note>
+This is a placeholder page. Python library documentation is automatically generated when you run \`fern generate --docs\` or \`fern generate --docs --preview\`.
+</Note>
+
```
thesandlord, line 1379: lets add another note or warning here saying this is not currently supported on local dev

thesandlord (approved):

PR #11544, opened 2026-01-14
fix(go): `alwaysSendRequiredParameters` fix for inlined request bodies

patrickthornton:
## Description
Exactly what it says on the tin.

## Changes Made
Two-line fix to pass `alwaysSendRequiredParameters` along when visiting inlined request bodies and file uploads, which got forgotten before; so setting `alwaysSendRequiredParameters: true` would occasionally get ignored.

## Testing
Regenerating seed now! [Large diff is due to previous SDK versions getting merged without regenerating snapshots, so you'll see some old stuff there.]


musicpulpite (approved):
💯

PR #11556, opened 2026-01-14
chore(cli): upgrade CLI to 3.37.6 for fern-java-model support

tstanmay13:
## Description
Linear ticket: Closes 

fern-java-model requires a more updated version of the CLI to release 

## Testing

### DIFF
--- a/packages/ir-sdk/fern/fern.config.json
+++ b/packages/ir-sdk/fern/fern.config.json
@@ -1,4 +1,4 @@
 {
   "organization": "fern",
-  "version": "0.62.4"
+  "version": "3.37.6"
 }


tjb9dc (approved):

packages/ir-sdk/fern/fern.config.json
```diff
@@ -1,4 +1,4 @@
 {
   "organization": "fern",
-  "version": "0.62.4"
+  "version": "3.37.6"
```
tjb9dc, line 3: can you upgrade this via `fern upgrade`? I think we are missing some feature flags that may be relevant for backcompat (then again they might not be)

tstanmay13: yes let me do that

PR #11563, opened 2026-01-15
feat(cli): Initialize CLI v2

amckinney:
## Description

This initialies the upcoming CLI v2 project by setting up a new CLI entrypoint and `v2` sub-command in `packages/cli/cli-v2/...`. 

## Changes Made

For now, this bootstraps the sub-command layout, including the `auth` command with no-op implementations for the additional nested sub-commands. Note that the top-level `v2` command is hidden.

A separate CLI library, [commander](https://www.npmjs.com/package/commander), was explored in [this](https://github.com/fern-api/fern/commit/2cf1972f881d9d082da664bc3655520b2dc67b45) commit, but it turns out that the automated `--help` text, developer experience, and lack of [shell completion](https://yargs.js.org/docs/#api-reference-completioncmd-description-fn) isn't quite as good as [yargs](https://www.npmjs.com/package/yargs) (which is what we use today). For now, we'll continue to use `yargs`, but other CLI libraries might be explored later.

This also splits out some of the logging utilities defined in the `@fern-api/cli` package out into `@fern-api/cli-logger` so that it can be independently consumed by `@fern-api/cli-v2`. For now, we create a lighter `Context` type that we can iterate on rather than carrying the baggage of the old `CliContext`. The package reorganization allows us to use the `TtyAwareLogger` abstraction as-is.

## Testing

A variety of manual tests are included below -- a test harness for the CLI v2 will be added in a follow-up.

```sh
$ pnpm fern-dev:local --help

Commands:
  fern diff                         Diff two versions of an API
  
  # fern v2 is NOT listed here.

...
```

```sh
$ pnpm fern-dev:local v2 auth --help
Authenticate fern

Commands:
  fern auth login   Log in to fern
  fern auth logout  Log out of fern
  fern auth token   Print the user's authentication token

Options:
  --help       Show help                                               [boolean]
  --version    Show version number                                     [boolean]
  --log-level  Set log level
          [string] [choices: "debug", "info", "warn", "error"] [default: "info"]
```

```sh
$ pnpm fern-dev:local v2 auth token
unimplemented
```


dsinghvi (approved):
👨‍🍳

packages/cli/cli-logger/src/TtyAwareLogger.ts
```diff
@@ -6,10 +6,13 @@ import IS_CI from "is-ci";
...
 import { Log } from "./Log";
-import { TaskContextImpl } from "./TaskContextImpl";
+
+interface Task {
+    printInteractiveTasks: ({ spinner }: { spinner: string }) => string;
+}
 
 export class TtyAwareLogger {
-    private tasks: TaskContextImpl[] = [];
+    private tasks: Task[] = [];
```
dsinghvi, line 15: should we continue to use this TtyAwareLogger, pretty much hasnt been touched since 2022 and most people are scared to touch it 😅 

wonder if there is something better/ootb that we should consider

dsinghvi: not related or a blocker for this PR, just thought to mention as you touched this file

amckinney: Good call, I'll carry it over as-is for now, but will track an issue to simplify it. Even if we define a separate logger in `cli-v2`, pushing this into the `cli-logger` package is a worthwhile refactor.

PR #11594, opened 2026-01-16
feat(cli): Add fern.yml with source code info

amckinney:
## Description

Linear ticket: Closes https://linear.app/buildwithfern/issue/FER-8259/cli-v2-implement-yaml-source-code-info

This introduces the initial schema for the `fern.yml` configuration file with source code info tracking, backed by [yaml](https://www.npmjs.com/package/yaml) and [zod](https://zod.dev/) (v4). This change is primarily focused on wiring up _YAML source code info_, so only the following subset of configuration is included for now:

```yaml
# fern.yml
edition: 2026-01-01
org: acme
sdks:
  autorelease: true
  targets:
    node:
      lang: typescript
```

With this, the CLI can provide far better diagnostics which are compatible with all IDEs. For example, suppose you provided the following invalid config:

```yaml
# fern.yml
edition: 2026-01-01
org: acme
sdks:
  autorelease: xyz
  targets:
    node:
      lang: 12345
```

```sh
$ fern v2 check
../fern.yml:5:16: sdks.autorelease must be a boolean
../fern.yml:8:13: sdks.targets.node.lang must be a string
```

## Changes Made

This introduces the [yaml](https://www.npmjs.com/package/yaml) library instead of [js-yaml](https://www.npmjs.com/package/js-yaml) because it includes source code information tracking. The new `@fern-api/yaml-loader` and `@fern-api/source` packages are used to map the YAML nodes to the respective Zod schema definition. The generic `Sourced` type is used to map YAML paths to their respective `<line, column>` numbers. 

From there, the `FernYmlConverter` will be able to refer to specific source paths if it encounters any validation errors beyond what `zod` is capable of (i.e. does a particular SDK version actually exist)?

With the help of TypeScript's [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) type, the UX for this is shown below:

```typescript
const loader = new YamlConfigLoader({ cwd });
const result = await loader.load(path, FernYmlSchema);
if (result.success) {
    console.log(result.sourced.edition);        // Plain value
    console.log(result.sourced.edition.$loc);   // Source location
}
```

We intentionally do _not_ attempt to track source code info in the converted `FernYml` because it will soon be typed in terms of internal abstractions (e.g. `Workspace`), and it will become impossible to reliably map these constructs back to a single source.

Note that `$ref` support, along with source code tracking between referenced files, will be added in a follow-up.

## Testing

### DIFF
--- a/packages/cli/cli-v2/src/__test__/loadFernYml.test.ts
+++ b/packages/cli/cli-v2/src/__test__/loadFernYml.test.ts
@@ -1,0 +1,277 @@
+import { mkdir, rm, writeFile } from "fs/promises";
+import { tmpdir } from "os";
+import { join } from "path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { loadFernYml } from "../config/loadFernYml";
+import { ValidationError } from "../errors/ValidationError";
+
+const SAMPLE_FERN_YML = `edition: 2026-01-01
+org: acme
+
+cli:
+  version: 3.38.0
+
+sdks:
+  autorelease: true
+  defaultGroup: public
+  targets:
+    node:
+      lang: typescript
+      version: "1.4.0"
+`;
+
+describe("loadFernYml", () => {
+    let testDir: string;
+    let nestedDir: string;
+
+    beforeEach(async () => {
+        testDir = join(tmpdir(), `fern-test-${Date.now()}`);
+        nestedDir = join(testDir, "nested", "deep");
+        await mkdir(nestedDir, { recursive: true });
+        await writeFile(join(testDir, "fern.yml"), SAMPLE_FERN_YML);
+    });
+
+    afterEach(async () => {
+        await rm(testDir, { recursive: true, force: true });
+    });
+
+    it("finds fern.yml in the current directory", async () => {
+        const fernYml = await loadFernYml({ cwd: testDir });
+
+        expect(fernYml.edition).toBe("2026-01-01");
+        expect(fernYml.org).toBe("acme");
+    });
+
+    it("crawls up the directory tree to find fern.yml", async () => {
+        const fernYml = await loadFernYml({ cwd: nestedDir });
+
+        expect(fernYml.edition).toBe("2026-01-01");
+        expect(fernYml.org).toBe("acme");
+    });
+
+    it("preserves source location for the root object via source property", async () => {
+        const fernYml = await loadFernYml({ cwd: testDir });
+
+        expect(fernYml.source.$loc.absoluteFilePath.toString()).toContain("fern.yml");
+        expect(fernYml.source.$loc.line).toBe(1);
+        expect(fernYml.source.$loc.column).toBe(1);
+    });
+
+    it("preserves source location for top-level values via source property", async () => {
+        const fernYml = await loadFernYml({ cwd: testDir });
+
+        expect(fernYml.source.edition.$loc.line).toBe(1);
+        expect(fernYml.source.edition.$loc.column).toBe(10);
+        expect(fernYml.source.org.$loc.line).toBe(2);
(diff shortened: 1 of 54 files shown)


packages/cli/cli-v2/src/__test__/loadFernYml.test.ts
```diff
@@ -0,0 +1,277 @@
...
+                const output: string[] = [];
+                for (const issue of validationError.issues) {
+                    output.push(issue.toString());
+                }
+
+                expect(output).toEqual(["fern.yml:7:13: sdks.targets.node.lang must be a string"]);
+            }
+        });
+    });
+});
```
dsinghvi, line 277: these tests are 👨‍🍳 so awesome

packages/cli/config/src/FernYml.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export interface OutputConfig {
+    path?: string;
+    git?: GitOutputConfig;
+}
+
+export interface GitOutputConfig {
+    repository: string;
+}
```
dsinghvi, line 61: do you need to define these types here? i would have expected these to be derived from the zod schemas

packages/cli/yaml/loader/src/ValidationIssue.ts
```diff
@@ -0,0 +1,41 @@
+import type { SourceLocation } from "@fern-api/source";
+
+/**
+ * Represents a validation issue found when parsing a YAML configuration file.
+ *
+ * Use `toString()` to format as `<file>:<line>:<col>: <message>` for CLI output.
+ */
+export class ValidationIssue {
```
dsinghvi, line 8: should we create a first class attribute called docs url which points to the right page to learn more?

dsinghvi (approved):
This is looking awesome and the impl makes sense to me 👌

packages/cli/config/src/FernYml.ts
amckinney: The rationale is that the `zod` shape is purely for validating the configuration at the _edge_. We can define a separate type for additional post-processing (e.g. validate that the SDK version actually exists, create other internal abstractions based on the config, etc). All of this will happen during the `FernYmlConverter` step, which still has access to the source code information.

This is similar to what we do today with the [GeneratorsConfiguration](https://github.com/fern-api/fern/blob/5673d57158ee7581ab7632cb1b384241e3d39953/packages/cli/configuration/src/generators-yml/GeneratorsConfiguration.ts#L19)) and the [convertGeneratorsConfiguration](https://github.com/fern-api/fern/blob/5673d57158ee7581ab7632cb1b384241e3d39953/packages/cli/configuration-loader/src/generators-yml/convertGeneratorsConfiguration.ts#L44) mapper.

There's a chance that this ends up being more boilerplate without much value though -- in that case, we can easily drop this, but I'd rather start with it upfront to see what we can do to consolidate _all_ of the `fern.yml` validation upfront so that all of the downstream components are guaranteed to act with a fully hydrated, valid config.

Thoughts?

packages/cli/yaml/loader/src/ValidationIssue.ts
amckinney: That's a great question -- I think that attribute will make sense for _most_ errors, but I don't think it should belong in the `ValidationIssue` built around YAML. Instead, I think we can make this an error related to the `FernYmlConverter` defined in the `@fern-api/cli-v2` package.

I'll track this as an issue and eventually follow-up with a PR more focused on CLI errors in general so that each sub-command can get some nice handling for free (i.e. if we throw an error, call `process.exit(1)`, etc).

dsinghvi: that makes sense 👍

PR #11619, opened 2026-01-20
fix(python): respect pyproject_python_version in CI workflow and classifiers

tjb9dc:
## Description

Refs customer support thread - fixes issue where `pyproject_python_version` configuration was being ignored in generated CI workflows and pyproject.toml classifiers.

**Link to Devin run**: https://app.devin.ai/sessions/90140afe6afd4e4391022f989c2405b6
**Requested by**: @tjb9dc

## Problem

When users set `pyproject_python_version: ">=3.9,<3.14"` in their generator config, the generated CI workflow still used Python 3.9 (hardcoded) and the pyproject.toml classifiers still included Python 3.8 regardless of the configured constraint. This caused CI failures when dependencies didn't support the hardcoded version.

## Changes Made

- Added `parse_python_version_constraint()` function in `pyproject_toml.py` to parse version constraints and extract:
  - Minimum Python version (for CI workflows)
  - List of supported versions (for classifiers)
- Supports constraint formats: `>=3.9,<3.14`, `^3.8`, `~3.9`, `>=3.9`, `3.9`
- Updated `_get_github_workflow()` and `_get_github_workflow_legacy()` to use the parsed minimum version
- Updated `PoetryBlock` to generate classifiers dynamically based on the version range
- Added version 4.48.2 changelog entry

## Human Review Checklist

- [ ] Verify the version constraint parsing logic handles edge cases correctly (e.g., `>3.9` vs `>=3.9` are treated the same, malformed inputs default to 3.8)
- [ ] Note: `ALL_PYTHON_VERSIONS` is hardcoded to 3.8-3.13 and will need updating when Python 3.14 releases
- [ ] Verify the f-string conversion in workflow YAML doesn't break existing functionality
- [ ] No unit tests added for `parse_python_version_constraint()` - consider if tests should be added
- [ ] Default `min_python_version` is `"3.9"` (not `"3.8"`) because pytest-xdist requires Python 3.9

## Testing

- [x] Ran `poetry run ruff check --fix` and `poetry run ruff format` to fix linting issues
- [x] Ran `poetry run pre-commit run --all-files` to verify all pre-commit hooks pass
- [x] All CI checks passing (python-sdk, pydantic, fastapi seed tests)
- [ ] Unit tests added/updated (not added - parsing function could benefit from tests)

### DIFF
--- a/generators/python/src/fern_python/cli/abstract_generator.py
+++ b/generators/python/src/fern_python/cli/abstract_generator.py
@@ -310,7 +310,7 @@ class AbstractGenerator(ABC):
         logger.debug("Finished writing files for GitHub repository.")
 
     def _get_github_workflow_legacy(
-        self, output_mode: GithubOutputMode, write_unit_tests: bool, min_python_version: str = "3.8"
+        self, output_mode: GithubOutputMode, write_unit_tests: bool, min_python_version: str = "3.9"
     ) -> str:
         workflow_yaml = f"""name: ci
 on: [push]
@@ -395,7 +395,7 @@ jobs:
         return workflow_yaml
 
     def _get_github_workflow(
-        self, output_mode: GithubOutputMode, write_unit_tests: bool, min_python_version: str = "3.8"
+        self, output_mode: GithubOutputMode, write_unit_tests: bool, min_python_version: str = "3.9"
     ) -> str:
         # new workflow that supports automated OIDC-attestation signing and publishing to PyPI
         workflow_yaml = f"""name: ci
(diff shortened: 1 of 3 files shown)


generators/python/src/fern_python/cli/abstract_generator.py
```diff
@@ -304,8 +309,10 @@ def _write_files_for_github_repo(
         project.add_source_file("tests/custom/test_client.py", client_test_content)
         logger.debug("Finished writing files for GitHub repository.")
 
-    def _get_github_workflow_legacy(self, output_mode: GithubOutputMode, write_unit_tests: bool) -> str:
-        workflow_yaml = """name: ci
+    def _get_github_workflow_legacy(
+        self, output_mode: GithubOutputMode, write_unit_tests: bool, min_python_version: str = "3.9"
```
tjb9dc: why is this defaulted to 3.9 but the other is defaulted to 3.8? Furthermore, do we need defaults in both places?

tjb9dc: Actually I remember why. pytest-xdist requires 3.9. Change it back.

PR #11638, opened 2026-01-21
fix(cli): handle duplicate summary collision in OpenAPI examples

stephencme:
## Description

Fixes a bug where multiple OpenAPI examples with the same `summary` value cause key collisions. When examples are stored using the summary as the key, JavaScript object deduplication causes one example to be silently lost.

## Changes Made

- Added `getIdForExample` helper function in `AbstractMediaTypeObjectConverter.ts` to determine unique example identifiers
- Added `usedExampleNames` Set to track which example names have been used
- When a summary would cause a collision, the code now disambiguates using `${summary} (${key})` format (e.g., "Pine Tree (pine2)")
- Edge case: if the disambiguated name also collides, falls back to the original key
- Integrated with recursive `$ref` resolution - collision detection now runs on resolved examples
- Added test case with two inline examples having the same "Pine Tree" summary to validate the fix
- Added unit tests for `getIdForExample` collision handling logic
- Updated CLI versions.yml with version 3.47.7

## Testing

- [x] Unit tests added for `getIdForExample` helper function
- [x] Snapshot tests updated to include new `/trees/pines` endpoint with duplicate summary examples
- [x] Lint checks pass

- Refactored collision detection logic into a dedicated `getIdForExample` helper function per PR review feedback
- Added unit tests for `getIdForExample` covering: no summary, summary with no collision, summary collision, double collision, and multiple sequential collisions
- Merged with main to integrate with recursive `$ref` resolution feature - collision detection now runs on resolved examples
- Bumped version to 3.47.7 since main already contains 3.47.6

---

Link to Devin run: https://app.devin.ai/sessions/8f4e17a2d39943bca190f55ffb88b5a4
Requested by: @stephencme


packages/cli/api-importers/v3-importer-commons/src/converters/abstract/AbstractMediaTypeObjectConverter.ts
```diff
@@ -137,9 +137,29 @@ export abstract class AbstractMediaTypeObjectConverter extends AbstractConverter
...
+                    exampleName = summary;
+                } else {
+                    // Collision detected - disambiguate with key
+                    const disambiguatedName = `${summary} (${key})`;
+                    // If disambiguated name also collides (edge case), fall back to just the key
+                    exampleName = usedExampleNames.has(disambiguatedName) ? key : disambiguatedName;
+                }
+            } else {
+                exampleName = key;
+            }
```
dsinghvi: can you move this into a helper function like `getIdForExample` ?

dsinghvi (approved):

packages/cli/api-importers/v3-importer-commons/src/converters/abstract/AbstractMediaTypeObjectConverter.ts
```diff
@@ -117,6 +117,32 @@ export abstract class AbstractMediaTypeObjectConverter extends AbstractConverter
...
+            if (!usedExampleNames.has(summary)) {
+                return summary;
+            }
+            // Collision detected - disambiguate with key
+            const disambiguatedName = `${summary} (${key})`;
+            // If disambiguated name also collides (edge case), fall back to just the key
+            return usedExampleNames.has(disambiguatedName) ? key : disambiguatedName;
+        }
+        return key;
+    }
```
dsinghvi, line 144: ideally this function is now unit tested (since the bug was here)

PR #11659, opened 2026-01-21
feat(python): add sse_event_in_data_payload config option for SSE event parsing

aditya-arolkar-swe:
## Description

Refs: Slack thread request from @aditya-arolkar-swe

Adds a new `sse_event_in_data_payload` config option to the Python SDK generator that controls how SSE events are parsed.

**Link to Devin run**: https://app.devin.ai/sessions/e0bff2772f0f4bb7a123722f298d133a

## Changes Made

- Added `sse_event_in_data_payload` config option to `SDKCustomConfig` (default: `true`)
- Modified SSE endpoint code generation in `endpoint_response_code_writer.py`:
  - When `true` (default): Uses `_sse.json()` to parse just the data field (existing behavior)
  - When `false`: Uses `asdict(_sse)` from `dataclasses` to parse the whole event object, allowing access to all SSE fields (event, data, id, retry)
- Added version 4.49.0 changelog entry
- Added new test fixture `server-sent-events/sse-full-event` to verify the new config option

## Testing

- [x] Seed test passed for `server-sent-events` fixture with both configurations
- [x] Pre-commit hooks passed (ruff, formatting)

## Human Review Checklist

- [ ] Verify the generated code in `seed/python-sdk/server-sent-events/sse-full-event/src/seed/completions/raw_client.py` shows correct `asdict(_sse)` usage (line 65, 136)
- [ ] Verify the import `from dataclasses import asdict` is correctly added (line 5)
- [ ] Confirm default `true` maintains backward compatibility for existing users


generators/python/src/fern_python/generators/sdk/custom_config.py
```diff
@@ -127,6 +127,10 @@ class SDKCustomConfig(pydantic.BaseModel):
 
     custom_pager_name: Optional[str] = None
 
+    # When true (default), SSE events are parsed from just the data field using _sse.json().
+    # When false, SSE events are parsed from the whole event object using asdict(_sse).
+    sse_event_in_data_payload: bool = True
```
tjb9dc, line 132: I really think this should live on the IR and be read at generation time, not runtime. What about the case where an API has events with both styles of event? (likely to happen with Anduril)

tjb9dc: To populate the IR, we could:

1. Add this option to the SSE extension
2. Once we support OAS 3.2, infer this based on the shape of the event spec

PR #11665, opened 2026-01-22
fix(cli): include name field in webhook example FDR conversion

stephencme:
## Description

Refs: https://github.com/fern-api/fern/pull/11656

Fixes webhook examples missing the `name` field during IR-to-FDR conversion. Previously, only the `payload` was included in webhook examples, causing the frontend to always display "Example 1", "Example 2", etc. instead of user-specified names.

**Link to Devin run:** https://app.devin.ai/sessions/ca54bf14d45f44fb88be412e1a7dcdbb
**Requested by:** @stephencme

## Changes Made

- Added `name` field to webhook example FDR conversion for v2Examples (both userSpecified and autogenerated)
- Added `name` field to legacy webhook examples using `example.name?.originalName`

## Testing

- [x] Lint checks pass
- [ ] Unit tests added/updated (no new tests - existing tests should still pass)

## Review Checklist

- [ ] Verify the FDR `ExampleWebhookPayload` schema supports the `name?: string` field
- [ ] Confirm that showing names for ALL webhook examples (including autogenerated) is the desired behavior - note this differs from endpoints where autogenerated names are only shown when disambiguation is needed
- [ ] Check if the fallback behavior is correct: v2Examples fall back to the example key, legacy examples have no fallback (returns `undefined` if `example.name` is null)


stephencme:
Duplicate of https://github.com/fern-api/fern/pull/11653

PR #11669, opened 2026-01-22
fix(csharp): bugs around defaults/`Optional<>` wrapper interaction

patrickthornton:
## Description
Fixes a bug where a `using XYZ.Core` would be missing in interface files where it would be needed.
Fixes a bug where certain types that needed a `.Value` to be parsed into a string, like `Optional<DateTime>`, wouldn't get one.
Fixes a bug where `??` would be used directly on an `Optional<>` - uses `.GetValueOrDefault` in this case.
Fixes a bug where the `Action` references in the raw response templates would get name collision occasionally (just clarifies to `System.Action`).
Fixes a bug where `Optional<>` would occasionally wrap twice.

## Testing
The `request-parameters:with-defaults` fixture now tests both defaults and the optional wrapper; tested that the other optional-wrapper fixtures aren't changed by this.


generators/csharp/sdk/src/endpoint/request/WrappedEndpointRequest.ts
```diff
@@ -376,15 +376,13 @@ export class WrappedEndpointRequest extends EndpointRequest {
...
+        // - When experimental flag is enabled: add .Value for optional<T>, nullable<T>, or Optional<T?> where T is a struct
         // - When experimental flag is disabled (legacy): add .Value for optional structs (which become T?)
         // Note: strings are reference types and don't need .Value (string? doesn't have .Value)
         const needsDotValue =
             (allowOptionals ?? true) &&
             isStruct &&
-            (this.context.generation.settings.enableExplicitNullableOptional
-                ? (isOptional && isNullable) || isNullable
-                : isOptional);
+            (this.context.generation.settings.enableExplicitNullableOptional ? isOptional || isNullable : isOptional);
```
patrickthornton: this one might warrant a double check

generators/csharp/base/src/asIs/WithRawResponseTask.Template.cs
```diff
@@ -73,12 +73,12 @@ internal Awaiter(TaskAwaiter<WithRawResponse<T>> awaiter)
         /// <summary>
         /// Schedules the continuation action.
         /// </summary>
-        public void OnCompleted(Action continuation) => _awaiter.OnCompleted(continuation);
+        public void OnCompleted(System.Action continuation) => _awaiter.OnCompleted(continuation);
```
Swimburger: When using `System`, you'll need to prefix with `global::` to avoid name collisions with `System` namespaces and types

Swimburger (changes requested):

generators/csharp/base/src/asIs/WithRawResponseTask.Template.cs
patrickthornton: oops, independently fixed this. the `csharp-system-collision` fixture did its job

Swimburger (approved):

PR #11762, opened 2026-01-26
fix(java): resolve path parameter 'key' conflict with lambda variable

tjb9dc:
## Description

Refs: Customer reported Java SDK compilation failure when path parameters are named `key`

When an endpoint has a path parameter named `key`, the generated code fails to compile because the lambda variable in `requestOptions.getQueryParameters().forEach((key, value) -> {...})` shadows the path parameter.

**Link to Devin run**: https://app.devin.ai/sessions/0d6147e0fa0f4c3f89cea590c4eae5c7
**Requested by**: @tjb9dc

## Changes Made

- Renamed lambda variables from `(key, value)` to `(_key, _value)` in `HttpUrlBuilder.java` to avoid shadowing path parameters
- Added minimal OpenAPI spec fixture `java-path-param-key-conflict` with a single `POST /{key}/{value}` endpoint that tests both problematic parameter names
- Added version 3.34.4 changelog entry in `versions.yml`

## Bug Reproduction

The generated code was producing this compilation error:
```
error: variable key is already defined in method getItem(String,...)
    requestOptions.getQueryParameters().forEach((key, value) -> {
                                                 ^
```

## Testing

- [x] Seed test confirms compilation failure before fix
- [x] Seed test passes after fix (`1/1 test cases passed`)
- [x] 3-commit structure verified: (1) fixture + buggy output, (2) generator fix, (3) fixed output

- Simplified test fixture per review feedback: now uses a single `POST /{key}/{value}` endpoint instead of multiple endpoints, testing both `_key` and `_value` lambda variable renames in one endpoint
- Updated ir-to-jsonschema and Swift SDK generator snapshots for the new fixture

## CI Status

The `java-sdk (leftovers)` and `seed-test-results (java-sdk)` checks are failing due to an unrelated compile error in `@fern-api/go-formatter`. All individual java-sdk test jobs pass, including the new `java-path-param-key-conflict` fixture.

## Human Review Checklist

- [ ] Verify lambda variable rename (`_key`, `_value`) doesn't conflict with any other generated code patterns
- [ ] Confirm changelog entry accurately describes the fix


tstanmay13 (approved):

test-definitions/fern/apis/java-path-param-key-conflict/openapi.yml
```diff
@@ -0,0 +1,79 @@
+openapi: 3.0.3
+info:
+  title: Path Parameter Key Conflict Test API
+  version: 1.0.0
+  description: Tests that path parameters named 'key' don't conflict with lambda variables in generated Java code
+paths:
```
tjb9dc, line 6: let's also add `value` as a path param. and instead we only need one endpoint:
`POST /{key}/{value}`

PR #11856, opened 2026-01-28
feat(python): dynamic ci python version and pypi classifiers

seymurkafkas:
Python SDK: Dynamic Python Version Support

Problem: Generated SDKs had hardcoded Python versions in CI workflows and PyPI classifiers, causing mismatches when users specified custom pyproject_python_version constraints.
Solution: CI Python versions and PyPI trove classifiers are now dynamically derived from the pyproject_python_version constraint.

Tested with multiple constraints (>=3.9,<3.14, >=3.10,<3.14, >=3.11,<3.14, >=3.12,<3.14) — everything works correctly. Also verified Poetry 1.5.1+ compatibility with Python 3.9-3.13.

### DIFF
--- a/generators/python/fastapi/Dockerfile
+++ b/generators/python/fastapi/Dockerfile
@@ -4,6 +4,7 @@ FROM python:3.11.13 AS base
 ENV PYTHONPATH=${PYTHONPATH}:${PWD}
 ENV _TYPER_STANDARD_TRACEBACK=1
 
+# Keep in sync with the poetry-core version in pyproject.toml
 RUN pip3 install poetry==1.8.5
 RUN poetry config virtualenvs.create false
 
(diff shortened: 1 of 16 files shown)


generators/python/fastapi/Dockerfile
```diff
@@ -4,6 +4,7 @@ FROM python:3.11.13 AS base
 ENV PYTHONPATH=${PYTHONPATH}:${PWD}
 ENV _TYPER_STANDARD_TRACEBACK=1
 
+# Keep in sync with the poetry-core version in pyproject.toml
```
tjb9dc, line 7: can we add some sort of pre-merge test that this is true?

tjb9dc (approved):

generators/python/fastapi/Dockerfile
seymurkafkas: Note that the docker build will fail if this constraint is not true, meaning it will be caught early on before a merge. Given that, should we still add the test case?

PR #12044, opened 2026-02-03
feat(internal): Add new cursor-based pagination locators to IR

tjb9dc:
## Problem
Some customers use cursor pagination schemes where the server provides the next cursor as an entire URL, or entire path, also encoding other state in the URL. This is sometimes referred to as HATEOS (Hypermedia as the Engine of Application State).

While we've gotten away with a custom pager for some users, (1) this won't work for large SDKs with heterogenous pagination schemes, and (2) we've seen enough of this that it's worth supporting first-class.

## Solution
This PR introduces the IR change to capture more information about what the generator should do with the cursor. Before, it needed to be a property on the request somewhere. Now, it is a union, where it may also be the full `uri` or the full `path` that the generated SDK should use verbatim.

There will be follow-on work for the CLI to populate this (likely by the `$requestUri` and `$requestPath` directives), and for each generator to support it.


packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
```diff
@@ -408,6 +408,8 @@ types:
       cursor: CursorPagination
       offset: OffsetPagination
       custom: CustomPagination
+      nextUri: NextUriPagination
+      nextPath: NextPathPagination
```
amckinney: nit: Should these be called `uri` and `path` given that the `next` prefix is part of the nested properties contained within each object?

amckinney (approved):

packages/ir-sdk/fern/apis/ir-types-latest/definition/http.yml
tjb9dc: ooh good idea, also would give us a more natural spot to add `lastUri` like for Payroc

PR #12060, opened 2026-02-04
fix(cli): use yauzl for large ZIP files (>2 GiB)

tstanmay13:
## Description

Fixes extraction of large generated ZIP files (>2 GiB) that were failing with `ERR_FS_FILE_TOO_LARGE` errors.

**Link to Devin run**: https://app.devin.ai/sessions/1371d60945fe49bb9af12141449a7a8e
**Requested by**: @tstanmay13

## Changes Made

- Replaced the `decompress` library with `yauzl` for ZIP extraction
- Removed `decompress` and `@types/decompress` dependencies
- Added `yauzl` (^3.2.0) and `@types/yauzl` dependencies
- Added new `extractZipToDirectory` helper function that streams ZIP entries individually
- Updated `downloadZipForTask` and `downloadAndExtractZipToDirectory` to use the new extraction function
- Added changelog entry in `versions.yml` for version 3.64.3
- Added unit tests for `extractZipToDirectory` function
- Added `yazl` and `@types/yazl` as dev dependencies for creating test zips programmatically

**Root cause**: Node.js has a hard 2 GiB limit for reading files into buffers. The `decompress` library reads the entire ZIP file into memory, which fails for large SDK outputs. The user was seeing this error with a 3.24 GiB Python SDK output.

**Solution**: `yauzl` streams ZIP entries individually instead of loading the entire file into memory, bypassing the 2 GiB limit. It uses pure JavaScript for CRC32 validation and is OS-agnostic (works on Windows, macOS, and Linux).

- Reverted from `yauzl-promise` to callback-based `yauzl` because `yauzl-promise` depends on `@node-rs/crc32` (native Rust binding) which breaks CLI bundling with esbuild
- Added unit tests covering: basic file extraction, directory extraction, nested structures, implicit directory creation, and empty zips

## Testing

- [x] Lint checks pass
- [x] Unit tests added for `extractZipToDirectory`
- [ ] Manual testing with large ZIP file needed

## Human Review Checklist

- [ ] **Changelog inconsistency**: The `versions.yml` entry mentions `yauzl-promise` but the code uses `yauzl` - should be updated to say `yauzl`
- [ ] **Path validation**: The implementation doesn't explicitly validate for zip slip attacks (path traversal) - verify this is acceptable since ZIPs come from trusted S3 source
- [ ] **Test coverage**: Unit tests cover happy paths but not error cases (corrupted zips, permission errors)


tstanmay13:
Can you use a OS agnostic dep like yauzl instead

packages/cli/generation/remote-generation/remote-workspace-runner/src/RemoteTaskHandler.ts
```diff
@@ -390,5 +392,56 @@ async function downloadAndExtractZipToDirectory({
     const outputZipPath = path.join(tmpDir.path, "output.zip");
     await pipeline(request.data, createWriteStream(outputZipPath));
 
-    await decompress(outputZipPath, outputPath);
+    // Use yauzl for OS-agnostic ZIP extraction that handles files larger than 2 GiB
+    // yauzl streams entries individually instead of loading the entire file into memory
```
amckinney: I'd remove these comments.

packages/cli/generation/remote-generation/remote-workspace-runner/src/RemoteTaskHandler.ts
```diff
@@ -237,7 +237,9 @@ async function downloadZipForTask({
     // Force remove the directory to handle read-only files (e.g., .git/objects)
     await forceRemoveDirectory(absolutePathToLocalOutput);
     await mkdir(absolutePathToLocalOutput, { recursive: true });
-    await decompress(outputZipPath, absolutePathToLocalOutput);
+    // Use yauzl for OS-agnostic ZIP extraction that handles files larger than 2 GiB
+    // yauzl streams entries individually instead of loading the entire file into memory
```
amckinney: Let's remove this comment too.

packages/cli/generation/remote-generation/remote-workspace-runner/src/RemoteTaskHandler.ts
```diff
@@ -390,5 +392,56 @@ async function downloadAndExtractZipToDirectory({
...
+            }
+            if (!zipfile) {
+                reject(new Error("Failed to open zip file"));
+                return;
+            }
+
+            zipfile.readEntry();
+            zipfile.on("entry", (entry) => {
+                const outputPath = path.join(outputDir, entry.fileName);
+                if (entry.fileName.endsWith("/")) {
```
amckinney, line 410: Is this the right way to confirm a directory entry w/ this API? Is there not a more clear signal?

amckinney (approved):

fern-support:
would simpler logic like this work import { open } from "yauzl-promise";
import { mkdir } from "fs/promises";
import { createWriteStream } from "fs";
import path from "path";

async function extractZipToDirectory(
  zipPath: string,
  outputDir: string
): Promise<void> {
  const zip = await open(zipPath);

  for await (const entry of zip) {
    const outputPath = path.join(outputDir, entry.filename);

    if (entry.filename.endsWith("/")) {
      await mkdir(outputPath, { recursive: true });
      continue;
    }

    await mkdir(path.dirname(outputPath), { recursive: true });

    const readStream = await entry.openReadStream();
    await new Promise<void>((resolve, reject) => {
      readStream
        .pipe(createWriteStream(outputPath))
        .on("finish", resolve)
        .on("error", reject);
    });
  }
} 

can you try it and make sure the tests still succeed.

PR #12067, opened 2026-02-04
feat(cli): improve .fernignore to prevent generation of new files

aditya-arolkar-swe:
## Description

Refs: User request from @aditya-arolkar-swe

Improves `.fernignore` to prevent generation of new files, not just prevent modification or deletion of existing files during SDK regeneration. This allows users to exclude files and directories from being created in the first place.

**Link to Devin run:** https://app.devin.ai/sessions/afdef0b07370432ea61a1a91874209d0

## Changes Made

- Added `git clean -fd -- ...fernIgnorePaths` command after `git reset` in `LocalTaskHandler.ts` to remove newly generated files that match `.fernignore` patterns
- This simple approach leverages git's existing pathspec matching to clean up files that shouldn't have been generated
- Added test case "Prevent initial generation of files listed in .fernignore"

- **Simplified implementation**: Per feedback, replaced the complex file filtering approach (copyDirectoryWithExclusions, shouldExcludeFile, etc.) with a simple `git clean` command that removes files matching fernignore patterns after they're copied
- **Removed changes to ClonedRepository.ts and GitHub.ts**: The simpler approach only requires changes to LocalTaskHandler.ts
- **No new dependencies**: No longer uses minimatch or any additional packages
- **Updated changelog wording**: Changed "initial generation" to "new files" and clarified that previously it prevented "modification or deletion" per review feedback

## Testing

- [x] Unit tests added/updated - Added test case "Prevent initial generation of files listed in .fernignore" 

## Human Review Checklist

- [ ] Verify `git clean -fd -- <paths>` correctly removes files/directories matching the fernignore patterns without affecting other files
- [ ] Note: This feature only applies when `.fernignore` exists in the output directory (the `copyGeneratedFilesWithFernIgnore*` code paths)
- [ ] Verify test correctly exercises the new behavior (that `src/Client.ts` and `src/core/` would be generated without the .fernignore exclusion)

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -2,7 +2,7 @@
 - version: 3.64.0
   changelogEntry:
     - summary: |
-        Improve `.fernignore` to prevent initial generation of files. Previously, `.fernignore` only prevented overwriting existing files during SDK regeneration.
+        Improve `.fernignore` to prevent generation of new files. Previously, `.fernignore` only prevented modification or deletion of existing files during SDK regeneration.
       type: feat
   createdAt: "2026-02-04"
   irVersion: 63
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.64.0
+  changelogEntry:
+    - summary: |
+        Improve `.fernignore` to prevent initial generation of files. Previously, `.fernignore` only prevented overwriting existing files during SDK regeneration.
+      type: feat
+  createdAt: "2026-02-04"
+  irVersion: 63
+
 - version: 3.63.0
   changelogEntry:
     - summary: |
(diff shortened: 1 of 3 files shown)


packages/cli/ete-tests/src/tests/generate/fernignore.test.ts
```diff
@@ -54,6 +58,39 @@ describe("fern generate --local", () => {
...
+        await expectPathDoesNotExist(absolutePathToCiYml);
+
+        // Run generation again to ensure it's stable
+        await runFernCli(["generate", "--local", "--keepDocker"], {
+            cwd: pathOfDirectory
+        });
+
+        // CI file should still not exist
+        await expectPathDoesNotExist(absolutePathToCiYml);
+    }, 360_000);
```
aditya-arolkar-swe: The test isn't good because it is checking if CI.yml gets generated which only happens when output is github but this test outputs to local-file-system. Let's test that another file and a folder that usually gets generated does not generate when fernignored. 

Update the .fernignore to this
```
src/client.ts
src/core/
```
and adjust the relevant logic appropriately.

packages/cli/cli/versions.yml
```diff
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.64.0
+  changelogEntry:
+    - summary: |
+        Improve `.fernignore` to prevent initial generation of files. Previously, `.fernignore` only prevented overwriting existing files during SDK regeneration.
```
tjb9dc: i think more friendly to say “new” files than initial. Should also manage previously it prevented modification or deletion of files

tjb9dc (approved):

PR #12130, opened 2026-02-06
feat(cli): add libraries schema and `fern docs md generate` command stub

paarthfern:
Add support for library documentation generation in docs.yml:
- New `libraries` config section with input (git URL), output (path), and lang (python|cpp)
- New `- library: <name>` navigation item type for referencing libraries
- New `fern docs md generate [--library <name>]` command stub
- 41 tests for configuration parsing and command logic

This is the CLI schema and stub for library docs v2. Actual generation will be implemented in a follow-up iteration.

## Description
Linear ticket: Closes 

## Testing

tested locally by running something like `FERN_NO_VERSION_REDIRECTION=true node ~/fern/fern/packages/cli/cli/dist/local/cli.cjs docs md g
enerate --library python-sdk` 
on a sample spec of:
```
libraries:
  python-sdk:
    input:
      git: https://github.com/django/django
    output:
      path: ./static/django-docs
    lang: python
```


fern/apis/docs-yml/definition/docs.yml
```diff
@@ -151,6 +151,23 @@ types:
...
+          Configuration for library documentation generation. Each library entry defines
+          a source repository and output location for generated MDX documentation.
+
+          Example:
+          ```yaml
+          libraries:
+            my-sdk:
+              input:
+                git: https://github.com/acme/sdk-python
+                subpath: src/sdk
```
amckinney, line 166: We should make sure whatever naming we have is consistent with the other sub-directory functionality (i.e. generate an SDK into a sub-directory of a repository). @patrickthornton thoughts?

fern/apis/docs-yml/definition/docs.yml
```diff
@@ -1167,12 +1185,13 @@ types:
     docs: |
       Configuration for generating Python library documentation from source code.
       The docs are generated from a GitHub repository and merged into the navigation.
+    availability: deprecated
```
amckinney, line 1188: Not actionable here, but do we have a plan for just removing this config entirely later (rather than hanging onto it as a deprecated, unused shape?

fern/apis/docs-yml/definition/docs.yml
```diff
@@ -1184,6 +1203,74 @@ types:
...
+      git:
+        type: string
+        docs: |
+          GitHub URL to the repository containing the library source code.
+          Example: https://github.com/django/django
+      subpath:
+        type: optional<string>
+        docs: |
+          Optional path within the repository to the library source.
+          Example: django/core
```
amckinney, line 1236: I think we need to support local paths too, right? Even if we only need to define a single input type to start, it would help to outline whatever [discriminated] union we plan to use.

fern/apis/docs-yml/definition/docs.yml
```diff
@@ -1184,6 +1203,74 @@ types:
...
+        docs: |
+          The name of the library to reference (must match a key in the `libraries` section).
+      title:
+        type: optional<string>
+        docs: |
+          Override the display title for this library reference.
+      slug:
+        type: optional<string>
+        docs: |
+          Override the URL slug for this library reference.
```
amckinney, line 1288: To confirm, the plan is to include the library within our standard Fern documentation as a navigation item? I was under the impression this would be a totally independent site, but it makes sense to navigate to it from the main documentation.

packages/cli/cli/src/cli.ts
```diff
@@ -1542,10 +1543,46 @@ function addDocsCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
...
         return yargs;
     });
 }
 
+function addDocsMdCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
+    cli.command("md", "[Beta] Commands for generating markdown documentation from library source code", (yargs) => {
+        addDocsMdGenerateCommand(yargs, cliContext);
+        return yargs;
+    });
+}
```
amckinney, line 1558: Would you mind hiding this command to start just so that it goes undiscovered and more clearly communicates that it's in beta?

fern/apis/docs-yml/definition/docs.yml
paarthfern: Yeah, we're consistent with the subpath convention in the sdk world

fern/apis/docs-yml/definition/docs.yml
paarthfern: Yeah, I'll track this, will remove this down the line

amckinney (commented):
Overall looks good, just some comments on the configuration/naming but once we nail that down good to merge.

fern/apis/docs-yml/definition/docs.yml
paarthfern: Yes that was the plan to begin with - sorry wasn't clear earlier

amckinney (approved):
lgtm, thanks for updating the config 👏

fern/apis/docs-yml/definition/docs.yml
paarthfern: done - but an undiscriminated union seems cleaner here

paarthfern:
> lgtm, thanks for updating the config 👏

ofc! thanks for reviewing

fern/apis/docs-yml/definition/docs.yml
patrickthornton: naming here is a bit of a crapshoot right now; under `output` to local-file-system it's just called `path`, but there are some config options that let you customize a `package-path` that does a similar thing. i would say `path` aligns more with current convention

paarthfern: guess I was a little mistaken, let me fix this in a future PR

PR #12149, opened 2026-02-07
feat(cli): refactor versioning and release procedure

aditya-arolkar-swe:
## Description

Unified release management system for any software component in the Fern monorepo. Provides interactive setup, automatic semantic versioning, and unified CI validation.

Link to Devin Session: https://app.devin.ai/sessions/11f70c1030c245198ac2daf571346c05
Follow-up changes: https://app.devin.ai/sessions/a99bcbc335bd4ce58ad04279900375c2
Requested by: @aditya-arolkar-swe

### Single Command Interface
```bash
# Prepare release for configured software, or configure new software with versioning
pnpm release <software>
```

### Key Features
- **Generalized**: Works for any software in the monorepo, not just CLI
- **Interactive Setup**: Prompts for configuration, creates changelog structure
- **Unified CI**: Single workflow validates all configured software dynamically
- **Auto-Severity**: Version bumps derived from change type (no manual severity field)
- **Location-Independent**: Schema paths calculated dynamically for any directory structure
- **Branch-Agnostic Releases**: Release script detects and pushes to the current branch; CI workflow controls the branch via `actions/checkout` (defaults to `main`, supports "break glass" releases from other branches via `workflow_dispatch`)

### Version Bump Logic
Change type automatically determines version bump:
- `fix`, `chore` → **patch** (1.0.0 → 1.0.1)
- `feat`, `internal` → **minor** (1.0.0 → 1.1.0)
- `break` → **major** (1.0.0 → 2.0.0)

### Setting Up New Software
```bash
pnpm release python-sdk
# Prompts for:
# - Full software name: Fern Python SDK Generator
# - Path to versions.yml (relative to repo root): generators/python/versions.yml
# - Changelog folder (default: generators/python/changes):
# - Software directory (default: generators/python):
#
# Creates:
# - Entry in release-config.json with key "python-sdk" and name "Fern Python SDK Generator"
# - Changelog structure with .template.yml and .gitkeep
# - Local changes only (use --create-pr to auto-create PR)
```

### Preparing Releases
```bash
pnpm release cli
# 1. Stashes uncommitted changes
# 2. Pulls latest for the current branch (branch controlled by caller / CI workflow)
# 3. Reads & validates unreleased changelog files against schema
# 4. Calculates next version from change types
# 5. Updates versions.yml
# 6. Moves changelog files to versioned directory
# 7. Commits changes
# 8. Checks for conflicts (auto-rollback if detected)
# 9. Prompts for confirmation
# 10. Pushes to the current branch
```

### CI Validation
The unified workflow automatically:
1. Reads `release-config.json` to discover all configured software
2. Checks which software has changed files in the PR
3. Validates that a changelog entry exists in `changes/unreleased/`
4. Provides helpful error messages with format examples

## Changelog Entry Format

```yaml
# <software>/changes/unreleased/my-change.yml
# yaml-language-server: $schema=../path/to/fern-changes-yml.schema.json

- summary: |
    Description of your change
  type: feat  # Options: fix, chore, feat, internal, break
  # Severity auto-derived from type - no manual field needed
```

## ⚠️ Admin Permissions Needed
 - Authorization flow to allow `pnpm release` to push to main, perhaps through a scoped token (?)
 - Make `.github/workflows/validate-changelogs.yml` a required CI workflow

### Configuration Files
- **`release-config.json`** - Central registry of all software configured for releases (key = command identifier, name = full descriptive name)
- **`release-config.schema.json`** - JSON schema for config validation and IDE support
- **`fern-changes-yml.schema.json`** - Changelog entry schema (type-based severity)

### Core Scripts (all TypeScript)
- **`scripts/release.ts`** - Main entry point; routes to setup or release flow
- **`scripts/release-setup.ts`** - Interactive setup with dry-run default
- **`scripts/release-config.ts`** - Configuration utilities (load, save, getters)
- **`scripts/release-workflow.ts`** - Batch release script for CI (auto-answers "yes" to prompts)
- **`scripts/validate-changelogs.ts`** - PR changelog validation script

### CI Workflows
- **`.github/workflows/validate-changelogs.yml`** - Unified validation for all software; reads `release-config.json` dynamically
- **`.github/workflows/release-software.yml`** - Auto-triggers releases when changelog files change on main; uses commit-subject matching to avoid release loops

### Generated Files (per software)
- **`<software>/changes/unreleased/.template.yml`** - Template with dynamic schema path
- **`<software>/changes/unreleased/.gitkeep`** - Ensures directory tracking with helpful comment

### Updates Since Last Revision
- Added runtime schema validation for changelog entries with clear error messages
- Added exhaustive default case to `getSeverityFromType()` to catch invalid types at compile time
- Fixed order-of-operations bug in `release-setup.ts` - now checks for uncommitted changes **before** modifying files
- Fixed shell injection vulnerability in `createPullRequest()` - now uses `--body-file` for PR body instead of inline string
- Generalized `fern-changes-yml.schema.json` title/description (was CLI-specific, now generic)
- **Converted `release-all.js` and `validate-changelogs.js` to TypeScript** - now use TypeScript with `tsx` for better type safety
- **Refactored to use `release-config.ts` APIs** - both validation and release scripts now import shared configuration utilities instead of manually parsing JSON
- **Reverted RC release functionality** (commit `58991fe`) — removed RC release config, reverted schema and templates to simpler non-RC versions, while preserving improvements to `release-software.yml` (commit-subject-based loop detection, renamed batch script to `release-workflow.ts`)
- **Removed hardcoded `main` branch references in `release.ts`** — the release script now detects the current branch via `git rev-parse --abbrev-ref HEAD` and operates on it, rather than force-checking out `main`. This allows the CI workflow (`release-software.yml`) to control the branch, enabling "break glass" releases from non-main branches via `workflow_dispatch` in the future.

## Testing
- [ ] Manual testing of release flow needed
- [ ] Verify ACTIONS_PAT secret is configured with main branch bypass access
- [ ] Test that runtime validation catches malformed changelog entries
- [ ] Verify `validate-changelogs.yml` runs successfully with the new Install step

## Review Checklist
- [ ] Verify race condition handling in release flow is acceptable (see [comment #17](https://github.com/fern-api/fern/pull/12149#issuecomment-3874311978))
- [ ] Confirm old workflows (`cli-release.yml`, `validate-changelog.yml`) should remain active or be deprecated
- [ ] Consider adding path filter to `validate-changelogs.yml` to avoid running on every PR (now includes full `pnpm install` via Install action)
- [ ] Verify `ACTIONS_PAT` secret is set up with appropriate permissions
- [ ] Review validation logic in `validateChangelogEntry()` — no automated tests exist for edge cases
- [ ] Confirm `release-workflow.ts` auto-answering "yes" is acceptable for CI context
- [ ] Verify `validate-changelogs.yml` Install step doesn't significantly slow down PR validation
- [ ] Verify `release-software.yml` commit-subject regex (`^chore\(.+\):\ release\ .+`) correctly matches release commits and avoids false positives
- [ ] Confirm branch-agnostic release behavior in `release.ts` is safe — running `pnpm release` locally from a feature branch would attempt to release from that branch rather than `main`


amckinney:
```
# Developer B creates: packages/cli/cli/changes/unreleased/fix-bug-b.yml
- summary: |
    Fix bug B
  type: fix
  severity: patch
```

The `type` implies the `severity` -- we shouldn't need the user to specify both, otherwise it'll be too easy to mess things up (i.e. a `feat` with `patch`).

.github/workflows/auto-release-cli.yml
```diff
@@ -0,0 +1,169 @@
...
+  TURBO_TEAM: "buildwithfern"
+
+jobs:
+  prepare-release:
+    # Run if: PR merged with !autorelease OR publish workflow completed successfully
+    if: |
+      github.repository == 'fern-api/fern' && (
+        (github.event_name == 'pull_request' &&
+         github.event.pull_request.merged == true &&
+         contains(github.event.pull_request.title, '!autorelease')) ||
```
amckinney: I'm not sure if including in the title is the right UX here -- thoughts on using a comment checkbox that users can toggle (similar to our `seed` selector [here](https://github.com/fern-api/fern/pull/12149#issuecomment-3863441414))? For reference, [here](https://github.com/fern-api/fern/blob/1123940a9554095176f7dead5e9187a5f37e7dd1/.github/workflows/seed-pr-comment.yml#L1) is the workflow.

.github/workflows/auto-release-cli.yml
```diff
@@ -0,0 +1,169 @@
...
+         github.event.workflow_run.conclusion == 'success')
+      )
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          ref: main
+          fetch-depth: 0
+          # Uses GITHUB_TOKEN by default (automatically provided)
```
amckinney: nit: This is implied; we can remove this or we could make it explicit like we do [here](https://github.com/fern-api/fern/blob/1123940a9554095176f7dead5e9187a5f37e7dd1/.github/workflows/seed-pr-comment.yml#L215C11-L215C45).

.github/workflows/auto-release-cli.yml
```diff
@@ -0,0 +1,169 @@
...
+      - name: Check for existing release PR
+        id: check_release_pr
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Get all open PRs and filter for release/cli-* pattern
+          OPEN_RELEASE_PRS=$(gh pr list \
+            --state open \
+            --json number,headRefName \
+            --jq '[.[] | select(.headRefName | startswith("release/cli-"))] | length')
```
amckinney: The release PRs are automatically merged, right? Is this purely to precent concurrent release PRs beyond the concurrency control already configured in this workflow? If not, can you speak more to this?

.github/workflows/auto-release-cli.yml
```diff
@@ -0,0 +1,169 @@
...
+          echo "" >> $GITHUB_STEP_SUMMARY
+          echo "📋 View existing release PRs:" >> $GITHUB_STEP_SUMMARY
+          gh pr list --state open --json number,title,url,headRefName --jq '.[] | select(.headRefName | startswith("release/cli-")) | "- [#\(.number): \(.title)](\(.url))"' >> $GITHUB_STEP_SUMMARY
+
+      - name: Check for unreleased changes
+        if: steps.check_release_pr.outputs.has_release_pr == 'false'
+        id: check_unreleased
+        run: |
+          # Check for both .yml and .yaml files, excluding hidden files (starting with .)
+          if [ -z "$(find packages/cli/cli/changes/unreleased -maxdepth 1 -type f \( -name "*.yml" -o -name "*.yaml" \) ! -name ".*" 2>/dev/null)" ]; then
```
amckinney: We'll eventually want to use this for the SDK generators too (and whatever else is using the `versions.yml` approach). Do you think it's worth generalizing a bit upfront so that the path to the `unreleased` directory is configurable / reusable so it's simpler to make that change?

.github/workflows/auto-release-cli.yml
```diff
@@ -0,0 +1,169 @@
...
+          # Capture the version from versions.yml
+          VERSION=$(grep -m 1 'version:' packages/cli/cli/versions.yml | awk '{print $2}')
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+
+      - name: Commit and push branch
+        if: steps.check_release_pr.outputs.has_release_pr == 'false' && steps.check_unreleased.outputs.has_changes == 'true'
+        run: |
+          git add packages/cli/cli/versions.yml
+          git add packages/cli/cli/changes/
+          git commit -m "chore(cli): release ${{ steps.prepare_release.outputs.version }}"
```
amckinney: mega nit:
```suggestion
          git commit -m "chore(cli): Release ${{ steps.prepare_release.outputs.version }}"
```

.github/workflows/auto-release-cli.yml
```diff
@@ -0,0 +1,169 @@
...
+          git push origin "${{ steps.create_branch.outputs.branch_name }}"
+
+      - name: Create release PR
+        if: steps.check_release_pr.outputs.has_release_pr == 'false' && steps.check_unreleased.outputs.has_changes == 'true'
+        id: create_pr
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          PR_URL=$(gh pr create \
+            --title "chore(cli): release ${{ steps.prepare_release.outputs.version }}" \
```
amckinney:
```suggestion
            --title "chore(cli): Release ${{ steps.prepare_release.outputs.version }}" \
```

.github/workflows/enforce-versions-yml-protection.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Check if PR is from release branch
+        id: check_branch
+        run: |
+          if [[ "${{ github.head_ref }}" == release/* ]]; then
+            echo "is_release_branch=true" >> $GITHUB_OUTPUT
+          else
+            echo "is_release_branch=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Fail if versions.yml was modified
```
amckinney: This is awesome 👏  -- only concern is if we want to change the content of our release changelog later on, how would we do that?

.github/workflows/validate-changelog-entry.yml
```diff
@@ -0,0 +1,53 @@
...
+            echo "✅ No CLI changes detected"
+            exit 0
+          fi
+
+          echo "📝 CLI changes detected:"
+          echo "$CLI_CHANGES"
+          echo ""
+
+          # Check if any files in packages/cli/cli/changes/unreleased/ were added or modified
+          CHANGELOG_CHANGES=$(echo "$CHANGED_FILES" | grep '^packages/cli/cli/changes/unreleased/.*\.ya?ml$' || true)
```
amckinney: Similar to above w.r.t hard-coding the CLI's `unreleased` directory. This might apply elsewhere, too.

packages/cli/cli/changes/unreleased/.template.yml
```diff
@@ -0,0 +1,9 @@
+# Template for changelog entries
+# Copy this file and rename it to describe your change (e.g., add-auth-feature.yml)
+# yaml-language-server: $schema=../../../../../fern-changes-yml.schema.json
```
amckinney, line 3: Related to above w.r.t. how to resolve the JSON schema file from arbitrary `unreleased` directory locations. I suppose this is a one-time setup for each thing, so not a huge deal if we just hard-code it everywhere it's defined.

scripts/release-cli.ts
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env tsx
+// biome-ignore-all lint/suspicious/noConsole: CLI script requires console output for user feedback
+import { existsSync, mkdirSync, readdirSync, readFileSync, renameSync, writeFileSync } from "fs";
+import { join } from "path";
+import { parse as parseYaml, stringify as stringifyYaml } from "yaml";
+
+interface ChangelogEntry {
+    summary: string;
+    type: "fix" | "chore" | "feat" | "internal" | "break";
+    severity: "major" | "minor" | "patch";
```
amckinney: Based on the above, I would resolve this from `type`.

1. `fix` -> `patch`
2. `chore` -> `patch`
3. `feat` -> `minor`
4. `internal` -> `patch`
5. `break` -> `break`

Thoughts?

amckinney (commented):
Nice work, this is coming together! A handful of comments on the high-level UX, but most of the workflow implementation looks solid to me.

amckinney:
Discussed offline -- there's also an opportunity for race conditions in the current design.
1. A CLI release PR is opened (w/ a computed version of `1.5.1`).
2. Another commit is merged that would cause a minor release (i.e. `1.6.0`).
3. The release PR is merged and includes the unintended commit.

We can address this by adding some sort of synchronization in the release process. We previously got this for free by tying the commit to the release, but there are ways to achieve the same outcome with other strategies.

@aditya-arolkar-swe to follow-up with a revised solution that removes the race condition.

.github/workflows/release-software.yml
```diff
@@ -0,0 +1,88 @@
...
+          ref: main
+          fetch-depth: 0
+
+      - name: Skip if triggered by own release commit
+        id: check-author
+        run: |
+          COMMIT_AUTHOR_EMAIL=$(git log -1 --format='%ae')
+          echo "Last commit author email: $COMMIT_AUTHOR_EMAIL"
+          if [[ "$COMMIT_AUTHOR_EMAIL" == "github-actions[bot]@users.noreply.github.com" ]]; then
+            echo "Triggered by automated release commit — skipping to avoid loop."
```
amckinney, line 70: If we ever add other automation that commits with the same author email, then this would accurately represent whether or not its triggered by this release commit. I can foresee this becoming an issue, so I think it's worth making a mild adjustment upfront.

---

We might want to use "trailers" here, something like:
```sh
$ git commit -m "Release v1.2.3

Generated-By: fern-release-workflow"
```

Then, instead of checking the author name, we check whether or not it has the `Generated-By` trailer set to whatever ID this workflow is responsible for.

```sh
TRAILERS=$(git log -1 --format='%(trailers:key=Generated-By,valueonly)')
if [[ "$TRAILERS" == "fern-release-workflow" ]]; then
  echo "Skipping — triggered by our own release workflow."
  exit 0
fi
```

Thoughts?

.github/workflows/release-software.yml
```diff
@@ -0,0 +1,88 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Install
+        if: steps.check-author.outputs.skip != 'true'
+        uses: ./.github/actions/install
+
+      - name: Run releases for all configured software
+        if: steps.check-author.outputs.skip != 'true'
+        run: npx tsx scripts/release-all.ts ${{ inputs.software || 'all' }}
```
amckinney: nit: If `all` is an argument now, I'd name this script `scripts/release-workflow.ts`

scripts/release.ts
```diff
@@ -0,0 +1,449 @@
...
+            validateChangelogEntry(entries[i], file, i);
+        }
+
+        changes.push({ filename: file, entries });
+    }
+
+    return changes;
+}
+
+function determineNextVersion(currentVersion: string, changes: UnreleasedChange[]): string {
```
amckinney, line 133: Is there some way for me to create a release candidate? Can we add that as an option the "changes" config? Maybe something additional like:

```yaml
rc: true
```

scripts/release.ts
```diff
@@ -0,0 +1,449 @@
...
+    console.log(`📦 Moving unreleased files to ${nextVersion}/...`);
+    moveUnreleasedFiles(fullUnreleasedDir, versionDir, changes);
+    console.log(`   ✅ Files moved to ${getChangelogFolder(config)}/${nextVersion}/\n`);
+
+    // Commit changes
+    console.log("💾 Committing changes...");
+    try {
+        execSync(`git add ${config.versionsFile}`, { stdio: "inherit" });
+        execSync(`git add ${getChangelogFolder(config)}`, { stdio: "inherit" });
+        execSync(`git commit -m "chore(${softwareName}): release ${nextVersion}"`, { stdio: "inherit" });
```
amckinney, line 330: This is where we'd add those "trailers" if we move forward with that approach.

scripts/release.ts
```diff
@@ -0,0 +1,449 @@
...
+    // Checkout main and pull latest
+    console.log("🔄 Checking out main and pulling latest...");
+    try {
+        execSync("git checkout main", { stdio: "inherit" });
+        execSync("git pull origin main", { stdio: "inherit" });
+        console.log("   ✅ On main with latest changes\n");
+    } catch (error) {
+        console.error("❌ Error updating main branch:", error);
+        process.exit(1);
+    }
```
amckinney: **Important:** We shouldn't manually checkout the branch here -- we should control that from the workflow. That way it's theoretically possible for me to trigger a release from a particular branch if I need to break glass. We can configure the `release-software.yml` file so that it is triggered on commits to `main` and `actions/checkout` will do whatever we need it to do.

.github/workflows/release-software.yml
aditya-arolkar-swe: Issue with this is that the commit and message gets pushed straight to main, where we'd want the commit message to follow our linting standard of `type(software-scope): message`. 

Updated it to pattern match the type of commit this creates, so a `chore(<anything>): release <anything>` being the previous commit wouldn't trigger a loop.

scripts/release.ts
aditya-arolkar-swe: Good point, just added that top level config which will auto generate RC versions too.

scripts/release.ts
aditya-arolkar-swe: Hmm I guess I couldn't imagine a scenario where we want to release from a branch, but open to that 'break glass' possibility. Removed references to `main` in this file, and allowing the `release-software.yml` to specify the branch (can be done in `workflow_dispatch` in the future as needed).

amckinney (approved):
With the opt-in flow, this looks great. Excited to see how this goes.

PR #12162, opened 2026-02-09
fix(cli): add cycle detection to v1 example generation to prevent OOM for recursive schemas

iamnamananand996:
## Description

Refs: Reported by @iamnamananand996

Fixes a Node.js OOM crash when running `fern ir` on OpenAPI specs that contain self-referencing or cyclically-referencing schemas (e.g. `JsonschemaSchema` with ~14 fields pointing back to itself, or `BulkItemizedSchedule` → `BulkSchedule` → `BulkMultiScheduleConfiguration` → `BulkItemizedSchedule` cycles).

The v1 example generation code recursively expands type references without tracking visited types, causing exponential memory growth (14^N for a type with 14 self-refs). A modest spec (~600 KB, 231 schemas) would exhaust 4 GB of heap.

Link to Devin run: https://app.devin.ai/sessions/b3f73869667846edad15983ac994106a

## Changes Made

- Thread `visitedTypes?: Map<string, number>` through `generateTypeReferenceExample`, `generateTypeDeclarationExample`, and `generateContainerExample`
- In `generateTypeReferenceExample`, when a named type is encountered: check visit count → return failure if count ≥ 2 → increment → recurse → decrement (DFS backtracking with depth-limited recursion)
- **K=1 depth-limited recursion**: allows one level of recursion before cutting off, producing richer examples while maintaining O(N) complexity
- Updated `example-depth` snapshots in both `ir-generator-tests` and `v3-importer-tests`, and `simple-fhir` snapshot
- Added changelog entry as version `3.70.1`
- Added unit tests for cycle detection (5 test cases)

- Reverted K=2 back to K=1 after testing showed K=2 produced excessively large IR output (~443 MB for a 600 KB spec). K=1 is the safe choice.
- All snapshots regenerated for K=1 behavior.

## Testing

- [x] Unit tests added in `packages/commons/ir-utils/src/__test__/generateTypeReferenceExample.test.ts`:
  - **14-self-ref type** (simulates `JsonschemaSchema`): completes in <1s, recursive fields contain one level of nested objects
  - **All-required recursive properties**: correctly returns failure
  - **Triangle cycle** (A→B→C→A with optional back-edge): traverses one full cycle before cutting off
  - **Backtracking correctness**: same type in sibling branches both expand independently
  - **O(N) complexity**: time scales linearly with field count (5→50 fields)
- [x] Manual testing: ran `fern ir` against the reporter's config; 74 MB IR generated at ~484 MB peak heap (vs OOM at 4 GB before)
- [x] TypeScript compilation passes locally
- [x] Lint passes locally (`pnpm run check`)
- [x] `v3-importer-tests` and `ir-generator-tests` pass locally after snapshot regeneration

## ⚠️ Reviewer Checklist

1. **K=1 threshold**: The `count >= 2` check means "block on second visit" = "allow 1 revisit" = K=1. Verify this matches the intended behavior.
2. **Visit count tracking**: The increment uses `(count ?? 0) + 1` and decrement uses `(count ?? 1) - 1`. Verify this asymmetry doesn't cause issues.
3. **simple-fhir snapshot**: Verify the snapshot size is reasonable for K=1 (should be much smaller than the K=2 version).
4. **v2 example generation**: Only v1 (`packages/commons/ir-utils/src/examples/v1/`) is patched. Investigation confirmed v2 (`ExampleConverter` in `v3-importer-commons`) already has its own cycle detection via `seenRefs: Set<string>`.
5. **Unit tests use synthetic IR types**: Tests construct `TypeDeclaration` objects directly rather than going through the full OpenAPI→IR pipeline.
6. **test-ete CI failure**: This is unrelated to the PR changes (remote `fern-v2 sdk generate` service issue). The `test` check passes.


packages/commons/ir-utils/src/examples/v1/generateTypeReferenceExample.ts
```diff
@@ -42,14 +45,21 @@ export function generateTypeReferenceExample({
             if (typeDeclaration == null) {
                 return { type: "failure", message: `Failed to find type declaration with id ${typeReference.typeId}` };
             }
+            const visited = visitedTypes ?? new Set<string>();
+            if (visited.has(typeReference.typeId)) {
+                return { type: "failure", message: `Detected recursive type ${typeReference.typeId}` };
```
tjb9dc, line 51: What does the failure here cause in practice? Remember that it is perfectly legal to have (indirectly) self-referencing types

tjb9dc: should we try for one depth? So that the examples are interesting enough? Or will that still cause an explosion

iamnamananand996: lets have k=1 that should be enough to have nested example ??

tjb9dc (approved):

PR #12163, opened 2026-02-09
fix(cli): add cycle detection to v1 example generation to prevent OOM for recursive schemas

iamnamananand996:
## Description

Refs: Reported by @iamnamananand996

Fixes a Node.js OOM crash when running `fern ir` on OpenAPI specs containing self-referencing or cyclically-referencing schemas. The specific case involved `JsonschemaSchema` (~14 fields referencing itself), which caused exponential memory growth (14^N) during v1 example generation—a ~600 KB spec would exhaust 4 GB of heap.

Link to Devin run: https://app.devin.ai/sessions/b3f73869667846edad15983ac994106a

## Changes Made

- Thread `visitedTypes?: Map<string, number>` through `generateTypeReferenceExample`, `generateTypeDeclarationExample`, and `generateContainerExample`
- In `generateTypeReferenceExample`, when a named type is encountered: check visit count → return failure if count ≥ 2 → increment → recurse → decrement (DFS backtracking with depth-limited recursion)
- **K=1 depth-limited recursion**: allows one level of recursion before cutting off, producing richer examples while maintaining O(N) complexity
- Updated `example-depth` and `simple-fhir` snapshots
- Added changelog entry as version `3.70.1`
- Added unit tests for cycle detection covering 5 scenarios (see Testing below)

Changed from K=0 (immediate cutoff using `Set<string>`) to **K=1** (one level of recursion using `Map<string, number>` with `count >= 2` threshold):

- `TreeNode` example now produces `{"value": "value", "left": {"value": "value"}, "right": {"value": "value"}}` instead of just `{"value": "value"}`
- Triangle cycles (A→B→C→A) now traverse one full cycle before cutting off
- `simple-fhir` snapshot grew significantly (~10K → ~68K lines) due to richer recursive examples
- Unit tests updated to verify K=1 behavior

## Testing

- [x] Manual testing: ran `fern ir` against the reporter's config; 77 MB IR generated at ~484 MB peak heap (vs OOM at 4 GB before)
- [x] Snapshots updated for `example-depth` and `simple-fhir` fixtures
- [x] Unit tests added in `packages/commons/ir-utils/src/__test__/generateTypeReferenceExample.test.ts`:
  - **14-self-ref type** (simulates `JsonschemaSchema`): completes in <1s, recursive fields now contain one level of nested objects
  - **All-required recursive properties**: correctly returns failure (can't generate a valid example)
  - **Triangle cycle** (A→B→C→A with optional back-edge): generates example traversing one full cycle, then cuts off
  - **Backtracking correctness**: same type in sibling branches both expand independently (verifies decrement works)
  - **O(N) complexity**: time scales linearly with field count (5→50 fields), not exponentially
- [x] TypeScript compilation passes locally
- [x] Lint passes locally (`pnpm run check`)

## ⚠️ Reviewer Checklist

1. **K=1 produces larger snapshots**: The `simple-fhir.json` snapshot grew from ~10K to ~68K lines. This is expected with K=1 but worth verifying the growth is bounded and not exponential for pathological cases.
2. **Visit count tracking**: The increment uses `(count ?? 0) + 1` and decrement uses `(count ?? 1) - 1`. Verify this asymmetry doesn't cause issues (it shouldn't—the `?? 1` in decrement is a defensive fallback).
3. **v2 example generation**: Only the v1 path (`packages/commons/ir-utils/src/examples/v1/`) is patched. Investigation confirmed v2 (`ExampleConverter` in `v3-importer-commons`) already has its own cycle detection via `seenRefs: Set<string>` with copy-on-branch semantics.
4. **Unit tests use synthetic IR types**: The tests construct `TypeDeclaration` objects directly rather than going through the full OpenAPI→IR pipeline. Real-world edge cases (alias chains, unions containing recursive types) may not be fully covered.
5. **No seed tests selected yet**: Seed tests have not been run via the PR checkbox selector. Consider running at least TypeScript and Python seeds to validate generated SDK output is acceptable with the new richer recursive examples.


iamnamananand996:
duplicated

PR #12222, opened 2026-02-10
feat(java): add SSE event-level discrimination support

tstanmay13:
## Description
Linear ticket: Closes [FER-8560: \[Java, Bloomberg\] SSE discrimination](https://linear.app/buildwithfern/issue/FER-8560/java-bloomberg-sse-discrimination)

Add support for SSE event-level discrimination in Java SDK generator, mirroring Python. This enables proper parsing of SSE streams where the discriminator is at the SSE level can be at the event field rather than inside the JSON data payload.

The new `fromSseWithEventDiscrimination()` method is ONLY used when:
1. The SSE payload is a discriminated union
2. AND the discriminator property is an SSE envelope field (event, id, retry, data)

All other SSE endpoints continue using `fromSse()` unchanged, ensuring no regressions for existing customers.

## Changes Made

- 
- 

## Testing


generators/java/generator-utils/src/main/java/com/fern/java/generators/SseEventParserGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+import java.nio.charset.StandardCharsets;
+
+public final class SseEventParserGenerator extends AbstractFileGenerator {
+
+    public SseEventParserGenerator(AbstractGeneratorContext<?, ?> generatorContext) {
+        super(generatorContext.getPoetClassNameFactory().getSseEventParserClassName(), generatorContext);
+    }
+
+    @Override
+    public GeneratedResourcesJavaFile generateFile() {
```
tjb9dc, line 32: this is a huge nit but I'm surprised this is our pattern for as-is files. Could be a lot less code.

tstanmay13: Yea this is, can look up how to reduce some bloat with as-is files as a backlog taks

tjb9dc (approved):

PR #12235, opened 2026-02-10
feat(cli): wire library docs generation end-to-end

paarthfern:
## Description

  Wires the `fern docs md generate` command end-to-end: from FDR API → IR download → local MDX generation → docs resolver integration. Library docs now render as a first-class navigation section in `fern docs dev` as well as generate/preview

  ## Changes Made

  - **CLI command** (`generateLibraryDocs.ts`): authenticate, start server-side parsing via FDR API, poll for completion, download IR from S3, run local MDX generator
  - **Navigation YAML** (`NavigationBuilder.ts`): serialize nav tree as `_navigation.yml` automatically during generation
  - **Docs resolver** (`DocsDefinitionResolver.ts`): read `_navigation.yml` + MDX files from the library output directory, convert into FernNavigation nodes with stable `fullSlug` URLs
  - **Root overview page**: derive root module page from nav structure and set as section overview
  - **Dedup fix**: filter out redundant child pages whose slug matches their parent section
  - **MDX escaping fixes**: escape JSX chars in table cells, handle indented code fences in list items, only wrap unindented code blocks in `<CodeBlock>`

  ## Testing

  - [x] Unit tests added/updated (library-docs-generator: 338 tests, generateLibraryDocs CLI tests)
  - [x] Manual testing completed (`fern docs dev` with nemo-rl library):
  
   the following doc started config: https://github.com/fern-api/docs-starter/compare/main...library-docs-testing
produced the following preview link: https://fern-preview-9a530bd6-9178-4f46-a9fa-a2c62cf2b5a7.docs.dev.buildwithfern.com/library-reference/nemo-rl

when `fern docs md generate --library nemo_rl` was run


packages/cli/cli/src/commands/docs-md-generate/__test__/generateLibraryDocs.test.ts
```diff
@@ -1,11 +1,106 @@
...
+
+import { createFdrService } from "@fern-api/core";
+import { generate } from "@fern-api/library-docs-generator";
+// Must import after vi.mock so mocks are in place
+import { askToLogin } from "@fern-api/login";
+import { type GenerateLibraryDocsOptions, generateLibraryDocs } from "../generateLibraryDocs.js";
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
```
amckinney: nit: I'd get rid of these useless Claude comments. This applies elsewhere.

packages/cli/cli/src/commands/docs-md-generate/__test__/generateLibraryDocs.test.ts
```diff
@@ -1,11 +1,106 @@
...
+                });
+            }
+        ),
+        logger: { info: (m: string) => logs.push(m) }
+    };
+}
+
+function makeMockProject({
+    libraries,
+    absoluteFilePath = "/home/user/project/fern"
```
amckinney, line 53: Does this actually read/write anything from this filepath? Should we use a temporary directory instead?

packages/cli/cli/src/commands/docs-md-generate/__test__/generateLibraryDocs.test.ts
```diff
@@ -120,77 +215,338 @@ describe("generateLibraryDocs", () => {
...
+            const outputPath = "./static/sdk-docs";
 
-            const subpathInfo = config.input.subpath != null ? ` (subpath: ${config.input.subpath})` : "";
-            const message = `Would generate library '${name}' from ${config.input.git}${subpathInfo} to ${config.output.path} (lang: ${config.lang})`;
+            const resolved = path.resolve(fernFolder, outputPath);
 
-            expect(message).toBe(
-                "Would generate library 'my-sdk' from https://github.com/acme/monorepo (subpath: packages/sdk) to ./static/sdk-docs (lang: python)"
-            );
+            expect(resolved).toBe("/home/user/project/fern/static/sdk-docs");
```
amckinney: Is this test necessary here? It's largely testing the behavior of `path.resolve`, which shouldn't be necessary, right?

packages/cli/cli/src/commands/docs-md-generate/__test__/generateLibraryDocs.test.ts
```diff
@@ -120,77 +215,338 @@ describe("generateLibraryDocs", () => {
...
+        });
+
+        it("resolves parent-relative path", () => {
+            const fernFolder = "/home/user/project/fern";
+            const outputPath = "../docs/sdk";
+
+            const resolved = path.resolve(fernFolder, outputPath);
+
+            expect(resolved).toBe("/home/user/project/docs/sdk");
         });
```
amckinney: Similar to above -- I don't think we need this sort of tests here. We should focus on what _we_ are defining, not another library's responsibility.

packages/cli/cli/src/commands/docs-md-generate/generateLibraryDocs.ts
```diff
@@ -58,13 +83,117 @@ export async function generateLibraryDocs({ project, cliContext, library }: Gene
...
+                    typeof fdr.docs.v2.write.startLibraryDocsGeneration
+                >[0]["githubUrl"],
+                language: config.lang === "python" ? "PYTHON" : "CPP",
+                config: {
+                    branch: undefined, // TODO: make optional in FDR API definition
+                    packagePath: gitInput.subpath,
+                    title: name,
+                    slug: name
+                }
+            });
```
amckinney: Accessing the arguments this way feels unsafe (i.e. `[0]["orgId"]` and `[0]["githubUrl"]`). Can we clean this up so that it flows a little simpler/safer?

packages/cli/cli/src/commands/docs-md-generate/generateLibraryDocs.ts
```diff
@@ -58,13 +83,117 @@ export async function generateLibraryDocs({ project, cliContext, library }: Gene
...
+            if (!startResponse.ok) {
+                return context.failAndThrow(
+                    `Failed to start generation for library '${name}': ${JSON.stringify(startResponse.error)}`
+                );
+            }
+
+            const jobId = startResponse.body.jobId;
+            context.logger.info(`Generation job started (${jobId}). Polling for completion...`);
+
+            // 2. Poll for completion
```
amckinney: nit: Any time we have these comments, there might be an opportunity to add helper functions (i.e. one to poll for completion, one to start server-side rendering, another to fetch the IR from S3, etc). 

Sometimes it makes sense to hoist logic out into a single place, but I think we can clean this up a bit. Mind taking a pass to see what we can do?

packages/cli/docs-resolver/src/DocsDefinitionResolver.ts
```diff
@@ -1582,6 +1592,174 @@ export class DocsDefinitionResolver {
...
+        parentSlug: FernNavigation.V1.SlugGenerator
+    ): Promise<FernNavigation.V1.SectionNode | null> {
+        const libraryConfig = this.parsedDocsConfig.libraries?.[item.libraryName];
+        if (libraryConfig == null) {
+            this.taskContext.logger.warn(`Library '${item.libraryName}' not found in libraries config. Skipping.`);
+            return null;
+        }
+
+        const outputDir = resolve(this.docsWorkspace.absoluteFilePath, libraryConfig.output.path);
+        const navFilePath = join(outputDir, RelativeFilePath.of("_navigation.yml"));
```
amckinney: Is the `_navigation.yml` filepath new here? Can you speak to why we have this `_` prefix naming convention, where it's expected to exist on disk, whether or not there should be one or more of them at different locations, etc?

packages/cli/cli/src/commands/docs-md-generate/__test__/generateLibraryDocs.test.ts
paarthfern: No real I/O occurs against this path — generate() is mocked, so it's only used for path resolution in the function under test

amckinney (approved):

PR #12246, opened 2026-02-11
feat(php): generate interfaces for all SDK clients

patrickthornton:
## Description
Equivalent of https://github.com/fern-api/fern/pull/11430 in PHP, and for the same reason; it helps mocking and DI.

## Changes Made
PHP generator update to generate standalone interface files like the following along all clients:
```php
  namespace Seed;

  interface SeedClientInterface
  {
      /**
       * @param ?array{
       *   baseUrl?: string,
       *   maxRetries?: int,
       *   timeout?: float,
       *   headers?: array<string, string>,
       *   queryParameters?: array<string, mixed>,
       *   bodyProperties?: array<string, mixed>,
       * } $options
       */
      public function foo(?array $options = null): void;
  }
```
The `SeedClient` then `implements SeedClientInterface`.

## Testing
Tested on `exhaustive` and `folders` (the latter of which has a root-level client interface); regenerating all seed fixtures now.


fern-support (approved):

fern-support (changes requested):
it would be better to have it with config `params` not for all SDK's by default.

generally popular SDK's does not Generates Client Interfaces

fern-support (approved):

PR #12250, opened 2026-02-11
fix(typescript): change _Unknown union discriminant from void to string

tjb9dc:
## Description

Refs: Requested by @tjb9dc
Link to Devin run: https://app.devin.ai/sessions/77f412e870f24f78a9fe1edb6995e0eb

Fixes a bug where `_Unknown.discriminant` was typed as `void`, breaking the `_visit` function because `void` is not assignable to `string` (the type expected by the `_other` visitor callback). This occurs when both `includeOtherInUnionTypes: true` and `includeUtilsOnUnionMembers: true` are enabled.

## Changes Made

1. **`UnknownSingleUnionType.ts`** — Override `getDiscriminantValueType()` to return `StringKeyword` for regular type unions. This is the targeted fix that only affects regular unions, not endpoint error unions.

2. **`UnknownSingleUnionTypeGenerator.ts`** — Removed the `as any` cast wrapping the spread in `getNonDiscriminantPropertiesForBuilder()`. Since both the input and `_Unknown` now use `string` for the discriminant, the types align and the cast is unnecessary.

3. **`GeneratedUnionImpl.ts`** — Added `needsCastInVisitBranches()` helper that checks whether the `_Unknown` discriminant is non-void (i.e., `string`). When it is, known case branches in the `_visit` switch cast the value to the specific variant type. This is needed because `_Unknown.type: string` overlaps with string literals, preventing TypeScript from narrowing. **Endpoint error unions** (where `_Unknown` discriminant remains `void`) do **not** get these casts since narrowing works correctly there.

- [x] Updated `generators/typescript/sdk/versions.yml` with v3.48.1 changelog entry

## ⚠️ Checklist for Human Review

1. **Verify that `needsCastInVisitBranches()` correctly distinguishes union types.** It returns `true` only when `includeOtherInUnionTypes` is enabled AND the unknown discriminant type node is not `VoidKeyword`. This means endpoint error unions (void discriminant) skip the casts while regular type unions (string discriminant) get them.

2. **The exact user scenario (regular unions with both `includeOtherInUnionTypes` + `includeUtilsOnUnionMembers`) may not be fully covered by seed fixtures.** The trace fixtures test the `_Unknown: string` change, and endpoint error unions test the cast removal, but a fixture with `_visit` on regular unions with `_Unknown: string` would provide more confidence.

3. **Verify the `_unknown` builder spread works without `as any`** — the types should align now that both input and `_Unknown` use `string` for the discriminant.

## Testing

- [x] Seed tests pass for `trace/exhaustive`, `trace/no-custom-config`, `trace/serde`, `trace/serde-no-throwing`
- [x] Seed tests pass for `error-property/union-utils`, `error-property/no-custom-config`
- [x] `pnpm run check` passes (biome lint)

- Addressed reviewer feedback (Swimburger): Removed unnecessary type casts from endpoint error union `_visit` branches. The casts are now conditional on the unknown discriminant being non-void, so endpoint error unions (which keep `void`) no longer have the casts.


generators/typescript/sdk/versions.yml
```diff
@@ -1,4 +1,17 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.48.1
+  changelogEntry:
+    - summary: |
+        Fix `_Unknown` union variant discriminant type from `void` to `string` when
+        `includeOtherInUnionTypes` is enabled. The `void` type was not assignable to the
+        `string` expected by the `_other` visitor callback, breaking the `_visit` function.
+        The `_unknown` factory function no longer uses an unnecessary `as any` cast, and
+        known case branches in `_visit` now include explicit casts to the specific variant
+        type to preserve type safety.
```
tjb9dc: Way too verbose

generators/typescript/model/union-generator/src/unknown-single-union-type/AbstractUnknownSingleUnionType.ts
```diff
@@ -19,7 +19,7 @@ export abstract class AbstractUnknownSingleUnionType<
     }
 
     public getDiscriminantValueType(): ts.TypeNode {
-        return ts.factory.createKeywordTypeNode(ts.SyntaxKind.VoidKeyword);
+        return ts.factory.createKeywordTypeNode(ts.SyntaxKind.StringKeyword);
```
tjb9dc: we need to use whatever the type the discriminator is for the other members. Not sure we can always assume that it's a string, or can we?

Swimburger (changes requested):
the type cast on error discriminants is unnecessary

Swimburger (approved):

PR #12287, opened 2026-02-12
feat(internal): filter optional header parameters from dynamic snippet generation

tjb9dc:
## Description

Refs: Requested by @tjb9dc

Excludes optional/nullable header parameters from generated code snippets across all language generators. Previously, all headers (including optional ones) were included in dynamic snippets, which cluttered the generated examples with parameters that aren't required.

[Link to Devin run](https://app.devin.ai/sessions/e7c745bd21ec451ba70b4f9c2b125249)

## Changes Made

- Added `filterRequiredParameters()` helper to `AbstractDynamicSnippetsGeneratorContext` that filters out parameters where `isOptional()` or `isNullable()` returns true on the type reference
- Updated `needsRequestParameter()` in the abstract context to only consider required headers when deciding if a request wrapper is needed
- Applied the filter at **constructor/global headers** (`ir.headers`) in each of the 8 language generators
- For **per-request headers** (`request.headers`): pass full headers list with `ignoreMissingParameters: true` to allow user-provided optional header values to be included while avoiding lookup failures
- Generators updated: TypeScript, Python, Java (2 constructor sites), Go, C#, Ruby, PHP, Rust
- Swift generator was not modified (it has no header processing logic)
- Fixed 5 pre-existing TS compile errors in `AbstractDynamicSnippetsGeneratorContext.ts` by adding `return` before `assertNever()` calls (surfaced because turbo cache was invalidated by the main changes)

Addressed review comment ([#discussion_r2800466633](https://github.com/fern-api/fern/pull/12287#discussion_r2800466633)):

- **Reverted** `filterRequiredParameters` at per-request `associateByWireValue` call sites across all 8 generators
- Now passing full `request.headers ?? []` with `ignoreMissingParameters: true` instead
- This fixes a bug where filtering parameters before `associateByWireValue` caused false "not recognized" critical errors when users provided values for optional headers
- Constructor header filtering remains unchanged (safe because it iterates over parameters, not user values)

Additionally:
- Fixed 5 pre-existing `assertNever` TS2366/TS7030 compile errors in `AbstractDynamicSnippetsGeneratorContext.ts` (`isFileUploadRequestBody`, `resolveEnvironmentName`, `includeRequestBodyInWrappedRequest`, `fileUploadHasBodyProperties`, `fileUploadHasFileProperties`) by adding `return` before `assertNever()` — these were exposed when the turbo cache was invalidated by changes to the same file

## Review Checklist

- [ ] **Filtering semantics**: The filter excludes both `optional` AND `nullable` type references. Verify this is the desired behavior — a header that is nullable but technically "required" will also be excluded. See `filterRequiredParameters()` in `AbstractDynamicSnippetsGeneratorContext.ts`.
- [ ] **`ignoreMissingParameters` behavior**: Per-request headers now use `ignoreMissingParameters: true`, which silently skips user-provided header values that don't match any parameter. This could mask typos in header names.
- [ ] **`needsRequestParameter()` change**: Endpoints with only optional headers won't generate request wrappers. Verify this is acceptable.
- [ ] **No test fixtures added**: Relying on existing seed tests to validate. Consider whether a dedicated fixture with optional headers would be valuable.

## Testing

- [x] Local compile verified (`pnpm turbo run compile --filter=@fern-api/browser-compatible-base-generator --force` and downstream `@fern-api/go-formatter`)
- [x] All 14 required CI checks pass
- [ ] Non-required seed test failures (`csharp-sdk`, `java-model`, `fastapi`) are pre-existing `TS2366` errors in `logger/src/console.ts`, `ir-utils`, etc. on `main` — unrelated to this PR
- [ ] Unit tests added/updated — no new tests; relying on existing seed test infrastructure to validate


tjb9dc (commented):
Looks good! But we also need new changelogs for all the SDKs (in versions.yml). These should all be type `fix` and be a PATCH version bump. For the summary just put `Filter out optional headers when generating dynamic snippets.`

PR #12300, opened 2026-02-12
chore(cli): support graphql json origin

chdeskur:
- refactors some of the graphql introspection code for better unit testing
- support graphql origins that provide the json schema directly (instead of requiring introspection)


packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
```diff
@@ -53,47 +50,244 @@ async function fetchGraphQLSchemaFromIntrospection(url: string, path: string, lo
...
+            body: JSON.stringify({
+                query: introspectionQuery
+            })
+        });
+
+        if (!resp.ok) {
+            if (resp.status === 401 || resp.status === 403) {
+                const authHint = url.includes("github.com")
+                    ? "Set GITHUB_TOKEN environment variable with a GitHub personal access token"
+                    : "Set GRAPHQL_TOKEN or API_TOKEN environment variable with an authentication token";
```
amckinney, line 127: Not actionable in this change, but are these environment variables standard? This is part of our API contract, so we might to come back to this to make sure these environment variables are intuitive for the introspection use case (i.e. I imagine `API_TOKEN` could be used for many other tools / use cases).

packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
```diff
@@ -53,47 +50,244 @@ async function fetchGraphQLSchemaFromIntrospection(url: string, path: string, lo
...
+    }
+
+    // Both attempts failed - provide comprehensive error message
+    const errorMessage =
+        `Failed to fetch GraphQL schema from ${url}.\n\n` +
+        `Attempt 1 (POST introspection): ${postResult.error}\n` +
+        `Attempt 2 (GET direct fetch): ${getResult.error}\n\n` +
+        `Please ensure the URL either:\n` +
+        `1. Accepts GraphQL introspection queries via POST, or\n` +
+        `2. Returns introspection results directly via GET`;
```
amckinney, line 291: Not actionable here, but let's come back to cleaning up this error message -- if (1) or (2) cannot be satisfied, is there anything else the user can do to provide the schema explicitly themselves? Or must they actually do either of these server-side changes to move forward?

packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
```diff
@@ -53,47 +50,244 @@ async function fetchGraphQLSchemaFromIntrospection(url: string, path: string, lo
...
+    // If it's wrapped in { data: ... }, unwrap it
+    if ("data" in data && data.data && typeof data.data === "object" && "__schema" in data.data) {
+        return data.data as IntrospectionQuery;
     }
 
+    throw new Error("Data does not contain valid GraphQL introspection result");
+}
+
+// Try GraphQL POST introspection approach
+export async function tryGraphQLIntrospection(
```
amckinney, line 92: Is this meant to be used anywhere outside `fetchGraphQLSchemaWithAutoDetection`? If not, we can remove the `export` so it's clear what the primary entrypoint is. The same might apply to other exported functions in this file.

packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
```diff
@@ -53,47 +50,244 @@ async function fetchGraphQLSchemaFromIntrospection(url: string, path: string, lo
...
-                `GraphQL introspection failed: ${resp.status} ${resp.statusText}. ` +
-                    `This endpoint requires authentication. ${authHint}`
-            );
-        }
-        throw new Error(`GraphQL introspection failed: ${resp.status} ${resp.statusText}`);
+    return headers;
+}
+
+// Helper function to check if JSON response contains introspection data
+export function isIntrospectionResult(data: unknown): boolean {
```
amckinney, line 57: nit: Let's drop in a comment for future readers.

```suggestion
// Helper function to check if JSON response contains introspection data.
// For details, see https://graphql.org/learn/introspection
export function isIntrospectionResult(data: unknown): boolean {
```

amckinney (approved):

packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
chdeskur: it's used in the fixture

packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
chdeskur: if origin fails to parse, they have to manually update the spec

these errors are only applicable for the origin setting

packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
chdeskur: there is no strict convention - i believe it's just the authentication for whatever service is providing the graphql endpoint (hence trying github auth, then bearer)

best solution down the line is probably to add a config for them to define what kind of auth to use. i'd say out-of-scope for now? just to get this shipped for the customer that's using it

packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
amckinney: Sweet, it might be worth following up to tell the user that this is unique to `origin` (i.e. "Failed to resolve GraphQL schema from origin <origin>. ..."). Not urgent, something we can come back to later as needed.

packages/cli/cli/src/commands/upgrade/updateApiSpec.ts
amckinney: Ideally we focus testing primarily on the entrypoint -- the internal functions are largely implementation details, so it's less valuable to test those individually. In other words, if we find ourselves exporting functions purely for the sake of testing (but not real consumption), that's often a code smell.

PR #12336, opened 2026-02-13
feat(go): add `omitEmptyRequestWrappers` config flag for empty request struct behavior

tstanmay13:
## Description

Refs: [Slack thread](https://buildwithfern.slack.com/archives/C06B0FSAQFQ/p1770249107657079)

Adds a new `omitEmptyRequestWrappers` configuration flag to control whether the Go SDK generator creates empty request struct parameters for endpoints. This was reported by a Polytomic user who noticed `Get` methods taking an empty `Request` struct argument that's never referenced (e.g., `modelsync.ExecutionsGetRequest{}`).

The fix is gated behind the flag for backward compatibility — existing users are unaffected unless they opt in via `"omitEmptyRequestWrappers": true` in their generator config.

**Link to Devin run:** https://app.devin.ai/sessions/33187a2cad704a4abd6dc89f64f5e1da
**Requested by:** @tstanmay13

## Root Cause

`needsRequestParameter()` in `sdk.go` had two issues:
1. It was unaware of **service-level headers** — it only checked `endpoint.Headers` but not the service's headers, so it couldn't account for whether service headers contributed fields to the request struct.
2. Its **default return value was `true`** — when an `SdkRequest` wrapper existed but none of the explicit checks (path params, query params, headers, body) matched, it still generated a request parameter, producing an empty struct.

**Why TS/Python don't have this bug:** Those generators flatten wrapper properties into individual function parameters, so an empty wrapper naturally produces zero parameters. Go uses a single struct parameter, making the empty case visible.

## Changes Made

- Added `omitEmptyRequestWrappers` boolean config flag through the full config chain: `customConfig` (JSON) → `cmd.Config` → `generator.Config` → `fileWriter` → function call sites
- Default is `false` for backward compatibility
- Added `serviceHeaders []*ir.HttpHeader` parameter to `needsRequestParameter()` and `shouldSkipRequestType()` (always active, not behind flag)
- Added `len(serviceHeaders) > 0` check alongside existing `len(endpoint.Headers) > 0` check
- Changed the default `return true` to `return !omitEmptyRequestWrappers` at end of `needsRequestParameter()` — when the flag is enabled and no visible fields exist, the request parameter is omitted
- Threaded `serviceHeaders` through the full snippet call chain: `WriteClient` → `NewGeneratedClient` → `newGeneratedEndpoint` → `newEndpointSnippet` → `getEndpointParameters`
- Updated `generators/go/sdk/versions.yml` with v1.23.8 (feat)
- Added `exhaustive:omit-empty-request-wrappers` seed fixture with generated output (160 files)

## Testing

- [x] All 88 CI checks pass (including all go-sdk, go-model, go-fiber seed tests)
- [x] `exhaustive:omit-empty-request-wrappers` seed fixture added with generated output — validates flag parsing, config threading, and no-regression for existing endpoints
- [ ] The exhaustive fixture does not contain endpoints that trigger the "empty wrapper omitted" code path specifically (no endpoint has a wrapper with zero user-facing fields). The fixture validates backward compatibility and config plumbing.

## ⚠️ Areas for review

1. **Positional parameter ordering**: Multiple functions (`newFileWriter`, `NewConfig`, etc.) had a new boolean parameter added. Verify the ordering is correct at all call sites.
2. **Default value (`false`)**: The flag defaults to `false`, meaning existing users keep the old behavior. Confirm this is the intended backward-compatible approach.
3. **`serviceHeaders` check is always active**: The `len(serviceHeaders) > 0` check runs regardless of the flag — only the final `return !omitEmptyRequestWrappers` is gated. This means service headers always contribute to the decision, but the "empty wrapper" fallback behavior is controlled by the flag.
4. **Fixture coverage**: The `exhaustive:omit-empty-request-wrappers` fixture validates config parsing and no-regression, but doesn't exercise the actual "empty wrapper omitted" code path end-to-end since exhaustive doesn't have endpoints that trigger that condition.


tstanmay13:
self +1 because it is gated behind a flag as well

fern-support (approved):

PR #12337, opened 2026-02-13
feat(cli): improve library docs generation with clean spinner UI

paarthfern:
Replace verbose polling logs with interactive task spinners for better UX:
- Use clean library name spinners
- Process libraries parallely with individual spinners
- Move verbose logs to debug level only
- Update tests to support new interactive task pattern

Provides cleaner, professional UI similar to `fern generate` command.

### DIFF
--- a/packages/cli/cli/src/commands/docs-md-generate/generateLibraryDocs.ts
+++ b/packages/cli/cli/src/commands/docs-md-generate/generateLibraryDocs.ts
@@ -6,7 +6,7 @@ import { resolve } from "@fern-api/fs-utils";
 import { generate } from "@fern-api/library-docs-generator";
 import { askToLogin } from "@fern-api/login";
 import { Project } from "@fern-api/project-loader";
-import { TaskContext } from "@fern-api/task-context";
+import { InteractiveTaskContext, TaskContext } from "@fern-api/task-context";
 import chalk from "chalk";
 
 import { CliContext } from "../../cli-context/CliContext.js";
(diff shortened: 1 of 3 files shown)


packages/cli/cli/src/commands/docs-md-generate/generateLibraryDocs.ts
```diff
@@ -74,56 +74,87 @@ export async function generateLibraryDocs({ project, cliContext, library }: Gene
...
-            });
+    await cliContext.runTask(async (context) => {
+        const results = await Promise.all(
+            Object.entries(librariesToGenerate).map(async ([name, config]) => {
+                if (config == null) {
+                    return false;
+                }
+                if (!isGitLibraryInput(config.input)) {
+                    context.failAndThrow(
+                        `Library '${name}' uses 'path' input which is not yet supported. Please use 'git' input.`
```
amckinney, line 85: Let's track an issue to support this if we haven't already -- should be a trivial addition.

packages/cli/cli/versions.yml
```diff
@@ -1,4 +1,14 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.78.0
+  changelogEntry:
+    - summary: |
+        Improve library docs generation UI with clean spinner interface.
+        Replace verbose polling logs with interactive task spinners for better UX.
+        Show "Initiating Library Generation" message and clean library name spinners
```
amckinney: Is this part outdated with the latest changes?

amckinney (approved):

packages/cli/cli/versions.yml
paarthfern: woops - good catch

PR #12345, opened 2026-02-13
fix(ruby): various changes for RuboCop compliance

patrickthornton:
## Description
A couple tweaks to make RuboCop happy:

In the regular SDK, change the enforced style from `snake_case` to `normalcase`, which is strictly more permissive and allows for certain constructions like `:recaptcha_v2, :sha256` that would otherwise demand underscores between the `v` and `2`.

In the generated snippets:
- Use double-quoted strings.
- Omit trailing semicolons.
- Drop empty parentheses on no-argument method calls.
- Prevents certain kinds of RuboCop checks for dynamic snippets `(FrozenStringLiteralComment, FirstHashElementIndentation)` (less relevant for short snippets instead of full SDK examples).
- Use `%w[]` syntax for arrays of simple strings.

Changes the Docker build command in `seed.yml` to use `@fern-api/ruby-sdk` instead of `@fern-api/fern-ruby-sdk`.

## Changes Made
Ruby-v2.

## Testing
Seed regenerating now!

### DIFF
--- a/generators/ruby-v2/ast/src/ast/TypeLiteral.ts
+++ b/generators/ruby-v2/ast/src/ast/TypeLiteral.ts
@@ -113,11 +113,11 @@ export class TypeLiteral extends AstNode {
     public write(writer: Writer): void {
         switch (this.internalType.type) {
             case "str": {
-                // For now, just write the string as a Ruby string literal
-                if (this.internalType.value.includes("'") || this.internalType.value.includes("#")) {
-                    writer.write(`"${this.internalType.value.replaceAll('"', '\\"')}"`);
+                const value = this.internalType.value.replaceAll("\r\n", "\n");
+                if (value.includes('"') && !value.includes("'") && !value.includes("#")) {
+                    writer.write(`'${value}'`);
                 } else {
-                    writer.write(`'${this.internalType.value}'`);
+                    writer.write(`"${value.replaceAll("\\", "\\\\").replaceAll('"', '\\"')}"`);
                 }
                 break;
             }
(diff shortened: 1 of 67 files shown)


generators/ruby-v2/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
```diff
@@ -3,13 +3,13 @@
...
 
-client = Acme::Client.new(token: '<YOUR_API_KEY>');
+client = Acme::Client.new(token: "<YOUR_API_KEY>")
 
 client.service.get_metadata(
   shallow: false,
-  tag: ['development', 'public'],
-  x_api_version: '0.0.1'
-);
+  tag: %w[development public],
```
iamnamananand996, line 10: I think we can just verify here. Does it work with special characters??

generators/ruby-v2/ast/src/ast/TypeLiteral.ts
```diff
@@ -190,6 +190,24 @@ export class TypeLiteral extends AstNode {
                     writer.write("[]");
                     break;
                 }
+                // Use %w[] for arrays where every element is a simple string with no
```
iamnamananand996, line 193: ooo it is handled here.

generators/ruby-v2/base/src/project/RubocopFile.ts
```diff
@@ -61,7 +61,7 @@ Layout/LineLength:
   Enabled: false
 
 Naming/VariableNumber:
-  EnforcedStyle: snake_case
+  EnforcedStyle: normalcase
```
iamnamananand996, line 64: It shouldn't update too many files for the customer; it can be done through a config, if we really need it for some customer, because by default, the `Ruby` ecosystem uses `snake_case` for `local variables` and `methods`

iamnamananand996 (approved):

PR #12346, opened 2026-02-13
fix(java, python): handle optional access tokens and type compatibility in OAuth generators

jsklan:
## Description
When an API spec defines the OAuth token response with optional `access_token` and `expires_in` fields, the Java and Python SDK generators produced code that failed to compile or raised runtime errors. This PR fixes both generators to handle optional OAuth response properties correctly, and fixes the Python wire test infrastructure for OAuth client credentials.

## Changes Made
- **Java OAuth generator**: Unwrap `Optional<String>` access tokens with `orElseThrow()` and use `Long` literals (`$LL`) for the default `expires_in` fallback to fix type compatibility
- **Python OAuth generator**: Check if `access_token` is optional and emit a `RuntimeError` guard when the value is `None`
- **Python wire test setup**: Use `client_id`/`client_secret` constructor params for OAuth client credentials instead of the invalid `_token_getter_override`
- **Python wire test generator**: Expect 2 requests when testing the OAuth token endpoint itself (one for auth, one for the actual call)

## Testing
- [x] Generated Java SDK compiles with optional `access_token` and `Long`-typed `expires_in`
- [x] Generated Python SDK passes mypy with optional `access_token`
- [x] Python wire tests pass with correct OAuth constructor params and request count


generators/java/sdk/src/main/java/com/fern/java/client/generators/OAuthTokenSupplierGenerator.java
```diff
@@ -184,8 +196,9 @@ public GeneratedJavaFile generateFile() {
             if (isOptional) {
                 // Handle optional expires_in with default fallback
                 // In Java, optional fields return Optional<T>, so use .orElse()
+                // Cast the default value to Long to handle type compatibility
                 getMethodSpecBuilder.addStatement(
-                        "this.$L = $L(authResponse.get$L().orElse($L))",
+                        "this.$L = $L(authResponse.get$L().orElse($LL))",
```
tjb9dc: is this compatible with APIs that use ints for expires_in?

tjb9dc (approved):

generators/java/sdk/src/main/java/com/fern/java/client/generators/OAuthTokenSupplierGenerator.java
jsklan: good catch. Fixed and pushed.

PR #12499, opened 2026-02-17
fix(cli): if asset outside of fern folder, use _up_ alias for "../"

Ryan-Amirthan:
## Description
This PR allows you to reference assets outside of the fern folder in your fern configuration

Linear ticket: Closes FER-8634

## Changes Made
Given we compute paths relative to the fern folder, we need a way to indicate a file is outside of it. We cannot use "../" to indicate that. When Fern uploads docs assets to S3 it creates presigned URLs with signatures based on the original path. The issue here is when any HTTP Client hits that url, it automatically normalizes URLs that contain "../". We are now using "_up_/" in the S3 signature and using it when retrieving assets from S3 to avoid HTTP normalization.

## Testing
Added tests for path normalization as well as the publish docs flow. 
Tested on a test fern config repo:

- does work when assets are outside of the fern folder.
- works when assets are inside fern folder (normal behavior).
- works when theres a "conflict" of file name and path inside and outside the fern folder


packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -143,7 +153,7 @@ export async function publishDocs({
 
                     const obj = {
                         filePath: CjsFdrSdk.docs.v1.write.FilePath(
-                            convertToFernHostRelativeFilePath(filePath.relativeFilePath)
+                            convertToFernHostRelativeFilePath(sanitizeRelativePathForS3(filePath.relativeFilePath))
```
chdeskur: ideally we can just convert this once?

Ryan-Amirthan: > ideally we can just convert this once?

@chdeskur question is can we just call `sanitizeRelativePathForS3` once right?

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -43,6 +43,12 @@ export async function calculateFileHash(absoluteFilePath: AbsoluteFilePath | str
     return createHash("sha256").update(new Uint8Array(fileBuffer)).digest("hex");
 }
 
+export function sanitizeRelativePathForS3(relativeFilePath: RelativeFilePath): RelativeFilePath {
+    // Replace ../ segments with _up_/ to prevent HTTP client normalization issues
```
thesandlord: @Ryan-Amirthan can we make this something more unique like `_fern_up_` or `_fern_dot_dot_` 

im worried _up_ is too generic, someone might actually use it

thesandlord (changes requested):

thesandlord (approved):

packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
Ryan-Amirthan: changed to _dot_dot_

PR #12559, opened 2026-02-19
feat(internal): Add replay step to post-generation pipeline

tstanmay13:
## Description
Linear ticket: Closes 

Introduces the PostGenerationPipeline in @fern-api/generator-cli, a step-based pipeline (ReplayStep to GithubStep) that runs after SDKgeneration for self-hosted GitHub repos. The workspace runner now delegates to this pipeline instead of handling GitHub operations inline. 

Replay (off by default, gated by `replay: { enabled: true }` in generators.yml) preserves user customizations across SDK regenerations by detecting patches via `.fern/replay.lock`, applying them with 3-way merge, and creating PRs with conflict resolution guidance when needed. Replay commands are also hidden and not discoverable. 

## Testing


fern/apis/generators-yml/definition/generators.yml
```diff
@@ -36,6 +37,11 @@ types:
         type: optional<map<string, group.GeneratorGroupSchema>>
       reviewers:
         type: optional<reviewers.ReviewersSchema>
+      replay:
+        type: optional<replay.ReplayConfigSchema>
+        docs: |
+          Configuration for SDK customization replay.
+          Automatically preserves user customizations across SDK regenerations.
```
amckinney, line 44: Do we still need this? Is the existence of the `replay.lock` sufficient (i.e. running `fern replay init`)?

packages/commons/github/src/ClonedRepository.ts
```diff
@@ -248,6 +254,52 @@ export class ClonedRepository {
         await this.git.push("origin", branch, { "--set-upstream": null });
     }
 
+    public async forcePush(): Promise<void> {
+        await this.git.cwd(this.clonePath);
+        const currentBranch = await this.getCurrentBranch();
+        await this.git.push("origin", currentBranch, ["--force"]);
+    }
```
amckinney, line 264: When do we need to `forcePush`? Do we _need_ to do this? This could be particularly destructive, especially if there's any opportunity to rewrite history.

packages/commons/github/src/ClonedRepository.ts
```diff
@@ -248,6 +254,52 @@ export class ClonedRepository {
...
+    public async getCommitTreeHash(commitSha: string): Promise<string> {
+        await this.git.cwd(this.clonePath);
+        const result = await this.git.raw(["rev-parse", `${commitSha}^{tree}`]);
+        return result.trim();
+    }
+
+    public async createAndPushTag(tagName: string, commitSha: string): Promise<void> {
+        await this.git.cwd(this.clonePath);
+        await this.git.tag(["-f", tagName, commitSha]);
+        await this.git.push("origin", `refs/tags/${tagName}`, ["--force"]);
```
amckinney, line 308: Do we _need_ `--force` here, too?

packages/generator-cli/src/replay/replay-init.ts
```diff
@@ -0,0 +1,302 @@
...
+ * 1. Clone the SDK repo
+ * 2. Run bootstrap() to scan history and create lockfile
+ * 3. Ensure .fernignore has replay entries
+ * 4. Commit with [fern-replay] prefix (recognized by isGenerationCommit)
+ * 5. Push (branch + PR in pull-request mode, direct push in push mode)
+ *
+ * The [fern-replay] commit message prefix ensures the bootstrap commit
+ * is NOT treated as a user customization patch by ReplayDetector.
+ */
+export async function replayInit(params: ReplayInitParams): Promise<ReplayInitResult> {
```
amckinney, line 52: Who is expected to run this command? Is a user expected to run `fern replay init` themselves, or is this automated in some way behind the scenes? This might affect what flows/modes we need to support for this functionality.

packages/generator-cli/src/api/replay-forget.ts
```diff
@@ -0,0 +1,18 @@
+import { type ForgetResult, forget } from "@fern-api/replay";
```
amckinney: nit: Should we push all these replay-specific behavior in a `replay/...` directory rather than just `api`? That way it's marginally more discoverable.

fern/apis/generators-yml/definition/generators.yml
amckinney: Discussed offline -- this feature will be helpful in the future for the user to have a frictionless opt-out experience, but we don't need it quite yet. Simply hiding the `fern replay` command and using it behind the scenes should be sufficient for now.

packages/cli/cli/src/cli.ts
```diff
@@ -680,6 +696,11 @@ function addGenerateCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext)
                 .option("output", {
                     type: "string",
                     description: "Custom output directory (currently only supported with --preview for SDK generation)"
+                })
+                .option("replay", {
+                    boolean: true,
+                    default: true,
+                    description: "Run replay after generation (use --no-replay to skip)"
                 }),
```
amckinney, line 692: Can we hide this flag for now?

packages/cli/cli/src/cli.ts
```diff
@@ -2023,3 +2046,441 @@ function writeBytes(stream: WriteStream, data: Uint8Array): Promise<void> {
...
+    cli.command("replay", "Commands for managing SDK customizations", (yargs) => {
+        addReplayInitCommand(yargs, cliContext);
+        addReplayStatusCommand(yargs, cliContext);
+        addReplayForgetCommand(yargs, cliContext);
+        addReplayResetCommand(yargs, cliContext);
+        return yargs;
+    });
+}
+
+interface ResolvedGithubConfig {
```
amckinney: nit: Let's put this and the accompanying `resolveGroupGithubConfig` in a separate file (i.e. `resolveGroupGithubConfig.ts`) -- the `cli.ts` shouldn't define any custom helpers/interfaces like this.

packages/commons/github/src/ClonedRepository.ts
amckinney: Discussed offline -- we can remove these.

packages/generator-cli/src/replay/replay-init.ts
amckinney: Does this command actually need to create a pull request or push directly? I imagine the initial onboarding to simply be:

**Ad hoc onboarding**
1. A user runs `fern replay init` to capture specific commits after one or more custom edits.
2. The user pushes directly the `replay.lock` to their repository.

**Automated onboarding**
1. We create an empty `replay.lock` alongside the rest of the generated files. No additional push / PR.
2. Any future custom edits are automatically tracked.

--

If the above is accurate, can this just be solely responsible for:
1. Run `bootstrap`
2. Generate the `replay.lock`
3. Done

packages/generator-cli/src/replay/replay-init.ts
```diff
@@ -0,0 +1,302 @@
...
+        entries.push({ level: "info", message: `\nPR created: ${result.prUrl}` });
+        entries.push({ level: "info", message: "Merge the PR to enable Replay for this repository." });
+    } else if (result.branch) {
+        entries.push({ level: "info", message: `\nPushed to branch: ${result.branch}` });
+    }
+
+    return entries;
+}
+
+function buildPrBody(result: BootstrapResult): string {
```
amckinney, line 214: Based on the conversation above, do we even need this functionality to start?

packages/generator-cli/src/api/replay-reset-remote.ts
```diff
@@ -0,0 +1,84 @@
...
+    reset: ResetResult;
+    /** Branch that was pushed to (undefined if dry-run or nothing to reset) */
+    branch?: string;
+}
+
+/**
+ * Reset Replay on a remote GitHub repository.
+ * Clones the repo, deletes replay.lock, commits with [fern-replay] prefix, and pushes directly.
+ */
+export async function replayResetRemote(params: ReplayResetRemoteParams): Promise<ReplayResetRemoteResult> {
```
amckinney: When do we expect to use this?

packages/generator-cli/src/api/replay-status-remote.ts
```diff
@@ -0,0 +1,83 @@
...
+    generationsCount: number;
+    /** Full lockfile data for detailed display (null if not initialized) */
+    fullLock: GenerationLock | null;
+}
+
+/**
+ * Get Replay status from a remote GitHub repository.
+ * Clones the repo into a temp directory, runs status, then cleans up.
+ */
+export async function replayStatusRemote(params: ReplayStatusRemoteParams): Promise<ReplayStatusRemoteResult> {
```
amckinney: When do we plan for this to be used?

packages/generator-cli/src/api/replay-status.ts
```diff
@@ -0,0 +1,58 @@
...
+    /** Full lockfile data for detailed display (null if not initialized) */
+    fullLock: GenerationLock | null;
+}
+
+/**
+ * Get the current Replay status for an SDK repository.
+ * Returns both the simplified status view and full lockfile data
+ * for the CLI's detailed display.
+ */
+export function replayStatus(params: ReplayStatusParams): ReplayStatusResult {
```
amckinney: When will this be used?

amckinney (commented):
General structure makes sense to me, but my high-level feedback is:
1. What are the core features required for the MVP to start testing this in our remote generation service?
2. Is there anything we can cut to reduce scope to start?

---

I think the initial rollout could be:
1. The operator (us or the user) manually runs `fern replay init` in their repository to generate the `replay.lock`.
2. The user regenerates their SDK and adds customizations as needed.
3. Upon regeneration, we run `fern replay run` and all customizations are applied.
4. (2) and (3) repeat forever afterwards.

Does this mean we could really scope this down to the `fern replay init` (i.e. without push and/or PR) and `fern replay run` commands?

That's not to say the surrounding implementation won't _eventually_ exist, but we can start with less to build out the core experience and expand from there as needed.

Thoughts?

packages/generator-cli/build.mjs
```diff
@@ -6,7 +6,7 @@ async function main() {
     await tsup.build({
         entry: ["src/cli.ts", "src/api.ts"],
         format: ["cjs"],
-        dts: true,
+        dts: { compilerOptions: { composite: false } },
```
amckinney, line 9: Do we want this?

amckinney (commented):
Discussed offline -- 

1. We can remove the unused commands for now (e.g. `reset`, `status`, etc).
    * We'll likely restore these commands at a later time, but less is more to start.
2. We'll revisit the `fern replay init` command and whether or not it should open a pull request at all. We will remove the automated `push` behavior to the base branch entirely.
    * In the future, we might be able to remove this entirely, but this is largely a quality of life improvement for the operator (for now).
3. Let's make sure all the flags and commands are hidden before we merge.

**Long-term plan**

In the long term, the user won't have to run `fern replay init` at all -- we'll just automatically create a `replay.lock` if it doesn't already exist and everyone will be onboarded. At that point, we can remove the command and just rely on `fern replay run` and the upcoming `fern replay forget <SHA>`

packages/cli/cli/src/cli.ts
amckinney: @tstanmay13 Let's try to handle this one here too just since it's a lightweight change.

tstanmay13: Sounds good! Will do

packages/commons/github/src/ClonedRepository.ts
tstanmay13: On push we overwrite the remote tag. Both are needed because this tag (fern-generation-base--<generator>) is a moving pointer, it gets updated to point at the new generation base SHA every time the pipeline runs.

Without force, the second generation would fail with "tag already exists" both locally and on push. It will only happen for replay branches.

amckinney (approved):
Excited to see this in action!

PR #12648, opened 2026-02-23
chore(swift): replace IR-generated endpoint path tests with focused unit tests

Swimburger:
## Description

[Link to Devin run](https://app.devin.ai/sessions/41dc4ecb3e5b4c6698bf3aba7d145a59)

Replaces the `format-endpoint-path-for-swift.test.ts` test suite — previously 139 tests taking **~22s** — with **15 focused unit tests** that run in **<250ms**.

### Why it was slow
The test iterated over all 139 test-definition directories, calling `createSampleIr()` for each (~180ms each, ~25s total) plus ~17s of module import time for `@fern-api/test-utils`. The function under test (`formatEndpointPathForSwift`) is a 22-line pure string formatter that only reads `fullPath.head`, `fullPath.parts`, and `allPathParameters` — full IR generation was massive overkill.

### Why the old tests were low-signal
Analysis of the 139 snapshot outputs:
- **13** were empty (no endpoints at all)
- **87** only had static paths (no path parameters) — testing only leading-`/` and trailing-`/` trim
- **39** actually exercised path parameter interpolation

100 of 139 tests didn't exercise the core logic of the function.

### Solution
Replace the 139 IR-generated golden-file tests with 15 handwritten unit tests using a lightweight `makeEndpoint` helper that constructs minimal `HttpEndpoint`-shaped objects. The tests cover all code branches in both `parseEndpointPath` and `formatEndpointPathForSwift`:

| Category | Tests |
|----------|-------|
| Basic path formatting | static path, single param, multiple params |
| Leading slash handling | missing slash on head, empty head with param |
| Trailing slash handling | strip trailing slash, preserve root `/`, strip after param tail |
| Null/empty head | empty head with no parts, `undefined` head with parts |
| Null tail | `undefined` tail in path parts |
| Missing declarations | path parameter not in `allPathParameters` (silently dropped) |
| Consecutive params | back-to-back params with no separator |
| Complex paths | param followed by tail, deeply nested multi-segment paths |

## Changes Made

- **`format-endpoint-path-for-swift.test.ts`**: Rewritten with a `makeEndpoint` helper and 15 targeted test cases covering all branches in `parseEndpointPath` and `formatEndpointPathForSwift`. No longer imports `@fern-api/test-utils`, `@fern-api/fs-utils`, or reads from the filesystem.
- **Deleted `snapshots/formatted-endpoint-paths/`** directory (145 `.swift` files).

## Testing

- [x] All 15 unit tests pass locally (<250ms total)
- [x] TypeScript compilation passes (zero new errors)
- [x] Biome lint passes
- [x] CI passes (compile, test, lint, seed tests)

## Reviewer Checklist

- [ ] **Coverage tradeoff**: The old suite was a golden-file regression net across all 139 API definitions. The new suite is a targeted unit test of `formatEndpointPathForSwift`'s code paths. If `parseEndpointPath` or `formatEndpointPathForSwift` has edge cases only triggered by specific real IR shapes, they won't be caught. The 15 tests now cover all branches in `parseEndpointPath` (null head, null tail, missing path parameter in lookup map) and `formatEndpointPathForSwift` (leading slash, trailing slash, root path).
- [ ] **Unsafe cast in `makeEndpoint`**: Uses `as unknown as FernIr.HttpEndpoint` to construct partial mock objects. If the function under test ever accesses additional `HttpEndpoint` fields not present in the mock, tests will pass compilation but fail at runtime with undefined access.
- [ ] **Double-slash behavior**: The test for missing path parameter declarations documents that undeclared params are silently dropped, leaving surrounding slashes intact (e.g. `"/items//details"`). Verify this is expected behavior rather than a bug.


generators/swift/sdk/versions.yml
fern-support, line 1: no need to create a new version

kafkas (approved):

PR #12663, opened 2026-02-23
chore(internal): speed up ir-migrations and validator tests via shared module caching

Swimburger:
## Description

[Link to Devin run](https://app.devin.ai/sessions/38fddb5f6cf54bd29fa1b5b1ec682fe7)

Adds `isolate: false` + `setupFiles` (workerSetup) to `ir-migrations` and `validator` — the two test packages with the highest file counts (56 and 47 respectively). With `isolate: false`, test files running in the same vitest worker thread share the module cache instead of re-importing heavy dependencies for every file.

Previously, each of the ~56 ir-migrations test files independently imported `@fern-api/workspace-loader`, `@fern-api/ir-generator`, `@fern-api/ir-sdk`, etc. With ~8 workers handling ~7 files each, modules were loaded 7× per worker. Now they're loaded once per worker via `setupFiles` and cached for subsequent files.

### Performance (8-core machine, isolated runs)

| Package | Before | After | Speedup |
|---|---|---|---|
| **ir-migrations** (56 files) | **132.5s** | **31.5s** | **76% faster** |
| **validator** (47 files) | **53.9s** | **21.0s** | **61% faster** |

### How it works
- **`isolate: false`** — disables per-file module isolation within a worker, so files share the module cache
- **`setupFiles`** (`workerSetup.ts`) — pre-imports heavy shared dependencies once when a worker starts, warming the cache before any test file runs

### Note on earlier approach
An earlier iteration of this PR attempted per-fixture test file generation (splitting monolithic test files into 139+ individual files). That approach was reverted because it added complexity without sufficient benefit for packages with few test files. The current `isolate: false` + `setupFiles` approach is config-only and works best for packages with many test files sharing heavy dependencies — which is exactly the case for ir-migrations and validator.

## Changes Made
- Added `packages/cli/generation/ir-migrations/src/workerSetup.ts` — pre-imports `@fern-api/cli-source-resolver`, `@fern-api/ir-generator`, `@fern-api/ir-sdk`, `@fern-api/workspace-loader`, etc.
- Added `packages/cli/fern-definition/validator/src/workerSetup.ts` — pre-imports `@fern-api/configuration-loader`, `@fern-api/lazy-fern-workspace`, `@fern-api/task-context`, etc.
- Updated `vitest.config.ts` for both packages to use `setupFiles` and `isolate: false`

## Testing
- [x] All 103 ir-migrations tests pass
- [x] All 69 validator tests pass
- [x] Biome passes on all changed files
- [x] Performance benchmarked locally (see table above)

## Human Review Checklist
1. **`isolate: false` safety**: Test files in the same worker now share module state. If any test mutates module-level globals or singletons, it could cause interference. Both packages' tests are stateless (load fixture → generate IR → snapshot), but worth verifying no module-level mutation exists.
2. **workerSetup import completeness**: The listed imports should cover the heaviest shared deps. If a heavy dep is missed, it just means slightly less cache benefit (not a correctness issue). If a listed dep isn't actually used by the tests, it adds unnecessary startup cost.
3. **CI performance**: Benchmarks were on an 8-core VM. Verify the speedup holds on CI runners (which may have different core counts).


packages/cli/generation/ir-generator-tests/package.json
```diff
@@ -27,9 +27,10 @@
     "compile": "tsc",
     "compile:debug": "tsc --sourceMap",
     "depcheck": "knip --config ../../../../knip.json --no-config-hints",
-    "test": "vitest --passWithNoTests --run",
-    "test:debug": "pnpm run test --inspect --no-file-parallelism",
-    "test:update": "vitest --passWithNoTests --run -u"
+    "generate-tests": "node generateTests.mjs",
+    "test": "node generateTests.mjs && vitest --passWithNoTests --run",
+    "test:debug": "node generateTests.mjs && pnpm run test --inspect --no-file-parallelism",
+    "test:update": "node generateTests.mjs && vitest --passWithNoTests --run -u"
```
fern-support: can't generateTests.mjs be done automatically when vitest is invoked?

.gitignore
```diff
@@ -92,6 +92,9 @@ seed/python-sdk/**/output.prof
 
 **/vitest.config.ts.*.mjs
 
+# Generated test files (created by globalSetup for parallel execution)
+**/__generated__/
```
fern-support: would it not be better for us to keep the generated test files checked into Git? It'll be more transparent and annotations would work better

fern-support (changes requested):

packages/cli/generation/ir-generator-tests/src/dynamic-snippets/__test__/DynamicSnippetsConverter.test.ts
fern-support, line 1: we can delete this file, right?

packages/cli/generation/ir-generator-tests/src/ir/__test__/generateIntermediateRepresentation.test.ts
```diff
@@ -2,14 +2,14 @@
...
+import { AbsoluteFilePath } from "@fern-api/fs-utils";
 import path from "path";
 
-import { loadApisOrThrow } from "../../loadApisOrThrow.js";
-import { generateAndSnapshotIR, generateAndSnapshotIRFromPath } from "./generateAndSnapshotIR.js";
+import { generateAndSnapshotIRFromPath } from "./generateAndSnapshotIR.js";
+
+// Test-definitions tests are now generated as individual files by globalSetup
+// for full parallel execution. See __generated__/test-definitions/ and
+// __generated__/test-definitions-openapi/ directories.
```
fern-support: we can omit this comment

fern-support (approved):

PR #12664, opened 2026-02-23
feat(internal): use expect.assert() for type narrowing in tests

Swimburger:
## Description

Refs: Vitest v4 `expect.assert()` adoption

Replaces manual type narrowing patterns (`if`-blocks, `as`-casts, conditional expects) with vitest's `expect.assert()` across 10 test files. This leverages the TypeScript `asserts condition` return type so the compiler narrows discriminated unions after the assertion, removing the need for manual casts or guarded blocks.

[Link to Devin run](https://app.devin.ai/sessions/4c4e4bc3cae84f3c95b3598f0a31fa26) | Requested by @Swimburger

## Changes Made

- Replaced `if (x?.type === "value") { ...assertions... }` patterns with `expect.assert(x?.type === "value"); ...assertions...` — assertions now run unconditionally (previously silently skipped if the type didn't match)
- Replaced `as` type casts (e.g., `nav[0] as { type: "section"; ... }`) with `expect.assert()` for runtime-validated narrowing
- Replaced `expect(x).toBeDefined(); expect(x?.type).toBe("value")` pairs with `expect.assert(x != null); expect.assert(x.type === "value")`
- Removed `/* eslint-disable jest/no-conditional-expect */` and several `biome-ignore` comments that are no longer needed
- Files changed:
  - `convertGeneratorsConfiguration.test.ts`
  - `writers.test.ts` and `integration.test.ts` (library-docs-generator)
  - `convertSecurityScheme.test.ts`
  - `navigationUtils.test.ts`
  - `availability-inheritance.test.ts`
  - `graphql-operation-layout.test.ts`
  - `migrateFromV65ToV63.test.ts`
  - `openapi-from-flag.test.ts`
  - `generateTypeReferenceExample.test.ts`

## ⚠️ Items for human review

1. **Behavioral change — silent skip → hard failure**: Previously, `if (x.type === "foo") { expect(...) }` would silently skip inner assertions when the condition was false. Now `expect.assert()` will fail the test. This is intentional but means **any previously-hidden type mismatches will surface as test failures**.
2. **`integration.test.ts` line ~288**: `for (const page of pages) { expect.assert(page.type === "page") }` — the original `if` only ran assertions on pages that matched; the new code asserts *every* item is a page. Verify that `flattenNav` only returns page nodes in this context.
3. **Optional chain narrowing**: Patterns like `expect.assert(x?.type === "section")` rely on vitest correctly narrowing through optional chains. If any of these don't narrow as expected at the type level, downstream property accesses would produce TS errors (should be caught by CI).

## Testing

- [x] Local `pnpm run check` (biome) passes (only pre-existing unrelated warning in `ir.test.ts`)
- [ ] CI — affected test suites should pass with the new unconditional assertions

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,12 +1,4 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
-- version: 3.82.1
-  changelogEntry:
-    - summary: |
-        Use vitest `expect.assert()` for type narrowing in tests, replacing manual
-        `if`-block and `as`-cast patterns around discriminated unions.
-      type: chore
-  createdAt: "2026-02-23"
-  irVersion: 65
 - version: 3.82.0
   changelogEntry:
     - summary: |
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,4 +1,12 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.82.1
+  changelogEntry:
+    - summary: |
+        Use vitest `expect.assert()` for type narrowing in tests, replacing manual
+        `if`-block and `as`-cast patterns around discriminated unions.
+      type: chore
+  createdAt: "2026-02-23"
+  irVersion: 65
 - version: 3.82.0
   changelogEntry:
     - summary: |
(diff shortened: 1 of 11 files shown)


packages/cli/cli/versions.yml
fern-support, line 1: No need to create a new version for test changes

fern-support (changes requested):

fern-support (approved):

PR #12665, opened 2026-02-23
chore(internal): adopt vitest AbortSignal API for test cleanup

Swimburger:
## Description
Linear ticket: Closes

Refs: Vitest AbortSignal API (v3.2+)

Wires `AbortSignal` from the Vitest test context (`{ signal }`) into e2e + CLI tests that spawn child processes or make network requests. When a test times out, bails, or is interrupted via Ctrl+C, the signal fires and child processes are killed so they don't leak.

Link to Devin run: https://app.devin.ai/sessions/8248da0ce5ee4de8bf7f2ef37e20038b  
Requested by: @Swimburger

### Core: `logging-execa` library
- Added `signal?: AbortSignal` to `loggingExeca.Options` interface.
- Added private `wireSignal()` helper in `loggingExeca.ts` that listens for `abort` and calls `.kill()` on the execa child process (with cleanup via `.finally()`).
  - Proactively attaches a no-op `.catch(() => {})` to the child process promise as soon as signal is provided, preventing unhandled rejection errors in Vitest when a process is intentionally terminated — whether by signal or manual `.kill()`.
- Wired signal internally in both `runExeca()` and `loggingExeca()`, so all callers get automatic cleanup by simply passing `signal` in options — no manual wiring needed.

### Callers: `runFernCli.ts` (named args refactor)
- Refactored `runFernCli` from positional params `(args, options?, includeAuthToken?, signal?)` to a single named-args options object: `(args, { includeAuthToken?, signal?, ...execaOptions })`.
- **Removed `runFernCliWithoutAuthToken`** — callers now use `runFernCli(args, { includeAuthToken: false })`.
- Refactored `captureFernCli` similarly: `(args, { signal?, ...execaOptions })`.
- Added `signal?: AbortSignal` to `CliV2.Options` in `runCliV2.ts` and forwarded it to `loggingExeca`.
- Added `signal?: AbortSignal` to `runContainer.Args` (and the deprecated `runDocker.Args`) in `docker-utils`, threaded through the retry/pull path.

### Tests
- Updated ~33 test files to destructure `{ signal }` from test context and pass it to CLI runners via the named-args options object.
- Updated helper functions (`init.ts`, `diff.ts`, `generateIrAsString.ts`, `generateDynamicIrAsString.ts`, `generateFdrApiDefinitionAsString.ts`) to thread signal through.
- Updated `docker-utils` test (`runDocker.test.ts`) to pass `{ signal }` into `runContainer(...)`.
- Reverted `protoc-gen-fern.test.ts` — `createLoggingExecutable` returns a `LoggingExecutable` whose options extend execa v5 `Options`, which does not support `signal`.
- **Fixed `docsDev.test.ts` unhandled rejections**: Removed manual `.catch(() => {})` calls before `.kill()` — no longer needed since `wireSignal` now proactively swallows rejections for all signal-wired processes.

- **Centralized signal handling**: Moved `wireSignal()` into the `logging-execa` library and wired it inside `runExeca()` / `loggingExeca()`. This eliminates the previously duplicated `wireSignal` helpers in `runFernCli.ts`, `runCliV2.ts`, and `docker-utils/runDocker.ts`. Callers now simply pass `signal` in the options object — no manual child-process wiring required.
- **Named args refactor**: Refactored `runFernCli`, `captureFernCli` to use a single options object instead of positional parameters. Removed `runFernCliWithoutAuthToken` entirely. This resolves the "positional API awkwardness" noted in the previous review checklist.
- **Proactive rejection swallowing**: `wireSignal` now attaches `.catch(() => {})` immediately when signal is provided (not just when signal fires), ensuring both signal-triggered kills and manual `.kill()` calls have their rejections swallowed. This fixes the pre-existing `docsDev.test.ts` unhandled rejection issue.
- **Made `wireSignal` private**: No longer exported from `loggingExeca.ts` since all callers use `runExeca`/`loggingExeca` which wire it internally.

- [ ] Updated README.md generator (N/A)

## Testing
- [ ] Unit tests added/updated (N/A — this is test infrastructure)
- [x] Manual testing completed — `pnpm run check` (biome) locally
- [x] All required CI checks pass (compile, lint, biome, boundaries, depcheck, test, seed tests)

## Review Checklist

**High Priority:**
1. **Bulk refactoring correctness**: ~40 test files were converted from positional args to named args. Verify that all call sites were correctly converted (especially multi-line patterns and `runFernCliWithoutAuthToken` → `runFernCli(..., { includeAuthToken: false })`).
2. **Proactive `.catch()` safety**: `wireSignal` now attaches `.catch(() => {})` to ALL child processes that have a signal, not just when signal fires. Verify this doesn't accidentally swallow legitimate errors (note: callers still `await` the result, so errors are still surfaced — the `.catch()` only prevents *unhandled* rejections).
3. **`captureFernCli` return type**: Changed from `import("execa").ExecaChildProcess` to `ReturnType<typeof runExeca>`. Verify this doesn't break any callers that depend on the specific type.

**Medium Priority:**
4. **`writeDocsDefinition.test.ts`**: The `testFixture` function accepts `signal` but the test calls don't pass it from the test context (tests are skipped anyway, but worth noting for completeness).

**Low Priority:**
5. **`protoc-gen-fern.test.ts` reverted**: This test was updated to receive `{ signal }` but then reverted because `createLoggingExecutable` doesn't support passing `signal` through to execa. If future work adds signal support to `LoggingExecutable`, this test should be revisited.


packages/cli/generation/local-generation/docker-utils/src/__test__/runDocker.test.ts
```diff
@@ -31,7 +31,7 @@ beforeAll(async () => {
 }, 60_000);
 
 describe("runContainer", () => {
-    it("basic-writer", async () => {
+    it("basic-writer", async ({ signal }) => {
```
fern-support, line 34: Can we take advantage of signal here?

fern-support: You can update the docker-utils library in a backwards compatible way in this PR

packages/cli/ete-tests/src/utils/runCliV2.ts
```diff
@@ -131,10 +161,12 @@ export async function runCliV2(options: CliV2.Options): Promise<CliV2.Result> {
             reject: false // Don't throw on non-zero exit.
         };
 
-        const result = await loggingExeca(CONSOLE_LOGGER, "node", ["--enable-source-maps", cliPath, ...options.args], {
```
fern-support, line 134: can't we update loggingExeca to accept signal?

fern-support (changes requested):

packages/cli/ete-tests/src/utils/runFernCli.ts
```diff
@@ -6,7 +6,8 @@ import path from "path";
 export async function runFernCli(
     args: string[],
     options?: Options,
-    includeAuthToken: boolean = true
+    includeAuthToken: boolean = true,
+    signal?: AbortSignal
 ): Promise<loggingExeca.ReturnValue> {
```
fern-support, line 16: too many position args, we should turn this into named args

fern-support (approved):

PR #12680, opened 2026-02-23
fix(python): respect package_name config in wire test imports

jsklan:
## Description
Linear ticket: Closes 

Fixes an issue where wire tests weren't respecting the `package_name` custom config, causing import errors when users specify custom package names. This change centralizes the module path logic and ensures consistent import behavior across all wire test components.

## Changes Made
- **Config Schema**: Added `package_name` option to `SdkCustomConfig` for explicit package naming
- **Context Centralization**: Added `getModulePath()` method to `SdkGeneratorContext` that prioritizes `package_name` config over organization-based paths
- **Wire Test Generator**: Updated to use centralized module path logic instead of hardcoded organization name
- **Wire Test Setup**: Removed duplicate module path logic and unified import generation
- **Version Update**: Bumped to 4.59.1 with changelog entry documenting the fix

## Testing
- [x] Verified wire tests generate correct imports with custom package names
- [x] Confirmed backward compatibility with existing configurations

### DIFF
--- a/generators/python-v2/sdk/src/SdkCustomConfig.ts
+++ b/generators/python-v2/sdk/src/SdkCustomConfig.ts
@@ -50,6 +50,7 @@ export const SdkCustomConfigSchema = z.object({
     /** @deprecated Use `wire_tests.enabled` instead */
     enable_wire_tests: z.boolean().optional(),
     package_path: relativePathSchema.optional(),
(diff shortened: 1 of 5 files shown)


iamnamananand996 (approved):
This looks good, would be cool to add any one seed fixtures example for wiretests

PR #12693, opened 2026-02-24
feat(cli): add --json flag to fern check for structured output

aditya-arolkar-swe:
## Description

Adds a `--json` flag to `fern check` that outputs structured validation results as JSON to stdout, with all other output redirected to stderr. This enables programmatic consumption of check results by CI pipelines, editor integrations, and other tooling.

Also fixes API naming in multi-API workspaces: both the JSON output and `fern init` now use the workspace directory name (e.g. `atoms`, `unified`, `waves`) instead of the hardcoded `"api"` from `api.yml`.

Link to Devin run: https://app.devin.ai/sessions/2610e2382fbe4a4aa9e407dda2d6acc3
Requested by: @aditya-arolkar-swe

### New `--json` flag on `fern check`
When `--json` is passed, the command:
1. Redirects all stdout writes to stderr (logging, progress spinners, etc.)
2. Runs validation as normal
3. Restores stdout and writes a single JSON object as the only stdout output

### JSON output format
```json
{
  "success": false,
  "results": {
    "apis": [
      {
        "api": "api-name",
        "severity": "fatal",
        "rule": "valid-oauth",
        "message": "Failed to resolve endpoint POST /ext/api/as/token.oauth2"
      }
    ],
    "docs": [],
    "sdks": []
  }
}
```

- `api` field only appears when there are multiple API workspaces
- `rule` field populated from the validation rule that produced the violation
- `sdks` reserved for future use (currently empty)

### Rule name propagation across all validators
Previously, validation violations did not carry the name of the rule that produced them. This PR adds `name?: string` to `ValidationViolation` in all four validator packages and threads the rule name through:

- **fern-definition-validator**: Introduced `NamedRuleVisitors` wrapper, updated all 3 AST visitor creators (`createDefinitionFileAstVisitorForRules`, `createRootApiFileAstVisitorForRules`, `createPackageMarkerAstVisitorForRules`) and `validateFernWorkspace` to pair rule names with visitors
- **generators-validator**: Same pattern — `NamedRuleVisitors` + updated `createGeneratorsYmlAstVisitorForRules` and `validateGeneratorsWorkspace`
- **oss-validator**: Maps rule name onto violations after `rule.run()` returns
- **docs-validator**: Already had `name?: string` on violations — no changes needed

### Stdout clamping mechanism
Uses a minimal monkey-patch approach:
```typescript
const originalStdoutWrite = process.stdout.write;
process.stdout.write = process.stderr.write.bind(process.stderr);
// ... all validation runs, all output goes to stderr ...
process.stdout.write = originalStdoutWrite;
process.stdout.write(JSON.stringify(jsonResult, null, 2) + "\n");
```

This works because `CliContext` holds a reference to `process.stdout`, so patching `.write()` redirects all logger output to stderr automatically.

### Fix API naming in multi-API workspaces
Two-pronged fix for the issue where all APIs in multi-API projects (e.g. `fern/apis/atoms`, `fern/apis/unified`, `fern/apis/waves`) reported `"api"` as their name:

1. **Runtime fallback** (`validateAPIWorkspaceAndLogIssues.ts`): Use `workspace.workspaceName` (the directory name) over `rootApiFile.contents.name` when available. Fixes existing projects that already have `name: api` in their `api.yml`.

2. **Preventive fix** (`fern init`): When creating new APIs in multi-API projects, set the `api.yml` name to the workspace directory name instead of hardcoded `"api"`. Distinguishes multi-API workspaces (`directoryOfWorkspace !== absolutePathToFernDirectory`) from single-inlined workspaces to avoid breaking test snapshots.

## Testing
- [x] All 47 CI checks passing
- [x] Verified no new type errors introduced
- [x] Updated test snapshots for `name` field additions

## Human Review Checklist
- [ ] Verify `workspace.workspaceName` is correctly populated for all workspace types (single-inlined vs multi-API)
- [ ] Confirm the `directoryOfWorkspace !== absolutePathToFernDirectory` comparison correctly distinguishes workspace types
- [ ] Check that JSON output format meets requirements for downstream consumers


packages/cli/cli/src/commands/validate/validateWorkspaces.ts
```diff
@@ -25,7 +26,14 @@ export async function validateWorkspaces({
     errorOnBrokenLinks: boolean;
     isLocal?: boolean;
     directFromOpenapi?: boolean;
+    outputJson?: boolean;
 }): Promise<void> {
+    // When outputting JSON, clamp all stdout to stderr so nothing pollutes the JSON output.
+    const originalStdoutWrite = process.stdout.write;
+    if (outputJson) {
+        process.stdout.write = process.stderr.write.bind(process.stderr) as typeof process.stdout.write;
+    }
```
amckinney: Should this happen higher up the chain so that it's more widely applicable to other CLI commands? In its current state, we might need to replicate this behavior for other `--json` flags later (i.e. `docs` commands, etc).

packages/cli/cli/src/commands/validate/validateWorkspaces.ts
```diff
@@ -137,7 +145,89 @@ export async function validateWorkspaces({
         });
     });
 
+    if (outputJson) {
+        // Restore stdout before writing JSON
+        process.stdout.write = originalStdoutWrite;
```
amckinney: Based on the comment above, if we want to move forward with this monkey patching approach, we should make it very discoverable -- if others need to reroute `stdout` / `stderr` we should do that through some API. Something like:

```typescript
// Grab the handle and write to it.
const stdout = getStdout();
stdout.write(...);

// Behind the scenes, reset to `originalStdoutWrite` which is centralized somewhere.
stdout.reset();
```

The above will need a little thought on exactly what flow makes sense, but you get the idea. This will make it easier to reuse elsewhere.

packages/cli/cli/src/commands/validate/validateWorkspaces.ts
```diff
@@ -137,7 +145,89 @@ export async function validateWorkspaces({
...
+        sdks: CheckJsonViolation[];
+    };
+}
+
+function buildJsonResult({
+    apiResults,
+    docsResult,
+    hasErrors,
+    showApiNames
+}: {
```
amckinney: Let's isolate this function and the `interface` declarations to a separate file (e.g. `buildCheckJsonResult.ts` or something similar).

packages/cli/cli/src/cli.ts
```diff
@@ -1118,6 +1118,11 @@ function addValidateCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext)
                     boolean: true,
                     description: "Whether to use the new parser and go directly from OpenAPI to IR",
                     default: false
+                })
+                .option("json", {
+                    boolean: true,
+                    description: "Output results as JSON to stdout (all other output goes to stderr).",
```
amckinney: nit: I'd remove the parentheses since it's implied.
```suggestion
                    description: "Output results as JSON to stdout.",
```

packages/cli/fern-definition/validator/src/rules/valid-type-reference-with-default-and-validation/__test__/valid-type-reference-with-default-and-validation.test.ts
```diff
@@ -32,60 +32,70 @@ describe("valid-default-and-validation-type-reference", () => {
...
                 nodePath: ["types", "Object", "properties", "value", "type"],
                 relativeFilepath: "__package__.yml",
+                name: "valid-type-reference-with-default-and-validation",
                 severity: "fatal"
             },
             {
                 message: "Validation for 'min' must be an integer, but found '26.2'",
                 nodePath: ["types", "Object", "properties", "value", "type"],
                 relativeFilepath: "__package__.yml",
+                name: "valid-type-reference-with-default-and-validation",
```
amckinney, line 63: nit: Should we call this `id`? That way we leave `name` for a potential human readable description later.

packages/cli/yaml/generators-validator/src/ValidationViolation.ts
```diff
@@ -2,6 +2,7 @@ import { NodePath } from "@fern-api/fern-definition-schema";
 import { RelativeFilePath } from "@fern-api/fs-utils";
 
 export interface ValidationViolation {
+    name?: string;
```
amckinney: See above for `name` vs. `id` -- when do we _not_ have a `name` to assign here?

packages/cli/cli/src/commands/validate/validateWorkspaces.ts
aditya-arolkar-swe: Proposing in a follow-up commit to add this to `CliContext` like so:

```
export class CliContext {

    ...

    private readonly stdoutGuard = new StdoutGuard();

    ... 

    /**
     * Execute `fn` with stdout redirected to stderr so that log output,
     * spinners, and other human-readable messages don't pollute stdout.
     * Use this for any command that supports a --json flag.
     */
    public async withJsonOutput<T>(fn: () => T | Promise<T>): Promise<T> {
        return this.stdoutGuard.withRedirect(fn);
    }

    ...

```

packages/cli/cli/src/cli-context/StdoutGuard.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Usage:
+ *   const guard = new StdoutGuard();
+ *   const result = await guard.withRedirect(async () => {
+ *       // everything that writes to process.stdout goes to stderr here
+ *       return doWork();
+ *   });
+ *   // stdout is restored — safe to write JSON now
+ */
+export class StdoutGuard {
```
amckinney: This is sweet 👏 

Feel free to leave the naming as-is, but I'll leave some food for thought.

Go's naming principles always encourage me to match the class name w/ the method names -- if this thing "redirects", is it better call a StdoutRedirector? 

At that point, you could make it more generically named as a FileDescriptorRedirector, so the caller could specify whether or not they want to redirect stdout to stderr or vice versa. This could work for any other output stream, so it could even be a StreamRedirector.

amckinney (approved):
Nicely done, this looks great! Excited that we have this now.

PR #12697, opened 2026-02-24
fix(csharp): replace lodash-es template with Eta for template file processing

Swimburger:
## Description

Refs: [TypeScript equivalent PR #12696](https://github.com/fern-api/fern/pull/12696)
Requested by: @Swimburger
[Link to Devin run](https://app.devin.ai/sessions/9cb0154bf7944505b571dc5032d9f282)

Replaces `lodash-es`'s `template()` function with [Eta](https://eta.js.org/) for processing template files in the C# SDK generator (`@fern-api/csharp-base`). This mirrors the same change made for the TypeScript generator in #12696.

Lodash's template engine crashes on backticks because it internally uses `new Function()`, which interprets backticks as JS template literal syntax. Eta uses the identical `<% %>` / `<%= %>` syntax and is a modern, lightweight, TypeScript-native engine with zero dependencies.

## Changes Made

- **Engine swap (`CsharpProject.ts`)**: Replaced `import { template } from "lodash-es"` with `import { Eta } from "eta"`, instantiated with `{ autoEscape: false, useWith: true, autoTrim: false }`, and swapped all three call sites (`githubWorkflow`, `testCsProjContents`, `replaceTemplate` helper)
- **Dependency**: Added `eta` (`^4.5.1`) as a dev dependency to `@fern-api/csharp-base` (so it gets bundled into the generator CLI)
- **Versioning**: Added `2.22.3` changelog entry in `generators/csharp/sdk/versions.yml`
- `lodash-es` is **not** removed — it's still used by `PackageUtilities.ts` (`camelCase`, `upperFirst`)

## Testing

- [x] TypeScript compilation passes (`pnpm compile --filter @fern-api/csharp-base`)
- [x] CI checks all pass (biome, compile, lint, depcheck, test, test-ete, versions.yml validation)
- [x] No seed test snapshot changes — generated output is identical to lodash

## Human Review Checklist

- **`useWith: true`**: Uses JavaScript's `with(){}` under the hood to put template variables in scope (matching lodash's behavior). Same approach as the TypeScript PR. The alternative would be rewriting all template files to use `it.varName` syntax.
- **`autoTrim: false`**: Without this, Eta strips trailing newlines after `<% %>` tags, producing different whitespace than lodash. Confirmed by CI seed tests passing with no snapshot changes.
- **`.replaceAll("\\{", "{")` post-processing**: The `githubWorkflow` template still has this post-processing step. With lodash, escaped braces `\{` were needed because lodash used `${}` interpolation. With Eta using `<%= %>`, this is likely a no-op now but is harmless to keep. Worth verifying the CI YAML template doesn't contain literal `\{` sequences that should remain escaped.
- **Node version requirement**: Eta v4 requires Node ≥ 20. CI uses Node 22. Verify this aligns with the C# generator's Docker base image.

### DIFF
--- a/generators/csharp/base/package.json
+++ b/generators/csharp/base/package.json
@@ -20,9 +20,7 @@
   "main": "lib/index.js",
   "source": "src/index.ts",
   "types": "lib/index.d.ts",
-  "files": [
-    "lib"
-  ],
+  "files": ["lib"],
   "scripts": {
     "clean": "rm -rf ./lib",
     "compile": "tsc",
--- a/generators/csharp/base/package.json
+++ b/generators/csharp/base/package.json
@@ -20,7 +20,9 @@
   "main": "lib/index.js",
   "source": "src/index.ts",
   "types": "lib/index.d.ts",
-  "files": ["lib"],
+  "files": [
+    "lib"
+  ],
   "scripts": {
     "clean": "rm -rf ./lib",
     "compile": "tsc",
@@ -43,5 +45,8 @@
     "lodash-es": "catalog:",
     "typescript": "catalog:",
     "vitest": "catalog:"
+  },
+  "dependencies": {
+    "eta": "^4.5.1"
   }
 }
--- a/generators/csharp/base/package.json
+++ b/generators/csharp/base/package.json
@@ -40,11 +40,9 @@
     "@fern-fern/ir-sdk": "^62.3.0",
     "@types/lodash-es": "catalog:",
     "@types/node": "catalog:",
+    "eta": "^4.5.1",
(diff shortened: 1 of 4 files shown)


generators/csharp/base/package.json
```diff
@@ -43,5 +43,8 @@
     "lodash-es": "catalog:",
     "typescript": "catalog:",
     "vitest": "catalog:"
+  },
+  "dependencies": {
+    "eta": "^4.5.1"
   }
```
Swimburger: this should probably be a devdependency so it gets bundled into the C# generator CLI

fern-support (approved):

PR #12705, opened 2026-02-24
fix(internal): use generator-cli JS API instead of subprocess spawning

Swimburger:
## Description

Fixes an issue where `README.md` and `reference.md` were generated as empty files when running seed tests with `--local` flag. The root cause was that `generator-cli` was only installed inside Docker images, and the runtime `npm install -f -g` fallback fails on Nix-based environments with read-only global `node_modules`.

Instead of bundling or installing the CLI binary, this PR switches `GeneratorAgentClient` to call the `@fern-api/generator-cli` **JS API** directly (`generateReadme`, `generateReference`, `githubPush`, `githubPr`). Since these are regular JS imports, esbuild bundles them into each generator's CJS output automatically. No subprocess spawning, no PATH resolution, no binary bundling, no runtime installation — it just works in both Docker and `--local` environments.

Because the JS API is now bundled into every generator's CJS output, **`npm install -f -g @fern-api/generator-cli` has been removed from all Dockerfiles** — making Docker builds faster and eliminating a runtime dependency.

Also updates to `@fern-api/generator-cli@0.6.4` (latest), which has improved ESM/CJS support and proper TypeScript type declarations.

Link to Devin run: https://app.devin.ai/sessions/954147587dde4c49a92dfe54bf922220
Requested by: @Swimburger

## Changes Made

- **Rewrote `GeneratorAgentClient`** (265 → 63 lines): replaced all subprocess spawning, temp file writing, and runtime installation logic with direct JS API calls to `@fern-api/generator-cli`
  - `generateReadme()` → calls `generateReadmeApi({ readmeConfig })` using typed `GenerateReadmeParams`
  - `generateReference()` → calls `generateReferenceApi({ referenceConfig })` using typed `GenerateReferenceParams`
  - `pushToGitHub()` → calls `githubPush({ githubConfig })` or `githubPr({ githubConfig })` using typed `GithubPushParams`
  - Removed `writeConfig`, `getOrInstall`, `install`, `createCli` private methods
  - Removed `tmp-promise`, `@fern-api/fs-utils`, and `@fern-api/logging-execa` imports (no longer needed)
  - Cleaned up log messages to avoid leaking implementation details (no "JS API" or "CLI" references)
- **Removed dead code from `AbstractGeneratorCli.ts`**: deleted private `pushToGitHub` and `getRawGitHubConfig` methods that were never called (they spawned generator-cli as subprocess), along with their now-unused imports (`RawGithubConfig`, `resolveGitHubConfig`, `createLoggingExecutable`, `writeFile`, `tmp`)
- **Removed `generator-cli` from all 10 Dockerfiles**: Since the JS API is bundled into each generator's CJS output by esbuild, the `npm install -f -g @fern-api/generator-cli@0.5.3` line is no longer needed in any Dockerfile (TypeScript SDK, TypeScript Express, Rust SDK/Model, Ruby SDK/Model, C#, PHP, Go, Python, Java, seed)
  - For Java SDK, also removed the @octokit/request CVE patching block that depended on generator-cli's node_modules path
- **Added `@fern-api/generator-cli` as a dependency** of `generators/base` and as a root devDependency
- **Updated catalog version** to `0.6.4` (latest, has better ESM/CJS support and proper TypeScript type declarations)
- **Removed unused `tmp-promise` dependency** from `generators/base/package.json` (no longer needed since we don't write temp config files)
- **Updated `pnpm-lock.yaml`** to reflect dependency changes
- **Restored seed output** for `use-oxc` and `use-oxfmt` fixtures — `README.md` and `reference.md` now generated correctly via JS API
- [x] Updated README.md generator (if applicable) — N/A

## Testing

- [x] `pnpm compile` passes for `@fern-api/base-generator` and `@fern-typescript/abstract-generator-cli`
- [x] `pnpm biome check` passes on changed files
- [x] `knip` (depcheck) passes for `@fern-api/base-generator`
- [x] Built TS SDK generator successfully (`pnpm turbo run dist:cli --filter @fern-typescript/sdk-generator-cli`)
- [x] Ran seed test locally with `--local` for `simple-api:use-oxfmt` fixture — passed with README.md (254 lines) and reference.md (51 lines) properly generated via JS API
- [x] Verified no stale `generator-cli.cjs` binary in dist/ (bundling approach was removed)
- [x] CI: **272/272 checks passing** — all seed tests pass, all core checks pass

## CI Status

✅ **All 272 CI checks passing** (0 fail, 0 pending). All required seed-test-results checks pass for all generators (TypeScript, Java, Python, Go, Rust, Ruby, C#, PHP, Swift, OpenAPI, Postman, Pydantic, FastAPI).

## ⚠️ Human Review Checklist

1. **Version bump to 0.6.4** — Updated from 0.5.3 to 0.6.4 (latest). This version has better ESM/CJS support and proper TypeScript type declarations. Verify that the new version doesn't introduce breaking changes in the JS API.

2. **Type safety improved** — Methods now use proper typed parameters (`GenerateReadmeParams`, `GenerateReferenceParams`, `GithubPushParams`) from `@fern-api/generator-cli` instead of generic type parameters. This provides better compile-time type checking.

3. **Constructor parameters now ignored** — `skipInstall` and `selfHosted` parameters are still accepted for backward compatibility but no longer used. Callers still pass these values (e.g., `new GeneratorAgentClient({ logger, skipInstall: true })`). Consider deprecating these parameters in a follow-up.

4. **Return value change in `pushToGitHub`** — Previously returned `content.stdout` from the subprocess. Now returns `""` (empty string) since the JS API functions return `void`. ✅ **Verified**: No callers depend on the return value.

5. **Unused dependencies in `generators/base/package.json`** — `@fern-api/logging-execa` and `@fern-api/fs-utils` are no longer imported in `GeneratorAgentClient.ts` but remain in dependencies. They may be used elsewhere in the package. Consider auditing in a follow-up cleanup.

6. **Seed output formatting diffs** — Minor markdown formatting changes (indentation, blank lines, trailing newlines) in `README.md` and `reference.md` for `use-oxc` and `use-oxfmt` fixtures. ✅ **Verified**: Content is identical, only whitespace/formatting differs. This comes from `generator-cli@0.6.4` JS API producing slightly different output than the previous CLI subprocess approach.

7. **Java SDK CVE patching removed** — The Java Dockerfile previously patched `@octokit/request` to fix CVE-2025-25290. Now that generator-cli is bundled via esbuild, the @octokit dependency is bundled too. ⚠️ **Needs verification**: Confirm that `@fern-api/generator-cli@0.6.4` bundles a safe version of `@octokit/request` (>= 8.4.1 or >= 9.2.1).

8. **Catalog version bump affects other consumers** — The catalog version bump `0.3.0` → `0.6.4` also affects `packages/cli/generation/local-generation/local-workspace-runner` which already had `@fern-api/generator-cli: catalog:` as a dependency. ✅ **Verified**: This consumer uses the CLI as a subprocess, not as a library, so the version bump is safe.

9. **esbuild bundling behavior** — The JS API code from `@fern-api/generator-cli` is now bundled into each generator's CJS output by esbuild. This increases the generator binary size but eliminates all runtime dependencies. Verify that the bundled code doesn't have unexpected side effects (e.g., file system access, network calls during import).

10. **Non-JS generators still work** — Generators with non-Node entrypoints (Python uses `python -m src.fern_python.generators.sdk.cli`, Go uses `/fern-go-sdk`, Java uses `sh /init.sh`) should still work because they invoke the Node CLI (`/bin/python-v2`, `/bin/go-v2`, `/bin/java-v2`) which has the JS API bundled. ✅ **Verified in CI**: All generators produce correct README.md and reference.md files.

11. **Dead code removal from AbstractGeneratorCli** — Removed `pushToGitHub` and `getRawGitHubConfig` private methods. ✅ **Verified**: These were private methods with no callers (checked via grep and TypeScript compilation).


generators/base/src/GeneratorAgentClient.ts
```diff
@@ -1,258 +1,63 @@
...
-            const errorMessage = error instanceof Error ? error.message : String(error);
-            this.logger.debug(
-                `GeneratorAgentClient.generateReadme: FAILED to get/install generator-cli: ${errorMessage}`
-            );
-            throw error;
-        }
-
-        // Verify CLI is accessible by checking version
+    public async generateReadme({ readmeConfig }: GenerateReadmeParams): Promise<string> {
+        this.logger.debug("GeneratorAgentClient.generateReadme: Calling JS API...");
```
fern-support: We don't need to log Calling JS API. THe fact that it's running JS or a CLI is a detail that doesn't need to be logged. Apply across all log messages

fern-support (approved):

dsinghvi (approved):

PR #12764, opened 2026-02-25
fix(ci): reduce turbo remote cache timeout and combine test-ete build steps

Swimburger:
## Description

Refs: Previous fix attempts in #12669, #12689, and #12761

Both `test` and `test-ete` jobs in CI intermittently hang for 10–25 minutes and time out. The turbo process prints "Remote caching enabled" and then produces **zero output** for the entire duration. This has been happening on ~40% of recent main-branch CI runs since around Feb 18.

**Root cause:** Turbo v2.8.10 intermittently hangs when connecting to the Vercel remote cache. The `test-ete` job was particularly affected because it ran two separate turbo invocations sequentially (`dist:cli:dev` then `seed:build`), increasing the likelihood of hitting the hang.

[Link to Devin run](https://app.devin.ai/sessions/22d5e7ddc0f9447c969de04990b9a21c) | Requested by: @Swimburger

## Changes Made

- **Combined two separate turbo invocations** in `test-ete` (`dist:cli:dev` + `seed:build`) into a single `pnpm turbo run dist:cli:dev dist:cli --filter=@fern-api/cli --filter=@fern-api/cli-v2 --filter=@fern-api/seed-cli` call. Reduces exposure to the intermittent remote cache hang by halving the number of turbo invocations.
- **Reduced `TURBO_REMOTE_CACHE_TIMEOUT` from 60 to 10** across all 15 workflow files that use turbo remote caching. Forces faster fallback to local execution when the remote cache is slow.
- **Added `pnpm turbo daemon stop || true`** before turbo runs in both `test` and `test-ete` jobs. Defensive cleanup of any stale daemon processes.
- **Added `timeout-minutes: 20`** to the "Run tests" and "Run tests with snapshot updates" steps in the `test` job (the `test-ete` job already had step-level timeouts from #12689).

- Removed `TURBO_TOKEN`/`TURBO_TEAM` clearing and `TURBO_CACHE: "local:rw"` override from the `test-ete` build step per reviewer feedback. CI passes without it — the combined invocation + reduced timeout is sufficient.

## Testing

- [x] CI fully green — all 45 checks pass (including both `test` and `test-ete`) on 3 consecutive CI runs
- [ ] **Cannot be fully validated locally** — the fix targets intermittent remote cache hangs in CI. Continued observation of main-branch CI runs needed to confirm long-term stability.
- [x] All 15 workflow files updated consistently (verified via grep)

## Human Review Checklist

- [ ] **Verify the combined turbo invocation produces the same artifacts.** The `test-ete` build step now runs `dist:cli:dev dist:cli` in one turbo call instead of separate `dist:cli:dev` and `seed:build` calls. Confirm these task names resolve to the same packages and outputs — `seed:build` previously ran `pnpm turbo run dist:cli --filter=@fern-api/seed-cli`, which is now included in the combined call.
- [ ] **Is 10s too aggressive for `TURBO_REMOTE_CACHE_TIMEOUT`?** This applies to all 15 workflows. If the remote cache regularly takes 5–15s for legitimate responses, this could cause unnecessary cache misses and slower builds. Monitor rebuild times after merging.
- [ ] **The underlying turbo hang is not fully explained.** The combined invocation + reduced timeout has passed CI consistently, but the root cause (turbo hanging at remote cache initialization) is not resolved. If hangs recur, consider upgrading turbo or reporting upstream.

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/12764" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


.github/workflows/ci.yml
```diff
@@ -189,13 +194,16 @@ jobs:
...
+      - name: Stop any stale turbo daemon
+        run: pnpm turbo daemon stop || true
 
-      - name: Build seed CLI
+      - name: Build CLIs (dev + seed)
         timeout-minutes: 10
-        run: pnpm seed:build
+        env:
+          TURBO_TOKEN: ""
+          TURBO_TEAM: ""
```
fern-support: I don't like this

fern-support (changes requested):

.github/workflows/ci.yml
```diff
@@ -189,13 +194,15 @@ jobs:
...
-        run: pnpm turbo run dist:cli:dev --filter=@fern-api/cli --filter=@fern-api/cli-v2
+      - name: Stop any stale turbo daemon
+        run: pnpm turbo daemon stop || true
 
-      - name: Build seed CLI
+      - name: Build CLIs (dev + seed)
         timeout-minutes: 10
-        run: pnpm seed:build
+        env:
+          TURBO_CACHE: "local:rw"
```
fern-support: just remove this env section

fern-support (approved):

PR #12774, opened 2026-02-25
feat(cli): update cli-v2 DocsSchema to use original schema shape

amckinney:
## Description

Linear ticket: Closes [FER-8697](https://linear.app/buildwithfern/issue/FER-8697/cli-v2-update-fern-apicli-v2-docsschema-to-use-the-original-schema)

Updates `@fern-api/cli-v2` DocsSchema zod definitions to match the original `docs.yml` configuration shape from `@fern-api/cli` instead of reinventing it. The zod schemas now mirror the serialization types in `packages/cli/configuration/src/docs-yml/schemas/sdk/serialization/resources/docs/types/`.

[Link to Devin run](https://app.devin.ai/sessions/1c2c36dcf8154585b52c2c50bd7da3da) | Requested by @amckinney

## Changes Made

**Zod schema updates (`@fern-api/config`):**
- Converted all schema properties from camelCase to kebab-case (`featureFlag` → `"feature-flag"`, `displayName` → `"display-name"`, `skipSlug` → `"skip-slug"`, etc.)
- Added missing fields across schemas to match original (e.g., `audiences`, `announcement`, `openrpc`, `postman`, `"tag-description-pages"`, `"title-source"`, etc.)
- Added `WithPermissions` / `WithFeatureFlags` mixins to `TabVariantSchema`
- Restructured `ProductConfigSchema` with shared `ProductConfigBaseSchema` extended by `InternalProduct` / `ExternalProduct`
- Updated `AvailabilitySchema` enum values to kebab-case
- Updated `FeatureFlagSchema` to support `string | config | config[]`
- `ProgrammingLanguageSchema` and `LanguageSchema` are `z.string()` type aliases so the underlying `docs.yml` shape can evolve without these drifting out of sync
- Made `PageConfigurationSchema.path` required (was optional)
- Removed `navigation` from `VersionConfigSchema` and `ProductConfigSchema`

**Domain model updates (`@fern-api/cli-v2`):**
- Updated `DocsConfig`, `Navigation`, `Product`, `Version` interfaces to use kebab-case property names
- Simplified `DocsConfigConverter` — most fields are now direct passthrough; only navigation items still need discriminated union conversion
- Simplified `LegacyDocsWorkspaceAdapter` to single-line passthrough cast

**Changelog:**
- [x] Updated `packages/cli/cli/versions.yml` with version 3.89.0 entry

## Testing

- [x] Typecheck passes for `@fern-api/config` and `@fern-api/cli-v2`
- [ ] Manual testing with real docs configurations
- [ ] Unit tests (none added — schema changes only)

## Review Checklist

**⚠️ Items worth verifying:**

1. **`LegacyDocsWorkspaceAdapter` type safety** — The adapter now uses `as unknown as docsYml.RawSchemas.DocsConfiguration` to cast the zod schema directly to the legacy type. This bypasses type checking. Verify that the zod schema shape actually matches the legacy serialization types, especially for:
   - Kebab-case property names
   - Enum value formats
   - Optional vs required fields
   - Union type structures

2. **Removed fields** — Verify these removals match the original schema:
   - `VersionConfigSchema` no longer has `navigation` (was required)
   - `InternalProductConfigSchema` no longer has `navigation` or `default`
   - `LibraryReferenceConfigurationSchema` no longer has `icon` or `hidden`
   - `SectionConfigurationSchema` no longer has `flattened`

3. **`PageConfigurationSchema.path` is now required** — Was optional before. Could break existing configs.

4. **Orphaned file** — `packages/cli/cli-v2/src/docs/config/SnippetLanguage.ts` is no longer imported but wasn't deleted.

5. **Pre-existing typecheck errors** — The following errors exist on main and are unrelated to this PR:
   - `LegacyDevServer.ts:57` — `forceDownload` property
   - Various API adapter errors


packages/cli/config/src/schemas/docs/DocsSchema.ts
```diff
@@ -27,6 +27,48 @@ import { AnalyticsConfigSchema } from "./seo/AnalyticsConfigSchema.js";
...
+    "no",
+    "pl",
+    "ru",
+    "sv",
+    "tr"
+]);
+
+export type LanguageSchema = z.infer<typeof LanguageSchema>;
+
 export const DocsSchema = z.object({
```
amckinney, line 50: It might be best to actually just use a type alias here -- that way the underlying `docs.yml` shape can continue to update behind the scenes without these configuration shapes growing out of sync. Let's simplify this as much as we can.

amckinney:
This isn't what I want -- will handle this one on my own.

PR #12813, opened 2026-02-25
fix(internal): update generator-cli package.json to use dist instead of lib

Swimburger:
## Description

Fixes the `@fern-api/generator-cli` published package configuration so that `exports`, `main`, and `types` reference the `dist/` output directory (from `tsup` build) instead of the `lib/` directory (from `tsc`). Adds a `versions.yml` entry for `0.6.3` so the fix gets published.

Requested by: @Swimburger
[Link to Devin run](https://app.devin.ai/sessions/25f54108f1ef44a3941d8fec0b4bc1e5)

## Changes Made

- Updated `exports`, `main`, and `types` to point to `dist/` instead of `lib/`
- Removed `development` and `source` export conditions (pointed to unpublished `src/`, flagged by `publint`)
- Changed `compile` script from `tsc --build` to `node build.mjs` (tsup) so that `dist/` is produced during compile — `tsc --build` was already broken on `main` with pre-existing type errors on `ClonedRepository`
- Updated `turbo.jsonc` to include `dist/**` in compile outputs
- `bin` and `files` left unchanged (still `bin/cli` and `["bin", "dist"]`)
- Added `versions.yml` entry for `0.6.3` to trigger publish

- Reverted `bin` and `files` changes per [reviewer feedback](https://github.com/fern-api/fern/pull/12813#discussion_r2855602672)
- Removed `development` and `source` export conditions (publint reported them as errors since `src/` is not published)
- Switched `compile` from `tsc --build` to `node build.mjs` so downstream monorepo packages can resolve `dist/api.d.ts` during compilation

## Review Checklist

- [ ] **Compile script change**: `compile` now runs `tsup` (`node build.mjs`) instead of `tsc --build`. This means compile no longer does strict TypeScript type-checking — `tsc --build` was already broken on `main` with pre-existing errors, so this is not a regression, but worth noting.
- [ ] **Removed export conditions**: `development` and `source` conditions were removed. These were not functional (no `customConditions` in tsconfig) and pointed to unpublished `src/`. Verify no tooling in the monorepo relied on them.
- [ ] Minor: `versions.yml` changelog summary mentions "bin fields" but `bin` was not changed.

## Testing

- [x] Verified `compile` (`node build.mjs`) succeeds and produces expected output in `dist/`
- [x] `publint` passes with no errors
- [x] Grep confirmed no monorepo imports reference `@fern-api/generator-cli` via `lib/`
- [x] Verified downstream `@fern-api/local-workspace-runner` no longer errors on `Cannot find module '@fern-api/generator-cli'`
- [x] All CI checks pass (47/47)


packages/generator-cli/package.json
```diff
@@ -11,17 +11,17 @@
...
+      "default": "./dist/api.js"
     }
   },
-  "main": "lib/api.js",
+  "main": "dist/api.js",
   "source": "src/api.ts",
-  "types": "lib/api.d.ts",
+  "types": "dist/api.d.ts",
   "bin": {
-    "generator-cli": "bin/cli"
```
fern-support, line 22: this should stay bin/cli, bin and dist can be published, just not lib

fern-support (changes requested):

fern-support (approved):

PR #12834, opened 2026-02-26
chore(ci): stop fetching all branches and tags during CI

Swimburger:
## Description

[Link to Devin run](https://app.devin.ai/sessions/adddf1353c00449692a67476663620f1)

CI workflows using `fetch-depth: 0` were fetching **all ~750 branches and hundreds of tags** on every run. This replaces those full fetches with shallow clones (`fetch-depth: 1` default + `fetch-tags: false`) and uses either PR event data or targeted git/API calls for base SHA computation — instead of requiring full local history.

`tj-actions/changed-files` (used via `check-for-changed-files` action) handles its own history fetching via its `fetch_depth: 30` parameter, so no additional local history deepening is needed for change detection.

> **Scope note:** Publish workflows (`publish-cli.yml`, `publish-generator-cli.yml`) were initially optimized but reverted due to race condition risks with the Commits API in this high-traffic repo. Those can be revisited separately.

## Changes Made

- **`seed.yml`** (`changes` job): Removed `fetch-depth: 0`, added `fetch-tags: false`. For PRs, uses `${{ github.event.pull_request.base.sha }}` directly from the event — no git fetch or API call needed. For push-to-main, the existing `gh api` call for last successful workflow run SHA is unchanged.
- **`update-seed.yml`** (`changes` job): Removed `fetch-depth: 0`, added `fetch-tags: false`. For `workflow_dispatch` on non-main branches, uses GitHub's `/repos/.../compare/` API for merge-base (only reliable approach with shallow clones). For the fallback path (no successful run found), uses `git fetch --no-tags --depth=2 origin main` + `git rev-parse origin/main~1` to get the previous main commit.
- **`seed-pr-comment.yml`**: Replaced `fetch-depth: 0` with `fetch-tags: false` (defaults to shallow clone). This workflow only needs to checkout the PR branch, run seed tests, commit, and push — no full history needed.
- **`seed-dockers.yml`**: Reduced `fetch-depth: 200` to `fetch-depth: 2` + `fetch-tags: false`. This workflow only runs on push to main and just needs HEAD~1 for change detection.

## Testing

- [x] Local git behavior verified: shallow clone (depth 1, no tags) + `git fetch --depth=10 origin main` fetches only the specified branch (82 objects) — **not** all 750 remote branches. `git diff` works correctly between commits in the shallow history.
- [x] All CI checks passing on this PR (including the `changes (*)` matrix jobs that exercise the modified change-detection logic)

## Human Review Checklist

1. **`seed.yml` — PR base SHA vs merge-base**: The PR path now uses `github.event.pull_request.base.sha` (base branch tip at PR sync time) instead of `git merge-base`. These can differ if main has advanced. Verify that `tj-actions/changed-files` handles this correctly — it should, since it computes its own diff, but worth confirming the behavior matches the old `git merge-base` approach for change detection purposes.
2. **`seed-dockers.yml` fetch-depth 200 → 2**: The `check-for-changed-files` action's `fetch_depth: 30` should handle additional fetching if needed, but worth confirming this doesn't break change detection on push to main.
3. **`tj-actions/changed-files` fetch_depth sufficiency**: With the checkout now being shallow (depth 1), the action needs to fetch enough additional history to diff against the provided `base_sha`. If the base SHA is older than ~30 commits from HEAD, the action may fail. Worth confirming `tj-actions/changed-files` can fetch the specific `base_sha` commit from the server rather than relying on linear depth alone.
4. **GitHub compare API limits** (`update-seed.yml` workflow_dispatch only): The `/repos/.../compare/` endpoint has a [250-commit comparison limit](https://docs.github.com/en/rest/commits/commits#compare-two-commits). Long-lived branches with 250+ divergent commits could hit this. This only affects the manual `workflow_dispatch` trigger on non-main branches, not the common PR path.
5. **`seed-pr-comment.yml` push from shallow clone**: The workflow commits and pushes from a now-shallow clone. Standard `git push` should work fine from shallow clones, but worth a sanity check.

### DIFF
--- a/.github/workflows/seed-dockers.yml
+++ b/.github/workflows/seed-dockers.yml
@@ -55,7 +55,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 200
+          fetch-depth: 2
+          fetch-tags: false
 
       - name: Check for ts changes
         id: ts
--- a/.github/workflows/seed-pr-comment.yml
+++ b/.github/workflows/seed-pr-comment.yml
@@ -211,7 +211,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ steps.parse.outputs.pr-ref }}
-          fetch-depth: 0
+          fetch-tags: false
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install
--- a/.github/workflows/publish-cli.yml
+++ b/.github/workflows/publish-cli.yml
@@ -65,14 +65,8 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
         with:
-          fetch-tags: false
-
-      - name: Fetch tags and history for git describe
-        run: |
-          # Fetch all tags (lightweight refs only, no branch history)
-          git fetch --tags --no-recurse-submodules origin
-          # Deepen current branch enough to reach the nearest tag
-          git fetch --deepen=100 --no-tags origin ${{ github.ref_name }}
+          fetch-depth: 0
+          fetch-tags: true
 
       - name: Install
         uses: ./.github/actions/install
@@ -128,7 +122,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: main
-          fetch-tags: false
+          fetch-depth: 20
 
       - name: Install
         uses: ./.github/actions/install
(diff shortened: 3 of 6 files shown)


.github/workflows/seed-dockers.yml
```diff
@@ -55,7 +55,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 200
+          fetch-depth: 2
```
fern-support, line 58: is this 200 -> 2 safe?

.github/workflows/seed-pr-comment.yml
```diff
@@ -211,7 +211,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ steps.parse.outputs.pr-ref }}
-          fetch-depth: 0
```
fern-support, line 214: is this safe?

fern-support (changes requested):

fern-support (approved):

PR #12851, opened 2026-02-26
fix(csharp): resolve seed test failures for unions, trace, and oauth

Swimburger:
## Description

Refs: https://app.devin.ai/sessions/e18135edc6684a0787036c468596a531
Requested by: @Swimburger

Fixes four bugs in the C# SDK generator discovered by running the seed test suite with build/test scripts enabled.

### 1. Fix mock server test generation for `singleProperty` union variants
**File:** `MockEndpointGenerator.ts`

The `filterUnionExample` method had two issues with `singleProperty` union variants:
- **Incorrect type annotation:** The IR type has `singleProperty` extending `ExampleTypeReference` directly (with `shape` and `jsonExample` on the shape itself), but the code expected a nested `typeReference` field. This caused a `TypeError: Cannot read properties of undefined` crash (e.g. the `trace` fixture).
- **Wrong property wire name:** The code used `wireDiscriminantValue.wireValue` (e.g. `"integerValue"`) as the JSON property key, and `Object.assign` for object values. The correct behavior is to look up the union type definition and use the singleProperty's `name.wireValue` (typically `"value"`). Added a `getSinglePropertyWireName` helper that looks up the union type definition from the context to resolve the correct wire name. This produces `{"type": "integerValue", "value": 1}` instead of `{"type": "integerValue", "integerValue": 1}`.

### 2. Fix CS0555 build error for unions with optional member types
**File:** `UnionGenerator.ts` (doGenerate)

When a union member type is `optional<Foo>`, the inner class `Foo` wraps `Foo?`. The generated implicit operator `implicit operator Foo(Foo? value)` is a self-conversion, which C# forbids (CS0555). The fix detects when the underlying type name matches the inner class name and skips generating the implicit operator.

### 3. Fix discriminant leaking into `AdditionalProperties` during union deserialization
**File:** `UnionGenerator.ts` (generateJsonConverter)

For `samePropertiesAsObject` union variants, the entire JSON (including the `type` discriminant) was passed to `Deserialize<T>()`. Since the target type doesn't have a `type` property, it ended up in `AdditionalProperties`. The fix strips the discriminant property from the JSON before deserializing these variants.

**Per-variant guard (updated):** The stripping is now per-variant — before stripping, the generator checks whether each `samePropertiesAsObject` variant's object type has a property with the same wire name as the discriminant (including `extendedProperties` for inherited props). If it does, the full JSON is used (preserving the subtype's own property). If it doesn't, the stripped JSON is used. If all variants have the matching property, the `jsonWithoutDiscriminator` variable is never emitted. This is validated by the `UnionWithDuplicativeDiscriminants` fixture whose variants have their own `type` property.

### 4. Fix OAuth client credentials custom flow to pass additional parameters
**Files:** `RootClientGenerator.ts`, `OauthTokenProviderGenerator.ts`

Additional OAuth properties (custom properties and scopes defined in the OAuth scheme) weren't being passed to the `OAuthTokenProvider` constructor. Added two helper methods to extract these parameters from the OAuth scheme configuration and inject them as constructor parameters. The filter now skips only literal properties (matching Java's `isLiteralProperty()` approach) rather than skipping all non-primitives.

### 5. Update `allowedFailures` in seed.yml
Removed `unions:no-custom-config` and `unions-with-local-date` from the allowed failures list since they now pass after the union fixes above.

- [x] Updated `versions.yml` with changelog entries (version 2.22.5)

## Testing

- [x] Ran seed tests locally to verify which fixtures now pass:
  - `unions:no-custom-config` - ✅ PASSED (previously failed with CS0555)
  - `unions-with-local-date` - ✅ PASSED (previously failed with discriminant in AdditionalProperties)
  - `trace` - ✅ Generated successfully (previously crashed; still has pre-existing test failures unrelated to this PR)
  - `oauth-client-credentials-custom` - ✅ Builds successfully (previously missing constructor params)
- [x] Lint/typecheck passed on all generator changes
- [x] Updated seed test expected outputs for all affected fixtures
- [x] All 74 CI checks passing

## Human Review Checklist

- [ ] Verify the `isLiteralTypeReference` helper (duplicated in `RootClientGenerator.ts` and `OauthTokenProviderGenerator.ts`) correctly identifies literal types — it checks `typeReference.type === "container" && typeReference.container.type === "literal"`
- [ ] Review the self-conversion detection logic for edge cases (nested optionals, etc.)
- [ ] Verify seed output changes match the expected behavior for each fix
- [ ] Check performance impact of creating `JsonObject` on every union deserialization for discriminant stripping (only when `samePropertiesAsObject` variants exist and at least one needs stripping)

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/12851" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


seed/csharp-sdk/examples/no-custom-config/src/SeedExamples/Commons/Types/Types/EventInfo.cs
```diff
@@ -174,12 +174,17 @@ JsonSerializerOptions options
                 discriminatorElement.GetString()
                 ?? throw new JsonException("Discriminator property 'type' is null");
 
+            // Strip the discriminant property to prevent it from leaking into AdditionalProperties
+            var jsonObject = System.Text.Json.Nodes.JsonObject.Create(json);
+            jsonObject?.Remove("type");
+            var jsonWithoutDiscriminator =
```
fern-support, line 180: Is it possible for a subtype to also have the discriminant property? Do we detect for that?

fern-support (changes requested):

seed/csharp-sdk/examples/no-custom-config/src/SeedExamples/Commons/Types/Types/EventInfo.cs
fern-support: Let's guard against it. In the generator, we know which subtypes would have a matching property, if so, use the full JSON, if not, the stripped JSON, if all of the subtypes have a matching property, don't create a stipped JSON version.

fern-support (approved):

PR #12869, opened 2026-02-26
fix(generator-cli): allow pipeline run to accept inline JSON config and output result to stdout

tstanmay13:
## Description
Linear ticket: Closes 

- `--config` now accepts inline JSON (detected by `{` prefix) in addition to file paths
- Result JSON always written to stdout; logs remain on stderr

This lets fiddle pass config inline and capture the result from stdout, eliminating the temp files that caused path resolution bugs in Docker. The entire temp file dance (serialize -> write to disk -> pass path -> read from disk -> parse) is replaced by (serialize -> pass as arg) for config and (write to stdout -> capture) for result

## Testing

### DIFF
--- a/packages/generator-cli/src/cli.ts
+++ b/packages/generator-cli/src/cli.ts
@@ -18,6 +18,8 @@ import { loadReadmeConfig } from "./configuration/loadReadmeConfig.js";
 import { loadReferenceConfig } from "./configuration/loadReferenceConfig.js";
 import { type PipelineConfig, PostGenerationPipeline } from "./pipeline/index.js";
 
+const STDIN_MARKER = "-";
+
 void yargs(hideBin(process.argv))
     // eslint-disable-next-line turbo/no-undeclared-env-vars
     .scriptName(process.env.CLI_NAME ?? "generator-cli")
@@ -172,17 +174,12 @@ void yargs(hideBin(process.argv))
                         .option("config", {
                             string: true,
                             required: true,
-                            description: "Pipeline config as inline JSON string or path to JSON file"
+                            description: "The file or data to use for pipeline configuration"
                         })
                         .option("output-dir", {
                             string: true,
                             required: true,
                             description: "Path to SDK output directory"
-                        })
-                        .option("result-file", {
-                            string: true,
-                            required: false,
-                            description: "Path to write result JSON"
                         });
                 },
                 async (argv) => {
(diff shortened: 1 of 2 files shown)


packages/generator-cli/src/cli.ts
```diff
@@ -172,17 +172,12 @@ void yargs(hideBin(process.argv))
                         .option("config", {
                             string: true,
                             required: true,
-                            description: "Path to pipeline config JSON"
+                            description: 'Pipeline config: inline JSON, "-" to read from stdin, or a file path'
```
amckinney: nit: We can imply the in-line JSON and stdin support with "data" -- less important for `generator-cli`, but good practice for us to document what this would look like in the `fern` CLI.

```suggestion
                            description: 'The file or data to use for pipeline configuration'
```

packages/generator-cli/src/cli.ts
```diff
@@ -196,8 +191,14 @@ void yargs(hideBin(process.argv))
                     try {
                         const wd = cwd();
 
-                        // Read pipeline config from file
-                        const configContent = await readFile(resolve(wd, argv.config), "utf-8");
+                        // Read pipeline config: "-" = stdin, starts with "{" = inline JSON, otherwise file path
+                        const configValue = argv.config;
+                        const configContent =
+                            configValue === "-"
```
amckinney: nit: Should we make `-` a constant somewhere so that we can eventually reuse it elsewhere?

packages/generator-cli/src/cli.ts
```diff
@@ -196,8 +191,14 @@ void yargs(hideBin(process.argv))
...
+                                ? await readStdin()
+                                : configValue.trimStart().startsWith("{")
+                                  ? configValue
+                                  : await readFile(resolve(wd, configValue), "utf-8");
                         const config: PipelineConfig = JSON.parse(configContent);
```
amckinney: Do we need to catch some sort of error here if the content is not parsed as JSON?

amckinney (approved):

packages/generator-cli/src/cli.ts
```diff
@@ -196,9 +193,21 @@ void yargs(hideBin(process.argv))
...
-                        // Read pipeline config from file
-                        const configContent = await readFile(resolve(wd, argv.config), "utf-8");
-                        const config: PipelineConfig = JSON.parse(configContent);
+                        // Read pipeline config: "-" = stdin, starts with "{" = inline JSON, otherwise file path
+                        const configValue = argv.config;
+                        const configContent =
+                            configValue === STDIN_MARKER
+                                ? await readStdin()
+                                : configValue.trimStart().startsWith("{")
+                                  ? configValue
```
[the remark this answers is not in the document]
tstanmay13: fine because only consumer is fiddle

PR #12870, opened 2026-02-26
fix(cli): support `x-fern-server-name` extension on AsyncAPI server definitions

iamnamananand996:
## Description

Refs [Slack thread](https://buildwithfern.slack.com/archives/C0ACD0AC361/p1772124821042889)

Fixes `x-fern-server-name` extension being ignored on AsyncAPI server definitions, which caused SDK generation to fail with:
```
Encountered undefined server name undefined at channel marketData /. Expected environment UAT to contain url for uat
```

**Link to Devin run:** https://app.devin.ai/sessions/a5bc21c94b6b4f83a75353d58026201b
**Requested by:** @iamnamananand996

## Changes Made

The AsyncAPI-to-IR converter was ignoring `x-fern-server-name` in two places:

1. **`AbstractServerConverter.convert()`** — environment IDs were always the raw server map key (e.g. `uat`), never the overridden name (`market_data`). Added a static `getServerName()` helper that checks for `x-fern-server-name` on the server object and falls back to the map key.

2. **Channel converters (v2 + v3)** — when resolving which server a channel references, the raw key from `$ref: "#/servers/uat"` or the `servers: [uat]` array was used directly as `baseUrl`. Now both `ChannelConverter3_0` and `ChannelConverter2_X` look up the actual server object and apply the same `x-fern-server-name` resolution.

- Added `AbstractServerConverter.getServerName()` static method (used by both server and channel converters)
- Updated `ChannelConverter3_0.resolveChannelServersFromReference()` and added `resolveFirstServerName()` fallback
- Added `ChannelConverter2_X.resolveServerName()` private helper
- Added v3.90.1 changelog entry

## Testing

- [ ] No unit tests exist for the asyncapi-to-ir package (pre-existing). The fix is consistent with how the OpenAPI converter handles `x-fern-server-name` via `ServersConverter.getServerName()`.
- [ ] Pre-existing compile errors on `main` confirmed unrelated to this change.

## Review Checklist

- **Verify the `as Record<string, unknown>` cast approach** in `getServerName()` is acceptable vs reusing the `ServerNameExtension` class from `v3-importer-commons` (which expects `OpenAPIV3_1.ServerObject` types, not AsyncAPI server types).
- **Check for other callsites** that use raw server keys and may also need `x-fern-server-name` resolution.
- **Confirm v2 fallback path** — `resolveServerName(Object.keys(...)[0])` correctly propagates `undefined` when no servers exist (matches prior behavior).

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,4 +1,17 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.91.2
+  changelogEntry:
+    - summary: |
+        Fix `x-fern-server-name` extension being ignored on AsyncAPI server definitions.
+        When an AsyncAPI spec uses `x-fern-server-name` to override a server's name
+        (e.g. renaming `uat` to `market_data` to avoid naming conflicts with another API),
+        both the environment ID and channel base URL now use the overridden name. This
+        fixes SDK generation errors like "Expected environment UAT to contain url for uat"
+        when the generators.yml environment URLs reference the overridden server name.
+      type: fix
+  createdAt: "2026-02-27"
+  irVersion: 65
+
 - version: 3.91.1
   changelogEntry:
     - summary: |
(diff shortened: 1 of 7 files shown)


packages/cli/cli/versions.yml
```diff
@@ -1,4 +1,137 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.91.2
```
amckinney, line 2: Are all the updated versions.yml entries intentional? Should we consolidate this is one?

amckinney (approved):

packages/cli/cli/versions.yml
iamnamananand996: aah, no, looks like it came while resolving conflicts, lemme fix it.

PR #12880, opened 2026-02-26
feat(cli): Add `fern sdk update` command to CLI v2

amckinney:
## Description

Closes [FER-8567](https://linear.app/buildwithfern/issue/FER-8567/cli-v2-implement-the-sdk-update-command)

Implements the `fern sdk update` command in the new CLI v2 (`@fern-api/cli-v2`). This is the CLI v2 equivalent of the existing `fern generator upgrade` command. It updates SDK target versions in `fern.yml` to the latest available release from the Fern registry.

Link to Devin run: https://app.devin.ai/sessions/3db1fbe8d39f4f2c9f8b90701507c7d9
Requested by: @amckinney

## Changes Made

- New `packages/cli/cli-v2/src/sdk/updater/SdkUpdater.ts` — Core updater class following the namespace + class pattern (à la `LegacyGeneratorInvocationAdapter`):
  - Namespaced interfaces: `Config`, `TargetEntry`, `TargetUpdate`, `TargetUpToDate`, `SkippedMajorUpgrade`, `UpdateResult`
  - `collectTargetEntries()` — filters `Workspace` `Target[]` by name and skips `"latest"` versions
  - `resolveUpdates()` — calls `getLatestGeneratorVersion` for each entry; detects skipped major upgrades
  - `applyUpdates()` — writes updated versions back to YAML atomically (all-or-nothing)
- New `packages/cli/cli-v2/src/commands/sdk/update/command.ts` — `UpdateCommand` orchestration layer:
  - Reads targets from `workspace.sdks.targets` (no manual YAML parsing for reading)
  - Interactive mode (TTY): confirm prompt for single update, checkbox for multiple
  - Non-interactive mode (`--yes` or non-TTY): applies all updates automatically
  - `--target` flag to filter to a specific SDK target (default: update all)
  - `--major` flag to include major version upgrades (default: minor/patch only)
  - `--prerelease` flag to include RC / pre-release versions
  - `$ref` resolution for targets defined in separate files
  - Output showing applied updates, up-to-date targets, skipped major upgrades, and changelog URLs
- Wired `addUpdateCommand` into the `sdk` command group

### Updates since last revision (addressing PR review comments)
- Extracted core logic into `SdkUpdater` class with namespaced interfaces per review feedback
- Refactored `command.ts` to read from `Workspace.sdks.targets` instead of manual YAML map iteration
- Replaced `--channel` / `--include-major` flags with `--prerelease` / `--major`
- Implemented all-or-nothing writes (all version resolutions succeed before any YAML is written)
- Removed `versions.yml` entry per review feedback

1. **`SdkUpdater.context` is stored but unused** — The `context` field is set in the constructor but none of the public methods reference `this.context`; all context-dependent work is in the command layer. Intentional or should methods like `resolveUpdates` log via context?
2. **`findImageForTarget` casts target name to language key** — `name as keyof typeof LANGUAGE_TO_DOCKER_IMAGE` assumes target names always match language keys; custom target names would silently return `undefined` (changelog URL omitted, not a crash).
3. **`target.version == null` is dead code** — `Target.version` is typed as `string` (non-optional), so the null check in `collectTargetEntries` never triggers. Harmless but worth noting.
4. **All-or-nothing is at resolution level, not file write level** — Version resolutions are collected before any writes, but if the YAML `writeFile` itself fails partway, partial state could occur (unlikely but not truly atomic).
5. **No manual end-to-end testing was performed** — Command was verified to compile and register (`fern sdk update --help`), but was not tested against a real workspace with pinned SDK versions and registry access.

## Testing

- [x] Verified compilation produces no new errors
- [x] All CI checks passing (biome, lint, compile, test, test-ete, depcheck)
- [ ] Manual testing completed — requires a `fern.yml` workspace with pinned SDK versions and network access to the Fern registry

### DIFF
--- a/packages/cli/cli-v2/src/commands/sdk/update/command.ts
+++ b/packages/cli/cli-v2/src/commands/sdk/update/command.ts
@@ -1,18 +1,16 @@
-import { getLatestGeneratorVersion } from "@fern-api/configuration-loader";
 import { AbsoluteFilePath, dirname, doesPathExist, join, RelativeFilePath } from "@fern-api/fs-utils";
 import chalk from "chalk";
-import { readFile, writeFile } from "fs/promises";
+import { readFile } from "fs/promises";
 import inquirer from "inquirer";
-import { type Document, type Pair, parseDocument, Scalar, YAMLMap } from "yaml";
+import { parseDocument } from "yaml";
 import type { Argv } from "yargs";
 import { FERN_YML_FILENAME, REF_KEY } from "../../../config/fern-yml/constants.js";
 import type { Context } from "../../../context/Context.js";
 import type { GlobalArgs } from "../../../context/GlobalArgs.js";
 import { CliError } from "../../../errors/CliError.js";
 import { LANGUAGE_TO_DOCKER_IMAGE } from "../../../sdk/config/converter/constants.js";
-import { LANGUAGE_DISPLAY_NAMES, LANGUAGES, type Language } from "../../../sdk/config/Language.js";
+import { SdkUpdater } from "../../../sdk/updater/SdkUpdater.js";
 import { Icons } from "../../../ui/format.js";
-import { Version } from "../../../version.js";
 import { command } from "../../_internal/command.js";
 
 const CHANGELOG_URLS: Record<string, string> = {
@@ -26,37 +24,16 @@ const CHANGELOG_URLS: Record<string, string> = {
     "fernapi/fern-swift-sdk": "https://buildwithfern.com/learn/sdks/generators/swift/changelog"
 };
 
-interface TargetUpdate {
-    name: string;
-    language: Language;
-    image: string;
-    currentVersion: string;
-    latestVersion: string;
-}
-
-interface TargetUpToDate {
-    name: string;
-    language: Language;
-    version: string;
-}
-
-interface SkippedMajorUpgrade {
-    name: string;
-    language: Language;
-    currentVersion: string;
-    latestMajorVersion: string;
-}
-
 export declare namespace UpdateCommand {
     export interface Args extends GlobalArgs {
         /** The SDK target to update (e.g. typescript, python, go). */
         target?: string;
 
         /** Include major version upgrades. */
-        "include-major": boolean;
+        major: boolean;
 
-        /** Release channel (e.g. ga, rc). */
-        channel?: string;
+        /** Include pre-release versions (RC, alpha, beta, etc.). */
+        prerelease: boolean;
 
         /** Accept all defaults (non-interactive mode). */
         yes: boolean;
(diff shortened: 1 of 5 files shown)


packages/cli/cli-v2/src/commands/sdk/update/command.ts
```diff
@@ -0,0 +1,486 @@
...
+export declare namespace UpdateCommand {
+    export interface Args extends GlobalArgs {
+        /** The SDK target to update (e.g. typescript, python, go). */
+        target?: string;
+
+        /** Include major version upgrades. */
+        "include-major": boolean;
+
+        /** Release channel (e.g. ga, rc). */
+        channel?: string;
```
amckinney: Let's simplify this to two separate flags: `--prerelease` and `--major`. The `--prerelease` flag will permit RCs, alpha, beta, etc. For now we just support RCs. The `--major` flag is a more succinct form of `--include-major`.

packages/cli/cli-v2/src/commands/sdk/update/command.ts
```diff
@@ -0,0 +1,486 @@
...
+    name: string;
+    language: Language;
+    currentVersion: string;
+    latestMajorVersion: string;
+}
+
+export declare namespace UpdateCommand {
+    export interface Args extends GlobalArgs {
+        /** The SDK target to update (e.g. typescript, python, go). */
+        target?: string;
```
amckinney, line 30: By default, let's update _all_ generators when a `--target` is not specified. Otherwise, just update that particular generator. 

We should only write the `fern.yml` back if we can confidently update all targets (partial failures should be a no-op).

packages/cli/cli-v2/src/commands/sdk/update/command.ts
```diff
@@ -0,0 +1,486 @@
...
+    language: Language;
+    version: string;
+}
+
+interface SkippedMajorUpgrade {
+    name: string;
+    language: Language;
+    currentVersion: string;
+    latestMajorVersion: string;
+}
```
amckinney: We should define the core of this funcitonality in a `SdkUpdater.ts` class defined in `packages/cli/cli-v2/src/sdk/updater/SdkUpdater.ts`.

That will colocate the interfaces all in one-place, so that the CLI command implementation here remains focused on the orchestration and `stderr` / `stdout` output formatting (as desired).

Make sure to take inspiration from the other class structures that we define elsewhere (e.g. `packages/cli/cli-v2/src/sdk/adapter/LegacyGeneratorInvocationAdapter.ts` w/ namespacing, etc).

packages/cli/cli-v2/src/commands/sdk/update/command.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    private collectTargetEntries({
+        document,
+        targetsPath,
+        targetFilter
+    }: {
+        document: Document;
+        targetsPath: string[];
+        targetFilter: string | undefined;
+    }): Array<{ name: string; language: Language; image: string; currentVersion: string }> {
```
amckinney: Let's define an `interface` for this internal type. This can be defined in the namespace we use for the `SdkUpdater`.

packages/cli/cli-v2/src/commands/sdk/update/command.ts
```diff
@@ -0,0 +1,486 @@
...
+                continue;
+            }
+
+            entries.push({ name, language, image, currentVersion });
+        }
+
+        return entries;
+    }
+
+    private resolveLanguage(name: string, document: Document, targetsPath: string[]): Language | undefined {
```
amckinney: We should use the `Workspace` directly for these sort of resolutions.

packages/cli/cli/versions.yml
```diff
@@ -1,4 +1,16 @@
...
+  changelogEntry:
+    - summary: |
+        Add `fern sdk update` command to the CLI v2. Updates SDK target versions in
+        `fern.yml` to the latest available release from the Fern registry. Supports
+        interactive mode with target selection prompts, non-interactive mode via `--yes`,
+        filtering by `--target`, major version upgrades via `--include-major`, and
+        release channel selection via `--channel`.
+      type: feat
+  createdAt: "2026-02-26"
+  irVersion: 65
```
amckinney: Remove the changes from `versions.yml` entirely.

amckinney:
This isn't quite what I want -- closing this in favor of a separate session I'll drive myself.

PR #12907, opened 2026-02-26
chore(typescript): skip or target formatter install for source-only downloadFiles mode

Swimburger:
## Description

Refs performance analysis of the TypeScript SDK generator.

Optimizes the `downloadFiles` output mode by avoiding the full `installDependencies` call (~1.8s) when the output path only copies source files and doesn't need to compile. Uses a two-tier strategy: first check if tool binaries are already on PATH (e.g. globally installed in Docker), and if not, fall back to a targeted install of only the configured formatter/linter packages.

Link to Devin run: https://app.devin.ai/sessions/3e1c651017e8452591d37dc4bcaa323b
Requested by: @Swimburger

## Changes Made

- In `AbstractGeneratorCli.ts` `downloadFiles` handler, compute `needsBuild` by checking `outputSrcOnly` and `outputSourceFiles`. Three-way branching:
  - `needsBuild` → full `installDependencies` (unchanged)
  - Tools already on PATH → skip install entirely (new)
  - Tools not on PATH → `installCheckFixDependencies` with only the specific formatter/linter packages (e.g. `@biomejs/biome@2.4.3`, `prettier@3.7.4`, `oxfmt@0.35.0`, `oxlint@1.50.0`)
- Added `areCheckFixToolsAvailable()` to `PersistedTypescriptProject` — uses `execSync("which <binary>")` to probe PATH for each required tool binary
- Added `getCheckFixToolBinaries()` to `TypescriptProject` — returns the binary names (e.g. `["biome"]`, `["prettier", "oxlint"]`) that must be on PATH for check:fix to work without installing
- Added `getCheckFixPackages()` to `TypescriptProject` — returns exact package specifiers based on the configured `formatter` and `linter`
- Added `checkFixPackages` and `checkFixToolBinaries` fields to `PersistedTypescriptProject.Init` to carry that info through
- Added `installCheckFixDependencies()` to `PersistedTypescriptProject` — both yarn and pnpm branches pass `--ignore-scripts --prefer-offline` consistently
- When no formatter or linter is configured (`"none"`), the list is empty and the install is skipped entirely
- `shouldGenerateFullProject` intentionally still takes the full `installDependencies` path because it copies the entire project directory (including `node_modules`) to the output
- Extracted shared `TOOL_VERSIONS` constant so `getCheckFixPackages()` and `getCommonDevDependencies()` reference the same version strings — prevents drift when bumping tool versions
- Added version `3.51.6` to `versions.yml`

## Human Review Checklist

- [x] ~~**Version string duplication**~~: Resolved — `TOOL_VERSIONS` constant is now shared between `getCheckFixPackages()` and `getCommonDevDependencies()`
- [ ] **`which` portability**: `areCheckFixToolsAvailable` uses `execSync("which <binary>")`. Binary names are hardcoded strings from `getCheckFixToolBinaries()` so no injection risk, but `which` is not POSIX — `command -v` would be more portable if non-Linux environments matter.
- [ ] **Version mismatch on PATH**: If a tool exists on PATH but at a different version than `TOOL_VERSIONS`, `areCheckFixToolsAvailable` returns `true` and skips install. In Docker (controlled image) this is fine; locally it would use whatever version is installed. Verify this is acceptable.
- [ ] **`pnpm add` vs `pnpm install`**: The targeted path uses `pnpm add --save-dev` which modifies `package.json` in the temp directory. Verify this doesn't interfere with the subsequent `checkFix` command or output file copying. (The source-only output paths copy only from the `src/` directory, so the modified `package.json` should not be included in the output.)

## Testing

- [x] Generator build passes (`pnpm turbo run dist:cli --filter @fern-typescript/sdk-generator-cli`)
- [x] All CI checks green (75 pass, 0 fail) including `seed-test-results (ts-sdk)`, biome, compile, lint, test, test-ete


generators/typescript/utils/commons/src/typescript-project/PersistedTypescriptProject.ts
```diff
@@ -141,6 +152,60 @@ export class PersistedTypescriptProject {
...
+     * available on the system PATH (e.g. globally installed in Docker).
+     * When true, callers can skip installing packages entirely.
+     */
+    public areCheckFixToolsAvailable(logger: Logger): boolean {
+        if (this.checkFixToolBinaries.length === 0) {
+            return true;
+        }
+        for (const binary of this.checkFixToolBinaries) {
+            try {
+                execSync(`which ${binary}`, { stdio: "ignore" });
```
fern-support: does this need to be sync, can't we do async?

fern-support (changes requested):

fern-support (approved):

PR #12923, opened 2026-02-26
fix(typescript): use JSON.parse + skipValidation for IR parsing under 500MB

Swimburger:
## Description

Refs: Performance optimization for IR parsing across all generators.

Optimizes `parseIR` with two changes: (1) use V8's native `JSON.parse` instead of `stream-json` for files under 500 MB, and (2) skip schema validation since the IR is a trusted internal format generated by fern.

Link to Devin run: https://app.devin.ai/sessions/3e1c651017e8452591d37dc4bcaa323b
Requested by: @Swimburger

## Changes Made

- **Fast-path JSON.parse for files < 500 MB**: Checks file size via `fs.stat`, then uses `readFile` + `JSON.parse` for most real-world IRs. Falls back to `streamObjectFromFile` for files ≥ 500 MB to avoid hitting V8's `MAX_STRING_LENGTH` (~512 MB) limit.
- **`skipValidation: true`**: Skips the per-field schema validation in the IR SDK `parse()` call, treating the IR as trusted input.
- Enhanced console log to show which parsing strategy was used and the file size.

Benchmarks (synthetic JSON on this VM):

| Size | JSON.parse | stream-json | Speedup |
|------|-----------|-------------|---------|
| 1 MB | 4 ms | 35 ms | 8.1x |
| 10 MB | 44 ms | 263 ms | 5.9x |
| 50 MB | 327 ms | 1,274 ms | 3.9x |
| 100 MB | 736 ms | 2,667 ms | 3.6x |
| 400 MB | 3,842 ms | 10,882 ms | 2.8x |

## Important Review Items

- **`skipValidation: true` affects ALL generators** (TypeScript, Python, Go, Java, C#, PHP, Ruby, Swift, OpenAPI, Postman) since `parseIR` is shared in `@fern-api/base-generator`. If the IR ever has an unexpected shape (e.g., during version migrations or fern CLI bugs), errors will no longer be caught at parse time and will surface as confusing downstream failures instead.
- **500 MB threshold vs V8 limit**: `MAX_STRING_LENGTH` is 536,870,888 bytes (~512 MB). The 500 MB threshold gives ~12 MB of margin. JSON IR files are mostly ASCII so byte count ≈ character count, but worth confirming this margin is sufficient.
- **No unit tests for the threshold/fallback logic** — tested indirectly via CI seed tests only.

## Testing

- [x] `pnpm turbo run compile --filter @fern-api/base-generator` passes
- [x] `pnpm turbo run depcheck --filter @fern-api/base-generator` passes
- [x] `pnpm turbo run test --filter @fern-api/base-generator` passes (1 test)
- [x] `pnpm turbo run dist:cli --filter @fern-typescript/sdk-generator-cli` passes
- [ ] Unit tests added/updated — no direct tests for `parseIR`; relies on CI seed tests

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/12923" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/generators/base/src/utils/parseIR.ts
+++ b/generators/base/src/utils/parseIR.ts
@@ -1,4 +1,5 @@
 import { AbsoluteFilePath, streamObjectFromFile } from "@fern-api/fs-utils";
+import { readFile, stat } from "fs/promises";
 
 export declare namespace parseIR {
     export interface Args<IntermediateRepresentation> {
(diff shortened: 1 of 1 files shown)


fern-support (approved):

generators/base/src/utils/parseIR.ts
```diff
@@ -62,14 +63,37 @@ export declare namespace parseIR {
...
 export async function parseIR<IR>({ absolutePathToIR, parse }: parseIR.Args<IR>): Promise<IR> {
-    const irJson = await streamObjectFromFile(absolutePathToIR);
+    const fileStat = await stat(absolutePathToIR);
+    const useStreaming = fileStat.size >= STREAM_THRESHOLD_BYTES;
+
+    let irJson: unknown;
+    if (useStreaming) {
+        irJson = await streamObjectFromFile(absolutePathToIR);
+    } else {
+        const raw = await readFile(absolutePathToIR, "utf-8");
```
fern-support, line 86: are you sure getting the whole string into memory is the best way here?

fern-support (changes requested):

fern-support (approved):

PR #12950, opened 2026-02-27
fix(cli): support GET webhooks without request body in OpenAPI parser

### LINKED ISSUE #5964: API Reference GET Webhook
### Which Fern component?

Fern Docs

### How urgent is this?

P1 - High (Strongly needed)

### What's the issue?

in our webhooks we have a hook that can be delivered as GET or POST, 
I've created it with both methods but I get an error from fern:
`[docs]:   Skipping webhook GET callHook: Missing a request body`

GET Should not have a requestBody

### Workaround

creating the GET with a requestBody renders it but it is not then clear that the values are in the query string

### Logs & Additional Context

This is the API reference, you can see the callHook has both GET and POST methods:
```
openapi: 3.1.0
info:
  title: Webhooks
  version: 0.9.3
servers:
  - url: /
security:
  - basicAuth: []
webhooks:
  callHook:
    post:
      operationId: callHookPost
      summary: New Call
      description: This webhook is sent when a new call is created on the platform, it is configured at the appplication level
      requestBody:
        description: Information about a new call
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/CallHook'
      responses:
        '200':
          description: Return a 200 containing a JSON payload consisting of an array of verbs
    get:
      operationId: callHookGet
      summary: New Call
      description: This webhook is sent when a new call is created on the platform, it is configured at the appplication level
      parameters:
        - name: call_sid
          in: query
          required: true
          schema:
              $ref: '#/components/schemas/CallHook/properties/call_sid'
        - name: application_sid
          in: query
          required: true
          schema:
              $ref: '#/components/schemas/CallHook/properties/application_sid'
        - name: account_sid
          in: query
          required: true
          schema:
              $ref: '#/components/schemas/CallHook/properties/account_sid'
        - name: direction
          in: query
          required: true
          schema:
            $ref: '#/components/schemas/CallHook/properties/direction'
        - name: from
          in: query
          required: true
          schema:
              $ref: '#/components/schemas/CallHook/properties/from'
        - name: to
          in: query
          required: true
          schema:
              $ref: '#/components/schemas/CallHook/properties/to'
        - name: caller_id
          in: query
          required: true
          schema:
              $ref: '#/components/schemas/CallHook/properties/caller_id'
        - name: call_status
          in: query
          required: true
          schema:
            $ref: '#/components/schemas/CallHook/properties/call_status'        
        - name: sip_status
          in: query
          required: true
          schema:
            $ref: '#/components/schemas/CallHook/properties/sip_status' 
      responses:
        '200':
          description: Return a 200 containing a JSON payload consisting of an array of verbs
  registrationHook:
    post:
      operationId: registrationHookPost
      summary: Client Registration
      description: This webhook is sent when a SIP client tries to register, it is configured at the account level
      requestBody:
        description: |
          provides details of the authentication request.  The receiving server is responsible for authenticating the 
          request as per [RFC 2617](https://tools.ietf.org/html/rfc2617)
        content:
          application/json:
            schema:
              required:
                - method
                - realm
                - username
                - expires
                - nonce
                - uri
                - response
              type: object
              properties:
                method:
                  type: string
                  description: sip request method
                  example: REGISTER
                realm:
                  type: string
                  description: sip realm
                  example: mycompany.com
                username:
                  type: string
                  description: sip username provided
                  example: daveh
                expires:
                  type: number
                  description: expiration requested, in seconds
                  example: 3600
                scheme:
                  type: string
                  description: encryption protocol
                  example: digest
                nonce:
                  type: string
                  description: nonce value
                  example: InFriVGWVoKeCckYrTx7wg=="
                uri:
                  type: string
                  format: uri
                  description: sip uri in request
                  example: sip:mycompany.com
                algorithm:
                  type: string
                  description: encryption algorithm used, default to MD5 if not provided
                  example: MD5
                qop:
                  type: string
                  description: qop value
                  example: auth
                cnonce:
                  type: string
                  description: cnonce value
                  example: 6b8b4567
                nc:
                  type: string
                  description: nc value
                  example: 1
                response:
                  type: string
                  description: digest value calculated by the client
                  example: be641cf7951ff23ab04c57907d59f37d
      responses:
        '200':
          description: |
            Your callback should return this HTTP status code in all cases.
            if the request was authenticated and you wish to admit 
            the client to the network, this is indicated by setting the 'response' 
            attribute in the body to 'ok'
          content:
            application/json:
              schema:
                type: object
                required:
                  - status
                properties:
                  status:
                    type: string
                    description: indicates whether the request was successfully authenticated
                    enum:
                      - ok
                      - fail
                    example: ok
                  message:
                    type: string
                    description: a human-readable message
                    example: authentication granted
                  call_hook:
                    type: string
                    format: url
                    description: url of application to invoke when this device places a call
                  expires:
                    type: number
                    description: |
                      The expires value to grant to the requesting user.
                      If not provided, the expires value in the request is observed.
                      If provided, must be less than the requested expires value.
                  blacklist:
                    type: number
                    description: |
                      If provided, represents a period in seconds during which the source IP 
                      address should be blacklisted by the platform (0 means forever).
components:
  schemas:
    CallHook:
      required:
        - call_sid
        - application_sid
        - account_sid
        - direction
        - from
        - to
        - caller_id
        - call_status
        - sip_status
      properties:
        call_sid:
          type: string
          example: "d2515c3b-b79a-41a2-971a-445e769c823c"
          description: a unique identifier for the call.
        application_sid:
          type: string
          example: "72c5c38f-9bba-40ce-aa83-aaa6be55e1b5"
          description: a unique identifier for the jambonz application controlling this call
        account_sid:
          type: string
          example: "bad98250-b34d-459d-9e90-f97dfb9bc519"
          description: a unique identifier for the jambonz account associated with the application
        direction:
          type: string
          enum: [inbound, outbound ]
          description: >
            the direction of the call:
              * `inbound` - The call was originated outside jambonz
              * `outbound` - The call was originated by jambonz
        from:
          example: "+12125551212"
          type: string
          description: the calling party number
        to:
          example: "+14155551234"
          type: string
          description: the called party number
        caller_id:
          type: string
          example: "John Doe"
          description: the caller name, if known
        sip_status:
          type: number
          example: 100
          description: the most recent sip status code received or generated for the call
        call_status:
          type: string
          enum: [trying, ringing, early-media, in-progress, completed, failed, busy, no-answer ]
          description: >
            current status of the call:
              * `trying` - a new incoming call has arrived or an outbound call has just been sent
              * `ringing` - a 180 Ringing response has been sent or received
              * `early-media` - an early media connection has been established prior to answering the call (183 Session Progress)
              * `in-progress` - call has been answered
              * `completed` - an answered call has ended
              * `failed` - a call attempt failed
              * `busy` - a call attempt failed because the called party returned a busy status
              * `no-answer` - a call attempt failed because it was not answered in time
  securitySchemes:
    basicAuth:
      type: http
      scheme: basic
```

Swimburger:
## Description

Closes https://github.com/fern-api/fern/issues/5964

GET webhooks that use query parameters instead of a request body were silently dropped during OpenAPI parsing with the error `Skipping webhook GET callHook: Missing a request body`. This is because the `requestBody == null` check ran before the method check and unconditionally returned early.

## Changes Made

- Moved the HTTP method validation (`POST`/`GET` check) **before** the request body null check so GET webhooks aren't rejected prematurely
- Unsupported methods (e.g. PUT) now log a `warn` instead of `error` before skipping
- For webhooks without a request body, synthesize an object schema payload from query parameters instead of skipping
- Hoisted `signatureVerification` computation out of the request-body loop since it's now shared by both code paths
- Added unit tests for `convertWebhookOperation` covering GET/POST webhooks with and without request bodies and query parameters (7 test cases)
- Updated IR generator test snapshots (`exhaustive.json`, `imdb.json`) after merging main
- Added changelog entry (v3.91.4)

- Removed invalid `description: qp.description` from synthesized `ObjectPropertyWithExample` objects — the interface has no `description` field, so the value was silently stripped during serialization. Query parameter descriptions are carried inside `qp.schema` via the schema conversion pipeline. (Flagged by Devin Review)
- Changed unsupported HTTP method handling from `logger.error` to `logger.warn`, since the IR type `WebhookHttpMethod` only supports GET/POST and expanding it would require an IR schema change
- Updated IR generator test snapshots after merging main (unrelated `availability` field changes)
- Bumped version to 3.91.4 (3.91.2 and 3.91.3 were taken by other PRs merged to main)

## Human Review Checklist

- [ ] **POST webhooks without a body** now also go through the "synthesize from query params" path rather than immediately erroring. They'll still error if there are zero query params — confirm this fallback behavior is acceptable.
- [ ] **Query parameter descriptions**: The `description` property was removed from the `ObjectPropertyWithExample` literal since it's not part of the interface. Descriptions should be embedded in `qp.schema` by `convertParameters` — worth verifying they propagate correctly to downstream output (docs, SDK generators).
- [ ] **Downstream consumers** (docs rendering, SDK generators) will see the synthesized object payload as a normal webhook payload — verify they render it sensibly for GET webhooks.

## Testing
- [x] Unit tests added/updated — 7 test cases in `convertWebhookOperation.test.ts`
- [x] Manual testing completed — built local CLI and ran `fern check` against an OpenAPI spec with GET webhooks using query parameters. Previously errored with "Missing a request body"; now passes successfully.
- [x] IR generator test snapshots updated

---

Link to Devin run: https://app.devin.ai/sessions/e3b0262e7edb43ba9edfdf879cc2213e
Requested by: @Swimburger


packages/cli/api-importers/openapi/openapi-ir-parser/src/openapi/v3/converters/operation/convertWebhookOperation.ts
```diff
@@ -42,8 +43,8 @@ export function convertWebhookOperation({
         source
     });
 
-    if (operation.requestBody == null) {
-        context.logger.error(`Skipping webhook ${method.toUpperCase()} ${path}: Missing a request body`);
+    if (method !== "POST" && method !== "GET") {
```
fern-support, line 46: Is Post or Get required for webhooks?

fern-support: In that case, don't make it mandatory.

fern-support (approved):

fern-support (approved):

Issue #5926, opened 2025-02-07
webhooks, operationId required


sammachin:
### Which Fern component?

SDK Generator

### How urgent is this?

P2 - Medium (Would be helpful)

### What's the issue?

create an api spec using the oas webhook example
```
openapi: 3.1.0
info:
  title: Webhook Example
  version: 1.0.0
# Since OAS 3.1.0 the paths element isn't necessary. Now a valid OpenAPI Document can describe only paths, webhooks, or even only reusable components
webhooks:
  # Each webhook needs a name
  newPet:
    # This is a Path Item Object, the only difference is that the request is initiated by the API provider
    post:
      requestBody:
        description: Information about a new pet in the system
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Pet'
      responses:
        '200':
          description: Return a 200 status to indicate that the data was received successfully
components:
  schemas:
    Pet:
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string

```

### Workaround

When creating a webhook api spec I get an error `Skipping webhook callHook because no operation id present`

as I understand it operationId is an optional value

Swimburger:
This has been fixed in [v3.95.1](https://github.com/fern-api/fern/pull/12952). operationId is now optional for webhooks — when not provided, the webhook key name (e.g. newPet from webhooks.newPet) is used as the identifier. This applies to both the OpenAPI 3.1 parser and the legacy parser.

PR #12952, opened 2026-02-27
fix(openapi): make operationId optional for webhooks in OpenAPI 3.1 specs

Swimburger:
## Description

Refs https://github.com/fern-api/fern/issues/5926

The OpenAPI 3.1 importer (`openapi-to-ir`) was skipping webhooks that didn't have an `operationId`, even though `operationId` is optional per the OpenAPI specification. The older `openapi-ir-parser` already handles this but generated hash-based IDs (e.g. `newpetPost82bd5e50`) instead of using the webhook key name directly.

## Changes Made

**V3 parser (`openapi-to-ir`):**
- Use the webhook key name (e.g. `newPet` in `webhooks.newPet`) as the fallback `operationId` when one is not explicitly set, instead of skipping the webhook entirely
- Set the fallback `operationId` directly on the operation object before passing it to `WebhookConverter`, so that `evaluateMethodNameFromOperation()` uses the key name as-is (producing `.newPet`) rather than generating a method-prefixed name (`.postNewPet`)
- Added a null check for `webhookItem.post` to tighten type narrowing alongside the existing `"post" in webhookItem` guard

**Legacy parser (`openapi-ir-parser`):**
- Updated `generateWebhookOperationId()` to use Option A with fallback (per PR review feedback):
  - Return `sdkMethodName.methodName` if available (existing behavior)
  - Return the webhook key name directly if it's clean (no special characters stripped during sanitization)
  - Fall back to `sanitized + method + hash` only when special characters were removed, for uniqueness

**Other:**
- Added changelog entry (version 3.95.1)
- Added a `newPet` webhook **without** `operationId` to the existing `webhooks` test fixture in both `v3-importer-tests` and `openapi-ir-to-fern-tests`, with updated snapshots

## Testing

- [x] Snapshot tests added — a webhook without `operationId` added to existing `webhooks` fixture across both parsers (v3 and legacy)
- [x] All 4 affected test suites pass (`v3-sdks`, `baseline-sdks`, `openapi-ir`, `openapi-ir-to-fern`)
- [x] Type-checked locally; no new errors introduced (pre-existing errors confirmed on `main`)
- [x] Manually tested with a webhook-only OpenAPI 3.1 spec — `fern check` passes
- [x] All CI checks passing (47/47)

## Human Review Checklist

- **Breaking change in legacy parser webhook IDs**: The `generateWebhookOperationId()` change affects all webhooks without an explicit `operationId` in the legacy parser — not just new ones. Previously `newPet` would produce `newpetPost82bd5e50`; now it produces `newPet`. If any downstream consumers depend on the old hash-based IDs, this is a breaking change. Confirm this is acceptable.
- The fallback `operationId` is injected onto the operation via a shallow spread (`{ ...webhookItem.post, operationId }`). Confirm this doesn't cause issues with any downstream code that mutates or deeply inspects the operation object. The original `webhookItem.post` is untouched when `operationId` is already present.
- The legacy parser's `sanitized === path.toLowerCase()` comparison assumes `sanitizePathExpression()` lowercases the path. Verify this is always true, or edge cases could produce incorrect IDs.
- Webhook key names with special characters (e.g. `/pet`) used as a fallback operationId could produce unexpected IDs. The existing `/pet` fixture has an explicit `operationId: CreatePet` so this path isn't exercised with slashes.

---

[Link to Devin run](https://app.devin.ai/sessions/ab80e2cd531944dd85443bc46bbf4c68) | Requested by: @Swimburger


packages/cli/api-importers/v3-importer-tests/src/__test__/__snapshots__/v3-sdks/webhooks.json
```diff
@@ -260,6 +260,83 @@
           }
         }
       }
+    ],
+    "newPet": [
+      {
+        "id": ".postNewPet",
```
fern-support: this ID format I don't like

packages/cli/api-importers/v3-importer-tests/src/__test__/__snapshots__/baseline-sdks/webhooks.json
```diff
@@ -545,6 +545,226 @@
             }
           }
         ]
+      },
+      {
+        "id": "webhooks_.newpetPost82bd5e50",
```
fern-support: Whatever the name is will be used for docs and SDKs, it needs to be good looking, but not conflicting at the same time.

fern-support: Give me a a pro con list

fern-support: Option A, with a fallback

fern-support (approved):

PR #12987, opened 2026-02-27
fix(internal): create and validate workspace once per generator

patrickthornton:
## Description
Previously, the `validateWorkspace` step that ran during `fern generate` ran in the CLI before instantiating the workspace for the particular generator being run. This has two big problems:
1. That you can override API parsing on a per-generator basis, which means different generator runs will have different Fern definitions they need to validate by design. [This means it's theoretically impossible to share a validation step from language to language in all cases.]
2. `toFernWorkspace()` was called twice, once for the validation step, and once for the generation proper. This meant that the four-step `Spec -> OpenAPI IR -> Fern Definition -> IR -> SDK` pipeline, the `Spec -> OpenAPI IR -> Fern Definition` half happened twice on every generation (!), since that's what goes into creating a workspace.

Now the validation step is a callback that's called with the workspace created per-generator in the generation step, so only one workspace is ever used. This is more correct and will cut down on duplicate first-half-of-the-pipeline parsing.

## Changes Made
To the local and remote generation files in `packages/cli`.

## Testing
Tested that debug logs are no longer duplicated in generation runs.


packages/cli/cli/src/commands/generate/generateAPIWorkspace.ts
```diff
@@ -91,13 +92,14 @@ export async function generateWorkspace({
...
-        ossWorkspace: workspace instanceof OSSWorkspace ? workspace : undefined
-    });
+    const validateWorkspace = async (fernWorkspace: FernWorkspace): Promise<void> => {
+        await validateAPIWorkspaceAndLogIssues({
+            workspace: fernWorkspace,
+            context,
+            logWarnings: false,
+            ossWorkspace: workspace instanceof OSSWorkspace ? workspace : undefined
+        });
+    };
```
amckinney: nit: Given that this is the only instance of the `validateWorkspace` closure that we pass in, it might be better to define this as a top-level function/utility, then simply pass in a `validateWorkspace: boolean` flag that instructs the downstream component to run the function. Otherwise tracing back the code to see what closure was passed in would be tougher to debug.

Thoughts?

amckinney: Ah this might be the only place where we have access to the parameters we need (i.e. both the OSS and Fern Workspace).

amckinney (approved):

amckinney (approved):

PR #13020, opened 2026-02-28
chore(ci): add workflow to rerun failed tests on seed PRs

davidkonigsberg:
## Description

Replaces the hello-world stub (merged in #13021) with the full implementation of a GitHub Actions workflow that automatically finds open seed PRs with a small number of transient test failures and re-runs them. This addresses cases like [#13008](https://github.com/fern-api/fern/pull/13008) where `test-ete` fails due to rate limiting or other transient issues, blocking auto-merge of seed update PRs.

[Link to Devin run](https://app.devin.ai/sessions/7b1e4ea814434e2eb597ea647b71c549)

## Changes Made

- Replaced the hello-world stub in `.github/workflows/rerun-seed-tests.yml` with the full workflow logic:
  - Adds a `schedule` trigger (every 2 hours via cron) alongside the existing `workflow_dispatch`
  - Finds all open PRs labeled `seed`
  - For each PR, checks: (1) PR is approved, (2) all CI checks are completed, (3) exactly 1–3 checks failed
  - For qualifying PRs, re-runs failed jobs on workflow runs that have been attempted fewer than 3 times
  - Waits 60 seconds between each rerun trigger to avoid rate limiting
- Uses only `gh` CLI and GitHub API calls — no checkout or build steps needed
- Job timeout set to 30 minutes to accommodate sleep delays between reruns

- Added 60-second sleep between each rerun trigger to avoid rate limiting
- Bumped job `timeout-minutes` from 10 → 30 to accommodate the added sleep delays
- (Previous) Fixed `headRefSha` → `headRefOid` (correct field name for `gh pr list --json`)
- (Previous) Fixed pagination: paginated API responses now use `--jq '.check_runs[]' | jq -s '.'` to properly flatten multi-page results into a single array

## Testing

- [ ] Manual testing completed (trigger via Actions → "Rerun Failed Seed PR Tests" → "Run workflow" on `main` and verify it processes seed PRs correctly)

## Human Review Checklist

- **Rate-limit sleep placement**: The `sleep 60` runs after every rerun trigger including the last one in each PR's loop, adding an unnecessary wait at the end. This is harmless but worth noting.
- **Subshell error propagation**: The `while read` loop runs in a subshell (due to pipe from `echo "$PRS"`). If a command inside the loop fails unexpectedly, it won't cause the outer script to fail despite `set -e`. Verify this is acceptable behavior or consider using a `for` loop with process substitution instead.
- **Permissions**: Confirm `actions: write` on `GITHUB_TOKEN` is sufficient for `gh run rerun`. The existing `rerun.yml` workflow uses `github.token` similarly.
- **Failure thresholds**: The 1–3 failure range and max 3 attempts cap — verify these match the desired policy.
- **Schedule frequency**: Every 2 hours (`0 */2 * * *`) — confirm this is appropriate.
- **Timeout adequacy**: 30 minutes should be sufficient for processing several PRs with 60s sleeps, but could be tight if many seed PRs qualify simultaneously.

### DIFF
--- a/.github/workflows/rerun-seed-tests.yml
+++ b/.github/workflows/rerun-seed-tests.yml
@@ -11,9 +11,6 @@ permissions:
   pull-requests: read
   checks: read
 
-env:
-  DO_NOT_TRACK: "1"
-
 jobs:
   rerun-failed-seed-tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/rerun-seed-tests.yml
+++ b/.github/workflows/rerun-seed-tests.yml
@@ -28,7 +28,7 @@ jobs:
           MAX_FAILURES=3
 
           echo "Fetching open PRs with 'seed' label..."
-          PRS=$(gh pr list --label seed --state open --json number,headRefSha --jq '.[] | "\(.number) \(.headRefSha)"')
+          PRS=$(gh pr list --label seed --state open --json number,headRefOid --jq '.[] | "\(.number) \(.headRefOid)"')
 
           if [ -z "$PRS" ]; then
             echo "No open PRs with 'seed' label found."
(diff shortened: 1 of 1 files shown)


davidkonigsberg:
no longer needed

Issue #4425, opened 2024-08-26
[Bug] CLI incorrectly reports "misconfigured fern directory" when not logged in


zachkirsch:
When we run `fern` in CI without a token, we get an error:

```
Misconfigured fern directory: please see the docs at https://buildwithfern.com/learn/api-definition/introduction/what-is-the-fern-folder
Login required.
```

<img width="1067" alt="Screenshot 2024-08-26 at 11 41 56 AM" src="https://github.com/user-attachments/assets/05c10cb6-adee-4a19-b1c5-04b28f4b0e60">

`fern` is working fine locally so I don't think the fern directory is actually misconfigured.

neehhaa06:
Hello @zachkirsch , would you assign for this work?
Thanks! :)

PR #13033, opened 2026-02-28
fix(cli): improve "Misconfigured fern directory" error message with actionable diagnostics

Swimburger:
## Description

Refs https://github.com/fern-api/fern/issues/4425

Replaces the generic "Misconfigured fern directory" error with an actionable message that tells users exactly what was expected, and adds debug-level diagnostics explaining which workspace checks failed. The error message is intentionally format-agnostic (says "API spec(s)" rather than naming specific formats like OpenAPI/AsyncAPI) since `generators.yml` has supported many configuration variations over time.

**Link to Devin run:** https://app.devin.ai/sessions/b918846a24f84eb6b3e9bc477cfa4f6b
**Requested by:** @Swimburger

## Changes Made

- **Replaced error message** in both copies of `handleFailedWorkspaceParserResult.ts`: the old `"Misconfigured fern directory: please see the docs at ..."` is now a structured message listing what was expected (an `api` section in `generators.yml` or a `definition/` directory)
- **Added debug-level diagnostics** in `loadAPIWorkspace.ts` that log why workspace loading fell through (e.g. "No generators.yml was found", "generators.yml was found but does not contain an api section", "has an api section but no valid API specs could be resolved from it")
- Updated ETE test snapshots to match the new error message
- Added `versions.yml` entry (3.96.1)
- **Added unit tests** for the `MISCONFIGURED_DIRECTORY` failure path covering:
  - `loadAPIWorkspace` returns `MISCONFIGURED_DIRECTORY` when generators.yml has no api section
  - `loadAPIWorkspace` returns `MISCONFIGURED_DIRECTORY` when generators.yml is empty
  - `handleFailedWorkspaceParserResult` outputs the new actionable error (and not the old generic message)
  - Debug-level diagnostic messages are emitted with details about what was checked

## Important review notes

- There are **two copies** of `handleFailedWorkspaceParserResult.ts` (one in `workspace/loader/`, one in `lazy-fern-workspace/utils/`). Both were updated consistently — worth verifying they stay in sync.
- The `doesPathExist(DEFINITION_DIRECTORY)` call in the new diagnostic block is redundant with the check on line ~321, since we only reach the fallback if that check was false. This is harmless (error path, not hot path) but could be simplified.
- This improves the error message but does **not** change the underlying architecture where auth errors and workspace loading errors can co-occur (the full scenario from #4425). The two code paths remain independent.
- The `generators-empty-definitions` test fixture was created but is not directly exercised by a test — may want to either add a test or remove it.
- The diagnostic logic intentionally avoids inspecting the internal structure of the `api` section (e.g. checking specific spec types or `definitions.length`). It only distinguishes three cases: no generators.yml, no `api` key, or `api` present but nothing resolved. This keeps it resilient to past and future `generators.yml` schema changes.

## Human review checklist

- [ ] Verify both copies of `handleFailedWorkspaceParserResult.ts` produce identical error messages
- [ ] Confirm `generators-empty-definitions` fixture is intentionally unused or should have a test
- [ ] Verify the diagnostic hints are generic enough to handle all `generators.yml` configuration variations (old and new formats)

## Testing

- [x] Biome lint/format passes
- [x] ETE test snapshots updated
- [x] Unit tests added for `MISCONFIGURED_DIRECTORY` failure path (3 new tests)
- [x] Local CLI build + manual verification of new error output
- [x] CI passing (47/47 checks green)


fern-support (approved):

Swimburger:
This has been fixed in [#13033](https://github.com/fern-api/fern/pull/13033) (shipping in 3.96.1). The generic "Misconfigured fern directory" message is now replaced with a clear, actionable error:

No API definition found. Expected one of the following:
  - An api section in generators.yml pointing to your API spec(s)
  - A definition/ directory with Fern Definition files
For more information, see https://buildwithfern.com/learn/api-definition/introduction/what-is-the-fern-folder
We also added debug-level diagnostics that explain exactly which checks failed during workspace loading, so --log-level debug will show whether generators.yml was missing, had no api section, or had an api section that didn't resolve to any specs.

Issue #13065, opened 2026-03-02
[Python] Path parameters are not url encoded


ouvreboite:
### Which Fern component?

SDK Generator

### How urgent is this?

P2 - Medium (Would be helpful)

### What's the issue?

I'm an indirect consumer of fern, as I'm using langfuse's python SDK which is generated by fern.
It seems the path parameters are not properly encoded.

For example, in the generated SDK, the path paremeters are processed by the `jsonable_encoder` function
 ```python
_response = self._client_wrapper.httpx_client.request(
            f"api/public/datasets/{jsonable_encoder(dataset_name)}/runs/{jsonable_encoder(run_name)}",
```
(From [langfuse/langfuse-python](https://github.com/langfuse/langfuse-python/blob/9efcca02013d2fd6ce455087a5f53d76ac39c5c1/langfuse/api/datasets/raw_client.py#L395))

But jsonable_encoder does not do any processing for a string input, and returns it as-is:
```python
 if isinstance(obj, (str, int, float, type(None))):
        return obj
```
(From [fern-api/fern](https://github.com/fern-api/fern/blob/70aad1b5eb17b65b4a7e8939ecdbc5f8b111796f/generators/python/core_utilities/shared/jsonable_encoder.py#L64C4-L65C19))
So if a path paremeter contains a special character (ex; "#"), the call fails.

Instead it should url encoded it (when it's used as a path/search parameters), for example using `urllib.parse.quote()`

### Are you interested in contributing a fix?

No

PR #13083, opened 2026-03-03
fix(python): Path parameters are not url encoded

HoaX7:
## Description
Refs: https://github.com/fern-api/fern/issues/13065
- Correctly encodes path parameters in generated python sdk.

## Changes Made
- [x] Added `quote` static method to `urllib_parse.py`
- [x] Added `encode_path_parameter` to `core_utilities.py`
- [x] Added test cases

## Testing
- `pnpm seed test --generator python-sdk --fixture path-parameters`
- `cd seed/python-sdk/path-parameters`
- `poetry install && poetry run pytest tests -v`

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13083" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


iamnamananand996:
hey @HoaX7, looks good to me, 

can we also add the generated files from 

`pnpm seed test --generator python-sdk --fixture path-parameters`

and update this version changelog https://github.com/fern-api/fern/blob/main/generators/python/sdk/versions.yml

iamnamananand996 (approved):

seed/python-sdk/api-wide-base-path/src/seed/service/raw_client.py
```diff
@@ -42,7 +43,7 @@ def post(
         HttpResponse[None]
         """
         _response = self._client_wrapper.httpx_client.request(
-            f"test/{jsonable_encoder(path_param)}/{jsonable_encoder(service_param)}/{jsonable_encoder(endpoint_param)}/{jsonable_encoder(resource_param)}",
+            f"test/{urllib.parse.quote(str(jsonable_encoder(path_param)), safe='')}/{urllib.parse.quote(str(jsonable_encoder(service_param)), safe='')}/{urllib.parse.quote(str(jsonable_encoder(endpoint_param)), safe='')}/{urllib.parse.quote(str(jsonable_encoder(resource_param)), safe='')}",
```
patrickthornton: could we extract into some kind of shared utility or redesign how this string is constructed? code like this looks kind of odd for the user. (i recognize that it already looked pretty odd before, but i'd rather overhaul it to make it clean instead of sticking to a cumbersome design)

generators/python/sdk/versions.yml
```diff
@@ -10,6 +10,16 @@
...
+- version: 4.61.4
+  changelogEntry:
+    - summary: |
+        Fix path parameters not being URL encoded. Path parameters containing
+        special characters are now correctly URL-encoded using `urllib.parse.quote`
+        before being embedded into request URLs.
+      type: fix
+  createdAt: "2026-03-03"
+  irVersion: 65
+
```
patrickthornton, line 272: changelog changes should always add a new, latest version!

HoaX7: hey @patrickthornton could you please verify if I need to bump the `irVersion`? all the previous summaries had `irVersion: 65` so i didnt change this.

Edit: I misunderstood. there was a merge conflict. I have moved the version summary to the top.

seed/python-sdk/api-wide-base-path/src/seed/service/raw_client.py
HoaX7: sure, i'll create a `encodeUrl` utility and pass the url string to that function instead of doing it this way.

HoaX7: @patrickthornton I've now moved the logic to a new file `path_encoder.py`, let me know if any more changes are needed

generators/python/core_utilities/shared/path_encoder.py
```diff
@@ -0,0 +1,8 @@
+import typing
+import urllib.parse
+
+from .jsonable_encoder import jsonable_encoder
+
+
+def encode_path_parameter(value: typing.Any) -> str:
+    return urllib.parse.quote(str(jsonable_encoder(value)), safe="")
```
amckinney, line 10: Note that if we pass in `None` as `value`, we'll encode `None` in the path (e.g. `/user/None`). Should we return some sort of internal error that the user can act upon rather than getting a 404? That way it's on us in the SDK to make sure we prevent this from happening.

HoaX7: do we need this check? there is no check in `jsonable_encoder` which was previously used directly.

Edit: I've added the check. with message

```python
def encode_path_parameter(value: typing.Any) -> str:
    if value is None:
        raise ValueError(
            "Path parameter value cannot be 'None', this would result in a malformed URL."
        )
    return urllib.parse.quote(str(jsonable_encoder(value)), safe="")
```

PR #13164, opened 2026-03-05
feat(cli-v2): add `fern api merge`, `split`; drop `parentpath`

patrickthornton:
## Description
Adds the `fern api` subcommand, in which now lie four subcommands:
```
fern api

Configure, compile, edit, and inspect your API specs

Commands:
  fern api check    Validate API definitions
  fern api compile  Compile your API specs
  fern api merge    Flatten overlay and override files into the base API spec
  fern api split    Extract changes from a modified spec into an overlay or override file, restoring the spec to
                    its git HEAD state
```
`compare` is the old command for working with overrides, which requires you to duplicate your spec, edit the second one, remember which one is the edited one and the non-edited one, and then you can generate an overrides file from the difference; we've cut it in favor of a new workflow. `merge` and `split` are a new workflow, where `merge` merges API files with their overrides files, and `split` splits off any git-tracked changes to an API file into an overrides file, restoring the source API file to its git HEAD state. The idea is that `merge` and `split` are dual or inverse operations, which we explicitly test for.

These functions can work with overrides or overlays; by default, `split` generates overlays, which I'd prefer for their indexing and removal syntax. [Note: worth making sure we can stipulate multiple overlays for a single API file as well as multiple overrides.]

The full interface is:
```
  fern api merge

  Flatten overlay and override files into the base API spec

  Options:
    --api       Filter by API name                                        [string]
    --remove    Remove overlay/override files and their references
                from fern.yml after merge                    [boolean, default: false]

  fern api split

  Extract changes from a modified spec into an overlay or override file,
  restoring the spec to its git HEAD state

  Options:
    --api          Filter by API name                                     [string]
    --output    Custom output path for the new overlay/override file   [string]
    --format       Output format: 'overlays' (OpenAPI Overlay 1.0.0)
                   or 'overrides' (Fern deep-merge)
                                      [string, choices: "overlays" | "overrides", default: "overlays"]
```
`split` always merges into the overrides file if it already exists. If `--output` isn't specified, it defaults to `[API-FILE-NAME]-overrides.yml` or `[API-FILE-NAME]-overlays.yml`.

There's also a simplification here which could be wrong; the `parentPath` parameter passed between functions I think can be cut because `yargs` deals with that automatically.

## Testing
New test suite for all commands + the merge/split round-trip.


packages/cli/cli-v2/src/commands/auth/login/command.ts
```diff
@@ -82,7 +82,6 @@ export function addLoginCommand(cli: Argv<GlobalArgs>, parentPath?: string): voi
                     description: "Use device code flow (for environments where browser cannot open automatically)"
                 })
                 .example("$0 auth login", "Log in via browser")
-                .example("$0 auth login --device-code", "Log in via device code"),
-        parentPath
```
amckinney, line 86: Thanks for dropping this -- meant to take care of it in the [last PR](https://github.com/fern-api/fern/pull/13154) where it was no longer needed.

packages/cli/cli-v2/src/commands/api/overrides/split/gitUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const execFileAsync = promisify(execFile);
+
+let cachedRepoRoot: string | undefined;
+
+/**
+ * Returns the file content from git HEAD for the given absolute path.
+ * Throws a CliError if the file is not tracked by git.
+ */
+export async function getFileFromGitHead(absolutePath: AbsoluteFilePath): Promise<string> {
```
amckinney: If these `gitUtils` are scoped to the `SplitCommand`, let's just make them private methods there. That way everything is all in one place and we don't have separate files lingering around. If it's a common utility, it should be hoisted up outside the `commands/...` directory.

packages/cli/cli-v2/src/commands/api/split/diffSpecs.ts
```diff
@@ -0,0 +1,104 @@
+import { isDeepStrictEqual } from "node:util";
+
+// biome-ignore lint/suspicious/noExplicitAny: OpenAPI specs can have any shape
+type OpenAPISpec = Record<string, any>;
+
+/**
+ * Generates an overrides object by comparing original and modified specs.
+ * Only includes paths/properties that have been added or changed in the modified spec.
+ */
+export function generateOverrides(original: OpenAPISpec, modified: OpenAPISpec): OpenAPISpec {
```
amckinney, line 54: Same thing as below w.r.t. moving all `compare`-specific functionality into the `CompareCommand` class.

packages/cli/cli-v2/src/commands/api/overrides/compare/command.ts
```diff
@@ -0,0 +1,79 @@
...
+import { loadSpec, serializeSpec } from "../_shared/loadSpec.js";
+import { generateOverrides } from "./diffSpecs.js";
+
+interface DiffArgs extends GlobalArgs {
+    original: string;
+    modified: string;
+    output?: string;
+}
+
+export function addCompareCommand(cli: Argv<GlobalArgs>): void {
```
amckinney: Let's follow the convention used in other CLI commands (i.e. this could all be contained within a single `CompareCommand` type that namespaces its args, private methods scoped to this command, etc.

packages/cli/cli-v2/src/commands/api/overrides/compare/command.ts
```diff
@@ -0,0 +1,79 @@
...
+    command<DiffArgs>(
+        cli,
+        "compare <original> <modified>",
+        "Compare two OpenAPI specs and generate an overrides file",
```
amckinney: Should we just call the command `diff`? That way it's more consistent with the standard unix command, etc.

packages/cli/cli-v2/src/commands/api/overrides/compare/command.ts
```diff
@@ -0,0 +1,79 @@
...
+async function handleDiff(context: Context, args: DiffArgs): Promise<void> {
+    // resolveOutputFilePath always returns non-null when given a non-null string
+    const originalPath = context.resolveOutputFilePath(args.original) as AbsoluteFilePath;
+    const modifiedPath = context.resolveOutputFilePath(args.modified) as AbsoluteFilePath;
+
+    context.stdout.info(chalk.dim(`  Comparing ${originalPath} with ${modifiedPath}`));
+
+    const [original, modified] = await Promise.all([loadSpec(originalPath), loadSpec(modifiedPath)]);
+
+    const overrides = generateOverrides(original, modified);
```
amckinney: Oh this command is meant to generate the overrides file? I don't think this is `compare` or `diff`.

packages/cli/cli-v2/src/commands/api/overrides/merge/command.ts
```diff
@@ -0,0 +1,116 @@
...
+        (yargs) =>
+            yargs
+                .option("api", {
+                    type: "string",
+                    description: "Filter by API name"
+                })
+                .option("spec", {
+                    type: "string",
+                    description: "Filter by spec file path"
+                })
```
amckinney: Thoughts on starting without this for now? We can always add it later as needed. I wonder if we should just support merging all the specs defined in an API to start.

packages/cli/cli-v2/src/commands/api/overrides/merge/command.ts
```diff
@@ -0,0 +1,114 @@
...
+                .option("cleanup", {
```
amckinney: Thinking out loud, what about the following atlernatives:
1. `--clear`
2. `--clean`
3. `--delete`
4. `--prune`

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,134 @@
...
+                    description: "Filter by API name"
+                })
+                .option("spec", {
+                    type: "string",
+                    description: "Filter by spec file path"
+                })
+                .option("merge-into", {
+                    type: "string",
+                    description: "Merge the extracted diff into this existing override file"
+                })
```
amckinney: Can we just specify `--merge` as a boolean flag since we already know where the overrides file is? We don't need them to re-specify.

packages/cli/cli-v2/src/commands/api/overrides/split/command.ts
```diff
@@ -0,0 +1,134 @@
...
+        (yargs) =>
+            yargs
+                .option("api", {
+                    type: "string",
+                    description: "Filter by API name"
+                })
+                .option("spec", {
+                    type: "string",
+                    description: "Filter by spec file path"
+                })
```
amckinney: Similar to above (re: starting without this).

packages/cli/cli-v2/src/commands/api/overrides/split/command.ts
```diff
@@ -0,0 +1,134 @@
...
+                .option("merge-into", {
+                    type: "string",
+                    description: "Merge the extracted diff into this existing override file"
+                })
+                .option("output", {
+                    type: "string",
+                    alias: "o",
```
amckinney: I was just thinking about this earlier -- let's start without aliases to start. I'd rather add them in a follow-up right before release so we can ensure it's all consistent and/or wait to add aliases until later. This is part of the API surface, so we can always add aliases later, but once we do they need to last forever.

amckinney (commented):
Nice, looking good so far -- mostly just style comments on maintaining consistency with the current CLI structure, but some other notes on flag naming / functionality too.

packages/cli/cli-v2/src/commands/api/overrides/split/gitUtils.ts
patrickthornton: agreed!

packages/cli/cli-v2/src/commands/api/split/diffSpecs.ts
patrickthornton: this gets used by split too, so moving to a shared dir

packages/cli/cli-v2/src/commands/api/split/command.ts
patrickthornton: i'm accounting for the potential of there being multiple overrides files, but this UX could definitely be improved (having both --merge-into and --output is confusing)

packages/cli/cli-v2/src/commands/api/overrides/merge/command.ts
patrickthornton: will go with --delete!

packages/cli/cli-v2/src/commands/api/overrides/compare/command.ts
patrickthornton: FYI for the future: 'compare' was an outdated CLI-v1 command we're cutting in favor of merge/split

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,282 @@
...
+
+            if (format === "overlays") {
+                await this.splitAsOverlay(context, editor, entry, originalContent, currentContent, args.output);
+            } else {
+                await this.splitAsOverrides(context, editor, entry, originalContent, currentContent, args.output);
+            }
+
+            // Restore spec to git HEAD after overlay/overrides are safely written
+            await writeFile(entry.specFilePath, originalRaw);
+            context.stderr.info(chalk.dim(`  Restored ${entry.specFilePath} to git HEAD`));
```
amckinney: Let's update these `stderr` logs to debug so users won't see anything by default. In general, we should make sure that the command output is as clean as possible and doesn't emit any unnecessary noise.

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    private async getFileFromGitHead(absolutePath: AbsoluteFilePath): Promise<string> {
+        try {
+            const repoRoot = await this.getRepoRoot(absolutePath);
+            const relativePath = path.relative(repoRoot, absolutePath);
+
+            const { stdout } = await execFileAsync("git", ["show", `HEAD:${relativePath}`], {
+                cwd: repoRoot,
+                encoding: "utf8",
+                maxBuffer: 50 * 1024 * 1024
```
amckinney: Let's define this constant somewhere.

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,282 @@
...
+            }
+            throw error;
+        }
+        const existing = parseSpec(raw, outputPath);
+        const merged = mergeWithOverrides({ data: existing, overrides });
+        await writeFile(outputPath, serializeSpec(merged, outputPath));
+        return true;
+    }
+
+    private async getFileFromGitHead(absolutePath: AbsoluteFilePath): Promise<string> {
```
amckinney: Why do we need to resolve it with respect to the `git` root? Can't we just follow the relative path based on the current working directory configured on the `Context`? We should make sure this works for any file on the user's machine, regardless of the `git` repository in which they're operating within.

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,282 @@
...
+    private async tryMergeIntoExistingFile(
+        outputPath: AbsoluteFilePath,
+        // biome-ignore lint/suspicious/noExplicitAny: OpenAPI specs can have any shape
+        overrides: Record<string, any>
+    ): Promise<boolean> {
+        let raw: string;
+        try {
+            raw = await readFile(outputPath, "utf8");
+        } catch (error: unknown) {
+            if (error instanceof Error && "code" in error && (error as NodeJS.ErrnoException).code === "ENOENT") {
```
amckinney: Is there a cleaner way to recognize these sort of exceptions? We shouldn't need to type `error: unknown` above (or at least this is not very common / idiomatic).

packages/cli/cli-v2/src/commands/api/utils/fernYmlApiEditor.ts
```diff
@@ -0,0 +1,269 @@
...
+}
+
+/**
+ * Stateful editor for mutating API spec override references in fern.yml.
+ *
+ * Follows the same pattern as FernYmlEditor: loads fern.yml, resolves $ref
+ * if the `api` section points to another file, and writes changes back to
+ * the correct file with formatting preserved.
+ */
+export class FernYmlApiEditor {
```
amckinney: Can we not just use the [FernYmlEditor](https://github.com/fern-api/fern/blob/main/packages/cli/cli-v2/src/config/fern-yml/FernYmlEditor.ts) for this? I think it'd be better than having an entirely separate abstraction just for a subset of the spec.

amckinney (commented):
The stderr indentation looks a little off when playing with this myself:

```sh
$ fern api merge
  /Users/alex/code/openapi/openapi.yml: no overrides or overlays configured, skipping.
No specs had overrides or overlays to merge.
```

The successful output looks great though 👏 :
```sh
$ fern api merge
✓ Merged 1 override(s) → /Users/alex/code/demo2/openapi/openapi.yml
```

We should also make sure all filepaths reported are _relative_ paths -- not absolute. I actually need to change this in some of the `fern sdk generate` output, too so we can handle that in a follow-up if not here.

The automatic removal from `fern.yml` is also slick 🎉 :
```sh
$ fernie api merge --remove
✓ Merged 1 override(s) → /Users/alex/code/demo2/openapi/openapi.yml
  Deleted /Users/alex/code/demo2/openapi/overrides.yml
  fern.yml:6: removed reference to overrides.yml
```

I'd consider unindenting the dim output on this one (even though we use indentation elsewhere like the `--strict` flag suggestion).

```sh
$ fernie api merge --remove
✓ Merged 1 override(s) → /Users/alex/code/demo2/openapi/openapi.yml
Deleted /Users/alex/code/demo2/openapi/overrides.yml
fern.yml:6: removed reference to overrides.yml
```

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,282 @@
...
+import { Icons } from "../../../ui/format.js";
+import { command } from "../../_internal/command.js";
+import { FernYmlApiEditor } from "../utils/fernYmlApiEditor.js";
+import type { SpecEntry } from "../utils/filterSpecs.js";
+import { filterSpecs } from "../utils/filterSpecs.js";
+import { loadSpec, parseSpec, serializeSpec } from "../utils/loadSpec.js";
+import { deriveOutputPath, type SplitFormat } from "./deriveOutputPath.js";
+import { generateOverlay, generateOverrides, hasChanges, type OverlayOutputAction } from "./diffSpecs.js";
+
+const execFileAsync = promisify(execFile);
```
amckinney, line 31: Can we just use `exec` here?

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,282 @@
...
+        output?: string;
+        format?: SplitFormat;
+    }
+}
+
+export class SplitCommand {
+    private cachedRepoRoot: string | undefined;
+
+    public async handle(context: Context, args: SplitCommand.Args): Promise<void> {
+        this.cachedRepoRoot = undefined;
```
amckinney: Discussed offline -- remove this.

packages/cli/cli-v2/src/commands/api/split/command.ts
patrickthornton: discussed offline; git's the ground truth for `split`

packages/cli/cli-v2/src/commands/api/split/command.ts
patrickthornton: i'm seeing that we're on `strict`, so `useUnknownInCatchVariables` is on, which forces some kind of `error: unknown` somewhere (at least to my understanding). i've extracted this to a helper cause it shows up in a few places but i think the dirtiness is somewhat inevitable:
```ts
/** Narrows an unknown caught value to a Node.js system error with an error code. */
function isNodeError(error: unknown): error is NodeJS.ErrnoException {
    return error instanceof Error && "code" in error;
}

/**
 * Returns true if the given error is a Node.js filesystem error with code ENOENT (file not found).
 */
export function isEnoentError(error: unknown): boolean {
    return isNodeError(error) && error.code === "ENOENT";
}
```

packages/cli/cli-v2/src/commands/api/utils/fernYmlApiEditor.ts
patrickthornton: agreed!

packages/cli/cli-v2/src/commands/api/split/command.ts
patrickthornton: i think this is better for security; it skips the shell, which (1) we don't need and (2) would require me to do lots of escaping and shell-injection-prevention stuff

packages/cli/cli-v2/src/commands/api/split/command.ts
patrickthornton: moved to just passing this as a parameter to the getFileFromGitHead function; it is worth precomputing so we don't call `git rev-parse` more often than once

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,282 @@
...
+                outputArg != null
+                    ? resolvePathOrThrow(context, outputArg)
+                    : deriveOutputPath(entry.specFilePath, "overlays");
+        }
+
+        if (await this.tryMergeOverlayIntoExistingFile(outputPath, overlay.actions)) {
+            context.stderr.info(
+                `${Icons.success} Merged ${chalk.bold(String(overlay.actions.length))} action(s) into existing ${chalk.cyan(outputPath)}`
+            );
+        } else {
```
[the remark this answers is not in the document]
patrickthornton: not sure if we care about this

packages/cli/cli-v2/src/commands/api/split/command.ts
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        const entries = filterSpecs(workspace, { api: args.api });
+
+        if (entries.length === 0) {
+            context.stderr.info(chalk.dim("No matching OpenAPI/AsyncAPI specs found."));
+            return;
+        }
+
+        const format: SplitFormat = args.format ?? "overlays";
```
amckinney: nit: `overlay`

amckinney (approved):

PR #13172, opened 2026-03-05
feat(cli): pass previous_version and language to AI prompt for AUTO versioning

Swimburger:
## Description

Refs Task 8 — Previous Version in AI Prompt

The `handleAutoVersioning()` flow in `LocalTaskHandler.ts` already extracts the previous version from the git diff, but this value was never forwarded to the AI prompt. This PR passes both `previous_version` and `language` to the `AnalyzeSdkDiff` BAML function so the AI has calibration context when generating commit messages and version bump decisions (e.g. "bumping from 0.1.2" reads differently than "bumping from 12.3.4").

Link to Devin session: https://app.devin.ai/sessions/798d52e1677248a183c57f0a841f4576
Requested by: @Swimburger

## Changes Made

- **BAML function signature** (`diff_analyzer.baml`): `AnalyzeSdkDiff` now accepts `diff`, `language`, and `previous_version` (previously only `diff`)
- **Prompt updates**: Added `Previous version: {{previous_version}}` and `SDK language: {{language}}` template variables, plus a guideline about not including the previous version literally in the commit summary line
- **`LocalTaskHandler.ts`**: Added `generatorLanguage` to the `Init` interface/class; updated the `AnalyzeSdkDiff` call to pass `this.generatorLanguage ?? "unknown"` and `previousVersion ?? "0.0.0"`
- **`runGenerator.ts`**: Passes `generatorInvocation.language` through to `LocalTaskHandler`
- **`sdkDiffCommand.ts`**: Updated second call site to pass all three params (defaults to `"unknown"` / `"0.0.0"` since this command lacks language/version context)
- **BAML client regenerated** via `pnpm run generate`
- **`versions.yml`**: Added v4.7.0 changelog entry

- Removed 0.x pre-stable API guideline from the prompt per review feedback (SDKs start at 0 by default at Fern, so the distinction was incidental)
- Removed corresponding test assertion for the 0.x guideline

## Important review notes

1. **BAML client `.js` extension removal**: Running `baml-cli generate` changed all internal imports from `./types.js` to `./types` across the regenerated `baml_client/` files. This is a side effect of the current BAML generator version, not an intentional change. These files are `@ts-nocheck` so they won't cause compile errors, but **verify this doesn't affect runtime module resolution**.

2. **`previousVersion ?? "0.0.0"` fallback**: In `handleAutoVersioning()`, there's an early return when `previousVersion == null` (line ~144). The `?? "0.0.0"` at the call site (line ~180) is a defensive fallback that should be unreachable in normal flow. Worth confirming this is the desired behavior vs. passing the already-validated `previousVersion` without a fallback.

3. **`sdkDiffCommand.ts` hardcoded defaults**: The `sdk-diff` CLI command doesn't have language or version context, so it passes `"unknown"` and `"0.0.0"`. Verify this is acceptable or whether the command should accept these as CLI flags.

## Human review checklist

- [ ] Confirm `.js` → extensionless import change in `baml_client/` files doesn't break runtime resolution
- [ ] Confirm `"unknown"` / `"0.0.0"` defaults in `sdkDiffCommand.ts` are acceptable for that command's use case

## Testing

- [x] Unit tests added (`LocalTaskHandler.prompt.test.ts` — 7 tests, all passing)
- [x] Tests verify BAML function signature, prompt template variables, and call-site parameter passing via source code inspection
- [ ] Manual testing completed (AUTO versioning requires AI service config + real diffs — not feasible in unit tests)

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13172" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/cli/ai/baml_src/diff_analyzer.baml
```diff
@@ -59,6 +63,14 @@ function AnalyzeSdkDiff(diff: string @description("The git diff to analyze for g
     - Do not use "Fern regeneration" in commit messages - use "SDK regeneration" instead
     - NEVER include the literal version "505.503.4455" in the commit message - if you see this placeholder
     in the diff, describe changes generically (e.g., "added X-Fern-SDK-Version header")
+    - The previous version is provided for context only. Do not include it
+    literally in the commit message summary line.
+    - For 0.x versions: the API is considered pre-stable. MAJOR changes should
+    still bump MAJOR (not MINOR), but note in the message that the API may
+    change further.
```
Swimburger: We don't want to make this distinction. SDKs start at 0 by default at Fern, so this could just be incidental.

fern-support (approved):

PR #13176, opened 2026-03-06
feat(cli): add language context to BAML prompt for SDK diff analysis

Swimburger:
## Description

Refs Task 12

Adds language context to the `AnalyzeSdkDiff` BAML prompt so the AI can apply language-specific breaking change rules when analyzing SDK diffs. The prompt now uses Jinja conditionals to include only the rules for the language being analyzed, keeping the prompt focused and improving version bump accuracy.

**Link to Devin Session:** https://app.devin.ai/sessions/4e805cd69cf54aaaa282754418bd916a
**Requested by:** @Swimburger

## Changes Made

- Added `extractLanguageFromGeneratorName()` utility in `VersionUtils.ts` that maps generator names (e.g. `fernapi/fern-typescript-node-sdk`) to normalized language identifiers (`typescript`, `python`, `java`, `go`, `ruby`, `csharp`, `php`, `swift`, `rust`, `kotlin`, or `unknown`)
  - Uses word-boundary regex (`\bgo\b`) for "go" to avoid false positives on "django", "mongo", "cargo"
  - Guards `java` with `!name.includes("javascript")` to avoid misclassifying JavaScript generators
- Updated `runGenerator.ts` to use `extractLanguageFromGeneratorName` as a fallback when `generatorInvocation.language` is undefined: `generatorLanguage: generatorInvocation.language ?? extractLanguageFromGeneratorName(generatorInvocation.name)`
- Added comprehensive, per-language breaking change rules to the BAML prompt using Jinja conditionals (`{% if language == "typescript" %}` ... `{% elif language == "python" %}` ... etc.) — only the rules for the current language are rendered into the prompt
  - Each of the 10 supported languages has **15–20+ tailored rules** organized into MAJOR / MINOR / PATCH sections
  - Rules cover: optional field handling, enum exhaustiveness, method signatures, visibility scopes, generics, discriminated unions, type narrowing/widening, builders, checked exceptions, pointer types, `#[non_exhaustive]`, sealed classes, functional options patterns, Pydantic models, keyword arguments, and more
  - SDK-specific patterns included: `RequestOptions` fields, SDK version headers, retry/timeout internals, serialization/deserialization changes, dependency version updates
  - `{% else %}` fallback block handles `"unknown"` language with generic static/dynamic typing heuristics
- Regenerated BAML client (`inlinedbaml.ts`)
- Added `versions.yml` entry for CLI v4.11.0

## Testing

- [x] `VersionUtils.test.ts`: 20 test cases covering all supported languages, case insensitivity, unknown generators, "go" false-positive prevention, "java"/"javascript" false-positive prevention, and integration verification of the wiring in `runGenerator.ts`
- [x] `LocalTaskHandler.prompt.test.ts`: 16 tests verifying conditional blocks for each language, MAJOR/MINOR/PATCH section structure, SDK-specific rule patterns, and language-specific content for TypeScript, Python, Java, Go, Rust, and unknown fallback
- [x] All 36 tests pass
- [x] `pnpm run check` (biome lint) passes cleanly
- [x] **E2E testing completed** with Anthropic Claude API — same diff (making a response field optional) classified differently per language:
  - TypeScript → **MAJOR** (correct: `T | undefined` breaks existing property access)
  - Python → **MINOR** (correct: None propagation, callers rarely type-check strictly)
  - Unknown → **MAJOR** (correct fallback for TS-style diff)

## Human Review Checklist

- [x] **Confirm the per-language breaking change rules are accurate** for how Fern actually generates each SDK. These rules were authored by AI — e.g., does the Java SDK generate exhaustive switch statements? Does the Go SDK use pointer fields for optional responses? Does the Swift SDK require exhaustive switches? Inaccurate rules could cause wrong version bumps.
- [x] **Token efficiency**: each language block is now ~40 lines with MAJOR/MINOR/PATCH sections. Only one block renders per call (via Jinja conditionals), but confirm this level of detail is worth the token cost vs. a shorter set of rules.
- [x] The `extractLanguageFromGeneratorName` function is only used as a fallback when `generatorInvocation.language` is `undefined` — verify this is the desired behavior vs. always extracting from name
- [x] JavaScript generators (e.g., `fernapi/fern-javascript-sdk`) will map to "unknown" due to the `!name.includes("javascript")` guard on Java matching. Is this intentional, or should there be a separate JavaScript handler?

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13176" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/cli/ai/baml_src/diff_analyzer.baml
```diff
@@ -93,6 +93,16 @@ function AnalyzeSdkDiff(
 
     --- End Examples ---
 
+    Language-specific breaking change rules:
```
Swimburger: That's all? There should be a lot more rules. ALso, we should only include the rules for the language currently analyzed to shorten the prompt

fern-support (approved):

PR #13190, opened 2026-03-06
feat(typescript): optimize TypeScript SDK generation performance

lifanzou:
## Summary

Performance optimizations for the TypeScript SDK generator targeting the two main bottlenecks: redundant context construction and O(n²) export file generation.

### Cache shared context parameters and consolidate iterations

1. **Cache shared context parameters**: `SdkContextImpl` takes ~70 constructor parameters, most of which are identical across all ~150-200+ `withSourceFile()` calls during generation. These shared params are now computed once and cached, so each context creation only needs to spread the cache with the per-file `sourceFile` and `importsManager`.

2. **Consolidate per-endpoint service iterations**: Endpoint type schemas and inlined request body schemas were generated in separate `forEachService` loops. They are now consolidated into a single `generatePerEndpointSerdeFiles()` pass.

### Optimize ExportsManager with bulk text-based export generation

Profiling against the Stripe OpenAPI spec (587 endpoints, 9,814 types, serde enabled) revealed that `ExportsManager.writeExportsToProject()` was the dominant bottleneck — **349 seconds, 92.2% of total generation time**. The root cause: each of ~40,000 export entries called `addExportDeclaration()` individually, which is O(n) per call (ts-morph AST insertion), making the total O(n²).

**Fix**: Replace individual `addExportDeclaration()` calls with bulk text-based generation. All export lines are built as plain strings and written via a single `replaceWithText()` call per file, reducing the complexity from O(n²) to O(n).

Additionally, exports are now grouped by resolved source file (using a `Map<SourceFile, string[]>`) to correctly handle path variants (e.g., `"src"` vs `"/src"`) that map to the same `index.ts` file.

### Benchmark: writeExportsToProject (isolated ts-morph)

| Total Exports | Old (addExportDeclaration) | New (replaceWithText) | Speedup |
|---|---|---|---|
| 2,000 (10 files × 200) | 751ms | 9ms | **83x** |
| 10,000 (50 files × 200) | 3,480ms | 37ms | **94x** |
| 20,000 (100 files × 200) | 6,848ms | 67ms | **102x** |

### Benchmark: End-to-end generation — Square API (4,467 generated files)

Using Square's real OpenAPI spec (4.7MB, ~1,500 schemas, ~250 endpoints) via `fern-api/fern-testing-square`:

| | Baseline (main) | Optimized | Speedup |
|---|---|---|---|
| Generator CLI only | **46.0s** | **38.8s** | **1.19x** |
| Full seed pipeline | **78.7s** | **64.9s** | **1.21x** |

### Benchmark: End-to-end generation — Synthetic XL API (6,455 generated files)

Synthetic spec with 100 services, 2,100 schemas, 1,000 endpoints, serde + neverThrowErrors:

| | Baseline (main) | Optimized | Speedup |
|---|---|---|---|
| Generator CLI only | **1,067s (17.8 min)** | **57.4s** | **18.6x** |

The O(n²) → O(n) export optimization produces increasingly dramatic gains as API size grows. At Square scale (moderate), we see ~20% improvement. At Stripe/XL scale (large), the quadratic bottleneck dominates and we see 18x+ improvement.

All benchmarks produce **100% identical output** with zero regressions.

## Test plan
- [x] `pnpm compile` — TypeScript compilation passes
- [x] `pnpm seed test --generator ts-sdk --fixture exhaustive --local --skip-scripts` — all 22 configs pass, zero output changes
- [x] Verified no diff in `seed/` directory after generation
- [x] Isolated ts-morph benchmark confirms 83-102x speedup for writeExportsToProject
- [x] Square API (fern-testing-square): 4,467 files generated, identical output, ~20% faster
- [x] Synthetic XL API: 6,455 files generated, identical output, 18.6x faster

### DIFF
--- a/generators/typescript/sdk/generator/src/SdkGenerator.ts
+++ b/generators/typescript/sdk/generator/src/SdkGenerator.ts
@@ -1935,6 +1935,7 @@ export class SdkGenerator {
                 coreUtilitiesManager: this.coreUtilitiesManager,
                 dependencyManager: this.dependencyManager,
                 fernConstants: this.intermediateRepresentation.constants,
+                // Shared by reference intentionally — all contexts accumulate exports into the same manager
                 exportsManager: this.exportsManager,
                 versionGenerator: this.versionGenerator,
                 versionDeclarationReferencer: this.versionDeclarationReferencer,
--- a/generators/typescript/sdk/generator/src/SdkGenerator.ts
+++ b/generators/typescript/sdk/generator/src/SdkGenerator.ts
@@ -240,6 +240,11 @@ export class SdkGenerator {
     private testDirectory: Directory;
     private packagePathDirectory: Directory;
 
+    // Cached shared params for SdkContextImpl creation — everything except sourceFile/importsManager
+    private cachedSharedContextParams:
+        | Omit<SdkContextImpl.Init, "sourceFile" | "importsManager" | "isForSnippet">
+        | undefined;
+
     constructor({
         namespaceExport,
         intermediateRepresentation,
@@ -613,8 +618,7 @@ export class SdkGenerator {
         let exportSerde: boolean = false;
         if (this.config.includeSerdeLayer) {
             this.generateTypeSchemas();
-            this.generateEndpointTypeSchemas();
-            this.generateInlinedRequestBodySchemas();
+            this.generatePerEndpointSerdeFiles();
             const serializationDirectory = this.rootDirectory.getDirectory("src/serialization");
             if (serializationDirectory != null && serializationDirectory?.getDescendantSourceFiles().length > 0) {
                 exportSerde = true;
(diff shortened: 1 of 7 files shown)


Swimburger:
Code looks correct — the O(n²) → O(n) fix in `ExportsManager` is sound, output is identical, and the consolidation of the two `forEachService` loops is fine (the dropped `{ generated: boolean }` return values were unused at the call site).

One concern: the benchmarks may overstate real-world impact. Testing against the GitHub and Stripe specs (two of the largest real customer specs) shows only ~1s improvement end-to-end. The dramatic speedups rely on either a synthetic spec engineered to hit the O(n²) knee.

The isolated `writeExportsToProject` benchmark (83–102x) is the honest story — the fix is real and correct. But customers are unlikely to see the 18x or even 1.2x improvements suggested by the end-to-end numbers. Worth toning down those claims, or noting that GitHub/Stripe show ~1s improvement at current scale.

Swimburger (approved):

PR #13193, opened 2026-03-06
feat(rust): add WebSocket connector pattern and README snippets

iamnamananand996:
## Description

Refs: Requested by @iamnamananand996
Link to Devin Session: https://app.devin.ai/sessions/5d39b24dcd814ff2ba711bf26fe88851

When `enableWebsockets: true` is set in the Rust SDK generator config, the generated SDK now:
1. **Generates WebSocket connector structs** (e.g., `RealtimeConnector`) that wrap `base_url` and delegate to the underlying `XxxClient::connect()` methods
2. **Adds connector fields to the root client** so WebSocket channels are accessible via `client.realtime.connect(...)` — matching the Python and Java generators' access patterns
3. **Includes a Websockets section in README.md** with usage examples showing the root-client-based connection pattern

Previously, WebSocket clients were standalone (required hardcoded URLs), and README generation produced no WebSocket documentation.

**Refactored shared logic to eliminate duplication risks** (addresses former checklist items #1 and #4):

- Extracted `buildConnectParams(channel)` — single method that derives the connect parameter list (path params → headers → query params) from IR. Both `generateConnectMethod()` and `generateConnectorStruct()` now call this instead of duplicating the iteration logic.
- Extracted `getConnectorName(clientName)` — single source of truth for the `${clientName.replace(/Client$/, "")}Connector` naming pattern. Used by `getConnectorInfo()`, `generateWebSocketModFile()`, and `generateConnectorStruct()`.

## Changes Made

- **`features.yml`**: Added `WEBSOCKETS` feature definition for the README feature framework
- **`WebSocketChannelGenerator.ts`**:
  - Added `getConnectorInfo()` to expose connector metadata to `RootClientGenerator`
  - Added `generateConnectorStruct()` to generate `XxxConnector` wrappers with delegating `connect()` methods
  - Added `buildConnectParams()` — shared parameter derivation used by both `generateConnectMethod()` and `generateConnectorStruct()` to keep signatures in sync
  - Added `getConnectorName()` — single source of truth for connector naming used across all three call sites
  - Updated `generateWebSocketModFile()` to re-export connector types
  - Updated `generateChannelFile()` to include connector struct in each channel file
- **`RootClientGenerator.ts`**:
  - Added `wsConnectors` property populated from `WebSocketChannelGenerator.getConnectorInfo()`
  - Updated `generateFields()` to add connector fields alongside HTTP sub-clients
  - Updated `generateConstructor()` to initialize connectors with `config.base_url.clone()`
  - Updated `generateImports()` to import connector types from `crate::api::websocket`
  - Added `getUniqueWsConnectors()` for collision avoidance with HTTP sub-clients
- **`ReadmeSnippetBuilder.ts`**:
  - Made `defaultEndpointId` optional so the constructor no longer throws for WebSocket-only SDKs
  - Added `buildWebSocketSnippets()` using `ClientConfig` + root client pattern with AST expressions
  - Added `getExampleWebSocketChannel()` mirroring Python's `_get_example_websocket_channel()`
  - Added helper methods for channel name mapping and message method name derivation
  - Updated `buildUsageSnippets()` and `getEndpointsForFeature()` to handle undefined `defaultEndpointId`
- **`SdkGeneratorCli.ts`**: Modified `generateReadme()` to proceed with README generation when WebSocket channels exist, even if there are no HTTP endpoints
- **`seed.yml`**: Added `websocket-bearer-auth` and `websocket-inferred-auth` fixtures with `enableWebsockets: true`
- **Seed snapshots**: Updated for all 3 websocket fixtures (`websocket`, `websocket-bearer-auth`, `websocket-inferred-auth`)

## ⚠️ Human Review Checklist

1. ~~**Connector parameter forwarding**~~: ✅ **Addressed** — Both `generateConnectMethod()` and `generateConnectorStruct()` now call the shared `buildConnectParams(channel)` method, ensuring parameter lists stay in sync automatically.

2. **README snippet placeholder values**: The `ClientConfig` in WebSocket snippets uses `base_url: " ".to_string()` (a single space) and always includes `api_key: Some("your-api-key".to_string())` regardless of auth requirements. These are artifacts of reusing `getClientConfigStruct("error")`. The hardcoded `api_key` appears even in the basic `websocket` fixture which has no authentication. Consider making this auth-aware as a follow-up.

3. **Collision detection scope**: `getUniqueWsConnectors()` only checks field name equality with HTTP sub-clients. Module-level collisions (if a WS connector and HTTP sub-client have different field names but same module path) are not detected. Current IR structure makes this scenario unlikely, but worth noting.

4. ~~**Connector re-export duplication risk**~~: ✅ **Addressed** — All three call sites (`getConnectorInfo()`, `generateWebSocketModFile()`, `generateConnectorStruct()`) now use the shared `getConnectorName(clientName)` method as the single source of truth.

5. **`defaultEndpointId` optional change**: Verified safe — all usages handle `undefined` correctly (`buildUsageSnippets` and `getEndpointsForFeature` both explicitly check for null/undefined and return empty arrays). WebSocket-only SDKs confirmed working.

6. **Subpackage iteration order**: `getExampleWebSocketChannel()` iterates `Object.keys(this.context.ir.subpackages)` and picks the first match. If the IR serialization order changes, a different channel may be selected for the README snippet. This is consistent with how Python handles it but worth noting.

## Testing

- [x] Built Rust SDK generator locally (`pnpm turbo run dist:cli --filter @fern-api/rust-sdk`)
- [x] Ran local seed tests for `websocket`, `websocket-bearer-auth`, `websocket-inferred-auth` — all 3 pass
- [x] Ran `simple-api` seed test for regression check — passes
- [x] Verified refactor produces identical output (git diff shows only `.ts` source changed, no snapshot files)
- [x] Verified generated output:
  - WebSocket-only SDK (`websocket`): Root client has `empty_realtime: EmptyRealtimeConnector` and `realtime: RealtimeConnector` fields, README uses `client.empty_realtime.connect()`
  - Mixed HTTP + WebSocket SDK (`websocket-bearer-auth`): Root client has both HTTP sub-clients and connector fields, README shows WebSocket section with `client.realtime_no_auth.connect()`
  - Connector structs generated with proper delegation to underlying client `connect()` methods
  - Connectors re-exported from websocket `mod.rs` and imported into root client
- [ ] End-to-end test against real API definition (`immix-fern-config`) pending CI completion

### DIFF
--- a/generators/rust/sdk/src/readme/ReadmeSnippetBuilder.ts
+++ b/generators/rust/sdk/src/readme/ReadmeSnippetBuilder.ts
@@ -40,13 +40,12 @@ export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
 
         this.endpointsById = this.buildEndpointsById();
         this.prerenderedSnippetsByEndpointId = this.buildPrerenderedSnippetsByEndpointId(endpointSnippets);
-        if (this.context.ir.readmeConfig?.defaultEndpoint != null) {
-            this.defaultEndpointId = this.context.ir.readmeConfig.defaultEndpoint;
-        } else if (endpointSnippets.length > 0) {
-            this.defaultEndpointId = this.getDefaultEndpointId();
-        } else {
-            this.defaultEndpointId = undefined;
-        }
+        this.defaultEndpointId =
+            this.context.ir.readmeConfig?.defaultEndpoint != null
+                ? this.context.ir.readmeConfig.defaultEndpoint
+                : endpointSnippets.length > 0
+                  ? this.getDefaultEndpointId()
+                  : undefined;
         this.crateName = this.context.getCrateName();
     }
 
@@ -607,15 +606,14 @@ export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
             return [];
         }
 
-        const { clientName } = names;
         const clientMessages = channel.messages.filter((m) => m.origin === "client");
         const serverMessages = channel.messages.filter((m) => m.origin === "server");
 
         // Get the subpackage access path (e.g., "market_data" or "realtime")
         const subpackageName = subpackage.name.snakeCase.safeName;
 
-        // Build connect params from IR
-        const connectParams: string[] = ['"wss://api.example.com"'];
+        // Build connect params from IR (without the url, since connectors provide it from config)
+        const connectParams: string[] = [];
         for (const pathParam of channel.pathParameters) {
             connectParams.push(`"${pathParam.name.snakeCase.safeName}"`);
         }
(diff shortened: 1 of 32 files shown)


generators/rust/sdk/src/readme/ReadmeSnippetBuilder.ts
```diff
@@ -39,10 +40,13 @@ export class ReadmeSnippetBuilder extends AbstractReadmeSnippetBuilder {
 
         this.endpointsById = this.buildEndpointsById();
         this.prerenderedSnippetsByEndpointId = this.buildPrerenderedSnippetsByEndpointId(endpointSnippets);
-        this.defaultEndpointId =
-            this.context.ir.readmeConfig?.defaultEndpoint != null
-                ? this.context.ir.readmeConfig.defaultEndpoint
-                : this.getDefaultEndpointId();
```
amckinney, line 45: I actually prefer the original ternary rather than `else if` chains. You're cool though since this is a small detail. Although, shouldn't we be using `this.getDefaultEndpointId` as the default case rather than `undefined`?

amckinney (approved):

PR #13198, opened 2026-03-06
fix(openapi): improve array parameter validation in example endpoint factory

nicasmar:
## Summary
- Add validation to ensure array items are primitives for query, header, and global header parameters
- Update isExamplePrimitive function to handle path parameters (which cannot accept arrays)
- Improve error messages to distinguish between primitive and array validation issues


aditya-arolkar-swe:
Scoped the issue to not generating examples for query parameters or headers when they were arrays of primitives. Added a function to check this and only skip example generation when it's NOT either a primitive NOR an array of primitives. Examples now generate properly from the fixtures.

aditya-arolkar-swe (approved):

PR #13245, opened 2026-03-09
feat(cli): add `fern check` rule to detect colliding component schemas

iamnamananand996:
## Description

Closes https://github.com/fern-api/fern/issues/11686

When multiple OpenAPI specs are configured in `generators.yml`, two specs can define `components/schemas` with the same name (e.g., both define `User`). During the IR merge in `mergeSchemaMaps`, the second schema silently overwrites the first. This PR adds a new `fern check` validation rule to detect and report these collisions.

Link to Devin Session: https://app.devin.ai/sessions/66062ea64ac74d1980bb9ea74ff41e50

## Changes Made

- Added new `no-component-schema-collisions` OSS validation rule in `packages/cli/workspace/oss-validator/src/rules/no-component-schema-collisions/`
- Registered the rule in `getAllRules.ts`
- Added CLI versions.yml entry (4.17.0, feat)
- Added 4 test fixtures covering: collision detected, no collision (different names), collision with `resolve-schema-collisions: true` (skipped), single spec (skipped)

- **Severity changed from `error` to `warning`** per reviewer feedback — avoids breaking existing users who may have colliding schemas in their specs today
- Version bumped to 4.17.0 (from 4.16.0) to resolve merge conflict with `.fernignore` autodiscovery feature on main
- E2E verified locally: built the CLI and confirmed the rule reports collisions correctly via `fern check --json`

## Testing

- [x] Unit tests added (4 test cases, all passing)
- [x] Biome lint/format passing
- [x] E2E verified with local CLI build against multi-spec fixture with colliding `User` schema

## Items for Reviewer Attention

1. **Severity is `warning`, not `error`**: Collisions won't block CI. If we want to promote this to `error` in a future release, it can be changed in one line.
2. **`resolveSchemaCollisions` skip logic**: The rule skips the *entire* check if *any* spec has `resolve-schema-collisions: true`. If only one of N specs enables it, collisions between the remaining specs won't be caught. Confirm this is the desired behavior vs. per-spec opt-out.
3. **AsyncAPI / namespaced schemas not covered**: The rule only checks OpenAPI spec component schemas at the root level. `mergeSchemaMaps` also merges namespaced and AsyncAPI schemas — collisions there are not detected by this rule.
4. **OSSWorkspace only**: The rule runs via the OSS validator, which requires the workspace to be an `OSSWorkspace` (OpenAPI-only projects). Projects that also have a `definition/` directory load as `LazyFernWorkspace` and won't trigger this rule.
5. **Redundant content-based file type check**: The specs are already filtered to `type === "openapi"`, but the rule also reads file contents to check for `"openapi"` / `"swagger"` strings. This is inherited from the sibling `no-schema-title-collisions` rule but could be simplified.

### DIFF
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -1,4 +1,16 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 4.16.0
+  changelogEntry:
+    - summary: |
+        Add `fern check` validation rule to detect component schema collisions
+        across multiple OpenAPI specs. When two specs define
+        `components/schemas` with the same name, one silently overwrites the
+        other during merge. The new `no-component-schema-collisions` rule now
+        reports this as an error, unless `resolve-schema-collisions` is enabled.
+      type: feat
+  createdAt: "2026-03-09"
+  irVersion: 65
+
 - version: 4.15.5
   changelogEntry:
     - summary: |
--- a/packages/cli/cli/versions.yml
+++ b/packages/cli/cli/versions.yml
@@ -6,7 +6,7 @@
         across multiple OpenAPI specs. When two specs define
         `components/schemas` with the same name, one silently overwrites the
         other during merge. The new `no-component-schema-collisions` rule now
-        reports this as an error, unless `resolve-schema-collisions` is enabled.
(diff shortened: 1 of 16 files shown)


amckinney (approved):
lgtm - let's update to a warning so we don't accidentally break anyone's API

PR #13279, opened 2026-03-09
feat(cli): remove FERN_SELF_HOSTED, rename OVERRIDE_FDR_ORIGIN to FERN_FDR_ORIGIN

thesandlord:
## Description

Refs discussion between @thesandlord and team about consolidating self-hosted env vars.

Removes the `FERN_SELF_HOSTED` environment variable from the CLI and replaces its three behaviors with more granular, independently useful flags.

Link to Devin Session: https://app.devin.ai/sessions/30e36ade33a6482ab5033fac90d53b6e
Requested by: @thesandlord

## Changes Made

- **Auth bypass** now triggers when `FERN_FDR_ORIGIN` (or legacy `OVERRIDE_FDR_ORIGIN`) is set — if you're pointing FDR at a custom origin, interactive login is skipped and `FERN_TOKEN` is read from env directly.
- **PostHog telemetry** — the conditional skip in `generateDocsWorkspace.ts` is removed. Telemetry opt-out is handled globally by the existing `FERN_DISABLE_TELEMETRY=true` env var in `getPosthogManager.ts` (no new env var needed).
- **`ValidFileTypes` validation skip** removed entirely (was only used in self-hosted, deemed unnecessary).
- **`OVERRIDE_FDR_ORIGIN` renamed to `FERN_FDR_ORIGIN`** across `fdr.ts`, `generateLibraryDocs.ts`, and `build.local.mjs`. The old name is still accepted as a fallback for backward compatibility.
- E2E tests updated to use `FERN_FDR_ORIGIN` instead of `FERN_SELF_HOSTED`.
- `versions.yml` entry added (4.21.0).

- Dropped `FERN_POSTHOG=false` env var per review feedback (consolidated on existing `FERN_DISABLE_TELEMETRY=true`).
- Reverted `getPosthogManager.ts` to its original state — no changes needed there.
- Updated `versions.yml` changelog to reference `FERN_DISABLE_TELEMETRY` instead of `FERN_POSTHOG`.

## ⚠️ Review Checklist

- [ ] **Companion fern-platform PR**: [fern-platform#8261](https://github.com/fern-api/fern-platform/pull/8261) updates self-hosted scripts (`run.sh`, `generate.sh`) to use `FERN_FDR_ORIGIN`, `FERN_DISABLE_TELEMETRY=true`, and drops `FERN_SELF_HOSTED`. Must ship after this CLI version is released.
- [ ] Verify backward compat: users setting `OVERRIDE_FDR_ORIGIN` (without `FERN_FDR_ORIGIN`) still get the auth bypass and FDR override.
- [ ] Confirm removing `ValidFileTypes` skip won't break self-hosted builds (were there file types that only failed in self-hosted?).
- [ ] `FERN_TOKEN` empty-string handling: `if (!fernToken)` treats `""` as falsy — matches previous behavior where `FERN_SELF_HOSTED=true` with `FERN_TOKEN=""` also failed.
- [ ] PostHog change: self-hosted builds will now send PostHog events unless `FERN_DISABLE_TELEMETRY=true` is set (previously always skipped by `FERN_SELF_HOSTED`). The companion PR sets this flag.

## Testing

- [x] E2E tests updated and passing (`pnpm run check`)
- [ ] Manual testing of self-hosted flow with `FERN_FDR_ORIGIN=http://localhost:8080 FERN_TOKEN=dummy FERN_DISABLE_TELEMETRY=true fern generate --docs`


packages/cli/cli/versions.yml
```diff
@@ -1,4 +1,22 @@
...
+- version: 4.21.0
+  changelogEntry:
+    - summary: |
+        Remove the `FERN_SELF_HOSTED` environment variable from the CLI. The
+        three behaviors it controlled are now handled independently:
+        - **Auth bypass**: when `FERN_FDR_ORIGIN` (or legacy `OVERRIDE_FDR_ORIGIN`)
+          is set, the CLI reads `FERN_TOKEN` directly instead of triggering
+          interactive login.
+        - **PostHog opt-out**: set `FERN_POSTHOG=false` to disable telemetry
+          (in addition to the existing `FERN_DISABLE_TELEMETRY=true`).
```
amckinney: Let's consolidate on `FERN_DISABLE_TELEMETRY` -- we don't need a PostHog-specific env var. Thoughts?

packages/cli/posthog-manager/src/getPosthogManager.ts
```diff
@@ -18,7 +18,8 @@ async function createPosthogManager(): Promise<PosthogManager> {
     try {
         const posthogApiKey = process.env.POSTHOG_API_KEY;
         const disableTelemetry = process.env.FERN_DISABLE_TELEMETRY === "true";
-        if (posthogApiKey == null || disableTelemetry) {
+        const posthogDisabled = process.env.FERN_POSTHOG === "false";
```
amckinney: Re: `versions.yml` comment -- I'd just use `FERN_DISABLE_TELEMETRY` above rather than rolling out a separate one.

packages/cli/cli/src/commands/docs-md-generate/generateLibraryDocs.ts
```diff
@@ -58,12 +58,12 @@ interface LibraryDocsClient {
...
  *
  * The base URL and auth mirror what `createFdrService` uses so that
- * env-var overrides (`DEFAULT_FDR_ORIGIN`, `OVERRIDE_FDR_ORIGIN`) keep
+ * env-var overrides (`DEFAULT_FDR_ORIGIN`, `FERN_FDR_ORIGIN`) keep
  * working.
  */
 function createLibraryDocsClient({ token }: { token: string }): LibraryDocsClient {
     const defaultOrigin = process.env.DEFAULT_FDR_ORIGIN ?? "https://registry.buildwithfern.com";
-    const baseUrl = process.env.OVERRIDE_FDR_ORIGIN ?? defaultOrigin;
+    const baseUrl = process.env.FERN_FDR_ORIGIN ?? process.env.OVERRIDE_FDR_ORIGIN ?? defaultOrigin;
```
amckinney, line 66: Nice - in CLI v2, I'll drop `OVERRIDE_FDR_ORIGIN` entirely.

amckinney (approved):

PR #13284, opened 2026-03-10
feat(go): bump default streaming buffer size to 1MB and add WithMaxStreamBufSize option

jsklan:
## Description

Refs #18140

Anduril customers are hitting buffer overflow errors on streaming/SSE endpoints where individual events exceed the 64KB default buffer limit. The error was confusingly reported as `io.EOF` instead of the actual `bufio.ErrTooLong`, making it hard to diagnose.

**Link to Devin Session:** https://app.devin.ai/sessions/90ede69b203e470aadbd0b9812f4a71f
**Requested by:** @jsklan

### 1. Bump default buffer size from 64KB → 1MB
- `generators/go-v2/base/src/asIs/core/stream.go_`: `defaultMaxBufSize = 1024 * 1024`
- This is the max per-event ceiling, not a pre-allocation — `bufio.Scanner` starts at `defaultInitBufSize` (4096 bytes) and grows as needed.

### 2. Add `WithMaxStreamBufSize` runtime option
- **Core layer** (`stream.go_`): New `WithMaxBufSize(size int) StreamOption` function with `size > 0` guard to prevent panics on invalid input
- **Internal layer** (`streamer.go_`): Added `MaxBufSize int` to `StreamParams`, threaded to `core.WithMaxBufSize`
- **Go v1 generator** (`sdk.go`): Added `MaxBufSize` to `RequestOptions` struct, generated `MaxBufSizeOption` struct, generated `WithMaxStreamBufSize` helper in `option` package, threaded into `StreamParams` in streaming endpoint code
- **Go v2 generator** (`Streamer.ts`): Added `MaxBufSize` field to `StreamParams` construction from `options.MaxBufSize`

SDK users can now do: `option.WithMaxStreamBufSize(10 * 1024 * 1024)` to increase beyond 1MB.

### 3. Apply configurable buffer to both stream reader types
- `ScannerStreamReader` (non-SSE streams) and `SseStreamReader` both now use `scanner.Buffer(defaultInitBufSize, maxBufSize)` so the configurable limit applies consistently across all streaming formats.

### 4. Fix misleading EOF error on buffer overflow
- `SseStreamReader.nextEvent()` was returning `io.EOF` when the scanner failed due to buffer overflow, swallowing the real `bufio.ErrTooLong` error. Added `scanner.Err()` check before falling back to `io.EOF`.

### 5. Seed fixture updates
- Updated `streaming`, `server-sent-events`, and `server-sent-event-examples` fixtures

- Extracted magic number `4096` into named constant `defaultInitBufSize` (per @patrickthornton review)
- Added `if size > 0` guard in `WithMaxBufSize` to silently ignore negative/zero values and prevent runtime panics

## Human Review Checklist
- [ ] **Naming consistency**: Core-level option is `WithMaxBufSize`, public API is `WithMaxStreamBufSize` — intentional for clarity at the SDK user level, but confirm this is preferred
- [ ] **`MaxBufSize` on all RequestOptions**: The field is added to `RequestOptions` globally (not conditionally for streaming endpoints only), matching the `MaxAttempts` pattern. Verify this is acceptable.
- [ ] **Idempotent classification**: `MaxBufSize` option is registered as `asIdempotentRequestOption` (same as `MaxAttempts`). Confirm this is correct for buffer sizing.
- [ ] **Default value semantics**: When `MaxBufSize` is 0 or negative, the `size > 0` guard in `WithMaxBufSize` silently ignores it and uses the 1MB default. Verify this is the intended behavior (vs. returning an error).

## Testing
- [x] Seed tests passed for `streaming`, `server-sent-events`, `server-sent-event-examples` fixtures
- [x] Generated code verified: `WithMaxStreamBufSize` helper generates correctly, `MaxBufSize` threads through `RequestOptions` → `StreamParams` → `core.WithMaxBufSize`
- [x] Default buffer change verified in generated `core/stream.go` files
- [x] Negative value guard tested (silently ignores invalid input)
- [x] Magic number extracted into named constant
- [ ] Manual end-to-end testing with large SSE events (blocked on environment access)

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13284" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


generators/go-v2/base/src/asIs/core/stream.go_
```diff
@@ -183,6 +194,7 @@ func newScannerStreamReader(
 	options *streamOptions,
 ) *ScannerStreamReader {
 	scanner := bufio.NewScanner(reader)
+	scanner.Buffer(make([]byte, slices.Min([]int{4096, options.maxBufSize})), options.maxBufSize)
```
patrickthornton: not a huge thing but the 4096 is a bit of a magic number

patrickthornton (approved):

PR #13287, opened 2026-03-10
fix(typescript, csharp, python, ruby, go, swift, rust, php): re-throw errors after logging for README/reference generation failures

iamnamananand996:
## Description

Refs [PR #13276](https://github.com/fern-api/fern/pull/13276)

Follow-up to #13276 (which added catch-and-log error handling). This PR makes README.md and reference.md generation failures **fatal** by replacing the catch-and-swallow pattern with `throw new Error(...)` using `extractErrorMessage` from `@fern-api/core-utils`. Previously, errors were caught and logged but swallowed, allowing generation to silently continue with missing documentation. Now failures break the generation flow so they surface as important issues to fix.

Additionally, this PR adds **graceful handling for APIs with no endpoints or missing endpoint examples**, so that generators skip documentation generation instead of throwing for legitimate no-endpoint cases (e.g. webhook-only or type-only APIs).

Link to Devin Session: https://app.devin.ai/sessions/415fec7ffce447359ec05a022645be95

## Changes Made

- Replaced catch-and-log blocks with `throw new Error(`Failed to generate X: ${extractErrorMessage(e)}`)` in every README.md and reference.md catch block across **all 9 SDK generators**:
  - **TypeScript** — README.md, reference.md, CONTRIBUTING.md
  - **C#** — README.md, reference.md
  - **Python-v2** — README.md, reference.md
  - **Ruby-v2** — README.md, reference.md
  - **Go-v2** — README.md, reference.md
  - **Swift** — README.md, reference.md
  - **Rust** — README.md, reference.md
  - **PHP** — README.md, reference.md
  - **Java-v2** — README.md, reference.md
- Uses `extractErrorMessage` from `@fern-api/core-utils` (already a dependency in all generators) instead of inline `e instanceof Error ? e.message : String(e)` pattern
- Removed all redundant `logger.warn` / `logger.debug` logging — the error context is embedded directly in the thrown `Error`
- Bumped `versions.yml` for all 8 generators that have version files (Java-v2 does not have one)
- **C# `buildReference`**: Changed return type to `ReferenceConfigBuilder | undefined`, returns `undefined` instead of throwing when no endpoints are found; caller (`generateReference`) now checks for null and skips with a debug log
- **Java-v2 `generateReadme`**: Added early return when `context.ir.dynamic?.endpoints` is empty, skipping README generation for APIs with no endpoints
- **Java-v2 `buildReference`**: Replaced `getExampleEndpointCallOrThrow(endpoint)` with inline example lookup that returns `undefined` for endpoints lacking examples, instead of throwing

- Switched from inline error message extraction to `extractErrorMessage` from `@fern-api/core-utils`, which handles additional edge cases (e.g., plain objects with a `.message` property).
- Catch blocks are now a single `throw` line — no intermediate `const errorMessage = ...` variable needed.
- Fixed seed test failures for C# and Java-v2 caused by APIs with no endpoints or missing endpoint examples:
  - C# `buildReference` now returns `undefined` instead of throwing for no-endpoint APIs
  - Java-v2 `generateReadme` skips generation when no dynamic endpoints exist
  - Java-v2 `buildReference` gracefully handles endpoints without examples (returns `undefined` per-endpoint instead of throwing)

## Important for Review

- **This is an intentional behavior change**: documentation generation failures that were previously non-fatal are now fatal. If the generator agent (external service for README/reference generation) is temporarily unavailable, SDK generation will now fail entirely instead of producing an SDK without docs.
- **Original error stack trace is lost**: since we throw a `new Error(...)` wrapping only the message string, the original error's stack trace is discarded. The caller will see a stack trace pointing to the catch block. Confirm this tradeoff is acceptable vs. re-throwing the original error.
- **Java-v2** has no `versions.yml` — confirm this is expected.
- **C# `buildReference` type signature change**: The return type changed from `ReferenceConfigBuilder` to `ReferenceConfigBuilder | undefined`. Verify no other callers of this function are affected.
- **Java-v2 `buildReference` duplicates example lookup logic**: Instead of calling `getExampleEndpointCallOrThrow`, the code now manually looks up examples inline. If `getExampleEndpointCallOrThrow` is updated in the future, this inline copy won't automatically inherit the changes. Consider refactoring to a shared `getExampleEndpointCallIfExists` method if this becomes a maintenance issue.
- **Java-v2 no-endpoints check uses `context.ir.dynamic?.endpoints`**: Verify this is the correct authoritative source for determining whether an API has endpoints. If an API has non-dynamic endpoints, would this incorrectly skip README generation?

## Testing

- [ ] Unit tests added/updated — _not applicable; pattern replacement with graceful edge-case handling_
- [x] Verified all 9 generators consistently use `extractErrorMessage` and throw `new Error(...)` on doc generation failure
- [x] Verified C# and Java-v2 gracefully handle no-endpoint and missing-example cases
- [x] Biome lint passes
- [x] CI passing for all generators (java-sdk, csharp-sdk, ts-sdk, python-sdk, go-sdk, swift-sdk, rust-sdk, php-sdk)

**Note**: Two required CI checks (fastapi, ruby-sdk-v2) failed due to infrastructure issues unrelated to these changes:
- `fastapi`: "Unexpected HTTP response: 403" (rate limiting)
- `ruby-sdk-v2`: "npm error code ECONNRESET" (network connectivity)


amckinney (approved):
Let's make sure that the user sees the error message -- if the error message itself is shown to the user then fine as-is, otherwise restore the log

PR #13292, opened 2026-03-10
fix(cli): expand allowed file types in docs validator

thesandlord:
## Description

Refs: https://buildwithfern.slack.com/archives/C08CH70C8LW/p1773160967763279

Self-hosted users uploading `.tar` files as docs assets (e.g., SDK bundles, proto archives) were getting blocked by the docs validator with:
```
[error] The file type of application/x-tar is not allowed: /fern/assets/sdks/lattice_sdk_go.tar
```

The docs validator uses an allowlist of MIME types. Several common file types were missing from this list.

## Changes Made

- Added the following MIME types to the `ALLOWED_FILE_TYPES` set in the docs validator:
  - **Archive files:** `application/x-tar`, `application/gzip`, `application/zip`, `application/x-bzip2`
  - **Image files:** `image/bmp`, `image/heif`
- Added unit tests for all archive types: `.tar`, `.zip`, `.tar.gz` (gzip), `.tar.bz2` (bzip2)
- Added test fixture files: `test.tar`, `test.tar.gz`, `test.tar.bz2`, `test.exe` (ELF binary for invalid type test)
- Updated the existing invalid file type test to use `test.exe` instead of `bird.zip` (since zip is now allowed)
- Bumped CLI `versions.yml` to `4.21.2` with changelog entry

## Testing

- [x] Unit tests added for all archive types (tar, zip, gzip, bzip2) and invalid type (ELF binary)
- [x] All 54 tests pass
- [x] Lint (`pnpm check`) passes

## Human Review Checklist

- [ ] Verify that `image/bmp` and `image/heif` are correctly recognized by the `file-type` library (no dedicated test fixtures for these)
- [ ] Confirm that allowing archive uploads does not conflict with the recent tar CVE fix (GHSA-qffp-2rhf-9h96) — that fix was about the `tar` npm package extraction vulnerability, not about upload validation

---

[Link to Devin session](https://app.devin.ai/sessions/1283eed1aae8427b843dd46036b84ca4)

Requested by: @thesandlord

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13292" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/cli/yaml/docs-validator/src/rules/valid-file-types/valid-file-types.ts
```diff
@@ -33,7 +33,15 @@ const ALLOWED_FILE_TYPES = new Set<MimeType>([
     "font/woff",
     "font/woff2",
     "font/otf",
-    "font/ttf"
+    "font/ttf",
+    // archive files
+    "application/x-tar",
+    "application/gzip",
+    "application/zip",
+    "application/x-bzip2",
```
amckinney, line 41: We should include `application/tar` and `application/tar+gzip`.

amckinney: Re: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types -- we should be good without these actually

amckinney (approved):

Issue #13329, opened 2026-03-11
Feature Request: enable logging for all language generators.


iamnamananand996:
### Which Fern component?

SDK Generator

### How important is this?

P2 - Medium (Would be helpful)

### What's the feature?

Reference: https://github.com/fern-api/fern/pull/12765

### Are you interested in contributing this feature?

No

PR #13342, opened 2026-03-11
feat(go): added configurable logging

HoaX7:
## Description
Refs: https://github.com/fern-api/fern/issues/13329

Adds configurable logging to the Go SDK generator, aligned with the existing Java ([#12765](https://github.com/fern-api/fern/pull/12765)) and Python ([#12219](https://github.com/fern-api/fern/pull/12219)) logging implementations. Silent by default — no log output unless explicitly configured.

**Link to Devin Session:** https://app.devin.ai/sessions/cc176160655749b48b6245048e394919
**Requested by:** @iamnamananand996

### Generator code (`generators/go/`)
- Added 5 new core logging files (embedded via `//go:embed`): `log_level.go`, `logger.go`, `console_logger.go`, `log_config.go`, `logging_http_client.go`
- `generator.go`: Added file generation calls for the 5 logging files
- `sdk.go` (WriteClient): Wraps `HTTPClient` with `LoggingHTTPClient` when logging is configured and not silent
- `sdk.go` (WriteRequestOptionsDefinition): Added `Logging *LogConfig` field to `RequestOptions` and `LoggingOption` struct
- `sdk.go` (WriteRequestOptions): Added `WithLogging()` helper in `option/` package

### Architecture (mirrors Java/Python)
- `ILogger` interface (custom backend) → `Logger` wrapper (level filtering + silent mode) → `LoggingHTTPClient` (wraps `HTTPClient`) → `ConsoleLogger` (default stderr backend)
- `LogConfigBuilder` pattern for configuration: `NewLogConfigBuilder().Level(...).Silent(false).Build()`
- Sensitive header redaction in request/response logs

### Seed fixtures (`seed/go-sdk/`)
- Updated all 122+ seed fixture directories with the new logging core files and updated `request_option.go` / `option/request_option.go`

### Bug fix
- Fixed import cycle bug: `writeRequestOptionStructs` was called with `useCore=true` inside `WriteRequestOptionsDefinition` (which generates `core/request_option.go`), causing the core package to reference `core.LogConfig` instead of `LogConfig` — a self-import cycle. Changed to `useCore=false`.

## Usage Example
```go
client := NewClient(
    option.WithLogging(core.NewLogConfigBuilder().
        Level(core.LogLevelDebug).
        Silent(false).
        Build()),
)
```

## Human Review Checklist
- [ ] Verify `useCore` parameter: both calls in `WriteRequestOptionsDefinition` now pass `false` — confirm `LoggingOption` in `core/request_option.go` correctly uses `*LogConfig` (not `*core.LogConfig`)
- [ ] Confirm seed fixtures match what the generator produces (CI regenerates seeds; committed fixtures were bulk-copied)
- [ ] Verify `LoggingHTTPClient` return type (`HTTPClient` interface) is compatible with `WriteClient` generated assignment
- [ ] Review sensitive headers list for completeness against Java/Python implementations
- [ ] Updated README.md generator (if applicable) — not updated in this PR

## Testing
- [x] Generator compiles (`go build ./...`)
- [ ] Seed tests pass in CI (import cycle fix pushed, awaiting CI)
- [ ] Manual testing


iamnamananand996 (approved):

generators/go/internal/generator/sdk/core/console_logger.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"sync"
+)
+
+// ConsoleLogger is the default ILogger implementation that writes to stderr.
+// It uses a simple format of "LEVEL - message".
+type ConsoleLogger struct {
+	mu     sync.Mutex
+	output io.Writer
+}
```
amckinney, line 15: Why not use the standard library [slog](https://go.dev/blog/slog) instead?

generators/go/internal/generator/sdk/core/logger.go
```diff
@@ -0,0 +1,98 @@
+package core
+
+// ILogger is the interface for logging SDK operations.
+// Implement this interface to provide custom logging backend.
+type ILogger interface {
```
amckinney, line 5: This isn't right -- this interface is not amenable to the most popular Go logging libraries (e.g. `zerolog`, `zap`, `slog`). Plus, the `I` prefix for interfaces in Go in is not idiomatic. 

We need to build something that supports arbitrary args in a way that's compatible with these libraries. I imagine something along the lines of:
```go
// LogLevel represents the severity of a log message.
type LogLevel int

const (
    LogLevelDebug LogLevel = iota
    LogLevelInfo
    LogLevelWarn
    LogLevelError
)

// Logger is the interface that callers implement to provide
// custom logging to the SDK.
type Logger interface {
    Log(level LogLevel, msg string, args ...any)
}
```

With this, users could build simple adapters for each library:

```go
// slog adapter
type SlogLogger struct{ Logger *slog.Logger }

func (l *SlogLogger) Log(level LogLevel, msg string, args ...any) {
    switch level {
    case LogLevelDebug:
        l.Logger.Debug(msg, args...)
    case LogLevelInfo:
        l.Logger.Info(msg, args...)
    case LogLevelWarn:
        l.Logger.Warn(msg, args...)
    case LogLevelError:
        l.Logger.Error(msg, args...)
    }
}

// zap adapter
type ZapLogger struct{ Logger *zap.SugaredLogger }

func (l *ZapLogger) Log(level LogLevel, msg string, args ...any) {
    switch level {
    case LogLevelDebug:
        l.Logger.Debugw(msg, args...)
    case LogLevelInfo:
        l.Logger.Infow(msg, args...)
    case LogLevelWarn:
        l.Logger.Warnw(msg, args...)
    case LogLevelError:
        l.Logger.Errorw(msg, args...)
    }
}
```

We could even build a lightweight functional option like so:

```go
// LoggerFunc adapts a plain function into a Logger.
type LoggerFunc func(level LogLevel, msg string, args ...any)

func (f LoggerFunc) Log(level LogLevel, msg string, args ...any) { f(level, msg, args...) }
```

```go
client := sdk.NewClient(sdk.WithLogger(sdk.LoggerFunc(
    func(level sdk.LogLevel, msg string, args ...any) {
        log.Printf("[%s] %s %v", level, msg, args)
    },
)))
```

HoaX7: Hi, thanks for your comment. I was told it needed to be similar to java loggers hence, the approach. Will make the changes.

HoaX7: hey @amckinney I've just noticed that this pr is deprecated, here is the correct one https://github.com/fern-api/fern/pull/13470. im closing this pr.

PR #13362, opened 2026-03-11
fix(seed): add fetch-depth: 2 so HEAD^1 is resolvable in PR affected detection

amckinney:
## Description

Refs: https://buildwithfern.slack.com/archives/C05J55L8JES/p1773066119051929

Fixes seed tests being falsely triggered on PRs that don't touch any generator code (e.g. [#13343](https://github.com/fern-api/fern/pull/13343) which only changed `@fern-api/cli-v2` but triggered all seed tests).

[Link to Devin Session](https://app.devin.ai/sessions/5eff335bf84540028b9c69fae3c2f98e)
Requested by: @amckinney

## Root Cause

The affected detection uses `HEAD^1` (first parent of the merge commit) as the base ref for `git diff`. However, with the default `fetch-depth: 1`, only the merge commit itself is fetched — its parents are **not available** in the shallow clone. This means `git rev-parse --verify HEAD^2` fails, the fallback uses `github.event.pull_request.base.sha` (which can be stale when main moves forward), and `git diff --merge-base` also fails — ultimately producing a diff that includes unrelated files from main.

## Changes Made

- **`fetch-depth: 2`** on the checkout step in `get-all-test-matrices` so the merge commit's parents (`HEAD^1`, `HEAD^2`) are resolvable locally — this is what makes the existing `HEAD^1` detection logic actually work
- **Moved `git fetch`** for the base commit into **only the fallback branch** (non-merge-commit case) since with `fetch-depth: 2` the parent is already available in the normal path
- Added a `NOTE` comment linking the `fetch-depth: 2` requirement to the `HEAD^1`/`HEAD^2` checks

Push/dispatch/workflow_call paths are **unchanged** in behavior.

Resolved merge conflicts with main, which had independently adopted the `HEAD^1` approach with a safety fallback. This PR now adds the missing `fetch-depth: 2` that makes that approach actually work in CI's shallow clone.

## Testing

- [x] CI passed on this PR — all 20 seed test jobs were correctly **skipped** (not triggered), validating the fix
- [ ] This is a CI-only change — the PR's own seed workflow run is the real test
- [ ] Verify the `Detect affected generators and fixtures` step output shows only this PR's changed files, not generator files

## Human Review Checklist

- [ ] **`fetch-depth: 2` correctness**: Confirm that `actions/checkout@v6` with `fetch-depth: 2` fetches the merge commit + both parents for all PR event types (`opened`, `synchronize`, `reopened`, `ready_for_review`)
- [ ] **`git fetch` only in fallback**: The `git fetch --no-tags --depth=1` now runs only when `HEAD^2` doesn't exist (non-merge-commit fallback). Verify this is safe — in the normal PR path, `HEAD^1` is already local from `fetch-depth: 2`
- [ ] **Push path unaffected**: The `else` branch (push/dispatch) still fetches the last successful run SHA separately and is not impacted by the `fetch-depth` change
- [ ] **No performance regression**: Checkout with `fetch-depth: 2` took ~6s in CI (comparable to depth=1). Detection step itself took ~2s. Total well under 1 minute


amckinney:
Fix the merge conflicts.

fern-support (approved):

PR #13370, opened 2026-03-11
feat(csharp): emit readonly struct for literal properties, add generate-literals flag

Swimburger:
## Description

When `generate-literals` is enabled (or the deprecated `experimental-readonly-constants`), literal properties now emit a self-contained `readonly struct` with a built-in `JsonConverter`. Two strategies are used depending on how the literal is defined in the IR:

1. **Named literal alias types** (IR type declarations like `SomeLiteral: literal<"value">`) → generate a **sibling struct file** named from the IR type (e.g., `SomeLiteral.cs`), reusable across multiple parents.
2. **Inline literal types** (no named IR type, e.g., `prompt: literal<"You are a helpful assistant">`) → generate a **nested `{PropertyName}Literal` struct** inside the parent record.

Both string and boolean literals are supported.

**Link to Devin Session:** https://app.devin.ai/sessions/2190fe676f334977bd40c4a9151f43e1

### Configuration

A new `generate-literals` flag replaces the experimental flag:

```yaml
# generators.yml
config:
  generate-literals: true   # New — preferred
  # experimental-readonly-constants: true  # Deprecated — still works, logs warning
```

When `experimental-readonly-constants` is used, a deprecation warning is logged:
> The "experimental-readonly-constants" option is deprecated. Use "generate-literals" instead.

Flag resolution order: `generate-literals` ?? `experimental-readonly-constants` ?? `false`

### Before (inline literal)
```csharp
[JsonPropertyName("type")]
public string Type
{
    get => "flushed";
    set => value.Assert(value == "flushed", string.Format("'Type' must be {0}", "flushed"));
}
```

### After — Named literal (sibling file)
```csharp
[JsonRequired]
[JsonPropertyName("context")]
public SomeLiteral Context { get; init; } = new();
```
with a sibling file `SomeLiteral.cs` (named from the IR type declaration) containing the `readonly struct`.

### After — Inline literal (nested struct)
```csharp
public record SendRequest
{
    [JsonRequired]
    [JsonPropertyName("prompt")]
    public SendRequest.PromptLiteral Prompt { get; init; } = new();

    [JsonConverter(typeof(PromptLiteralConverter))]
    public readonly struct PromptLiteral
    {
        public const string Value = "You are a helpful assistant";
        public static implicit operator string(PromptLiteral _) => Value;
        // ... ToString, GetHashCode, Equals, ==, !=, JsonConverter
    }
}
```

### Boolean literal example
```csharp
[JsonConverter(typeof(SuccessLiteralConverter))]
public readonly struct SuccessLiteral
{
    public const bool Value = true;
    public static implicit operator bool(SuccessLiteral _) => Value;
    // ... GetBoolean(), WriteBooleanValue(), hardcoded error message
}
```

### Collision detection
If a parent record has a property `{X}` and another named `{X}Literal`, the nested struct name would collide. The generator detects this and appends a numeric suffix (e.g., `TypeLiteral2`).

---

### Feature flag changes
- **`CsharpConfigSchema.ts`** — Added `"generate-literals": z.boolean().optional()` to the config schema
- **`generation-info.ts`** — Added `generateLiterals` getter that resolves `generate-literals` ?? `experimental-readonly-constants` ?? `false`. Marked `enableReadonlyConstants` as `@deprecated`.
- **`GeneratorContext.ts`** — Logs deprecation warning when `experimental-readonly-constants` is used (warning message does not include URL to avoid broken links)
- **`generateFields.ts` / `generateModels.ts`** — Changed from `enableReadonlyConstants` to `generateLiterals` for gating literal struct generation

### Generator changes
- **New `generateLiteralType.ts`** — emits raw C# `readonly struct` files for **named** literal types (sibling files), plus exports `generateNestedStringLiteralBody` / `generateNestedBoolLiteralBody` templates for nested inline literal structs. All System namespace references use `global::` prefix to avoid conflicts with customer-defined types.
- **`generateFields.ts`** — two code paths for literal properties: `resolveNamedLiteralType()` for named IR literal aliases, `extractInlineLiteral()` + `computeNestedStructName()` + `buildNestedLiteralStruct()` for inline literals with collision detection. Registers class reference BEFORE building nested struct. Fallback path restores old get/set + Assert accessors for edge cases. Both paths emit `[JsonRequired]` + `{ get; init; } = new()`. **Added defensive check** in `buildNestedLiteralStruct()` to throw for unsupported literal types (addresses Devin Review finding).
- **`WrappedRequestGenerator.ts`** — Fixed missing collision detection: computes `allPropertyPascalNames` and passes it to both `generateField()` calls
- **`name-registry.ts`** — Five fixes to collision resolution:
  1. Added `typeRegistry.has(fullyQualifiedName)` check for 3+ type collisions
  2. Changed `fullyQualifiedNameOf()` to use `+` separator for nested types (matching .NET IL)
  3. **Added cross-format collision check with loop-based permutation generation** to detect all possible nested variants of non-nested types (e.g., `A.B.C.D` checks for `A.B.C+D`, `A.B+C+D`, `A+B+C+D`) — fixes Graphite AI review finding about incomplete deeply nested type collision detection
  4. Added namespace conflict check for dotted form of nested type FQNs
  5. Added unit tests for deeply nested type collision scenarios (77 tests total)
- **`ClassReference.ts`** — Added `csharpQualifiedName` getter that converts internal `+` separators back to `.` for C# code output
- **`Class.ts`** — added `addRawBodyContent()` method for injecting pre-formatted code blocks
- **`Writer.ts`** — fixed `stringifyImports()` crash when `addNamespace()` creates empty refs array
- **`generateModels.ts`** — returns `GenerateModelsResult` with both `files` and `literalTypeFiles`. Generates struct files from IR alias type declarations. Registers struct names through name registry before generating raw files.
- **`ObjectGenerator.ts`** — simplified `generateFields()` call
- **`CsharpProject.ts`** — added `sourceRawFiles` array and `addSourceRawFile()` method
- **`GeneratorContext.ts`** — made `getLiteralValue()` public
- **`ModelGeneratorCli.ts` & `SdkGeneratorCli.ts`** — destructure new return type, add raw literal files

### Mock server test organization changes
- **`MockServerTestGenerator.ts`** — Reorganized test file structure to use service-name subfolders instead of `_` suffixes for name collisions:
  - Added `getServiceNamespaceSegments()` helper that uses `fernFilepath.allParts` to include the service name
  - Updated `getTestNamespace()` to build namespace from service segments
  - Updated `getDirectory()` to place tests in service-name subdirectories (e.g., `MockServer/Headers/SendTest.cs` instead of `MockServer/SendTest.cs`)
  - **Impact**: ALL csharp-sdk test fixtures now organize tests by service, even single-service APIs (which get a subfolder named after that one service)
- **Seed test file reorganization**: Hundreds of test files moved from `MockServer/XTest.cs` to `MockServer/ServiceName/XTest.cs` across all csharp-sdk fixtures

### Versioning
- Bumped `csharp-model` to 0.2.0
- Bumped `csharp-sdk` to 2.25.0
- Updated version changelogs

### Seed test fixtures
- `literal/no-custom-config` — updated snapshots
- `literal/readonly-constants` — updated to use `generate-literals: true`
- `websocket/with-websockets` — updated to use `generate-literals: true`
- Updated snapshots: old per-property struct files removed, new IR-named struct files added, inline literal properties now have nested structs
- **All csharp-sdk fixtures**: Test files reorganized into service-name subfolders (e.g., `MockServer/Headers/SendTest.cs`, `MockServer/Query/SendTest.cs` instead of `MockServer/SendTest_`, `MockServer/SendTest__`)

---

1. **Mock server test subfolder organization** — Tests are now organized into subfolders by service name (e.g., `MockServer/Headers/SendTest.cs`) instead of using `_` suffixes for name collisions (`MockServer/SendTest_`). This change affects ALL csharp-sdk test fixtures. Uses `fernFilepath.allParts` to include the service name itself in the path, ensuring each service gets its own subdirectory.

2. **Biome formatting fix** — Fixed code style in `MockServerTestGenerator.ts` to satisfy biome linter (multiline `join()` call collapsed to single line).

3. **New `generate-literals` configuration flag** — Added as the standard option for enabling literal struct generation. Replaces `experimental-readonly-constants` which is now deprecated.

4. **Deprecation warning for old flag** — When `experimental-readonly-constants: true` is used, the generator logs a warning. Warning does not include a documentation URL to avoid broken links.

5. **Seed config migration** — Updated all seed test configurations to use `generate-literals: true` instead of `experimental-readonly-constants: true`.

6. **Flag resolution logic** — New `generateLiterals` setting uses fallback chain: `generate-literals` ?? `experimental-readonly-constants` ?? `false`.

7. **Backward compatibility preserved** — The old `experimental-readonly-constants` flag continues to work (with deprecation warning).

8. **Added `using System;` directive to top-level literal struct files** — Ensures consistency with nested struct approach.

9. **Namespace safety with `global::` prefix** — All System namespace references in literal struct files use the `global::` qualifier to avoid conflicts with customer-defined types.

10. **Fixed nested struct name divergence bug** — Class reference registration now happens BEFORE building the nested struct to prevent name mismatches.

11. **Fixed deeply nested type collision detection** — Graphite AI review identified incomplete collision check that only examined one level of nesting. Fixed with loop-based permutation generation that checks all possible nested variants (e.g., `A.B.C.D` now checks for `A.B.C+D`, `A.B+C+D`, `A+B+C+D`). Added comprehensive unit tests.

12. **Added defensive check for unsupported literal types** — Devin Review identified missing error handling for unknown literal types in `buildNestedLiteralStruct()`. Added explicit `throw` for unsupported types to fail fast instead of silently generating invalid code.

---

## Testing

- [x] Seed tests updated and passing (`csharp-model:literal` with full build/test scripts)
- [x] `csharp-sdk:literal` tests now pass (both `no-custom-config` and `readonly-constants` fixtures)
- [x] Lint and typecheck passing
- [x] Unit tests added/updated (77 name-registry tests passing, including deeply nested collision scenarios)
- [x] All 7 required CI checks passing: compile, biome, lint, test, test-ete, depcheck, versions.yml
- [x] Non-required seed test failures (11 shards) are all transient GitHub Actions auth errors (`fatal: could not read Username`), unrelated to changes
- [ ] Manual C# compilation for csharp-sdk variant fully verified

---

## Review Checklist

**High-risk items to review:**

1. **Mock server test subfolder organization scope** — This change reorganizes ALL csharp-sdk test fixtures, not just literal-related ones. Every service now gets its own subdirectory under `MockServer/`, even in single-service APIs. The choice to use `fernFilepath.allParts` (includes service name) instead of `packagePath` (parent path only) means top-level services create a subfolder. Verify this is the intended behavior and doesn't break any customer workflows that rely on specific test file paths.

2. **Flag precedence logic** — The `generateLiterals` getter uses `generate-literals ?? experimental-readonly-constants ?? false`. If a user explicitly sets `generate-literals: false` but also has `experimental-readonly-constants: true`, the `false` from `generate-literals` wins. This seems correct (new flag takes precedence), but worth confirming.

3. **`global::` namespace prefix scope** — System namespace references in literal struct files now use `global::` prefix to avoid customer type conflicts. This was only applied to literal struct generation code. Verify whether other generated files need the same treatment.

4. **Backward compatibility test coverage** — All seed tests now use `generate-literals: true`. The backward compatibility path is only tested by type-level logic, not by actual seed fixture execution.

5. **Dead code cleanup** — The `enableReadonlyConstants` getter is marked `@deprecated` but still defined. Intentional for backward compatibility with external consumers, but worth noting.

6. **Deeply nested type collision detection** — Loop-based permutation generation checks all possible nested variants (O(n) where n = number of namespace segments). Implementation generates variants like `A.B.C+D`, `A.B+C+D`, `A+B+C+D` for input `A.B.C.D`. The check is bidirectional but relies on namespace collision firing first for most cases. Unit tests cover the expected scenarios (namespace collision fires first, and pure cross-format collision when namespace doesn't conflict).

7. **Name registry FQN format change** — Using `+` separator in registry keys is a structural change. The `csharpQualifiedName` getter converts back to `.` for output. Verify this doesn't break edge cases.

8. **Raw file generation** — Named literal struct files are generated as raw `File` objects (bypassing AST). Registers names through name registry first. Verify CSharpier formats these correctly.

9. **Cross-format collision detection** — Bidirectional check for nested vs. non-nested type ambiguity. The nested→dotted check uses `replaceAll("+", ".")`. The dotted→nested check generates all permutations via loop.

10. **WrappedRequestGenerator collision detection** — Verify `allPropertyPascalNames` computation is correct for both `inlinedRequestBody` and `fileUpload` callbacks.

11. **Nested struct generation for inline literals** — Verify `{PropertyName}Literal` naming convention, collision detection, and nested struct shape match named literal structs.

12. **Boolean literal differences** — 4 key differences from string: `bool Value`, `implicit operator bool`, `GetHashCode()` (no `StringComparison.Ordinal`), `GetBoolean()`/`WriteBooleanValue()`, hardcoded error message.

13. **`[JsonRequired]` placement** — Added via `fieldAttributes.unshift()`. Verify STJ respects `[JsonRequired]` on `init`-only properties for all target frameworks.

14. **Defensive check for unknown literal types** — The `buildNestedLiteralStruct()` function now throws for unsupported literal types. The error uses `(literal as Literal).type` cast which assumes the literal object conforms to the union type structure. Should be safe but relies on IR contract.

15. **Test file namespace changes** — Hundreds of test files now have updated namespaces (e.g., `SeedLiteral.Test.Unit.MockServer.Headers` instead of `SeedLiteral.Test.Unit.MockServer`). Verify no hardcoded namespace references break.

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13370" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


seed/csharp-sdk/literal/no-custom-config/src/SeedLiteral.Test/Unit/MockServer/SendTest___.cs
```diff
@@ -0,0 +1,103 @@
+using NUnit.Framework;
+using SeedLiteral;
+using SeedLiteral.Test.Utils;
+
+namespace SeedLiteral.Test.Unit.MockServer;
+
+[TestFixture]
+public class SendTest___ : BaseMockServerTest
```
Swimburger: what's with these single double and triple underscore tests?

Swimburger: Is this because we should be generating subfolders/namespaces for these tests?

generators/csharp/base/src/context/GeneratorContext.ts
```diff
@@ -78,6 +78,14 @@ export abstract class GeneratorContext extends AbstractGeneratorContext {
         this.readOnlyMemoryTypes = new Set<PrimitiveTypeV1>(
             ast.convertReadOnlyPrimitiveTypes(this.settings.readOnlyMemoryTypes)
         );
+
+        // Log deprecation warning if the old flag is used
+        if (this.customConfig["experimental-readonly-constants"] === true) {
+            this.logger.warn(
+                'The "experimental-readonly-constants" option is deprecated. Use "generate-literals" instead. ' +
+                    "See https://buildwithfern.com/learn/sdks/features/literal-types for more details."
```
Swimburger: this URL does not exist

Swimburger (commented):
Scan your code for dangerous namespace usings.
we usually prefix with global::System... to avoid conflicts with types and namespaces that our customers could have in their OpenAPI specs.
They sometimes have System, Type, etc.

seed/csharp-sdk/literal/no-custom-config/src/SeedLiteral.Test/Unit/MockServer/SendTest___.cs
Swimburger: use subfolder/namespaces

fern-support (approved):

PR #13371, opened 2026-03-11
feat(go): inferred auth

patrickthornton:
## Description
Equivalent of https://github.com/fern-api/fern/pull/11062 in Go; adds support for basic inferred authentication _during Fern generation_, i.e. allow Fern to generate authentication from the shape of endpoints rather than read off an explicit declaration.

The general-use `token_provider.go` core utility that gets written (the only difference between Payroc's use of it between OAuth and inferred is that inferred assumes 1-hour expiry whereas OAuth assumes we never auto-refresh):
```go
// Code generated by Fern. DO NOT EDIT.

package core

import (
	"sync"
	"time"
)

const (
	// tokenExpirationBufferMinutes is subtracted from the token expiration time
	// to ensure we refresh the token before it actually expires.
	tokenExpirationBufferMinutes = 2

	// DefaultExpirySeconds is used as a fallback when the token response doesn't include
	// an expiry value.
	DefaultExpirySeconds = 3600 // 1 hour
)

// TokenProvider manages access tokens with caching and automatic refresh.
type TokenProvider struct {
	mu          sync.Mutex
	accessToken string
	expiresAt   time.Time

	// defaultExpirySeconds controls behavior when the token response doesn't include
	// an expiry value. If 0, the token is treated as never expiring. If > 0, the token
	// is assigned this default expiry (e.g. 3600 for inferred auth).
	defaultExpirySeconds int

	// fetchMu ensures only one goroutine fetches a new token at a time
	fetchMu sync.Mutex
}

// NewTokenProvider creates a new TokenProvider.
//
// defaultExpirySeconds controls the behavior when the token response doesn't include an
// expiry value. Use 0 to treat tokens as never expiring (e.g. for OAuth), or a positive
// value to apply a default expiry (e.g. 3600 for inferred auth).
func NewTokenProvider(defaultExpirySeconds int) *TokenProvider {
	return &TokenProvider{
		defaultExpirySeconds: defaultExpirySeconds,
	}
}

// SetToken sets the cached access token and its expiration time.
// The expiresIn parameter is the number of seconds until the token expires.
func (t *TokenProvider) SetToken(accessToken string, expiresIn int) {
	t.mu.Lock()
	defer t.mu.Unlock()
	t.accessToken = accessToken
	if expiresIn > 0 {
		bufferSeconds := tokenExpirationBufferMinutes * 60
		effectiveExpiresIn := expiresIn - bufferSeconds
		if effectiveExpiresIn < 0 {
			effectiveExpiresIn = 0
		}
		t.expiresAt = time.Now().Add(time.Duration(effectiveExpiresIn) * time.Second)
	} else if t.defaultExpirySeconds > 0 {
		// Apply default expiry when no expiry info is provided.
		effectiveExpiresIn := t.defaultExpirySeconds - (tokenExpirationBufferMinutes * 60)
		if effectiveExpiresIn < 0 {
			effectiveExpiresIn = 0
		}
		t.expiresAt = time.Now().Add(time.Duration(effectiveExpiresIn) * time.Second)
	} else {
		// No expiration info and no default — token won't auto-refresh based on time.
		t.expiresAt = time.Time{}
	}
}

// GetToken returns the cached access token if it's still valid.
// Returns an empty string if the token is expired or not set.
func (t *TokenProvider) GetToken() string {
	t.mu.Lock()
	defer t.mu.Unlock()
	if t.accessToken == "" {
		return ""
	}
	if !t.expiresAt.IsZero() && time.Now().After(t.expiresAt) {
		return ""
	}
	return t.accessToken
}

// GetOrFetch returns a valid token, fetching a new one if necessary.
// The fetchFunc is called at most once even if multiple goroutines call GetOrFetch
// concurrently when the token is expired. It should return (accessToken, expiresInSeconds, error).
func (t *TokenProvider) GetOrFetch(fetchFunc func() (string, int, error)) (string, error) {
	// Fast path: check if we have a valid token
	if token := t.GetToken(); token != "" {
		return token, nil
	}

	// Slow path: acquire fetch lock to ensure only one goroutine fetches
	t.fetchMu.Lock()
	defer t.fetchMu.Unlock()

	// Double-check after acquiring lock (another goroutine may have fetched)
	if token := t.GetToken(); token != "" {
		return token, nil
	}

	// Fetch new token
	accessToken, expiresIn, err := fetchFunc()
	if err != nil {
		return "", err
	}

	t.SetToken(accessToken, expiresIn)
	return accessToken, nil
}

// NeedsRefresh returns true if the token needs to be refreshed.
func (t *TokenProvider) NeedsRefresh() bool {
	t.mu.Lock()
	defer t.mu.Unlock()
	if t.accessToken == "" {
		return true
	}
	if !t.expiresAt.IsZero() && time.Now().After(t.expiresAt) {
		return true
	}
	return false
}

// Reset clears the cached token.
func (t *TokenProvider) Reset() {
	t.mu.Lock()
	defer t.mu.Unlock()
	t.accessToken = ""
	t.expiresAt = time.Time{}
}
```

Block that gets written to `client.go` for inferred auth:
```go
	inferredAuthProvider := core.NewTokenProvider(
		core.DefaultExpirySeconds,
	)
	authOptions := *options
	authClient := auth.NewClient(
		&authOptions,
	)
	options.SetTokenGetter(func() (string, error) {
		return inferredAuthProvider.GetOrFetch(func() (string, int, error) {
			response, err := authClient.GetTokenWithClientCredentials(context.Background(), &fern.GetTokenRequest{
				XApiKey:      options.XApiKey,
				ClientId:     options.ClientId,
				ClientSecret: options.ClientSecret,
			})
			if err != nil {
				return "", 0, err
			}
			if response.AccessToken == "" {
				return "", 0, errors.New(
					"inferred auth response missing access token",
				)
			}
			expiresIn := core.DefaultExpirySeconds
			if response.ExpiresIn > 0 {
				expiresIn = response.ExpiresIn
			}
			return response.AccessToken, expiresIn, nil
		})
	})
```

When **both OAuth and inferred auth** are present (e.g. `any-auth`, `endpoint-security-auth`), this now matches [Java's behavior](https://github.com/fern-api/fern/pull/11062): both auth paths are supported simultaneously. The OAuth block generates shared `authOptions`/`authClient`, and the inferred auth block handles `SetTokenGetter` — making inferred auth the default token source (equivalent to Java's default builder path). Since Go's `SetTokenGetter` is a single slot, the last call wins, and inferred auth is intentionally last.

Generates wire tests for the inferred auth, too, if `enable_wire_tests` is true.

## Changes Made
- New unified `TokenProvider` replaces the old `OAuthTokenProvider` — supports both OAuth (no default expiry) and inferred auth (1-hour default expiry)
- New `authUtils.ts` with shared helpers: `getOAuthClientCredentialsScheme`, `getInferredAuthScheme`, `getClientAccessPath`, `resolveTokenEndpointBodyProperties`
- `ClientGenerator.ts` updated to generate both OAuth and inferred auth token fetching when both are configured, with inferred auth's `SetTokenGetter` as the final call
- When both OAuth and inferred auth are present, OAuth's `writeOAuthTokenFetching()` only emits the shared `authOptions`/`authClient` setup (no token provider or `SetTokenGetter`), and `writeInferredAuthTokenFetching()` reuses those shared variables and emits the sole `SetTokenGetter`
- Dynamic snippets: added `getConstructorInferredAuthArgs` support for inferred auth parameters
- New seed fixtures: `inferred-auth-implicit`, `inferred-auth-explicit`, `inferred-auth-implicit-api-key`, `inferred-auth-implicit-no-expiry`, `inferred-auth-implicit-reference`
- Updated `any-auth` and `endpoint-security-auth` seed fixtures to show inferred auth as the default token source when both auth types are present

### ⚠️ Items for human reviewer to verify
- **Seed fixture accuracy**: The `any-auth` and `endpoint-security-auth` seed fixtures were manually updated (not regenerated). Please regenerate to confirm they match actual generator output.
- **`ClientId` vs `ClientID` field names**: The inferred auth block in `any-auth`/`endpoint-security-auth` uses `ClientID` (matching OAuth). Verify this is correct for the generated request type — the standalone inferred auth fixtures use `ClientId` (camelCase).
- **No OAuth `TokenProvider` when both auth types present**: Unlike Java which generates both `OAuthTokenSupplier` and `InferredAuthTokenSupplier`, Go now only generates `inferredAuthProvider` when both are configured (since `SetTokenGetter` is single-slot). Confirm this trade-off is acceptable.
- **v1 generator (`sdk.go`)**: Lines 718-724 still skip inferred auth `SetTokenGetter` when OAuth is present. Left unchanged since v2 overwrites `client.go`, but verify this assumption holds for all generation paths.

## Testing
All five `inferred-auth` fixtures pass; the new wire tests pass.

Link to Devin session: https://app.devin.ai/sessions/6498a6dda4db4c1aab5722c02cc96680


seed/go-sdk/any-auth/client/client.go
```diff
@@ -72,6 +71,30 @@ func NewClient(opts ...option.RequestOption) *Client {
 			return response.AccessToken, expiresIn, nil
 		})
 	})
+	inferredAuthProvider := core.NewTokenProvider(
+		core.DefaultExpirySeconds,
+	)
+	options.SetTokenGetter(func() (string, error) {
```
iamnamananand996, line 53: Hey @patrickthornton, I see this function is names 2 times. line 53.

It may need to combine both in 1 function.

iamnamananand996 (approved):

PR #13426, opened 2026-03-12
feat(ruby): Added configurable logging

HoaX7:
## Description
Refs: https://github.com/fern-api/fern/issues/13329
Adds configurable logging support to the Ruby v2 SDK generator, matching the architecture of the Java and Python SDK
logging systems. Logging is silent by default and can be enabled via the logging parameter when initializing the root
client.

New Core Utility Files (6 files)
 - `LogLevel` — Module with DEBUG, INFO, WARN, ERROR constants and level parsing
 - `ILogger` — Interface module for custom logger backend implementations
 - `ConsoleLogger` — Default logger implementation using Ruby's built-in Logger (writes to stderr)
 - `LogConfig` — Configuration class with level, logger, and silent attributes
 - `Logger` — Main logger class with level filtering and silent-by-default behavior
 - `LoggingMiddleware` — HTTP middleware that logs requests/responses with sensitive header redaction

Generator Changes
 - `RootClientGenerator.ts` — Added logging parameter to client constructor
 - `RawClient` — Accepts logging_config, initializes logger and middleware
 - `SdkGeneratorContext.ts` — Includes logging files in core AsIs files
 - `AsIs.ts` — Added logging files with proper load ordering

Features
 - ✅ Silent by default (no log output unless explicitly enabled)
 - ✅ Four log levels: :debug, :info, :warn, :error
 - ✅ Automatic HTTP request/response logging at debug level
 - ✅ Sensitive header redaction (Authorization, API keys, cookies, etc.)
 - ✅ Custom logger support via ILogger interface
 - ✅ Flexible config: accepts Hash, LogConfig, or nil

Architecture Alignment
This implementation follows the same architecture as the Java and Python SDK logging systems, ensuring consistency across
all Fern-generated SDKs.

## Testing
 - [x] Build compiles successfully (yarn compile for both ruby-v2/base and ruby-v2/sdk)
 - [x] CLI bundle builds successfully
 - [ ] Manual testing


iamnamananand996 (approved):

generators/ruby-v2/base/src/asIs/internal/logging/log_config.Template.rb
```diff
@@ -0,0 +1,39 @@
...
+      #   silent - true (no output unless explicitly enabled)
+      class LogConfig
+        # @return [Symbol] the minimum log level (:debug, :info, :warn, :error)
+        attr_reader :level
+
+        # @return [ILogger] the logger implementation
+        attr_reader :logger
+
+        # @return [Boolean] whether logging is silent (disabled)
+        attr_reader :silent
```
amckinney: I wonder if `silent` should just be configured whether or not the user passes in a no-op logger implementation. Given that it's an interface (i.e. `ILogger`), we might not need this.

generators/ruby-v2/base/src/asIs/internal/logging/logger.Template.rb
```diff
@@ -0,0 +1,101 @@
...
+      #
+      # @example Create via LogConfig
+      #   config = LogConfig.new(level: :debug, silent: false)
+      #   logger = Logger.from(config)
+      #   logger.debug("request sent")
+      #
+      # @example Create directly
+      #   logger = Logger.new(level: :debug, logger: ConsoleLogger.new, silent: false)
+      #   logger.info("connected")
+      class Logger
```
amckinney: With all the feedback in mind -- I think we can drop this type in favor of just the `ILogger` interface backed by concrete implementations. The concrete `class` doesn't actually buy us anything in this case.

We can simplify to the following:
1. The root client exposes an `ILogger` parameter.
2. We provide a `ConsoleLogger` for users to use (opt-in).
3. If a logger is not specified, we default to a `NoOpLogger` that we define internally which implements the silent logging.

HoaX7: Hi @amckinney I've cleaned this up and added `NoOpLogger` as required. please review, thanks

PR #13457, opened 2026-03-12
feat: add claude-code option to page-actions configuration

fern-support:
## Description
Adds claude-code as a toggleable option in the page-actions configuration in docs.yml, allowing users to enable/disable the "Connect to Claude Code" button. The FDR SDK (fern-platform) already supports claudeCode — this wires it through the CLI schema.

## Changes Made

-  Added claude-code to PageActionOption enum and PageActionOptions schema in DocsYmlSchemas.ts
- Added claudeCode: boolean to ParsedPageActionsConfig interface
- Added parsing logic in parseDocsConfiguration.ts (defaults to true)
- Added claudeCode passthrough in DocsDefinitionResolver.ts
- Updated auto-generated SDK types (api + serialization)

## Testing

- Lint passes (pnpm run check)
- Manual review of pattern consistency with existing options (chatgpt, claude, cursor, etc.)

### DIFF
--- a/packages/cli/configuration/src/docs-yml/DocsYmlSchemas.ts
+++ b/packages/cli/configuration/src/docs-yml/DocsYmlSchemas.ts
@@ -59,6 +59,7 @@ export const PageActionOption = z.enum([
     "chatgpt",
     "claude",
     "cursor",
+    "claude-code",
     "vscode"
 ]);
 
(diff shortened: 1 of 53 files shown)


packages/cli/configuration/src/docs-yml/DocsYmlSchemas.ts
```diff
@@ -59,6 +59,7 @@ export const PageActionOption = z.enum([
     "chatgpt",
     "claude",
     "cursor",
+    "claude-code",
```
chdeskur, line 62: make sure we add this to docs.yml config, and then generate the docs.yml schema with the new option

PR #13519, opened 2026-03-13
feat(csharp): rework enum serialization to eliminate reflection

Swimburger:
## Description

Linear ticket: Closes [FER-4622](https://linear.app/buildwithfern/issue/FER-4622/c-rework-enumconversion-to-not-use-reflection)

Eliminates all reflection usage from C# enum JSON serialization by generating per-enum serializer classes with compile-time dictionary lookups.

[Link to Devin Session](https://app.devin.ai/sessions/75bc57809500450fbd083d4cc48e1ecc)
Requested by: @Swimburger

## Changes Made

**Regular enums (`EnumGenerator.ts` / `Enum.ts`):**
- Each enum now generates a companion `{EnumName}Serializer` class in the same file
- Two static readonly dictionaries (`_stringToEnum`, `_enumToString`) provide O(1) bidirectional mapping
- `Read` uses `_stringToEnum.TryGetValue()` — returns `default` for unknown values
- `Write` uses `_enumToString.TryGetValue()` — writes `null` for unknown values
- Replaces `[JsonConverter(typeof(EnumSerializer<T>))]` → `[JsonConverter(typeof(TSerializer))]`

**String enums (`StringEnumGenerator.ts`):**
- Each string enum now generates a nested `{EnumName}Serializer` class inside the record struct
- `Read`: `new MyEnum(stringValue)` — no reflection needed (previously used `Activator.CreateInstance`)
- `Write`: `writer.WriteStringValue(value.{propertyName})` — uses the actual property name (may be `Value_` if a member is named "Value")
- Replaces `[JsonConverter(typeof(StringEnumSerializer<T>))]` → `[JsonConverter(typeof(T.TSerializer))]`

**Removed generic serializer files from as-is includes:**
- `EnumSerializer.Template.cs` and `StringEnumSerializer.Template.cs` are no longer included in generated output
- Corresponding test template files (`EnumSerializerTests`, `StringEnumSerializerTests`) no longer included either

**Updated template test files** to use inline serializers instead of generic ones.

- **Refactored regular enum serializers to use dictionary lookups** instead of switch statements, per review feedback. Two `static readonly Dictionary` fields provide bidirectional mapping; `Read`/`Write` use `TryGetValue()`.
- **Resolved second merge conflict** in `generators/csharp/sdk/versions.yml` — bumped SDK version to `2.28.0` (main claimed `2.27.0` for gRPC changes).
- **Previous fixes retained:** undiscriminated-unions `Value_` property fix, biome formatting fixes, changelog validation fixes.

## Testing

- [x] Seed tests: all 6 `exhaustive` fixture variants pass (build + tests)
- [x] Seed tests: `undiscriminated-unions` fixture passes (build + tests) — validates the `Value` property rename edge case
- [x] Seed tests: `enum` fixture (both `plain-enums` and `forward-compatible-enums`) passes
- [x] Full CI: all 61 checks pass

## Human Review Checklist

- **Dictionary `Write` behavior for unknown values:** The dictionary-based `Write` method writes `null` for unmapped enum values (via `TryGetValue` returning `false`). The previous switch-based approach threw `ArgumentOutOfRangeException`. Verify this is the desired behavior for forward compatibility.
- **`QueryStringConverterTest.Template.cs` still uses switch-based serializer:** The test helper `TestEnumSerializer` in this file uses switch statements rather than dictionaries. This is a test-only helper (not generated code), but is inconsistent with the generated pattern. May want to align for clarity.
- **`StringEnumGenerator.ts` — serializer created after `valueProperty`:** The nested serializer class is intentionally defined after `valueProperty` (line ~227) so that the `Write` method can reference `valueProperty.name`. If anyone reorders this code and moves the serializer back before `valueProperty`, the hardcoded `value.Value` bug will return for enums with a "Value" member. Worth verifying in the `undiscriminated-unions/KeyType.cs` seed output that `value.Value_` is generated correctly.
- **`Enum.ts:writeSerializerClass()`** — The companion serializer is written via raw string generation rather than the AST builder. Verify the generated dictionary initializers handle edge cases (special chars in wire values, `default` fallback for unknown values on deserialization).
- **`Extensions.cs` `Stringify` still uses reflection** — Determined to be unused in generated code, so left as-is. Confirm this is acceptable.
- **`popScope(false)` calls** in `Enum.ts` — Used to close dictionary initializer blocks without trailing newline. Verify the generated C# formatting is correct (semicolons placed correctly after closing braces).

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13519" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


generators/csharp/base/src/asIs/test/Json/EnumSerializerTests.Template.cs
```diff
@@ -49,12 +49,36 @@ public class DummyObject
...
     [EnumMember(Value = "known_value1")]
     KnownValue1,
 
     [EnumMember(Value = "known_value2")]
     KnownValue2
-}
\ No newline at end of file
+}
+
+internal class DummyEnumSerializer : JsonConverter<DummyEnum>
```
fern-support, line 62: I think we can do better. Let's create two maps enum -> string and string -> enum and just do lookups instead of switch statements.

generators/csharp/base/src/asIs/test/Json/StringEnumSerializerTests.Template.cs
```diff
@@ -135,4 +135,18 @@ public override int GetHashCode()
     public static bool operator ==(DummyEnum value1, string value2) => value1.Value.Equals(value2);
 
     public static bool operator !=(DummyEnum value1, string value2) => !value1.Value.Equals(value2);
+
+    internal class DummyEnumSerializer : JsonConverter<DummyEnum>
```
fern-support, line 139: Do we need an inline serializer, what was the previous StringEnumSerializer coming short for?

fern-support (changes requested):

fern-support (approved):

PR #13539, opened 2026-03-13
fix(openapi): preserve endpoint security auth overrides

kdawgwilk:
## Problem
Fern docs currently behave inconsistently when an OpenAPI spec uses per-operation `security` and `generators.yml` uses:

```yml
api:
  auth:
    endpoint-security: {}
```

In that setup, the OpenAPI spec should remain the source of truth for endpoint auth.

In practice, the importer was still treating generator auth overrides like global auth overrides:
- it rewrote `spec.security`
- it removed operation-level `security`
- it synthesized default auth onto endpoints that should stay public

That breaks the docs / API Explorer behavior in exactly the shape we hit downstream:
- protected endpoints can lose their rendered auth metadata
- `security: []` token endpoints can be treated as authenticated
- public endpoints can inherit auth they should not have

## Fix
This change makes `endpoint-security` behave as documented:
- do not rewrite the OpenAPI spec's auth/security when the override mode is `endpoint-security`
- preserve operation-level `security` instead of replacing it with synthesized global auth
- do not apply default auth override behavior to endpoints when `endpoint-security` is configured

## Regression Coverage
Adds a focused fixture/test that exercises the docs-facing case end to end:
- protected endpoint with explicit `security`
- token endpoint with explicit `security: []`
- public endpoint with no auth

The test asserts both layers:
- IR endpoint `auth` / `security`
- FDR endpoint `auth` / `authV2` / `multiAuth`

## Testing
- `pnpm --filter @fern-api/register test -- packages/cli/register/src/ir-to-fdr-converter/__test__/openapi-from-flag.test.ts -t "endpoint-security auth"`
- `pnpm exec biome lint packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts packages/cli/api-importers/openapi-to-ir/src/3.1/paths/operations/OperationConverter.ts packages/cli/register/src/ir-to-fdr-converter/__test__/openapi-from-flag.test.ts`
- `pnpm turbo run compile --filter=@fern-api/openapi-to-ir --filter=@fern-api/register`

## Note
A generated IR fixture in `ir-generator-tests` was also updated because CI's `test` job was rewriting that tracked file and then failing the repo cleanliness check.

### DIFF
--- a/packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts
+++ b/packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts
@@ -1,3 +1,4 @@
+import { visitRawApiAuth } from "@fern-api/fern-definition-schema";
 import { AuthScheme, FernIr, IntermediateRepresentation } from "@fern-api/ir-sdk";
 import { constructHttpPath, convertApiAuth, convertEnvironments } from "@fern-api/ir-utils";
 import { AbstractSpecConverter, Converters, ServersConverter } from "@fern-api/v3-importer-commons";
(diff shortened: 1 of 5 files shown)


kdawgwilk:
Still would love to get this merged to fix issues we have been running into

kdawgwilk:
Alright doesn't seem like anyone is going to look at this 😞

PR #13559, opened 2026-03-15
fix(typescript): include root markdown files in output when outputSourceFiles is false

Sigmabrogz:
## Summary
Fixes an issue where , , and  are not included in the output artifact when generating TypeScript SDKs with  (the default mode that builds ES and CJS artifacts).

## Problem
Fixes #13549.
The TypeScript generator builds the artifacts and then calls , which delegates to . Because it solely bundles the  (which only contains the compiled  and  files), any documentation files generated in the root of the project directory (such as ) are silently dropped from the final uploaded/downloaded artifact. This is inconsistent with Python and Go, which maintain their doc files.

## Solution
Updated  to explicitly include all  files residing in the root project directory alongside the dist contents. This enables  to preserve standard documentation files in the output ZIP without changing the core TS artifact behavior.

### DIFF
--- a/generators/typescript/utils/commons/src/typescript-project/PersistedTypescriptProject.ts
+++ b/generators/typescript/utils/commons/src/typescript-project/PersistedTypescriptProject.ts
@@ -364,12 +364,20 @@ export class PersistedTypescriptProject {
         unzipOutput?: boolean;
         logger: Logger;
     }): Promise<void> {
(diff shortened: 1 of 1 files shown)


Sigmabrogz:
This addresses the issue with the missing root markdown documentation files (`README.md`, `reference.md`, `CONTRIBUTING.md`) in the TypeScript generator's artifact when `outputSourceFiles` is false.

PR #13571, opened 2026-03-16
fix(csharp): add global:: prefix for namespace/type name collisions (CS0426)

iamnamananand996:
## Description

Fixes a C# SDK compilation error (CS0426) when the client class name matches a namespace root segment. For example, with config `namespace: Candid.Net` and `client-class-name: Candid`, the C# compiler resolves `Candid.Net` as looking for a `Net` member on the `Candid` type instead of the `Candid.Net` namespace.

Reported by a customer upgrading from generator v2.4.2 to v2.23.6.

### Core: type-namespace conflict detection
- **`name-registry.ts`**: Added `hasTypeNamespaceConflict()` method that detects when a name is both a registered type name and a root-level namespace segment (has `""` as parent in `namespaceNames`). Excludes known built-in identifiers (`System`, `NUnit`, `OneOf`) to avoid false positives on framework namespaces.
- **`ClassReference.ts`**: Added `hasTypeNamespaceConflict` check to:
  1. The `shouldGlobal` computation — ensures the FQN includes `global::` when the type's namespace root is shadowed
  2. The partial qualification ambiguity check — falls back to FQN when the qualification's first segment is shadowed
  3. **New:** When the class name itself matches a root namespace segment (CS0118 — namespace used like a type), uses FQN to disambiguate
- **`Writer.ts`**: Updated `stringifyImports` to add `global::` prefix to `using` directives when the namespace's first segment has a type/namespace conflict

### Pre-qualified namespace properties (per reviewer feedback)
Instead of wrapping every namespace usage with a `qualifyNamespace()` helper, the `namespaces` object now exposes pre-qualified properties:
- **`generation-info.ts`**: Added `qualifiedRoot` and `qualifiedCore` lazy properties that include `global::` prefix when the root segment has a type-namespace conflict. Also added the `qualifyNamespace()` method that powers them.
- **Template files** (`MultipartFormTests.Template.cs`, `RetriesTests.Template.cs`, `JsonAssert.Template.cs`, `OptionalComparer.Template.cs`, `AdditionalPropertiesComparer.Template.cs`): Use `namespaces.qualifiedRoot` / `namespaces.qualifiedCore` instead of raw `namespaces.root` / `namespaces.core`
- **Endpoint generators** (`HttpEndpointGenerator.ts`, `BytesOnlyEndpointRequest.ts`, `ReferencedEndpointRequest.ts`, `WrappedEndpointRequest.ts`, `WebsocketClientGenerator.ts`): Use `this.namespaces.qualifiedCore` for `HeadersBuilder` / `QueryStringBuilder` references
- **`CsharpProject.ts`**: Removed the `qualifyNamespace` lambda from all 5 template rendering contexts (no longer needed)
- **`with-generation.ts`**: Removed the now-unused `qualifyNamespace` convenience method

### Regression test fixtures
- **`seed.yml`**: Added `csharp-namespace-conflict` fixture with `namespace: Seed.CsharpNamespaceConflict`, `client-class-name: Seed` — reproduces the exact Candid scenario
- **`seed.yml`**: Removed `csharp-namespace-collision:namespace-client-collision` from `allowedFailures` — the fix now resolves the CS0426 errors for the Contoso.Net variant added in #13597
- **`seed/csharp-sdk/csharp-namespace-conflict/client-class-name-matches-namespace-root/`**: Generated seed output showing `global::` prefixes throughout
- **`seed/csharp-sdk/csharp-namespace-collision/namespace-client-collision/`**: Regenerated with `global::` prefixes applied

### Changelog
- **`versions.yml`**: Added v2.31.1 changelog entry

## Testing

- [x] Ran `pnpm run check` (biome lint) — passes
- [x] Seed generation succeeds for `csharp-namespace-conflict` (client-class-name-matches-namespace-root)
- [x] Seed generation succeeds for `csharp-namespace-collision` (namespace-client-collision — previously in allowedFailures)
- [x] Seed generation succeeds for `csharp-namespace-collision` (fully-qualified-namespaces — no diff, no false positives)

1. **Indentation in `RetriesTests.Template.cs`**: Lines 188-193 and 376-377 show indentation changes in the diff — verify these don't break the generated test file formatting.
2. **`context.namespaces.qualifiedCore` in templates**: The `RetriesTests.Template.cs` uses `context.namespaces.qualifiedCore` (via the `context` variable which is the full `SdkGeneratorContext`). Verify this property is accessible through that path.
3. **`hasTypeNamespaceConflict(this.name)` in ClassReference.ts** (line 233): New check that fires when a class name itself matches a root namespace segment (CS0118). Verify this doesn't produce false positives for classes that happen to share a name with a namespace segment but aren't actually conflicting.
4. **`namespace-client-collision` removed from `allowedFailures`**: CI must now successfully compile the Contoso.Net variant. If CI fails on this, it means the fix is incomplete.
5. **Built-in exclusion completeness**: The `knownBuiltInIdentifiers` set excludes `System`, `NUnit`, `OneOf`. If new external library references are added in the future, they may need to be added too.

Link to Devin session: https://app.devin.ai/sessions/8b1f947060004cd58b51c594d21e6da4


generators/csharp/base/src/asIs/test/RawClientTests/MultipartFormTests.Template.cs
```diff
@@ -1055,9 +1055,9 @@ private static string GetBoundary(MultipartFormDataContent content)
                 .Value?.Trim('"') ?? throw new global::System.Exception("Boundary not found");
     }
 
-    private static <%= namespaces.core %>.MultipartFormRequest CreateMultipartFormRequest()
+    private static <%= qualifyNamespace(namespaces.core) %>.MultipartFormRequest CreateMultipartFormRequest()
     {
-        return new <%= namespaces.core %>.MultipartFormRequest
+        return new <%= qualifyNamespace(namespaces.core) %>.MultipartFormRequest
```
Swimburger: can we make it so `namespaces.core` is already qualified?

Swimburger (approved):

PR #13670, opened 2026-03-18
fix(cli): skip semver prerelease tags in getLatestRelease

jochs:
## Summary

- `getLatestRelease` now detects when GitHub's "latest release" has a semver prerelease tag (e.g. `0.0.0-dev-babaf54d`) and falls back to scanning recent releases for the first stable version
- Adds `isSemverPrerelease()` helper using a simple regex — no new dependencies
- Prevents `semver.inc("0.0.0-dev-babaf54d", "patch")` from computing wrong versions like `0.0.1` instead of the correct next version

## Context

GitHub's "get latest release" API returns the most recent non-draft, non-prerelease release — but "prerelease" refers to GitHub's `prerelease` flag, not the semver concept. Releases tagged `0.0.0-dev-babaf54d` that aren't flagged as GitHub prereleases but ARE semver prereleases get returned as "latest", causing `computeSemanticVersion` to compute incorrect next versions.

## Test plan

- [x] Existing `getLatestRelease` tests still pass
- [x] New test cases cover: fallback to `listReleases`, skipping GitHub drafts/prereleases during fallback, all-prereleases returning undefined, v-prefixed prerelease tags

### DIFF
--- a/packages/commons/github/src/getLatestRelease.ts
+++ b/packages/commons/github/src/getLatestRelease.ts
@@ -54,7 +54,7 @@ export async function getLatestRelease(
         const releases = await octokit.rest.repos.listReleases({
             owner,
             repo,
-            per_page: 15
+            per_page: 100
         });
 
         for (const release of releases.data) {
--- a/packages/commons/github/src/getLatestRelease.ts
+++ b/packages/commons/github/src/getLatestRelease.ts
@@ -51,22 +51,30 @@ export async function getLatestRelease(
 
         // The "latest" release is a semver prerelease — scan recent releases
         // for the first stable (non-semver-prerelease, non-draft, non-GitHub-prerelease) tag.
-        const releases = await octokit.paginate(octokit.rest.repos.listReleases, {
-            owner,
-            repo,
-            per_page: 15
-        });
-
-        for (const release of releases) {
-            if (release.draft || release.prerelease) {
-                continue;
+        // uses pagination which may cause a lot of API calls if there are many recent prerelease/draft versions,
+        //  but this is a rare edge case and the GitHub API doesn't provide a better way to do this
+        const results = await octokit.paginate(
+            octokit.rest.repos.listReleases,
+            {
+                owner,
+                repo,
+                per_page: 15
+            },
+            (response, done) => {
+                for (const release of response.data) {
+                    if (release.draft || release.prerelease) {
+                        continue;
+                    }
+                    if (!isSemverPrerelease(release.tag_name)) {
+                        done();
+                        return [release.tag_name];
+                    }
+                }
+                return [];
             }
-            if (!isSemverPrerelease(release.tag_name)) {
-                return release.tag_name;
-            }
-        }
+        );
 
-        return undefined;
+        return results[0];
     } catch {
         // No releases found (404) or other error — return undefined
         return undefined;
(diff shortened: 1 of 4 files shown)


packages/commons/github/src/getLatestRelease.ts
```diff
@@ -2,12 +2,26 @@ import { Octokit } from "octokit";
 
 import { parseRepository } from "./parseRepository.js";
 
+/**
+ * Returns true if the tag looks like a semver prerelease version.
+ * Matches patterns like "1.0.0-beta", "v0.0.0-dev-abc123", etc.
+ */
+function isSemverPrerelease(tag: string): boolean {
```
jsklan, line 10: `semver.prerelease()` from the `semver` package already does this — returns `null` for stable versions and handles `v` prefixes.
`semver` is already used across the monorepo (core-utils, cli, seed, etc.) so it's a familiar dependency. 
Can we also use that import here instead of the regex?

pnpm-lock.yaml
```diff
@@ -5258,12 +5258,6 @@ importers:
         specifier: ^17.4.1
         version: 17.7.2
 
-  packages/cli/cli/dist/prod-unminified:
-    dependencies:
-      '@boundaryml/baml':
-        specifier: 'catalog:'
-        version: 0.219.0
```
jochs: not sure why this is here... I'm gonna back it out I guess?

jsklan (approved):

packages/commons/github/src/getLatestRelease.ts
```diff
@@ -29,7 +43,30 @@ export async function getLatestRelease(
...
+        if (!isSemverPrerelease(tag)) {
+            return tag;
+        }
+
+        // The "latest" release is a semver prerelease — scan recent releases
+        // for the first stable (non-semver-prerelease, non-draft, non-GitHub-prerelease) tag.
+        const releases = await octokit.rest.repos.listReleases({
+            owner,
+            repo,
+            per_page: 15
```
amckinney: It looks like we arbitrarily only look at the last 15 releases -- if we have a lot of prereleases in that set, we'll fail to find a stable release. We don't have to try _every_ release, but we should at least set a higher limit to prevent that from happening (e.g. 100 releases)? Thoughts?

We could add pagination support here so we actually check every release, but that could lead to a lot of round trips for repositories with tons of releases, so I'd rather leave that out to start.

amckinney (approved):
Beyond the page limit, lgtm! Once that's addressed I'll merge.

packages/commons/github/src/getLatestRelease.ts
jsklan: Hey @jochs, I just chatted with @amckinney and we aligned on increasing the per_page and adding pagination. as the way forward. 

```
const releases = await octokit.paginate(octokit.rest.repos.listReleases, {
      owner,
      repo,
      per_page: 100
  });
```

jochs: if you're using pagination why do you want to up the limit?

jochs: it the vast majority of cases you're gonna take the first release, and pagination already handles fetching the next page, so wouldn't you want to go back down to 15 to avoid sending tons of unnecessary bytes over the wire?

jochs: that's what I'm gonna put up, but feel free to just change it

packages/commons/github/src/getLatestRelease.ts
```diff
@@ -29,7 +43,30 @@ export async function getLatestRelease(
...
+            return tag;
+        }
+
+        // The "latest" release is a semver prerelease — scan recent releases
+        // for the first stable (non-semver-prerelease, non-draft, non-GitHub-prerelease) tag.
+        const releases = await octokit.paginate(octokit.rest.repos.listReleases, {
+            owner,
+            repo,
+            per_page: 15
+        });
```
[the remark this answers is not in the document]
jochs: no

PR #13759, opened 2026-03-19
feat(cli-v2): integrate Sentry error reporting

FedeZara:
## Summary

- Integrate Sentry error reporting into CLI v2's `TelemetryClient`, initialized alongside PostHog and gated on the same telemetry opt-in setting.
- Add selective error filtering via `shouldReportToSentry` in `withContext.ts` — only internal/unexpected errors are reported; user-facing errors (validation, auth, CLI usage) and `FernCliError` are excluded.
- Wire `@sentry/esbuild-plugin` into the build pipeline to upload source maps to Sentry during CI builds, with automatic cleanup of `.map` files from the distributed bundle.
- Pass `SENTRY_DSN_CLI` and `SENTRY_AUTH_TOKEN` secrets in `publish-cli.yml` for dev, prod, and manual release jobs.

## Setup required

Two GitHub repository secrets must be created before this takes effect:

| Secret | Purpose |
|---|---|
| `SENTRY_DSN_CLI` | Sentry DSN for the `cli` project (baked into the CLI binary at build time) |
| `SENTRY_AUTH_TOKEN` | Org-scoped Sentry auth token for source map uploads (`project:releases` + `org:read` scopes) |

## Design decisions

- **Centralized in TelemetryClient**: Sentry shares the same `isTelemetryEnabled()` check, `distinctId`, and `flush()` lifecycle as PostHog.
- **Client instance over global API**: Uses `Sentry.NodeClient` instance methods (`captureException`, `flush`) rather than global `Sentry.*` functions to avoid global state.
- **Environment tag**: Build-time `SENTRY_ENVIRONMENT` (`"development"` / `"production"`) is orthogonal to the `ci` tag (runtime detection via `is-ci`).
- **Per-project DSN naming**: Workflow secrets use `SENTRY_DSN_CLI` (mapped to generic `SENTRY_DSN` env var) to support future per-generator DSNs.
- **FernCliError excluded with TODO**: `FernCliError` loses context (it's a blank marker error). A TODO documents the need to refactor it to carry cause/code for future reportability.

## Test plan

- [ ] Add `SENTRY_DSN_CLI` and `SENTRY_AUTH_TOKEN` secrets to the repository
- [ ] Verify dev publish (`publish-cli.yml` dev job) uploads source maps to Sentry
- [ ] Install published dev CLI and trigger an internal error to confirm it appears in Sentry with correct release, environment, tags, and readable stack trace
- [ ] Verify telemetry opt-out (`FERN_TELEMETRY_DISABLED=1`) prevents Sentry initialization

### DIFF
--- a/packages/cli/cli-v2/build-utils.mjs
+++ b/packages/cli/cli-v2/build-utils.mjs
@@ -67,6 +67,8 @@ export async function buildCli(config) {
         tsupOverrides = {}
     } = config;
 
+    const version = process.argv[2] || packageJson.version;
+
     // Build with tsup
     await tsup.build({
         entry: { cli: "src/main.ts" },
@@ -77,7 +79,7 @@ export async function buildCli(config) {
         clean: true,
         env: {
             ...env,
-            CLI_VERSION: process.argv[2] || packageJson.version
+            CLI_VERSION: version
         },
         ...tsupOverrides
     });
--- a/packages/cli/cli-v2/src/telemetry/TelemetryClient.ts
+++ b/packages/cli/cli-v2/src/telemetry/TelemetryClient.ts
@@ -1,4 +1,5 @@
 import { AbsoluteFilePath, doesPathExist, join, RelativeFilePath } from "@fern-api/fs-utils";
+import * as Sentry from "@sentry/node";
 import { mkdir, readFile, writeFile } from "fs/promises";
 import IS_CI from "is-ci";
 import * as os from "os";
@@ -13,11 +14,13 @@ import type { Tags } from "./Tags.js";
 
 export class TelemetryClient {
     private readonly posthog: PostHog | undefined;
+    private readonly sentry: Sentry.NodeClient | undefined;
     private readonly baseTags: Tags;
     private readonly accumulatedTags: Tags = {};
     private cachedDistinctId: string | undefined;
 
     constructor({ isTTY }: { isTTY: boolean }) {
+        const isTelemetryEnabled = this.isTelemetryEnabled();
         const apiKey = process.env.POSTHOG_API_KEY;
         this.baseTags = {
             version: Version,
@@ -27,8 +30,22 @@ export class TelemetryClient {
             tty: isTTY,
             usingAccessToken: process.env.FERN_TOKEN != null
         };
-        this.posthog =
-            apiKey != null && apiKey.length > 0 && this.isTelemetryEnabled() ? new PostHog(apiKey) : undefined;
+        this.posthog = apiKey != null && apiKey.length > 0 && isTelemetryEnabled ? new PostHog(apiKey) : undefined;
+
+        const sentryDsn = process.env.SENTRY_DSN;
+        if (sentryDsn != null && sentryDsn.length > 0 && isTelemetryEnabled) {
+            const sentryEnvironment = process.env.SENTRY_ENVIRONMENT;
+            if (sentryEnvironment == null || sentryEnvironment.length === 0) {
+                throw new Error("SENTRY_ENVIRONMENT must be set when SENTRY_DSN is configured");
+            }
+            this.sentry = Sentry.init({
+                dsn: sentryDsn,
+                release: `cli@${Version}`,
+                environment: sentryEnvironment,
+                defaultIntegrations: false,
+                tracesSampleRate: 0
+            });
+        }
     }
 
     /** Send a named event that inherits base + accumulated properties. */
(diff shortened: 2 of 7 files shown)


packages/cli/cli-v2/src/telemetry/TelemetryClient.ts
```diff
@@ -13,11 +14,13 @@ import type { Tags } from "./Tags.js";
 
 export class TelemetryClient {
     private readonly posthog: PostHog | undefined;
+    private readonly sentry: Sentry.NodeClient | undefined;
     private readonly baseTags: Tags;
     private readonly accumulatedTags: Tags = {};
     private cachedDistinctId: string | undefined;
 
     constructor({ isTTY }: { isTTY: boolean }) {
+        const enabled = this.isTelemetryEnabled();
```
amckinney: nit: we might have other enabled features in this constructor later, so let's give this a more unique name.
```suggestion
        const isTelemetryEnabled = this.isTelemetryEnabled();
```

packages/cli/cli-v2/src/telemetry/TelemetryClient.ts
```diff
@@ -27,8 +30,19 @@ export class TelemetryClient {
...
-        this.posthog =
-            apiKey != null && apiKey.length > 0 && this.isTelemetryEnabled() ? new PostHog(apiKey) : undefined;
+        this.posthog = apiKey != null && apiKey.length > 0 && enabled ? new PostHog(apiKey) : undefined;
+
+        const sentryDsn = process.env.SENTRY_DSN;
+        this.sentry =
+            sentryDsn != null && sentryDsn.length > 0 && enabled
+                ? Sentry.init({
+                      dsn: sentryDsn,
+                      release: `cli-v2@${Version}`,
```
amckinney: Let's just name this `cli@{Version}` for now since we'll eventually just refer to this as the primary CLI and it'll share the same version set as the original.

packages/cli/cli-v2/src/telemetry/TelemetryClient.ts
```diff
@@ -27,8 +30,19 @@ export class TelemetryClient {
...
+                      environment: process.env.SENTRY_ENVIRONMENT || "development",
```
amckinney: Can we instead fail hard if the environment variable isn't set? That way we ensure there are no silent failures, and each compiled CLI is guaranteed to have this env setup in its build step? Plus, this is already set in `build.dev.mjs` so we're good to go there.

packages/cli/cli-v2/src/telemetry/TelemetryClient.ts
```diff
@@ -27,8 +30,19 @@ export class TelemetryClient {
...
+                      defaultIntegrations: false,
+                      tracesSampleRate: 0
```
amckinney: Do we intend to change this at any point?

packages/cli/cli-v2/build-utils.mjs
```diff
@@ -67,6 +68,21 @@ export async function buildCli(config) {
...
+    const version = process.argv[2] || packageJson.version;
+
+    const esbuildPlugins = [];
+    if (process.env.SENTRY_AUTH_TOKEN) {
+        esbuildPlugins.push(
+            sentryEsbuildPlugin({
+                org: "buildwithfern",
+                project: "cli",
+                authToken: process.env.SENTRY_AUTH_TOKEN,
+                release: { name: `cli-v2@${version}` },
```
amckinney: Let's update `cli@${version}` here too.

amckinney (approved):
Only nits, nice work!

.github/workflows/publish-cli.yml
```diff
@@ -60,6 +60,8 @@ jobs:
       AUTH0_DOMAIN: ${{ secrets.AUTH0_DOMAIN }}
       AUTH0_CLIENT_ID: ${{ secrets.AUTH0_CLIENT_ID }}
       POSTHOG_API_KEY: ${{ secrets.POSTHOG_PROJECT_API_KEY }}
+      SENTRY_DSN: ${{ secrets.SENTRY_DSN_CLI }}
+      SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
```
amckinney: nit: Update the env/secret to `FERN_SENTRY_AUTH_TOKEN` -- we have other Sentry auth tokens so I want the secret name to be more explicit.

packages/cli/cli-v2/src/telemetry/TelemetryClient.ts
FedeZara: Here's the rationale behind these choices:

- `defaultIntegrations: false` -- Disables Sentry's built-in integrations (HTTP tracing, console breadcrumbs, unhandled rejection handlers, etc.). In a CLI, these add overhead and noise. We only want explicit captureException calls, not automatic instrumentation. No reason to change this.

- `tracesSampleRate: 0` -- Disables performance/transaction tracing entirely. Tracing is designed for long-running services (web servers, APIs) where you want to measure request latency. A CLI is a short-lived process where tracing adds no value -- we care about errors, not performance spans. No reason to change this either.

Let me know if you have different opinions on this

PR #13770, opened 2026-03-19
fix(csharp): add ReadAsPropertyName/WriteAsPropertyName to all remaining custom JsonConverters

Swimburger:
## Description

Refs https://github.com/fern-api/fern/pull/13763

Extends the `IStringEnum` dictionary-key serialization fix from #13763 to **all remaining** custom `JsonConverter` types in the C# generator. Without these overrides, `System.Text.Json` throws `NotSupportedException` at runtime when any of these types is used as a `Dictionary<TKey, TValue>` key.

### High priority (most likely dictionary key types)
- **`EnumSerializer<TEnum>`** — generic template (`EnumSerializer.Template.cs`) and inline per-enum serializer (`Enum.ts`)
- **`DateTimeSerializer`** — round-trip DateTime format
- **`DateAsDateTimeConverter`** — date-only (`yyyy-MM-dd`) format; extracted `DateFormat` const to avoid string duplication

### Low priority (unlikely dictionary key types, added for completeness)
- **Discriminated union `JsonConverter`** (`UnionGenerator.ts`) — uses the discriminant string as the property name
- **Literal type converters** (`generateLiteralType.ts`) — all four variants (nested/standalone × string/bool)

### Unit tests
- Added dictionary key serialize/deserialize tests to `EnumSerializerTests.Template.cs`, `DateTimeJsonTests.Template.cs`, and `DateOnlyJsonTests.Template.cs`
- Added `ReadAsPropertyName`/`WriteAsPropertyName` to the `DummyEnumSerializer` test helper

### Other
- Bumped `versions.yml` for `csharp-sdk` (→ 2.37.1) and `csharp-model` (→ 0.5.1)
- Regenerated all csharp-model and csharp-sdk seed snapshots
- Updated `allowedFailures` in `seed/csharp-model/seed.yml` and `seed/csharp-sdk/seed.yml` to account for pre-existing failures from main (`mixed-case`, `unions-with-local-date`, `unions:no-custom-config`, `property-access` — all CS9035 `required` member errors in union types, unrelated to this PR)

### Enum `WriteAsPropertyName` consistency fix

During self-review, the initial implementation had a bug where `WriteAsPropertyName` silently no-oped when the enum value wasn't found in the dictionary. Fixed to match each serializer's existing `Write` behavior:

- **`EnumSerializer.Template.cs`**: Uses direct dictionary access `_enumToString[value]`, which throws `KeyNotFoundException` on unknown values — matching its `Write` method.
- **`Enum.ts` inline serializer**: Uses `_enumToString.TryGetValue(value, out var stringValue) ? stringValue : value.ToString()` — matching its `Write` method's `TryGetValue`-with-fallback pattern. Falls back to `value.ToString()` instead of `null` since null is not a valid property name.

## ⚠️ Human review checklist

1. **Two different enum error-handling strategies are intentional**: `EnumSerializer.Template.cs` throws on unknown enum values (matching its `Write`), while the `Enum.ts` inline serializer falls back to `value.ToString()` (matching its `Write` which falls back to `null`). Verify this divergence is acceptable.

2. **`UnionGenerator.ReadAsPropertyName`** constructs a union via `new UnionType(stringValue, stringValue)` — passing the raw string as both the discriminant and the value. This is a lossy round-trip (inner variant data is not preserved), which may be acceptable for dictionary-key usage but worth confirming.

3. **`DateTimeSerializer.ReadAsPropertyName`** uses `DateTime.Parse(reader.GetString()!, ...)` with a null-forgiving operator. This is safe for property name tokens (which are always non-null strings), but differs from `DateAsDateTimeConverter.ReadAsPropertyName` which uses the more defensive `TryParse` pattern.

4. **Test coverage gap for generic `EnumSerializer.Template.cs`**: The `DummyEnumSerializer` in `EnumSerializerTests.Template.cs` uses the `TryGetValue`-with-fallback pattern (matching the inline `Enum.ts` serializer), but does not exercise the throw-on-missing behavior of the generic `EnumSerializer.Template.cs`. Consider whether a separate test with the generic serializer is needed.

5. **`allowedFailures` additions**: Verify that `mixed-case`, `unions-with-local-date`, `unions:no-custom-config`, and `property-access` are genuinely pre-existing failures from main (all are CS9035 `required` member errors in union types) and not regressions from this PR's union `ReadAsPropertyName`/`WriteAsPropertyName` changes. These same fixtures fail identically on main's CI.

## Testing
- [x] Lint/typecheck passing (`pnpm run check`)
- [x] Seed snapshots regenerated for csharp-model and csharp-sdk
- [x] Unit tests added for dictionary key serialization (DateOnly, DateTime, native enums)
- [x] CI: All required checks pass; non-required seed failures are pre-existing from main

Link to Devin session: https://app.devin.ai/sessions/8440309f1e9944208c6105bd1d08aa20
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13770" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


generators/csharp/base/src/asIs/DateAsDateTimeConverter.Template.cs
```diff
@@ -23,6 +23,29 @@ public override void Write(Utf8JsonWriter writer, DateTime value, JsonSerializer
...
+        throw new JsonException($"Unable to convert \"{dateString}\" to DateTime.");
+    }
+
+    public override void WriteAsPropertyName(
+        Utf8JsonWriter writer,
+        DateTime value,
+        JsonSerializerOptions options
+    )
+    {
+        writer.WritePropertyName(value.ToString("yyyy-MM-dd"));
```
Swimburger: we should probably reuse a const value for the pattern

fern-support (approved):

PR #13805, opened 2026-03-20
fix(csharp): enforce global:: prefix for System namespace references

Swimburger:
## Description

Prevents CS0118/CS0426 compile errors when customer projects define a type or namespace named `System` (e.g., root namespace `Foo.Net` with a class `System`). All `System.*` references in generated C# source files now use the `global::System.*` prefix, and a CI validation script enforces this going forward.

**Exclusions:** `.csproj` `<Using Include>` elements do **not** use `global::` (MSBuild doesn't support it), and dynamic snippets (user-facing code examples) skip `global::` to keep output clean. Fixtures using `experimental-dotnet-format` are excluded from CI validation because `dotnet format` strips `global::` as a post-processing optimization.

### Generator source
- **`Writer.ts`**: Always emits `global::` prefix when a `using` directive's namespace starts with `System`. Added `skipGlobalQualifier` option to conditionally skip this (used by dynamic snippets).
- **`ClassReference.ts`**: Always emits `global::` for inline `System.*` type references, respecting the writer's `skipGlobalQualifier` flag. Fixed a bug where System types going through the `typeQualification` partial-qualification path would lose the `global::` prefix — now routes these through `fqName` when `shouldGlobal` is true.
- **`AstNode.ts`**: Threads `skipGlobalQualifier` option through `toString()` / `toStringAsync()` into the Writer constructor.
- **`EndpointSnippetGenerator.ts`**: Passes `skipGlobalQualifier: true` when generating dynamic snippets so user-facing code stays clean. Added comments documenting why disabling all `global::` qualifiers (not just System) is acceptable for snippets.
- **`CsharpProject.ts`**: `.csproj` implicit `<Using Include>` directives remain plain `System.*` (no `global::` — MSBuild doesn't support it).
- **`HttpEndpointGenerator.ts`**: Replaced hardcoded `"catch (System.Text.Json.JsonException)"` strings with proper `context.System.Text.Json.JsonException` class references so the `global::` prefix is applied automatically.
- **`WebsocketClientGenerator.ts`**: Replaced hardcoded `System.IO.MemoryStream` / `System.Text.Encoding` strings with `this.System.IO.MemoryStream` / `this.System.Text.Encoding_UTF8` class references. This also adds proper `using global::System.Text;` / `using global::System.IO;` directives, allowing the generated code to use short names (`MemoryStream`, `Encoding.UTF8`).

### Template .cs files (10 files)
- Fixed hardcoded `using System.*` → `using global::System.*` and inline `System.*` → `global::System.*` in:
  - `QueryStringConverterTest.Template.cs` (also fixed indentation inconsistencies)
  - `EnumSerializerTests.Template.cs`
  - `StringEnumSerializerTests.Template.cs`
  - `AdditionalPropertiesComparer.Template.cs`
  - `WebSocketClientTests.Template.cs`
  - `DisconnectionInfoTests.Template.cs`
  - `WebSocketConnectionTests.Template.cs` (including inline `System.Diagnostics.Stopwatch`, `System.Net.WebSockets.ClientWebSocket`)
  - `TestWebSocketServer.Template.cs` (including inline `System.Net.Sockets.TcpListener`)
  - `E2eWebSocketTests.Template.cs` (including inline `System.Net.WebSockets.WebSocketCloseStatus`, `System.Net.WebSockets.WebSocketException`)
  - `QueryTests.Template.cs` (inline `System.Collections.IEnumerable` cast)

### CI enforcement
- Added a validation script in `seed/csharp-sdk/seed.yml` and `seed/csharp-model/seed.yml` that greps all generated `.cs` files for unqualified `System` references and fails the build if any are found
- **Check 1**: `using` directives — catches `using System.Linq;` etc.
- **Check 2**: inline references — catches `new System.Exception()`, `System.Threading.Tasks.Task`, etc. using regex: `(?<!using |global::|\.| )System\.(Collections|IO|Linq|Net|Text|Threading|Runtime|Globalization|Diagnostics|Reflection)`
- Both checks use `grep -v -P ':\s*//'` to exclude comments (the colon-anchored pattern is required because `grep -rn` output includes the `file:line:` prefix, so `^\s*//` would never match)
- Excludes `DynamicSnippets` directories (intentionally skip `global::` for user-facing code)
- Fixture exclusion via explicit `SKIP_GLOBAL_VALIDATION` variable at top of script (currently: `fully-qualified-namespaces`)

### Seed test outputs
- Regenerated all csharp-sdk and csharp-model seed fixtures

### versions.yml
- SDK: → 2.52.0
- Model: → 0.7.0

## Testing
- [x] Unit tests updated (`Type.test.ts` expectations updated for `global::System.Type`)
- [x] Snapshot tests updated (`Literal.test.ts.snap`, `DynamicSnippetsGenerator.test.ts.snap`)
- [x] Dynamic snippets snapshots confirmed to NOT contain `global::` (user-facing)
- [x] Seed tests regenerated and passing
- [x] WebSocket seed fixtures (`websocket:with-websockets`, `websocket:no-custom-config`) regenerated and passing
- [x] All required CI checks passing (61/61)
- [x] Validation script tested locally against regenerated output — 0 violations in both csharp-sdk and csharp-model

## ⚠️ Human Review Checklist
1. **`skipGlobalQualifier` disables ALL `global::` prefixes in snippets, not just System** — With `skipGlobalQualifier: true`, the flag skips `global::` for ambiguous type names and type-namespace conflicts too. This is acceptable because snippets use empty `allNamespaceSegments`/`allTypeClassReferences` sets, making the ambiguity and conflict checks no-ops anyway. The flag just additionally suppresses the unconditional System prefix. See comments in `EndpointSnippetGenerator.ts`.
2. **`ClassReference.ts` `shouldGlobal` addition (line 223)** — When a System type goes through the `typeQualification` partial-qualification path, we now fall through to `fqName` (with `global::`) instead of writing `typeQualification + nameToWrite` (without `global::`). This is the correct fix but it broadens the scope slightly — previously `trace` fixture wrote `System.Collections.Generic.KeyValuePair`, now it writes `global::System.Collections.Generic.KeyValuePair`. Verify the additional seed output changes in `examples`, `validation`, and `csharp-namespace-conflict` fixtures look correct.
3. **WebSocket `InjectTestMessage` now uses short names** — The generator now emits `using global::System.Text;` and `using global::System.IO;` directives for WebSocket files, so `InjectTestMessage` uses `MemoryStream` / `Encoding.UTF8` instead of the previous fully-qualified `System.IO.MemoryStream` / `System.Text.Encoding.UTF8`. Verify these using directives are correct and the short names resolve properly.
4. **`experimental-dotnet-format` incompatibility is fundamental** — The `dotnet format` post-processor adds `<Using Include="System" />` to `.csproj` and then strips `global::` from using directives as redundant. The current workaround (skip validation via `SKIP_GLOBAL_VALIDATION`) is pragmatic but means that fixture isn't validated for other potential regressions.
5. **Inline validation regex may have edge cases** — The regex `(?<!using |global::|\.| )System\.(Collections|IO|...)` only covers known System sub-namespaces. A bare `System.Foo` for some unlisted namespace would not be caught. The generator itself handles all System namespaces correctly via `ClassReference.ts`, so this is a CI safety net, not the primary enforcement.

Link to Devin session: https://app.devin.ai/sessions/1f4ff43edd984ad68434fac160bcd07a
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/13805" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


generators/csharp/base/src/project/CsharpProject.ts
```diff
@@ -151,7 +151,7 @@ export class CsharpProject extends AbstractProject<GeneratorContext> {
                 )
                 .replace(/<ImplicitUsings>enable<\/ImplicitUsings>/, `<ImplicitUsings>disable</ImplicitUsings>`)
                 .replace(/<LangVersion>12<\/LangVersion>/, `<LangVersion>11</LangVersion>`)
-                .replace(/<\/Project>/, `<ItemGroup><Using Include="System" /></ItemGroup></Project>`)
+                .replace(/<\/Project>/, `<ItemGroup><Using Include="global::System" /></ItemGroup></Project>`)
```
Swimburger: Does this need to be done inside of cproj?

generators/csharp/base/src/project/CsharpProject.ts
```diff
@@ -870,11 +870,11 @@ ${projectGroup.join("\n")}
         <UsePortableDateOnly>true</UsePortableDateOnly>
     </PropertyGroup>
     <ItemGroup Condition="'$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'netstandard2.0'">
-      <Using Include="System.Net.Http" />
-      <Using Include="System.Collections.Generic" />
-      <Using Include="System.Linq" />
-      <Using Include="System.Threading" />
-      <Using Include="System.Threading.Tasks" />
+      <Using Include="global::System.Net.Http" />
```
Swimburger: Does this need to be done inside of cproj? I don't think this is an issue

generators/csharp/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
```diff
@@ -3,7 +3,7 @@
 exports[`snippets (default) > examples > 'GET /metadata (allow-multiple)' 1`] = `
 "using Acme;
 using Acme.Service;
-using System.Collections.Generic;
+using global::System.Collections.Generic;
```
Swimburger: In snippets we don't want to use global:: because that's user facing, and it's ugly

fern-support (approved):

PR #13832, opened 2026-03-20
feat(cli): added md check command 

Ryan-Amirthan:
## Description
Adds a command `fern docs md check` that runs markdown validation on all markdown referenced in your navigation structure. 
Additionally, markdown validation sometimes does not give the right line number (due to frontmatter offset) or due to Fern not accounting for different shapes for the error object.

## Changes Made
markdown validation sometimes does not give the right line number (due to frontmatter offset) and also due to Fern not accounting for different shapes for the error object. This PR also normalizes them to a uniform `filepath:line:column` format.

## Testing
Tested manually
CleanShot 2026-03-22 at 15 58 21@2x


packages/cli/cli/src/commands/validate/validateMdx.ts
```diff
@@ -0,0 +1,127 @@
...
+function extractLineColumn(error: string): { line?: number; column?: number } {
+    const match = error.match(/(?:at )?(\d+):(\d+)/);
+    if (match && match[1] && match[2]) {
+        return {
+            line: parseInt(match[1], 10),
+            column: parseInt(match[2], 10)
+        };
+    }
+    return {};
+}
```
amckinney: Don't we control the error format? Why rely on a brittle regular expression when we can instead return some structured `interface` that this code can more clearly act upon to extract the line/column?

See the `fern check` implementation (i.e. `fern check --json`) for reference.

packages/cli/cli/src/commands/validate/validateMdx.ts
```diff
@@ -0,0 +1,127 @@
...
+                    position?: { start?: { line?: number; column?: number } };
+                };
+                if (errorObj.place?.line != null && errorObj.place?.column != null) {
+                    const adjustedLine = errorObj.place.line + frontmatterLineCount;
+                    errorMessage = `${adjustedLine}:${errorObj.place.column} ${errorMessage}`;
+                } else if (errorObj.position?.start?.line != null && errorObj.position?.start?.column != null) {
+                    const adjustedLine = errorObj.position.start.line + frontmatterLineCount;
+                    errorMessage = `${adjustedLine}:${errorObj.position.start.column} ${errorMessage}`;
+                }
+            }
```
amckinney: Depending if/how we refactor the `parseMarkdownToTree` error reporting, we might be able to simplify this error handling code quite a bit. Ideally there is some `ValidationViolation` class (or something similar) which tracks where it is meant to be written. 

That way all call-sites can simply access the `toString` method. For example, see the [ValidationIssue](https://github.com/fern-api/fern/blob/0203a8d10e97af46b2dbbf14c2d774dce9167d36/packages/cli/yaml/loader/src/ValidationIssue.ts#L8) class for reference.

--

If all of this is too difficult to refactor here, we can proceed with a simple helper function to replace this for now.

packages/cli/cli/src/commands/validate/validateMdx.ts
Ryan-Amirthan: true, should now be using the error object

packages/cli/cli/src/commands/validate/validateMdx.ts
Ryan-Amirthan: Will other files be calling this function to format MDX errors though?

packages/cli/cli/src/cli.ts
```diff
@@ -1917,6 +1919,49 @@ function addDocsBrokenLinksCommand(cli: Argv<GlobalCliOptions>, cliContext: CliC
     );
 }
 
+function addDocsLintCommand(cli: Argv<GlobalCliOptions>, cliContext: CliContext) {
```
amckinney: nit: addDocsMdCheckCommand

packages/cli/cli/src/commands/validate/validateMdx.ts
```diff
@@ -0,0 +1,119 @@
...
+                } else if (errorObj.place?.line != null) {
+                    line = errorObj.place.line + frontmatterLineCount;
+                    column = errorObj.place.column;
+                } else if (errorObj.position?.start?.line != null) {
+                    line = errorObj.position.start.line + frontmatterLineCount;
+                    column = errorObj.position.start.column;
+                }
+            }
+
+            errors.push(new MdxValidationError(filepath, message, line, column));
```
amckinney, line 95: We should report the files as relative file paths (not absolute filepaths). We can either update the underlying function to return relative paths, or compute the relative path using the current working directory (cwd) here.

amckinney (approved):

PR #13834, opened 2026-03-20
feat(cli): link github run id for dashboard

kennyderek:
Link GitHub run id for displaying in dashboard

### DIFF
--- a/packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
+++ b/packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
@@ -78,7 +78,7 @@ export function sanitizeRelativePathForS3(relativeFilePath: RelativeFilePath): R
 }
 
 interface CISource {
-    type: string;
+    type: "github" | "gitlab" | "bitbucket";
     repo?: string;
     runId?: string;
     runUrl?: string;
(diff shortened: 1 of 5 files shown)


packages/cli/generation/remote-generation/remote-workspace-runner/src/publishDocs.ts
```diff
@@ -77,6 +77,16 @@ export function sanitizeRelativePathForS3(relativeFilePath: RelativeFilePath): R
...
 
+interface CISource {
+    type: string;
+    repo?: string;
+    runId?: string;
+    runUrl?: string;
+    commitSha?: string;
+    branch?: string;
+    actor?: string;
+}
```
sbawabe, line 88: can we make this a discriminated union with specific `type` values rather than have everything be optional?

sbawabe: OR, if they all share the same metadata, specify expected values for `type`?

kennyderek: lgtm!     type: "github" | "gitlab" | "bitbucket";

sbawabe (approved):

PR #13887, opened 2026-03-21
fix(typescript): replace ts-morph AST mutations with string-based regex in fixImportsForEsm

lifanzou:
## Description
Refs FER-8691

Replace the ts-morph AST mutation approach in `fixImportsForEsm` with direct string-based regex replacement. This was the dominant bottleneck in TypeScript SDK generation — ts-morph's `setModuleSpecifier()` triggers an expensive node replacement cascade that degrades super-linearly with file count.

## Changes Made
- Replaced ts-morph `Project` + `setModuleSpecifier()` with regex on raw file text
- Added tsconfig `include`/`exclude` resolution (walking `extends` chain) to replicate ts-morph's file scoping
- Preserved identical modification logic: add `.js` extension, add `/index.js`, replace `.ts` with `.js`
- Preserved identical caching strategy (normalized path → modification type)
- Removed `ts-morph` import (no longer needed by this module)

## Performance (real Stripe OpenAPI spec via `seed run` — production pipeline, Docker, 2,299 generated files)

| Metric | Before | After | Speedup |
|--------|--------|-------|---------|
| `fixImportsForEsm` | **19.0s** | **0.18s** | **104x** |

Measured via `pnpm seed run --generator ts-sdk` against the real Stripe OpenAPI spec (7.2MB, ~587 endpoints) through the full production pipeline (Docker container, IR generation from OpenAPI, biome formatting). Timing captured via instrumented `console.log` in `postProcess()`.

## Root Cause
`setModuleSpecifier()` calls `replaceCompilerNode` → `_replaceCompilerNodeFromFactory` → `_clearInternals` → `#handleChildren`, which invalidates cached compiler info for all downstream AST nodes. With thousands of mutations, each subsequent call pays for all previous ones — super-linear degradation. V8 CPU profiling confirmed 99.3% of `fixImportsForEsm` time was in this mutation cascade, not in file parsing or type resolution.

## Testing
- [x] 196/196 seed test fixtures pass (`pnpm seed test --generator ts-sdk --skip-scripts`)
- [x] Zero git diff in seed output — generated code is byte-identical before and after
- [x] Profiled against real Stripe OpenAPI spec through production pipeline (`seed run`)


generators/typescript/utils/commons/src/typescript-project/fixImportsForEsm.ts
```diff
@@ -172,3 +170,68 @@ function determineModification(
...
+        includeDirs:
+            result.include.length > 0 ? result.include.map((inc) => path.resolve(projectPath, inc)) : [projectPath],
+        excludeDirs: result.exclude.map((excl) => path.resolve(projectPath, excl))
+    };
+}
+
+// Walk the tsconfig extends chain to collect include/exclude arrays.
+// TypeScript merges: child "include" overrides parent "include", same for "exclude".
+function resolveFromTsconfig(tsconfigPath: string): { include: string[]; exclude: string[] } {
+    if (!existsSync(tsconfigPath)) {
```
Swimburger: let's the async equivalent

generators/typescript/utils/commons/src/typescript-project/fixImportsForEsm.ts
```diff
@@ -172,3 +170,68 @@ function determineModification(
...
+        return { include: [], exclude: [] };
+    }
+
+    try {
+        const raw = JSON.parse(readFileSync(tsconfigPath, "utf-8"));
```
Swimburger: let's use the async equivalent

generators/typescript/utils/commons/src/typescript-project/fixImportsForEsm.ts
```diff
@@ -24,100 +25,97 @@ type ImportModificationType = (typeof ImportModification)[keyof typeof ImportMod
...
+    const BATCH_SIZE = 100;
+    for (let i = 0; i < tsFiles.length; i += BATCH_SIZE) {
+        const batch = tsFiles.slice(i, i + BATCH_SIZE);
+        await Promise.all(
+            batch.map(async (filePath) => {
+                const content = await readFile(filePath, "utf-8");
+                const newContent = fixImportsInSource(content, filePath, fileExistenceCache, importModificationCache);
+                if (newContent !== content) {
+                    await writeFile(filePath, newContent);
+                }
```
Swimburger, line 57: any chance we could use streams? I wonder if it would help with perf

Swimburger (changes requested):

generators/typescript/utils/commons/src/typescript-project/fixImportsForEsm.ts
lifanzou: | Approach | Median | Avg | Min | Max |
| :--- | ---: | ---: | ---: | ---: |
| buffered (readFile/writeFile) | 741ms | 856ms | 376ms | 1,359ms |
| stream (readline/writeStream) | 731ms | 1,236ms | 514ms | 3,519ms |

Using the larger IR in the original overstated profliing (10k files, 5 run average),  Looks like essentially no difference. This is probably because the individual files are all relatively small.

PR #13898, opened 2026-03-22
fix(swift): optimize snippet generation performance (~80s speedup)

lifanzou:
## Summary

- **Reuse `NameRegistry` across `DynamicSnippetsGeneratorContext` clones** — Every call to `generateSync()` cloned the context and rebuilt the entire symbol registry (all IR types + endpoints) from scratch. For Stripe's spec this was ~35ms × 1142 snippets = ~40s wasted. The registry is read-only during generation, so clones now share it.
- **Share a single `DynamicSnippetsGenerator` between README and Reference generation** — Previously each created its own instance and generated snippets for all endpoints independently, doubling the work.

### Benchmarks (Stripe OpenAPI spec — 587 endpoints, 1142 snippets)

| Phase | `main` | This PR | Speedup |
|-------|--------|---------|---------|
| `output (github)` | 104,006ms | 2,805ms | **37x** |

## Test plan

- [x] `pnpm seed test --generator swift-sdk --fixture exhaustive` passes with zero diff in generated output
- [x] Verified on Stripe's real OpenAPI spec locally on both `main` and this branch


lifanzou:
This pattern is applicable to multiple languages (Rust, java-v2, go-v2, ruby-v2, php) but I will need to benchmark the improvements for each individually.

patrickthornton (approved):

PR #13920, opened 2026-03-23
chore(deps): upgrade @fern-api/fdr-sdk to 1.1.13-c1ad12a2b8

dsinghvi:
## Description

Upgrades `@fern-api/fdr-sdk` from `0.145.12-b50d999d1` to `1.1.13-c1ad12a2b8`. The new SDK version uses oRPC/zod for type inference instead of Fern-generated types, introducing several breaking changes that are addressed here.

The final SDK version (`1.1.13-c1ad12a2b8`) includes the `DocsConfig.root` field typed as `RootNode` (via [fern-platform#8856](https://github.com/fern-api/fern-platform/pull/8856)), eliminating the need for `as FernNavigation.V1.RootNode` casts.

## Changes Made

- **Version bump**: Updated `pnpm-workspace.yaml` catalog, `package.json` overrides, and lockfile to `1.1.13-c1ad12a2b8`
- **oRPC response pattern migration**: Converted all FDR client call sites from `{ ok, body, error }` response pattern to try/catch (oRPC returns responses directly and throws on errors)
- **Enum string literals**: Replaced namespace-based enum access (e.g. `CjsFdrSdk.docs.v1.commons.PageActionOption.CopyPage`) with string literals (`"copyPage"`) — the new SDK defines these as `z.enum(...)` so there are no exported enum objects
- **API path restructuring**: `fdr.api.v1.register` → `fdr.api.register`, `fdr.sdks.versions.computeSemanticVersion` → `fdr.sdks.computeSemanticVersion`
- **Type renames**: `DynamicIrUpload` → `DynamicIRUpload`, removed `DocsV2Write.AuthConfig` explicit annotation
- **`uploadUrls` type change**: Adapted to new `Record<string, string>` shape (previously `Record<string, { fileId, uploadUrl }>`)
- **`finishDocsRegister` signature**: Changed from positional args to single object parameter
- **Navigation node types**: Added required `collapsed` property to all navigation node construction sites
- **Null coalescing**: Added `?? undefined` in places where the SDK now returns `T | null | undefined` instead of `T | undefined`
- **Removed RootNode casts**: Removed `as FernNavigation.V1.RootNode` assertions now that the SDK properly types the field
- **authConfig consistency**: Fixed preview path to use the shared `authConfig` variable (including `authType: "sso"` for private sites) instead of an inline object that omitted it
- **Unused imports**: Removed stale imports (`APIV1Write` from `ChangelogNodeConverter.ts`, `FernNavigation` from `dry.test.ts`) after type migration
- **Merge conflict resolution**: Resolved multiple rounds of merge conflicts with `main`, preserving the oRPC error pattern while incorporating upstream changes (e.g. `resolvedUrl` variable renames, new `resolveTarget`/`resolvePreviewUrlFromId` helpers, `CliError.Code` additions)

## ⚠️ Human Review Checklist

These items carry runtime risk and cannot be verified through type-checking alone:

1. **Placeholder values in `checkSdkDynamicIrExists`** (`publishDocs.ts` ~line 724): Uses `apiId: ""` and `irVersions: []` because the old API signature `{ orgId, snippetConfiguration }` no longer exists. **These placeholders will not work correctly at runtime — need correct values.**

2. **Placeholder values in `getSdkDynamicIrUploadUrls`** (`runRemoteGenerationForGenerator.ts` ~line 420): Same issue — uses `apiId: ""` and `irVersions: []` instead of the original `version` and `snippetConfiguration` params. **Will not work correctly at runtime.**

3. **Error handling in try/catch blocks**: Caught errors are cast as `Record<string, unknown>` and error types extracted via `errorObj?.error`. This depends on the oRPC error shape — **verify the error structure matches at runtime** (e.g. does it actually have an `.error` property?).

4. **`convertToFilePathPairs` now uses the key as `fileId`** (`publishDocs.ts` ~line 609): Previously used a dedicated `fileId` from the upload response object. Verify the server accepts file keys as IDs.

5. **`authConfig` includes `authType: "sso"`** (line 151 of `publishDocs.ts`): The original code included this field and it is now used consistently in both preview and production paths. Verify the new SDK type actually sends/accepts this field rather than silently dropping it.

## Testing

- [x] `pnpm compile` passes (170/170 packages, 0 errors)
- [x] `pnpm run check` (biome lint/format) passes
- [x] All required CI checks pass
- [ ] Unit tests added/updated — no new tests; existing tests updated for type compatibility
- [ ] Manual/runtime testing — not performed; changes verified through type checking only

Link to Devin session: https://app.devin.ai/sessions/b596dc8a05504f5880f5d49ca471d06c
Requested by: @dsinghvi


packages/cli/configuration-loader/src/docs-yml/parseDocsConfiguration.ts
```diff
@@ -357,19 +357,19 @@ function convertPageActionOption(
 ): CjsFdrSdk.docs.v1.commons.PageActionOption {
     switch (option) {
         case "copy-page":
-            return CjsFdrSdk.docs.v1.commons.PageActionOption.CopyPage;
+            return "copyPage";
```
chdeskur, line 399: why not use the VALUE of the enum?

packages/cli/docs-preview/src/runAppPreviewServer.ts
```diff
@@ -146,7 +146,7 @@ class SlugChangeTracker {
         if (docsDefinition.config.root) {
             // Convert V1 navigation root to latest version
             const migratedRoot = FernNavigation.migrate.FernNavigationV1ToLatest.create().root(
-                docsDefinition.config.root
+                docsDefinition.config.root as FernNavigation.V1.RootNode
```
chdeskur: why is `as` required here

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/valid-markdown-link.ts
```diff
@@ -45,7 +45,9 @@ export const ValidMarkdownLinks: Rule = {
 
         // TODO: this is a bit of a hack to get the navigation tree. We should probably just use the navigation tree
         // from the docs definition resolver, once there's a light way to retrieve it.
-        const root = FernNavigation.migrate.FernNavigationV1ToLatest.create().root(resolvedDocsDefinition.config.root);
+        const root = FernNavigation.migrate.FernNavigationV1ToLatest.create().root(
+            resolvedDocsDefinition.config.root as FernNavigation.V1.RootNode
```
chdeskur: it seems like we're coercing the `RootNode` type a lot

packages/cli/docs-preview/src/runAppPreviewServer.ts
dsinghvi: Can you explain which descendant types dont align since ideally we would like the types to align and not use any `as` conversion

fern-support:
Updated the PR with the following changes:

**Merged with main** (resolved 6 conflicting files)

**Updated `@fern-api/fdr-sdk`** from `1.1.13-31b7d20b66` → `1.1.13-c1ad12a2b8` (latest version from main)

**Addressed review comments on type casts:**
- Replaced the `as FernNavigation.V1.RootNode` assertions in `runAppPreviewServer.ts`, `valid-markdown-link.ts`, and `missing-redirects.ts` with an `isV1RootNode` type guard function that checks the `"type"` discriminant. The FDR SDK types `config.root` as `{}` via zod inference; the guard safely narrows this to `FernNavigation.V1.RootNode` without a blind cast.
- Note: `parseDocsConfiguration.ts` still has `as unknown as docsYml.ParsedDocsConfiguration["layout"]` — this remains necessary because the zod-inferred layout config type doesn't match the SDK's layout type annotations. This was already present in the PR with the inline comment "Cast needed until the fern-platform companion PR merges and the SDK is updated."

**Other fixes from the merge:**
- Updated `ir.apiName.originalName` → `getOriginalName(ir.apiName)` for IR v66 `NameOrString` compatibility
- Removed `definitionV2` and `docsUrl` parameters not present in the new SDK's `registerApiDefinition` type
- Added `?? undefined` to snippetsConfig fields that now return `string | null | undefined` in the new SDK

PR #14075, opened 2026-03-25
feat(cli): add image schema for custom container registry support

fern-support:
## Summary

Adds support for pulling generator Docker images from custom container registries during `fern generate --local`. Users can specify an `image` object with `name` and `registry` fields instead of the standard `name` field in `generators.yml`.

The `image.name` must be a recognized fern generator name (e.g., `fern-typescript-sdk`) and the `version` must match a published version. This allows IR version resolution to work normally without requiring a manual `ir-version` override. The intended use case is organizations that mirror fern's public Docker images into a private registry.

### Example usage

```yaml
groups:
  typescript:
    generators:
      # Default: pulls from Docker Hub (no change from today)
      - name: fern-typescript-sdk
        version: 3.60.9

      # Custom registry: same generator name/version, different registry
      - image:
          name: fern-typescript-sdk
          registry: ghcr.io/myorg
        version: 3.60.9
```

## Changes

- **Schema**: Add `GeneratorImageObjectSchema` (`{ name, registry }`) to support custom registries
- **CLI v1**: `GeneratorInvocationSchema` becomes an undiscriminated union of `DefaultGeneratorInvocationSchema` (uses `name`) and `CustomGeneratorInvocationSchema` (uses `image`), enforcing mutual exclusivity at the type level
- **CLI v2**: `image` on `SdkTargetSchema` is widened from `string` to `string | { name, registry }` union
- **Name normalization**: Custom image names go through the same Docker Hub org normalization as regular generators, ensuring IR version resolution works
- **Local generation**: Wire `containerImage` through the local generation runner to pull from the resolved registry
- **Remote generation guard**: Block custom `image` for remote generation (non-`--local`) to prevent arbitrary code execution via Fiddle
- **Generator upgrade**: Skip custom image generators during `fern upgrade` with a warning
- **ETE tests**: Replace Docker-only integration tests with proper end-to-end tests that spin up a local authenticated registry, push a real fern generator image, and run `fern generate --local` through the CLI — covering both successful generation and auth failure cases

## Test plan

- [x] Unit tests for config parsing (v1 `convertGeneratorsConfiguration`, v2 `SdkConfigConverter`)
- [x] Unit tests for `GeneratorPipeline` image reference construction
- [x] ETE test: `fern generate --local` succeeds with authenticated custom registry
- [x] ETE test: `fern generate --local` fails when not authenticated with custom registry
- [x] Backward compatibility: existing `name`-only invocations still work (default `fern init` test)


packages/cli/config/src/schemas/GeneratorImageSchema.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from "zod";
+
+export const GeneratorImageObjectSchema = z.object({
+    name: z.string(),
+    registry: z.string()
+});
```
amckinney, line 6: nit: For CLI v2 it'd be awesome if we treated `image` as a union so that users could still explicitly configure the image name without the regsitry. This is because `name` no longer exists in CLI v2 (in favor of `lang` or the default target name).

jsklan: fixed ✅

amckinney (approved):

packages/cli/cli-v2/src/migrator/converters/convertSdkTargets.ts
```diff
@@ -186,6 +189,14 @@ function convertGeneratorToTarget(options: ConvertGeneratorOptions): ConvertGene
         output: {}
     };
 
+    // If the generator uses a custom image with registry, set the image field
+    if ("image" in generator) {
+        target.image = {
+            name: generator.image.name,
+            registry: generator.image.registry
+        };
+    }
```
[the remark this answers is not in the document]
jsklan: this is an incorrect analysis. the SdkTargetSchema has no registry property

PR #14098, opened 2026-03-25
fix(openapi): preserve examples from schema-less response bodies

dsinghvi:
## Description

Refs https://github.com/fern-api/fern/issues/13922

When an OpenAPI spec defines response examples on a 2xx response that has **no schema** (e.g. a `POST /webhook/` 201 response with `examples` but no `schema` under `application/json`), the `ResponseBodyConverter` falls through to a synthetic empty-object fallback path. That fallback was constructing a brand-new `MediaTypeObject` with only `{ schema: { type: "object" } }`, discarding the original `examples`/`example` from the spec. This caused three downstream bugs:

1. **Empty response bodies** — response example values showed as `{}`
2. **Lost example names** — `displayName` wasn't set, so the frontend fell back to "Example N"
3. **Reversed example order** — the consolidation logic picked incorrect pairings when response examples were missing

## Changes Made

- **`ResponseBodyConverter.ts`** — In the 2xx synthetic fallback path, look up the original JSON media type object and spread its `examples` and `example` fields into the synthetic object so they flow through `parseMediaTypeObject` → example extraction → IR.
- **New assertion-based test file** (`schema-less-response-examples.test.ts`) — Verifies IR output with explicit assertions on `displayName`, response body values, and ordering instead of snapshot matching.
- **Three edge-case test fixtures:**
  - `schema-less-response-with-examples` — 201 response with `examples` but no `schema` (the primary bug)
  - `schema-less-response-no-json-content` — 2xx response with only `text/plain`, no JSON content types at all (verifies `originalJsonMediaTypeObject` is `undefined` and spread is a no-op)
  - `non-json-examples-with-json-fallback` — 201 response where `application/xml` has examples but JSON doesn't (verifies non-JSON examples don't leak into the JSON fallback)
- **Auto-generated snapshots** — The existing snapshot-based tests (`v3-sdks.test.ts`, `baseline-sdks.test.ts`) auto-discover all fixtures, so snapshots are committed alongside the assertion tests.
- **`versions.yml`** — Bumped CLI version to `4.44.1` for release.

## Testing

- [x] All 236 v3-importer tests pass (no regressions)
- [x] Assertion-based tests verify correct `displayName`, response body values, and ordering
- [x] Edge case: no JSON content types — synthetic fallback still works, spread is a no-op
- [x] Edge case: non-JSON examples — XML examples don't leak into JSON fallback response
- [x] `pnpm run check` (biome lint) passes cleanly

## Human Review Checklist

- [ ] The conditional spread `...(originalJsonMediaTypeObject?.examples != null && { examples: ... })` evaluates to `...false` (a no-op) when `originalJsonMediaTypeObject` is `undefined`. This is covered by the `schema-less-response-no-json-content` test, but worth confirming the pattern behaves as expected in all JS engines.
- [ ] The fix only looks at `jsonContentTypes` to find the original media type object. If a spec has examples *only* on a non-JSON content type (e.g. `application/xml`), they won't be carried over into the synthetic JSON fallback. This is intentional (the fallback produces a JSON response), but worth considering if that's the right behavior.
- [ ] The response type for schema-less responses is still a synthetic `{ type: "object" }` — the examples will have values that don't match this schema. Confirm downstream consumers (FDR, docs frontend) handle this gracefully.

Link to Devin session: https://app.devin.ai/sessions/373e40d189b54985801d6268377591fb
Requested by: @dsinghvi


dsinghvi:
Can you create assertions on the IR as tests and not commit the snapshot in this case? 

Can you add tests for both of the edge cases: 

```
Verify the conditional spread pattern (...(x != null && { y: x })) behaves correctly when originalJsonMediaTypeObject is undefined (no JSON content types at all) — undefined?.examples evaluates to undefined, so the spread is a no-op, but worth confirming
 Consider edge cases: what if a non-JSON content type has examples but JSON doesn't? (Current fix only looks at jsonContentTypes, which is intentional since the fallback produces a JSON respo
 ```
 
 Can you rename the fixtures to not reference customer names but the specific edge cases

dsinghvi:
Can you bump versions.yml for the Fern CLI so I can release?

PR #14116, opened 2026-03-26
feat(cli): use Fern GitHub App token for replay init

tstanmay13:
## Summary

- `fern replay init --group <name>` now automatically fetches a GitHub App installation token from Fiddle when `GITHUB_TOKEN` is not set, matching how `fern generate` works without requiring manual token setup
- Falls back gracefully: `--token` flag > `GITHUB_TOKEN` env > Fiddle installation token > actionable error message
- Updated error message to mention Fern GitHub App installation as a resolution path

**Requires**: Corresponding Fiddle backend endpoint (`POST /api/remote-gen/github/installation-token`) to be deployed first. Fiddle changes are in a separate PR.

## Test plan

- [ ] Deploy Fiddle endpoint first
- [ ] Run `fern replay init --group python` without `GITHUB_TOKEN` set, with `fern login` completed and Fern App installed on the repo — should succeed
- [ ] Run with `GITHUB_TOKEN` set — should use env var (no Fiddle call)
- [ ] Run with `--token` flag — should use flag directly
- [ ] Run without Fern App installed — should fail with actionable error message
- [ ] `pnpm compile` passes

### DIFF
--- a/packages/cli/cli/src/cli.ts
+++ b/packages/cli/cli/src/cli.ts
@@ -2,6 +2,7 @@
 import type { ReadStream, WriteStream } from "node:tty";
 import { fromBinary, toBinary } from "@bufbuild/protobuf";
 import { CodeGeneratorRequestSchema, CodeGeneratorResponseSchema } from "@bufbuild/protobuf/wkt";
+import type { FernToken } from "@fern-api/auth";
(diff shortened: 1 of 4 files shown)


fern-support:
Closing in favor of a server-side approach

PR #14120, opened 2026-03-26
fix(typescript): harden fixImportsForEsm regex and simplify file discovery

Swimburger:
## Description
Refs FER-8691

Follow-up to #13887 addressing several edge cases identified in the regex-based `fixImportsForEsm` implementation.

## Changes Made
- **Simplified file discovery**: Removed the tsconfig `include`/`exclude` parsing logic (~50 lines) and replaced it with a direct scan of the `src/` subdirectory. Generated TypeScript projects always use `include: ["src"]` in their tsconfig, so we scan `src/` directly — this avoids the complexity and fragility of parsing tsconfig (glob patterns, JSONC comments, `extends` chains/arrays, npm package references) while matching the original ts-morph scoping behavior.
- **Comment-safe matching**: Added `stripComments()` to blank out `//` and `/* */` comments (preserving character positions) before running import regexes. This prevents false matches on patterns like `// from "./old-path"`. Replacements are then applied to the original source using tracked offsets. String literals are intentionally **not** stripped — the `from`/`import` keyword requirement in the regexes is sufficient to avoid false positives from non-import strings.
- **Side-effect import support**: Updated `STATIC_IMPORT_REGEX` to also match `import "./path"` (side-effect imports with no `from` keyword), which the previous regex missed but ts-morph handled.
- **Extended file type support**: Added `.tsx`, `.mts`, `.cts`, `.jsx`, `.mjs`, `.cjs` to the file collection and processing sets, matching what ts-morph would have included via tsconfig resolution.

## Reviewer Checklist
- [ ] Verify the position-based replacement logic (`specifierStart` / `offset` tracking) is correct — this replaced the simpler `String.replace()` callback approach to support the stripped-vs-original two-pass strategy.
- [ ] Confirm the updated `STATIC_IMPORT_REGEX` (`(?:from|import)\s+`) does not accidentally match dynamic imports (`import(...)`) — the `\s+` requirement after `import` prevents matching `import(` since `(` is not whitespace.
- [ ] Verify that hardcoding `src/` as the scan directory is acceptable. This matches all current generated tsconfigs (`include: ["src"]`), but couples to that convention.
- [ ] Consider whether string literals like `const s = 'from "./utils"'` could cause false positive matches. The `from`/`import` keyword prefix makes this unlikely but not impossible in unusual generated code.

## Testing
- [x] TypeScript compilation passes (`pnpm run compile`)
- [x] Biome lint passes (`pnpm run check`)
- [x] Seed tests (`simple-api` fixture, 19/19 configs) pass with zero `.ts` file diffs
- [ ] Full seed test suite should be run to verify byte-identical output across all 196 fixtures (CI)

Link to Devin session: https://app.devin.ai/sessions/875263aabe0c4dcbbdacaabc7478ce1c
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14120" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/generators/typescript/express/cli/src/ExpressGeneratorCli.ts
+++ b/generators/typescript/express/cli/src/ExpressGeneratorCli.ts
@@ -96,7 +96,6 @@ export class ExpressGeneratorCli extends AbstractGeneratorCli<ExpressCustomConfi
         const typescriptProject = await expressGenerator.generate();
         const persistedTypescriptProject = await typescriptProject.persist();
         await expressGenerator.copyCoreUtilities({
-            pathToSrc: persistedTypescriptProject.getSrcDirectory(),
             pathToRoot: persistedTypescriptProject.getRootDirectory()
         });
 
--- a/generators/typescript/express/generator/src/ExpressGenerator.ts
+++ b/generators/typescript/express/generator/src/ExpressGenerator.ts
@@ -315,17 +315,8 @@ export class ExpressGenerator {
         });
     }
 
-    public async copyCoreUtilities({
-        pathToSrc,
-        pathToRoot
-    }: {
-        pathToSrc: AbsoluteFilePath;
-        pathToRoot: AbsoluteFilePath;
-    }): Promise<void> {
-        await this.coreUtilitiesManager.copyCoreUtilities({
-            pathToSrc,
-            pathToRoot
-        });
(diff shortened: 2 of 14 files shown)


lifanzou:
lgtm!

Swimburger:
Is this still relevant?

PR #14127, opened 2026-03-26
fix(csharp): wire auth headers and fix mock-server test generation

iamnamananand996:
## Description

Fixes a bug where the C# SDK generator accepted Basic Auth username/password constructor parameters but never set the `Authorization: Basic <base64>` header on outgoing HTTP requests. Python, Java, TypeScript, and Go generators already handled this correctly.

Additionally extends mock-server test generation to match auth headers (basic, bearer, and header auth schemes) for all authenticated endpoints, and fixes three seed test generation failures.

### Basic Auth header wiring
- `generators/csharp/sdk/src/root-client/RootClientGenerator.ts` — Detect `basic` auth scheme from IR, clone `clientOptions` into `clientOptionsWithAuth`, and emit `Authorization: Basic {base64(user:pass)}` header. Handles both unified and non-unified client option styles.
- Respects `isAuthMandatory`: when auth is optional, the header assignment is wrapped in a null guard (`if username != null && password != null`); when mandatory, the header is set unconditionally.
- `generators/csharp/sdk/versions.yml` — Add version 2.54.7 changelog entry

### Auth header matching in mock-server tests
- `generators/csharp/sdk/src/test-generation/mock-server/MockEndpointGenerator.ts` — For authenticated endpoints, emit `.WithHeader(...)` matchers for all auth schemes (basic, bearer, header). Header auth includes the scheme prefix when defined (e.g. `test_prefix X_API_KEY`).
- `packages/commons/mock-utils/index.ts` — Add `headers` field to `WireMockMapping.request` and generate auth header regex matchers (`.+`) for WireMock JSON mappings across all generators.

### Per-endpoint security filtering
- `MockEndpointGenerator.ts` and `mock-utils/index.ts` — Added `getEndpointAuthSchemeKeys()` helper that reads `endpoint.security` to determine which specific auth schemes each endpoint uses. When per-endpoint security is defined, only the relevant scheme headers are emitted; otherwise falls back to all schemes (global auth). This fixes the `endpoint-security-auth` fixture where endpoints like `getWithBearer` were incorrectly getting all auth headers (Bearer + ApiKey + Basic) instead of just Bearer.

### Bug fixes
- `generators/csharp/sdk/src/reference/buildReference.ts` — Skip endpoints with unsupported `uri`/`path` pagination types before generating reference docs. Previously, attempting to generate snippets for these endpoints threw `'uri' pagination is not supported in C#`. This fixes the `pagination-uri-path` seed fixture.
- `MockEndpointGenerator.ts` (header auth case) — Include `scheme.prefix` in the generated header value so the mock server matches the actual client request (e.g. `"test_prefix X_API_KEY"` instead of `"X_API_KEY"`). This fixes the `header-auth` seed fixture 404 error.

### Seed snapshots
- Regenerated seed snapshots for all affected `csharp-sdk` fixtures (basic-auth, bearer-token, header-auth, endpoint-security-auth, exhaustive, imdb, literal, pagination, trace, and many more)
- Added `seed/csharp-sdk/basic-auth/wire-tests/` — MockServer-based wire tests that verify the generated client sends requests with auth configured
- Updated `seed/csharp-sdk/seed.yml` — Registered `basic-auth` wire-test fixture

## Testing
- [x] Seed test snapshots regenerated — all 154/154 csharp-sdk test cases pass
- [x] `pagination-uri-path` fixture now succeeds (previously in `allowedFailures`)
- [x] `endpoint-security-auth` fixture now generates correct per-endpoint auth headers
- [x] Wire tests added for basic-auth fixture (MockServer-based)

## ⚠️ Human Review Checklist
- **Per-endpoint security key matching**: `getEndpointAuthSchemeKeys()` uses `Object.keys(securityItem)` to extract scheme keys from `endpoint.security` (typed as `Record<AuthSchemeKey, AuthScope[]>`). Verify these keys match `scheme.key` on the corresponding `AuthScheme` objects — a casing or naming mismatch would silently skip all scheme headers for that endpoint.
- **Symmetry between MockEndpointGenerator and mock-utils**: Both files now have identical `getEndpointAuthSchemeKeys()` logic. If one is updated in the future, the other must be updated too. Consider extracting to a shared utility.
- **Multiple `Authorization` headers in `any-auth`**: Fixtures using multiple auth schemes (e.g. bearer + basic) generate multiple `.WithHeader("Authorization", ...)` calls. Verify WireMock handles this correctly (last-write-wins vs. multi-value).
- **Header auth prefix matching**: The MockEndpointGenerator now emits `scheme.prefix + " " + headerValue`. Verify the generated C# client sends this exact format (see `RootClientGenerator.ts` line 265–266 where the prefix is interpolated with a space).
- **Optional auth null guard**: When `isAuthMandatory` is false, the Basic Auth header is only set if both username and password are non-null. If only one credential is provided, no `Authorization` header is sent (silent no-op). Verify this matches the desired behavior.
- **Optional auth path not seed-tested**: All three basic-auth seed fixtures have mandatory auth, so the `isAuthOptional` conditional branch is not exercised by any snapshot.
- **`pagination-uri-path` fix scope**: The filter in `buildReference.ts` skips `uri`/`path` pagination endpoints entirely from reference docs. This is consistent with `MockServerTestGenerator.shouldGenerate()` and `EndpointGenerator.generateInterfaceSignature()`, but confirm no reference doc entries are unexpectedly dropped.
- **Duplicate lookup**: `hasBasicAuth` uses `.some()` and then a second `.find()` retrieves the same scheme — functionally correct but could be a single lookup.

Link to Devin session: https://app.devin.ai/sessions/6dcbc8a751a44ba184aa9a1f4349ae38


Swimburger (approved):

Swimburger (changes requested):
I don't see any wire test changes in the C# SDK generator

Swimburger (approved):

PR #14147, opened 2026-03-26
fix(cli): resolve OAuth client-credentials from generators.yml in new OpenAPI parser

dsinghvi:
## Description

Fixes a bug where OAuth client-credentials auth schemes defined in `generators.yml` were being converted to bearer tokens instead of proper `AuthScheme.oauth` in the new OpenAPI parser (`openapi-to-ir`).

**Root cause**: `convertApiAuth` in `ir-utils` has a `// TODO: implement` stub for OAuth that falls back to `generateBearerAuth`. Unlike the legacy `ir-generator` version (which has full OAuth support with `EndpointResolver`/`PropertyResolver`), the `ir-utils` version was never implemented.

**Fix approach**: A standalone `resolveOAuthEndpointReferences()` function runs as a post-processing step in `OSSWorkspace.getIntermediateRepresentation()` **after all specs have been merged** via `mergeIntermediateRepresentation()`. This is critical because with multi-spec setups (e.g. separate `identity.yaml`, `channels.yaml`, etc.), the OAuth token endpoint may live in a different spec than the one declaring the auth scheme. The function:
1. Detects bearer auth placeholders that correspond to OAuth schemes in `authOverrides`
2. Finds the token endpoint in the merged IR by matching HTTP method + path
3. Resolves request/response property references from the endpoint's body schema (supports both `inlinedRequestBody` and `reference`-type bodies)
4. Constructs the full `AuthScheme.oauth` with `OAuthConfiguration.clientCredentials`
5. Returns a new IR (immutable — does not mutate the input)

## Changes Made
- Added standalone `resolveOAuthEndpointReferences()` in new file `packages/cli/api-importers/openapi-to-ir/src/resolveOAuthEndpointReferences.ts` (~320 lines)
- Exported function from `@fern-api/openapi-to-ir` package index
- Called from `OSSWorkspace.getIntermediateRepresentation()` after `mergeIntermediateRepresentation()` completes
- Added `oauth-client-credentials-openapi` test fixture with an OpenAPI spec + generators.yml auth override
- Added CLI `versions.yml` entry (v4.46.4)

Refactored `resolveOAuthEndpointReferences.ts` to reduce file size (~400 → ~320 lines) per review feedback:
- Replaced `parseHttpMethod` switch with `HTTP_METHOD_MAP` record lookup
- Merged `parseEndpointReference`, `findEndpointInIr`, and `findSubpackageForService` into a single `resolveEndpointFromReference` function
- Extracted shared `findObjectPropertyByWireValue` and `getObjectPropertiesForNamedType` helpers to deduplicate property lookup logic across `findRequestBodyProperty` and `findResponseProperty`
- Simplified `findRequestBodyProperty` by normalizing `inlinedRequestBody` / `reference` branches into a common property list before searching
- Combined `extractPropertyName` + default value handling into `extractWireName`
- Used object spread (`{ ...prop }`) for `RequestPropertyValue.body()` construction instead of manually listing all fields

## Testing
- [x] Built CLI locally and generated IR for the test fixture — verified output contains `_type: "oauth"` with `clientCredentials` configuration, resolved endpoint references, and correct request/response property mappings
- [x] Lint (`biome check`) passes clean
- [x] Compilation passes for `@fern-api/openapi-to-ir` and `@fern-api/lazy-fern-workspace`
- [ ] Unit tests added/updated (no automated test; test fixture added for manual/seed verification)
- [ ] Multi-spec scenario not covered by test fixture (fixture uses single `openapi.yml`; the user's scenario has 6 separate specs)

## Review Checklist
Reviewers should pay attention to:

1. **`{ ...prop }` spread in `RequestPropertyValue.body()`** — the refactored code uses object spread instead of manually listing `name`, `valueType`, `availability`, `docs`, `propertyAccess`, `v2Examples`. Verify that `ObjectProperty` fields are a superset of what `body()` accepts (extra fields are harmless, but missing fields would be a problem).
2. **Path matching in `resolveEndpointFromReference`** — matches by reconstructing `fullPath` as a string. Does not account for base paths. Could this break for APIs with non-trivial path structures?
3. **`findRequestBodyProperty` handles `inlinedRequestBody` and `reference` types** — but does not handle `fileUpload` or `bytes` body types. These are uncommon for OAuth token endpoints but worth noting.
4. **`findResponseProperty` only handles JSON responses with `named` types** — similar limitation for response property resolution.
5. **The `ir-utils` TODO stub is not removed** — other consumers of `convertApiAuth` from `ir-utils` would still get bearer for OAuth. This fix is scoped to the `OSSWorkspace` path only.
6. **`scopes` request property is hardcoded to `undefined`** — no attempt to resolve a scopes property from the request body even if configured in the schema.

Link to Devin session: https://app.devin.ai/sessions/148941f4549e49a496054f7758860483
Requested by: @dsinghvi

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14147" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/cli/api-importers/openapi-to-ir/src/resolveOAuthEndpointReferences.ts
```diff
@@ -0,0 +1,371 @@
...
+ * because the OAuth token endpoint may be defined in a different spec
+ * than the one that defines the auth scheme.
+ */
+export function resolveOAuthEndpointReferences({
+    ir,
+    authOverrides
+}: {
+    ir: IntermediateRepresentation;
+    authOverrides: RawSchemas.WithAuthSchema;
+}): void {
```
fern-support: can we have this return a new IR?

packages/cli/api-importers/openapi-to-ir/src/resolveOAuthEndpointReferences.ts
```diff
@@ -0,0 +1,401 @@
...
+} from "@fern-api/ir-sdk";
+
+/**
+ * Resolves OAuth endpoint references in the IR auth schemes.
+ *
+ * This must run AFTER all specs have been merged into a single IR,
+ * because the OAuth token endpoint may be defined in a different spec
+ * than the one that defines the auth scheme.
+ */
+export function resolveOAuthEndpointReferences({
```
dsinghvi, line 26: Are there existing utilities that you can take advantage of to reduce the number of lines in this file, if so can you bring those in?

dsinghvi: You dog

PR #14165, opened 2026-03-27
fix(docs): create tag description pages for packages with title override

jon-fern:
## Description

When a tag reference in `docs.yml` uses a `title` override (e.g., `activitiesWhatsAppMessages: title: WhatsApp Messages`), the tag description overview page was not being generated, making the section header unclickable.

## Changes Made

- In `#convertPackage`, fall back to `createTagDescriptionPageId(subpackage, pkg.package)` when no explicit `overviewAbsolutePath` is provided by the user
- Renamed `overviewPageId` → `explicitOverviewPageId` to clarify the distinction between a user-provided overview and a tag-description-derived one
- Added `packageLocator` parameter to `createTagDescriptionPageId` as a fallback key for `openApiTags` lookup — handles nested tags where `subpackage.name` (leaf) differs from the full camelCase key (e.g., `whatsAppMessages` vs `activitiesWhatsAppMessages`)
- Also pass `unknownIdentifier` as locator in `#convertUnknownIdentifier` for consistency
- Added `versions.yml` entry (v4.46.5)

## Root Cause

Tag references with a `title` property are parsed as `Package` type items (not `Item` type). The `#convertPackage` method never called `createTagDescriptionPageId`, so tag description overview pages were silently dropped. Plain string references (without `title`) went through `#convertUnknownIdentifier`, which correctly called `createTagDescriptionPageId`.

Additionally, for nested tags like `"Activities / WhatsApp Messages"`, `createTagDescriptionPageId` only checked `subpackage.name` against `openApiTags`. The `openApiTags` dictionary is keyed by `camelCase(tagId)` (e.g., `activitiesWhatsAppMessages`), which may not match the leaf subpackage name alone. The new `packageLocator` fallback addresses this.

## Human Review Checklist

- [ ] Verify the `else` branch (subpackage not found, lines 366–384) correctly uses `explicitOverviewPageId` — tag description pages can't be created without a resolved subpackage, so no fallback is needed there
- [ ] Confirm `createTagDescriptionPageId` side effects (mutating `#tagDescriptionContent`, `markdownFilesToNoIndex`, `markdownFilesToTags`) are safe in the `#convertPackage` context — they mirror how `#convertUnknownIdentifier` already calls this method
- [ ] Check that `camelCase(packageLocator)` in the fallback can't accidentally match an unrelated tag (low risk since it only fires when the direct `subpackageName` lookup misses)
- [ ] No new unit test was added for the nested-tag + title-override scenario; consider whether one should be added

## Testing

- [x] Existing unit tests pass (44 tests in docs-resolver)
- [x] `biome format` / `biome check` passes
- [x] Verified locally with `fern docs dev` against the [fern-close-forked](https://github.com/fern-support/fern-close-forked) repo — WhatsApp Messages tag description page appears and is clickable
- [ ] No new test added for the specific bug scenario

Link to Devin session: https://app.devin.ai/sessions/6bae235b53574bfb96243189c7541628
Requested by: @jon-fern


packages/cli/docs-resolver/src/ApiReferenceNodeConverter.ts
```diff
@@ -194,8 +195,12 @@ export class ApiReferenceNodeConverter {
...
         }
 
-        // Check if this subpackage corresponds to a tag with description
-        const tagInfo = this.openApiTags[subpackageName];
+        // Try matching by subpackage name first, then by the full package locator (camelCased).
+        // For nested tags like "Activities / WhatsApp Messages", the subpackage name is the leaf
+        // ("whatsAppMessages") but the openApiTags key is the full camelCase ("activitiesWhatsAppMessages").
+        const tagInfo =
+            this.openApiTags[subpackageName] ??
+            (packageLocator != null ? this.openApiTags[camelCase(packageLocator)] : undefined);
```
dsinghvi, line 203: is it fair to say that this `(packageLocator != null ? this.openApiTags[camelCase(packageLocator)] : undefined);` was the key thing?

dsinghvi: Does this mean a tag description page id is always added? what if that isn't enabled in the docs.yml then are we still going to create it?

dsinghvi (approved):

PR #14210, opened 2026-03-27
fix(typescript): send empty JSON body instead of undefined for optional request bodies

jsklan:
## Description

Fixes a bug where the TypeScript SDK generator produces `body: request` for endpoints with optional `unknown` request bodies. When the caller omits the body argument, `request` is `undefined` and no HTTP body is sent, but the server expects an empty JSON object `{}`. This causes wire test failures with: `"Request body is empty, expected a JSON object"`.

### Why `?? {}` instead of omitting `Content-Type`?

Per RFC 9110 (HTTP Semantics), a POST with no body is technically valid HTTP. The alternative fix would be to omit the `Content-Type: application/json` header when the body is undefined. However, `?? {}` is the correct approach because:

- **Content-Type contract**: When the SDK sets `Content-Type: application/json`, it makes a contract that the body is valid JSON. Sending `undefined`/no body while claiming JSON content type violates this contract — the header and body must agree.
- **Server compatibility**: Most API servers with a POST endpoint that *accepts* a JSON body will attempt to parse the body as JSON regardless. An empty/missing body with `Content-Type: application/json` causes parse failures (400 errors) in most frameworks.
- **`{}` is the correct "no data" in JSON**: An empty object is the natural representation of "no properties provided" in JSON, just as omitting the body would be for a truly bodyless request.
- **Less invasive**: Conditionally omitting `Content-Type` would require changes deeper in the fetcher pipeline and risks breaking servers that always expect the header on POST endpoints.

## Changes Made

- **Generator fix** (`GeneratedDefaultEndpointRequest.ts`): When `requestParameter.isOptional()`, the serialized body is wrapped with `?? {}` so that `body: request ?? {}` is emitted instead of `body: request`
- **Exhaustive fixture**: Added `postOptionalBody` endpoint to test definitions with `request: unknown` and an `EmptyBody` example, updated seed output across all 22 exhaustive fixture configurations
- **Changelog**: Added version `3.59.12` entry to `generators/typescript/sdk/versions.yml`

## Testing
- [x] `pnpm seed:local test --generator ts-sdk --fixture exhaustive` — all 22 configurations pass
- [x] `pnpm check:fix` — no lint issues
- [x] `pnpm test:update` — no snapshot changes needed

## ⚠️ Human Review Checklist
- [ ] **Scope of `?? {}`**: The fix applies to _all_ endpoints where `requestParameter.isOptional()` is true, not just `unknown` bodies. Verify this is correct for other optional body types (e.g., optional inlined request bodies). The code path is only reached for JSON request types, so this should be safe.
- [ ] **Fixture definition** (`test-definitions/.../endpoints/object.yml`): Confirm `request: unknown` with an empty example accurately reproduces the customer scenario
- [ ] The bulk of this diff is auto-generated seed output — only the fixture YAML, `GeneratedDefaultEndpointRequest.ts`, and `versions.yml` need close review

Link to Devin session: https://app.devin.ai/sessions/23449f05dbde4bffa573d88f2d6491c2
Requested by: @jsklan

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14210" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


jsklan:
Closing this PR in favor of a revised approach. The `?? {}` fix works for the original `unknown`-typed body case, but as Devin's review correctly flagged, it applies to **all** optional request bodies — sending `{}` for a typed optional body like `optional<ObjectWithRequiredField>` would trigger server-side validation errors (missing required fields).

After investigating how major server frameworks handle a POST with **no body and no Content-Type header** for optional body endpoints:

| Framework | Behavior | Result |
|-----------|----------|--------|
| Express.js | `body-parser` skips parsing when Content-Type doesn't match | 200 |
| FastAPI | Optional param (`= None`) receives `None` | 200 |
| ASP.NET Core 7+ | Nullable param implies `EmptyBodyBehavior.Allow` | 200 |
| Play Framework | No `Content-Length`/`Transfer-Encoding` → no parse attempted | 200 |
| React Router/Remix | No Content-Type → no JSON parse triggered | 200 |

The correct fix is to **omit both the body and the `Content-Type` header** when the optional body is not provided, rather than sending `{}`. This is semantically correct ("I have no body to send, so I won't claim I'm sending JSON") and works across all major server frameworks on modern versions.

A new PR with this approach is incoming.

PR #14270, opened 2026-03-30
feat(go): SSE parity; context cancellation, metadata, comments, terminators, testing

patrickthornton:
## Description
Brings Go SSE up to parity with other languages (TS/Python/Java). Namely:
- `core.NewStream` now accepts a `context.Context` argument to support context cancellation, so blocking `Recv()` calls will unblock when the context is cancelled or times out. This is an internal breaking change; it shouldn't affect the user since `streamer.go`'s `Stream` method already asked for a `context.Context` argument.
- The new `RecvEvent()` and `RecvEventRaw()` methods return SSE metadata (event, id, retry) alongside the data payload; the raw version returns it as bytes.
- Fully parse `id` and `retry`. The new `LastEventID()` method returns the most recent event ID (not currently useful; will be useful when implementing auto-reconnection).
- Fix support for SSE comment-lines like `: heartbeat`.
- Handle all terminators allowed for by the SSE spec: LF, CR, CRLF.
- Go SDKs with streaming endpoints now actually generate `stream_test.go` so users can unit test functionality. Also expanded `stream_test.go` to test new functionality.

Summary of expanded public interface:
```go
  // Existing (unchanged)
  stream.Recv() (T, error)
  stream.RecvRaw() ([]byte, error)
  stream.Close() error

  // New
  stream.RecvEvent() (StreamEvent[T], error)
  stream.RecvEventRaw() (StreamEventRaw, error)
  stream.LastEventID() string

  type SseEventMeta struct {
      ID    string  // SSE id: field
      Event string  // SSE event: field
      Retry int     // SSE retry: field (milliseconds, 0 if absent)
  }

  type StreamEvent[T any] struct {
      SseEventMeta
      Data T
  }

  type StreamEventRaw struct {
      SseEventMeta
      Data []byte
  }

  All three Recv variants can be interleaved freely on the same stream — LastEventID() tracks the most recent ID
   regardless of which method is used.
```

## Changes Made
Entirely to as-is files `stream.go`, `stream_test.go`, and `streamer.go` in go-v2.

## Testing
New tests in `stream_test.go`; the streaming fixtures pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,9 @@ seed/python-sdk/**/output.prof
 # Claude
 **/settings.local.json
 
+# Superpowers
+docs/superpowers/
+
 # Preview
 **/.preview/
 
--- a/generators/go-v2/base/src/AsIs.ts
+++ b/generators/go-v2/base/src/AsIs.ts
@@ -17,6 +17,7 @@ export enum AsIsFiles {
     PagerTest = "internal/pager_test.go_",
     CustomPagination = "core/custom_pagination.go_",
     Stream = "core/stream.go_",
+    StreamTest = "core/stream_test.go_",
     Streamer = "internal/streamer.go_",
     TokenProvider = "core/token_provider.go_"
 }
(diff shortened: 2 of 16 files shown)


patrickthornton:
@claude review once

jsklan (approved):

PR #14272, opened 2026-03-30
feat(cli): auto-infer discriminator context for SSE event spec shapes

dsinghvi:
## Description

When `x-fern-discriminator-context` is not explicitly set on a discriminated union, the system now auto-infers the context as `"protocol"` if every variant matches the SSE event spec shape (properties exclusively from `{event, data, id, retry}` with correct types, and the `event` field is required). Previously this always defaulted to `"data"`.

Inference is scoped to the **legacy OpenAPI importer** only (per review feedback — V3 importer and Fern definition paths were removed to be addressed separately).

## Changes Made

- **Legacy OpenAPI importer** (`convertDiscriminatedOneOf.ts`, new `inferDiscriminatorContext.ts`): Replaced hardcoded `"data"` fallback with `inferDiscriminatorContext()` / `inferDiscriminatorContextFromVariants()` calls. The SSE spec matching logic (property collection via allOf, type checking, `event` field requirement) is self-contained in `inferDiscriminatorContext.ts`.
- **Test fixture + unit test**: New OpenAPI fixture at `test-definitions-openapi/fern/apis/discriminator-context/` with three discriminated unions covering SSE-shaped → protocol, non-SSE → data, and explicit override → data. Programmatic assertions in `generateIntermediateRepresentation.test.ts`.
- Updated `versions.yml` with changelog entry (v4.50.0)

### Inference Rule
Infer `"protocol"` when **all** variants are objects whose properties (including allOf-inherited) are a subset of:

- `event` — `string` (required on every variant)
- `data` — any type
- `id` — `string`
- `retry` — `integer`

Explicit `x-fern-discriminator-context` always takes precedence. Variants with no properties, unresolvable `$ref`, or missing `event` field return `"data"`.

- **Inlined SSE spec matching** — removed the shared `sseSpecShape.ts` from `@fern-api/core-utils` and inlined the logic directly into `inferDiscriminatorContext.ts` since only the legacy OpenAPI importer consumes it (per @dsinghvi feedback). Removed corresponding exports from `core-utils/src/index.ts`.
- **Changelog updated** — `versions.yml` v4.50.0 summary now correctly references only the legacy OpenAPI importer.
- **Merged latest main** into branch.

## Human Review Checklist
- [x] Test fixture and unit test added covering SSE → protocol, non-SSE → data, and explicit override → data
- [x] Changelog entry (`versions.yml` v4.50.0) correctly scoped to legacy OpenAPI importer only
- [ ] `catch {}` in `inferDiscriminatorContext` / `inferDiscriminatorContextFromVariants` silently falls back to `"data"` on unresolvable `$ref` — verify this is acceptable vs. logging a warning
- [ ] `collectAllProperties` recursively resolves `$ref` and `allOf` — confirm no risk of infinite loops on circular schemas (the `catch {}` would contain a stack overflow, but worth verifying)

## Testing
- [x] Unit tests added (discriminator context inference fixture + programmatic assertions)
- [x] Lint (`biome check`) passes
- [x] Existing snapshot tests pass (329 tests, including new `discriminator-context` fixture)
- [x] CI green (295 checks pass)

Link to Devin session: https://app.devin.ai/sessions/10b06a8238d04a5385a799224445b7e1


dsinghvi:
Can you add unit tests and a fixture somewhere to test this, please assert the value is defaulted properly in the IR generated from the fixture

packages/cli/api-importers/openapi/openapi-ir-parser/src/schema/inferDiscriminatorContext.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    for (const schema of Object.values(mapping)) {
+        try {
+            const resolved = context.resolveSchemaReference({ $ref: schema });
+            if (!variantMatchesSseSpec(resolved, context)) {
+                return "data";
+            }
+        } catch {
+            return "data";
```
aditya-arolkar-swe, line 120: I guess if the reference resolution errors, we default back to data?

packages/cli/generation/ir-generator/src/converters/type-declarations/convertDiscriminatedUnionTypeDeclaration.ts
```diff
@@ -236,8 +236,159 @@ function unwrapNullableAndOptional(resolvedType: ResolvedType): ResolvedType {
...
+        return UnionDiscriminatorContext.Data;
+    }
+
+    // Infer from variant shapes
+    if (allVariantsMatchSseSpec({ union, file, typeResolver })) {
+        return UnionDiscriminatorContext.Protocol;
+    }
 
-    return discriminantContext === "protocol" ? UnionDiscriminatorContext.Protocol : UnionDiscriminatorContext.Data;
+    return UnionDiscriminatorContext.Data;
```
aditya-arolkar-swe: Why are we repeating the same SSE discriminator logic across 3 packages? Is it even needed in all these places if it's only for SDK generation?

dsinghvi: are there places we can share?

dsinghvi:
Can you remove any additions to the V3 importer for now, ill come back to address this for docs

packages/cli/generation/ir-generator/src/converters/type-declarations/convertDiscriminatedUnionTypeDeclaration.ts
dsinghvi: Let's remove this from the fern definition as well `ir-generator` b/c we don't care as much about fern definition users anymore

packages/commons/core-utils/src/sseSpecShape.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Returns true only when every variant passes `variantMatchesSseSpecShape`.
+ * Returns false for empty variant lists.
+ */
+export function allVariantsMatchSseSpecShape(variants: NormalizedVariantProperty[][]): boolean {
+    if (variants.length === 0) {
+        return false;
+    }
+
+    return variants.every(variantMatchesSseSpecShape);
+}
```
dsinghvi: what is using this file? if nothing can we delete

dsinghvi: Yeah lets do that

dsinghvi:
Can you pull latest from main and make sure the changelog is accurate

PR #14280, opened 2026-03-30
fix(typescript): support alphabetical body param sorting in webhook signature verification

iamnamananand996:
## Description

Linear ticket: Refs FER-8948

The TypeScript SDK generator's `WebhooksHelperGenerator` now handles the `bodySort: ALPHABETICAL` field on `WebhookPayloadFormat` in the IR. When a provider's webhook signing algorithm requires POST body parameters to be sorted alphabetically by key before computing the HMAC signature, the generated `verifySignature` method now supports this.

## Changes Made

- **`WebhooksHelperGenerator.ts`**: 
  - `buildHmacParameters`: When `bodySort` is set, the `requestBody` parameter type changes from `string` to `string | Record<string, string>`
  - `addPayloadConstruction`: When `bodySort` is set, generates a runtime `typeof` check that either uses the raw string or sorts Record keys alphabetically and concatenates `key + value` pairs
  - `buildJsDoc`: Adds documentation about the dual-type `requestBody` parameter when `bodySort` is present
- **`WebhooksHelperGenerator.test.ts`**: 3 new test cases (BODY-only with bodySort, NOTIFICATION_URL + BODY with bodySort, JSDoc generation)
- **`versions.yml`**: Added changelog entry for v3.61.0-rc.1 (`fix`, irVersion 66)

## Review Checklist

- [ ] **Sorting expression correctness**: The generated code is `Object.keys(requestBody).sort().map(key => key + requestBody[key]).join("")`. Params sorted alphabetically by key, each key+value concatenated (no separator between pairs). Verify this matches the expected signing spec.
- [ ] **Backwards compatibility**: When `bodySort` is `undefined` (all existing webhook configs), the generated code is byte-for-byte identical to before — no type change, no sorting logic. Existing snapshots are unchanged.
- [ ] **`bodySort` property access**: Accessed directly from `config.payloadFormat.bodySort` using the native IR SDK type — no local type extensions or casts.

## Testing

- [x] Unit tests added: 3 new tests covering BODY-only payload with bodySort, NOTIFICATION_URL + BODY payload, and JSDoc generation
- [x] Snapshot tests updated (2 new snapshots)
- [x] All 62 CI checks pass (compile, test, lint, biome, depcheck, seed tests, etc.)
- [ ] No e2e testing — changes validated via unit/snapshot tests only

Link to Devin session: https://app.devin.ai/sessions/e1333475cf904a6e98026ecfaa106ad9


generators/typescript/sdk/generator/src/__test__/__snapshots__/WebhooksHelperGenerator.test.ts.snap
```diff
@@ -44,6 +44,60 @@ export class WebhooksHelper {
...
+    }
+}
+"
+`;
+
+exports[`WebhooksHelperGenerator > HMAC verification > writes HMAC class with ALPHABETICAL bodySort and NOTIFICATION_URL + BODY payload (Twilio pattern) 1`] = `
+"/**
+ * Verify an HMAC webhook signature.
+ *
+ * Extract the signature from the "X-Twilio-Signature" header and pass it as the signatureHeader parameter.
```
fern-support: Avoid mentioning our customers in source code, like Twilio

fern-support (approved):

PR #14290, opened 2026-03-30
fix(cli): preserve null as valid example value in anyOf/oneOf validation

fern-support:
## Description
Linear ticket: Closes 

Fixes false validation errors for OpenAPI 3.1 specs using `anyOf: [{type: string}, {type: null}]` with `null` example values (e.g. Close customer's `POST /saved_search/` with `query: null`).

Two separate bugs were causing null examples to be rejected:

1. **`convertUnion()`**: The nullish coalescing operator (`??`) treats `null` as nullish, so explicit `null` examples were replaced by fallback values instead of being passed through to variant converters.
2. **`convertObject()`**: The `propertyIsOmittedFromExample` logic only recognized legacy `nullable: true` for null-allowing properties. It did not handle OpenAPI 3.1's `anyOf`/`oneOf` with `{type: "null"}` variants, so `null` property values were treated as omitted.

### Bug fixes
- Changed `??` to `!== undefined` checks in `convertUnion()` for both `containerExample` and `variantExample` so `null` is preserved as a valid example
- Added null-aware logic in `convertObject()` that detects null-allowing properties via `nullable: true`, `anyOf` with null type, or `oneOf` with null type
- Updated `propertyIsOmittedFromExample` to correctly distinguish between `null` (valid value) and `undefined` (omitted) when the schema allows null

### Refactoring: named type guards
- Extracted `isInlineSchema(schema): schema is SchemaObject` — type guard that returns true when a schema is an inline `SchemaObject` (not a `$ref`)
- Extracted `schemaAllowsNull(schema): boolean` — checks whether a schema explicitly allows null via `nullable: true`, `anyOf` with `{type: "null"}`, or `oneOf` with `{type: "null"}`
- Replaced inline `in` checks in `convertObject()` with calls to these named functions for readability
- Uses a safe `"nullable" in schema` guard inside `schemaAllowsNull` for OpenAPI 3.0 compat (`nullable` was removed from 3.1 type defs but still appears in real specs)

### Documentation
- Added a "prefer named type guard functions over inline `in` checks" rule to `CLAUDE.md`
- Updated existing type guard rule to emphasize custom type guards over inline narrowing

### Changelog
- Added `versions.yml` entry for v4.50.5 describing the fix

## Testing
- [x] Unit tests added/updated — 2 new tests in `ExampleConverter.test.ts` (null anyOf standalone + null in object property)
- [x] Integration test added — `null-anyof-repro.test.ts` reproduces exact Close customer scenario with real `OpenAPIConverterContext3_1`
- [x] Manual testing completed — verified against Close customer repo locally (`FERN_NO_VERSION_REDIRECTION=true`), `POST /saved_search/` warning is resolved, all 225 examples validate as valid (was 182 valid / 43 invalid before)
- All 42 tests pass, lint clean

## Human Review Checklist
- [ ] `schemaAllowsNull` only inspects non-`$ref` schemas in `anyOf`/`oneOf`. A `{type: "null"}` behind a `$ref` would not be detected (unlikely in practice but worth awareness).
- [ ] Confirm the `!== undefined` change in `convertUnion()` doesn't break cases where `this.example` is intentionally `undefined` (the fallback to `maybeResolveSchemaExample` still fires for `undefined`).
- [ ] The `"nullable" in schema` check inside `schemaAllowsNull` is an inline `in` check, but it's the implementation detail of a well-named function (not scattered business logic). Verify this is acceptable.
- [ ] Changelog version 4.50.5 — confirm no version conflict with other in-flight PRs.

Link to Devin session: https://app.devin.ai/sessions/97ad0be6d4fb478194c643aca3f5d397
Requested by: @fern-support


dsinghvi:
Can you update branch and fix changelog as well

PR #14355, opened 2026-03-31
feat(openapi): add x-fern-default extension for path, header, and query parameters

Swimburger:
## Description

Adds a new `x-fern-default` OpenAPI extension that allows specifying client-side default values for path, header, and query parameters. When present, generators can emit optional parameters that automatically send the default value when the caller omits them.

```yaml
parameters:
  - name: region
    in: path
    required: true
    schema:
      type: string
    x-fern-default: "us-east-1"
  - name: X-API-Version
    in: header
    schema:
      type: string
    x-fern-default: "2024-02-08"
```

This addresses the use case where an SDK should pin a version header or region path parameter by default but still allow callers to override it.

## Changes Made

- Register `x-fern-default` in `FernOpenAPIExtension` (`fernExtensions.ts`)
- Add `clientDefault` to OpenAPI IR types (`PathParameterWithExample`, `HeaderWithExample`, `QueryParameterWithExample`, `PathParameter`, `Header`, `QueryParameter`) in both Fern definitions and TS SDK types
- Extract `x-fern-default` from parameters in `convertParameters.ts`
- Wire `clientDefault` through example stripping in `generateIr.ts` for all three parameter types
- Map `clientDefault` → `default` in Fern definition output (`buildPathParameter.ts`, `buildHeader.ts`, `buildQueryParameter.ts`)
- Add `clientDefault: optional<types.Literal>` to Fern IR `PathParameter`, `HttpHeader`, and `QueryParameter` in `http.yml`
- Extract shared `convertDefaultToLiteral` utility (`convertDefaultToLiteral.ts`) used by both `convertHttpService.ts` (path params + headers) and `convertQueryParameter.ts` — converts Fern definition `default` to IR `Literal` (supports string and boolean; silently skips other types like numbers)
- Add `clientDefault: undefined` to all construction sites: AsyncAPI v2/v3, async-sync-operation, idempotency header, missing-path-parameter, response body/error converters, global headers extension, operation converters, abstract parameter converter, and channel converter
- IR SDK types regenerated (`pnpm ir:generate`) with serialization support for `clientDefault` on all three parameter types
- IR version bumped to 65.7.0
- Seed test fixture (`x-fern-default/`) with path parameter, string header, and string query parameter defaults
- Changelog entry (v4.54.0)
- All snapshot tests updated across packages: `openapi-from-flag`, `openapi-ir-to-fern-tests`, `v3-importer-tests`, and `ete-tests` (ir, writeDefinition, dependencies)

- **Addressed PR review comments:**
  - Fixed `fern.config.json` version mismatch (was `4.52.1`, now matches `versions.yml` at `4.54.0`)
  - Updated `fernExtensions.ts` doc comment to include "path" parameters (was "header or query")
  - Extracted duplicate `convertDefaultToLiteral` into a shared utility file (`convertDefaultToLiteral.ts`), imported by both `convertHttpService.ts` and `convertQueryParameter.ts`
  - Removed misleading `console.warn` for non-string/non-boolean defaults — native OpenAPI schema defaults (e.g., `default: 42` on a number param) were triggering a warning about "x-fern-default" even though the user never used that extension. Now silently returns `undefined` for unsupported types, since those values are already captured in the type's `PrimitiveTypeV2.default` metadata.

## Human Review Checklist

- [x] ~~**Redundant `clientDefault` on query params in `convertParameters.ts`**~~ — Fixed previously.
- [x] ~~**Silent drop for unsupported types**~~ — Fixed: `convertDefaultToLiteral` silently skips non-string/non-boolean types (correct behavior since native schema defaults already flow to `PrimitiveTypeV2.default`).
- [x] ~~**Schema defaults silently dropped**~~ — Fixed: restored `else if (queryParameterDefault != null)` fallback in `buildQueryParameter.ts`.
- [x] ~~**Duplicated `convertDefaultToLiteral` helper**~~ — Extracted to shared `convertDefaultToLiteral.ts`.
- [ ] **Native defaults also populate `clientDefault`**: Because the Fern definition only has one `default` field, native OpenAPI schema defaults (not just `x-fern-default`) flow through to `clientDefault` in the IR via `convertQueryParameter.ts`. This means generators consuming `clientDefault` will auto-send native defaults too. Verify this is acceptable, or whether a separate Fern definition field is needed to distinguish the two sources.
- [ ] **Boolean defaults on headers untested in seed tests**: The boolean header case was removed from the test fixture because the Python generator produces a mypy error (assigns `bool` to a `str` header field). The IR-level support is there, but no generator currently handles this correctly. May need a follow-up in generators.
- [ ] **Global headers not covered**: `buildGlobalHeaders.ts` is not modified — `x-fern-global-headers` entries won't get `clientDefault` support. The original Polytomic use case uses global headers, so this may need a follow-up.
- [ ] **Hand-edited SDK types**: The TS types under `openapi-ir/src/sdk/` were manually edited to add `clientDefault` for all three parameter types (path, header, query). Verify these stay in sync if the package is regenerated.
- [ ] **Path parameter default guard**: `convertPathParameter` uses `"type" in parameter` to distinguish `TypeReferenceDetailedSchema` from plain strings and `VariableReferenceSchema`. Variable-reference path params cannot have defaults — verify this is the desired behavior.

## Testing

- [x] Seed test fixture created (`x-fern-default/openapi.yml` + `generators.yml`) covering path parameter, string header, and string query parameter defaults
- [x] All seed snapshot tests pass (21/21 generators)
- [x] `openapi-from-flag` snapshots updated (12 snapshots)
- [x] `openapi-ir-to-fern-tests` snapshots updated (5 snapshots)
- [x] `v3-importer-tests` snapshots updated (1 snapshot)
- [x] `ete-tests` snapshots updated (ir, writeDefinition, dependencies)
- [x] IR generator compiles with regenerated types
- [x] CI green (306/306 checks passing)

Link to Devin session: https://app.devin.ai/sessions/dae61f87717a46d781e579e47b4758e5
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14355" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/cli/generation/ir-generator/src/converters/services/convertHttpService.ts
```diff
@@ -445,13 +449,27 @@ export function convertHttpHeader({
...
         env: typeof header === "string" ? undefined : header.env,
+        clientDefault: convertDefaultToLiteral(defaultValue),
         v2Examples: {
             userSpecifiedExamples: {},
             autogeneratedExamples: {}
         }
     };
 }
 
+function convertDefaultToLiteral(defaultValue: unknown): Literal | undefined {
```
Swimburger: we should log warnings when it can't be converted

packages/cli/generation/ir-generator/src/converters/services/convertQueryParameter.ts
```diff
@@ -35,6 +37,19 @@ export function convertQueryParameter({
     };
 }
 
+function convertDefaultToLiteral(defaultValue: unknown): Literal | undefined {
```
Swimburger: log warnings when failed to convert to literal.

fern-support (approved):

packages/cli/api-importers/v3-importer-tests/src/__test__/__snapshots__/baseline-sdks/defaults.json
```diff
@@ -332,7 +332,6 @@
                     "primitive": {
                       "v1": "BOOLEAN",
                       "v2": {
-                        "default": true,
```
fern-support, line 335: why are losing these `default` properties all over snapshots?

fern-support (changes requested):

fern-support (approved):

PR #14406, opened 2026-03-31
feat(typescript): support omitting username/password in basic auth when configured in IR

Swimburger:
## Description

Refs #14378

Adds conditional support for omitting username/password in basic auth for the TypeScript SDK generator. When `usernameOmit` or `passwordOmit` flags are set in the IR's `BasicAuthScheme`, the corresponding field is **completely removed from the generated SDK's `AuthOptions` type** — not just made nullable. Internally, the omitted field is treated as an empty string when encoding the auth header (`base64(username:password)` with the colon always present).

Default behavior (both required) is preserved when no omit flags are set.

### Generator (`BasicAuthProviderGenerator.ts`)
- `getAuthOptionsProperties`: conditionally **excludes** each field from the `AuthOptions` type entirely when its corresponding omit flag is set (field is removed from the end-user API)
- `generateCanCreateStatements`: per-field checks — omitted fields always satisfy `canCreate` (return `true`), required fields must be present
- `generateGetAuthRequestStatements`: refactored into `buildNullChecks` helper — omitted fields use `""` directly instead of reading from options; null checks are only generated for non-omitted fields; declarations and null checks are interleaved to preserve short-circuit evaluation order

### Core utility (`BasicAuth.ts`)
- Interface fields widened to `username?: string; password?: string` **unconditionally**
- `toAuthorizationHeader` uses `?? ""` fallbacks and returns `undefined` when both are empty strings

### Tests & fixtures
- New `basic-auth-pw-omitted` test fixture with `password: omit: true`
- Full seed output for `seed/ts-sdk/basic-auth-pw-omitted/`
- Updated `seed/ts-sdk/basic-auth/` output (reflects widened `BasicAuth.ts` interface)
- Unit tests added for username-only, password-only, neither, and both-empty cases

### Changelog
- `versions.yml`: new 3.62.2 entry (`irVersion: 66`)

- **Fixture renamed**: `basic-auth-optional` → `basic-auth-pw-omitted` across all directories, filenames, and file contents (kebab-case, camelCase, PascalCase variants). The name now accurately reflects what the fixture tests: password omission, not generic "optional" behavior.
- **Terminology fix**: Updated changelog, comments, and descriptions to use "omit" (field removed from SDK API) instead of "optional" (nullable/undefined). "Optional" in the codebase now only refers to TypeScript type optionality (`?` markers for non-mandatory auth).
- **Fixed short-circuit evaluation order**: `buildNullChecks` now interleaves declaration and null-check for each field, preserving the original short-circuit behavior where the password supplier is never evaluated if the username is null.
- **Resolved multiple merge conflicts** in `versions.yml` (bumped through 3.61.2 → 3.62.2 after main merged Error.cause, test import paths, and other features).

## Testing
- [x] Unit tests added (`BasicAuth.test.ts`: 5 cases covering all credential combinations)
- [x] Seed snapshot generated for `basic-auth-pw-omitted` fixture
- [x] Existing `basic-auth` seed output updated and verified
- [x] All required CI checks pass

## ⚠️ Human Review Checklist

1. **`BasicAuth.ts` interface is widened unconditionally**: Every generated TS SDK (not just those with omit flags) will now have `username?: string; password?: string` in the core utility. The generator controls what appears in `AuthOptions`, so the runtime behavior is guarded — but this does widen the type contract for downstream consumers who import `BasicAuth` directly.
2. **Behavioral change for empty strings**: `toAuthorizationHeader({username: "", password: ""})` now returns `undefined` instead of `Basic Og==`. This affects the `authHeader != null` guard in generated code.
3. **Both-fields-omitted path**: When both `usernameOmit` and `passwordOmit` are true, `canCreate()` always returns `true` and `getAuthRequest` hardcodes both to `""`, producing `Basic Og==` — but `toAuthorizationHeader` then returns `undefined` since both are empty. Verify this no-op edge case is acceptable.
4. **Snapshot formatting change**: The `AuthProviders.test.ts.snap` diff shows null-check blocks collapsed to single lines and `AuthOptions` type properties changed from computed names (`[USERNAME_PARAM]?:`) to literal names (`username?:`). Verify this is acceptable generated output.
5. **`seed/ts-sdk/basic-auth/` diff**: The existing (non-omit) fixture output changed due to the unconditional `BasicAuth.ts` widening. Confirm the generated client behavior is still correct for the required-fields case.
6. **versions.yml**: Bumped to 3.62.2, `irVersion: 66`. Verify version ordering is correct relative to main.

Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14406" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


seed/ts-sdk/basic-auth/src/auth/BasicAuthProvider.ts
```diff
@@ -56,7 +51,7 @@ export namespace BasicAuthProvider {
     export const AUTH_CONFIG_ERROR_MESSAGE_PASSWORD: string =
         `Please provide '${PASSWORD_PARAM}' when initializing the client` as const;
     export type Options = AuthOptions;
-    export type AuthOptions = { [USERNAME_PARAM]: core.Supplier<string>; [PASSWORD_PARAM]: core.Supplier<string> };
+    export type AuthOptions = { username: core.Supplier<string>; password: core.Supplier<string> };
```
Swimburger: we should continue using the [USERNAME_PARAM] const and pwd

fern-support (approved):

PR #14409, opened 2026-03-31
feat(csharp): remove omitted basic auth fields from SDK API, use empty string internally

Swimburger:
## Description

Refs #14378

When `usernameOmit` or `passwordOmit` flags are set in the IR's `BasicAuthScheme`, the generated C# SDK now **completely removes** the flagged field from the constructor and client options — it does not appear in the end-user API at all. Internally, the omitted field is treated as an empty string when encoding the `Authorization: Basic` header (e.g. `password: omit: true` → header encodes `username:`). Default behavior (both required) is preserved when no omit flags are set.

## Changes Made

- **`RootClientGenerator.ts`**:
  - Reads `basicScheme.usernameOmit` / `basicScheme.passwordOmit` directly (no cast needed — `@fern-fern/ir-sdk` v66 has typed fields)
  - **Constructor parameter removal**: omitted fields are excluded entirely from the `params` array (not just made optional). Only non-omitted fields appear in the generated constructor.
  - **Auth header encoding**: omitted fields produce clean empty strings in the interpolation (e.g. `$"{username}:"` when password is omitted) — no `{""}` wrapper or null-coalescing fallback.
  - Per-field condition logic: only non-omitted fields contribute to the null-check guard (e.g. when only `passwordOmit` is set, condition is `username != null`)
  - **if/else if control flow fix**: replaced `i === 0` index check with `isFirstBlock` tracking variable, so that when a both-omitted scheme is skipped via `continue`, the next scheme still correctly emits `if` instead of `else if`
  - **Both-fields-omitted case**: when both `usernameOmit` and `passwordOmit` are true, the scheme is skipped entirely (`continue`) — no meaningless `Authorization: Basic Og==` header is sent
- **`EndpointSnippetGenerator.ts` (dynamic snippets)**: Omitted fields are excluded from generated snippet constructor args. Uses `as unknown as Record<string, unknown>` cast (necessary — `@fern-api/dynamic-ir-sdk` lacks typed `usernameOmit`/`passwordOmit` fields)
- **`DynamicSnippetsConverter.ts`**: Passes `usernameOmit`/`passwordOmit` through to dynamic IR so downstream snippet generators can read the flags
- **`versions.yml`**: new 2.58.0 entry (`irVersion: 66`)
- **`snippet.json`**: Updated to reflect password omission — constructor calls now use `new SeedBasicAuthPwOmittedClient("USERNAME")` (single arg)
- **New `basic-auth-pw-omitted` test fixture** with `password: omit: true`, plus seed output at `seed/csharp-sdk/basic-auth-pw-omitted/`

- **Seed output fully regenerated**: `SeedBasicAuthPwOmittedClient.cs` now correctly removes the `password` parameter from the constructor and encodes the auth header as `$"{username}:"` (clean interpolation, no `{""}` wrapper).
- **`{""}` pattern removed**: Auth header encoding changed from `$"{username}:{""}"` to `$"{username}:"` — both in the generator logic and the seed output.
- **Biome formatting fix**: Removed trailing whitespace on the `writeTextStatement` line.
- **Merge conflict resolved**: Incorporated new 2.56.6 entry from main into `versions.yml`.
- **Dynamic snippets updated**: All `Example0.cs`–`Example6.cs` now exclude the `password` named argument. `BaseMockServerTest.cs` instantiates with only `"USERNAME"`.

## Testing

- [x] Seed snapshot generated for `basic-auth-pw-omitted` fixture
- [x] Existing seed fixtures unchanged (no regressions)
- [x] Lint passes (`biome check`)
- [x] All 20 seed-test-results pass
- [x] `test-ete` failure is pre-existing flaky timeout (`upgrade-generator.test.ts` 180s) — unrelated to this PR

## ⚠️ Human Review Checklist

1. **`isFirstBlock` control flow** (lines ~456–490 of `RootClientGenerator.ts`): Confirm that `continue` (both-omitted case) does **not** set `isFirstBlock = false`, so the next non-skipped scheme correctly gets `if`.
2. **Auth header interpolation correctness**: Verify that `$"${usernamePart}:${passwordPart}"` in the TypeScript template produces valid C# like `$"{username}:"` (not `$":"` or other malformed strings) when one field is omitted. The TS template uses `${}` for JS interpolation, while the output `$""` is C# interpolation — double-check there's no escaping issue.
3. **Dynamic snippets cast is intentional**: `EndpointSnippetGenerator.ts` uses `as unknown as Record<string, unknown>` because `@fern-api/dynamic-ir-sdk` lacks typed omit fields. This is a known limitation — a follow-up to add these fields to the dynamic IR schema would make this fully robust (see Devin Review comment).
4. **`DynamicSnippetsConverter.ts` schema gap**: `usernameOmit`/`passwordOmit` are passed through as extra properties on the `BasicAuth` object, but these fields are not yet in the dynamic IR YAML schema. This works for in-process / raw JSON paths (including seed tests) but would be stripped by schema-based deserialization. Deferring the IR schema update to a follow-up.

Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14409" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


seed/csharp-sdk/basic-auth-pw-omitted/src/SeedBasicAuthOptional/SeedBasicAuthPwOmittedClient.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace SeedBasicAuthPwOmitted;
+
+public partial class SeedBasicAuthPwOmittedClient : ISeedBasicAuthPwOmittedClient
+{
+    private readonly RawClient _client;
+
+    public SeedBasicAuthPwOmittedClient(
+        string? username = null,
+        string? password = null,
```
Swimburger: why is there still a password field when it should be omitted?

seed/csharp-sdk/basic-auth-pw-omitted/src/SeedBasicAuthPwOmitted/SeedBasicAuthPwOmittedClient.cs
```diff
@@ -31,7 +30,7 @@ public SeedBasicAuthPwOmittedClient(
         }
         var clientOptionsWithAuth = clientOptions.Clone();
         clientOptionsWithAuth.Headers["Authorization"] =
-            $"Basic {Convert.ToBase64String(global::System.Text.Encoding.UTF8.GetBytes($"{username}:{password}"))}";
+            $"Basic {Convert.ToBase64String(global::System.Text.Encoding.UTF8.GetBytes($"{username}:{""}"))}";
```
Swimburger: you can remvoe {""}

fern-support (approved):

PR #14412, opened 2026-03-31
feat(php): support omitting username/password from basic auth when configured in IR

Swimburger:
## Description

Refs #14378

Adds conditional support for omitting username or password in basic auth for the PHP SDK generator. When `usernameOmit` or `passwordOmit` flags are set in the IR's `BasicAuthScheme`, the omitted field is **completely removed from the generated SDK's constructor** (not just made optional). Internally, the omitted field is treated as an empty string when encoding the `Authorization` header. Default behavior (both required) is preserved when no omit flags are set.

## Changes Made

- **`RootClientGenerator.ts` — `getParameterForAuthScheme()`**: Reads `scheme.usernameOmit`/`scheme.passwordOmit` directly (IR SDK v66 has typed fields). When a field is omitted, it is excluded from the returned `ConstructorParameter[]` entirely — the end user never sees it.
- **`RootClientGenerator.ts` — `resolveBasicAuthScheme()` (new helper)**: Extracted helper that resolves a basic auth scheme into its null-check condition and a single `credentialExpr` string. Builds clean concatenation without redundant empty-string literals (e.g., `$username . ":"` not `$username . ":" . ""`). Returns `undefined` when both fields are omitted, allowing the caller to pre-filter skipped schemes via `.map().filter()`.
- **`RootClientGenerator.ts` — `getConstructorMethod()`**: Pre-filters basic auth schemes through `resolveBasicAuthScheme()`, then iterates only resolved schemes. Per-field condition logic: only non-omitted fields are null-checked.
- **`EndpointSnippetGenerator.ts`**: Dynamic snippets exclude omitted fields from generated code examples. Uses `as unknown as Record<string, unknown>` cast (the `@fern-api/dynamic-ir-sdk` package lacks typed `usernameOmit`/`passwordOmit` fields).
- **`WireTestGenerator.ts` — `buildAuthParamsForTest()`**: Now checks `basicScheme.usernameOmit`/`passwordOmit` before emitting auth params in wire test `setUp()`, so omitted fields are not passed to the generated `SeedClient` constructor. Uses `test-username`/`test-password` values to match WireMock mapping expectations from `mock-utils`.
- **Bug fix**: Password `@param` docs now correctly reference `password` name instead of `username` (copy-paste bug in existing code, visible in both `basic-auth` and `basic-auth-pw-omitted` fixtures).
- **`versions.yml`**: New 2.4.0 entry (`irVersion: 66`).
- **New `basic-auth-pw-omitted` test fixture** with `password: omit: true`, wire-tests only (no default output folder). Orphaned root-level seed output cleaned via `pnpm seed clean`.
- **Wire tests enabled for `basic-auth-pw-omitted`**: Added `wire-tests` output folder with `enable-wire-tests: true` in `seed.yml`. WireMock mappings assert `Authorization: Basic dGVzdC11c2VybmFtZTo=` (`base64("test-username:")`) at request time via `equalTo` matcher.
- **`basic-auth` wire-tests**: WireMock mappings now include `Authorization` header `equalTo` matching (from merged shared `mock-utils` changes). Wire test username corrected from `test-user` to `test-username`.
- **`test-definitions`**: Removed unnecessary `ir-version: v61` constraint for PHP (already consumes v66); updated Go constraint from `v61` to `v63`.

## Testing

- [x] Seed snapshot generated for `basic-auth-pw-omitted` fixture — constructor has only `$username`, dynamic snippets omit password
- [x] Wire tests enabled for `basic-auth-pw-omitted` — WireMock mappings assert correct `Authorization` header at request time
- [x] Wire test `setUp()` correctly omits password param for pw-omitted fixture
- [x] Wire test username aligned with mock-utils (`test-username`, not `test-user`)
- [x] Existing `basic-auth` fixture regenerated — no regressions (docs fix + WireMock header matching + username fix)
- [x] Orphaned seed output cleaned (`pnpm seed clean` — removed root-level `.fern/`, `.github/`, `src/`, `tests/` dirs)
- [x] `pnpm run check` (biome lint) passing locally
- [x] CI passing (compile, lint, biome, test, all 17 required seed-test-results)

## ⚠️ Human Review Checklist

1. **Wire test auth values must match mock-utils**: `buildAuthParamsForTest()` uses `test-username`/`test-password` — these must match what `mock-utils/index.ts` uses to generate WireMock `Authorization` header `equalTo` matchers. A mismatch causes WireMock to reject requests at stub-matching time (404), failing wire tests.
2. **Dynamic snippets cast**: `EndpointSnippetGenerator.ts` uses `as unknown as Record<string, unknown>` because `@fern-api/dynamic-ir-sdk` lacks typed omit fields. This is intentional but fragile — if IR field names change, the cast will silently produce incorrect behavior. Updating the dynamic IR SDK types is out of scope for this PR.
3. **`resolveBasicAuthScheme()` credential expression**: Three branches — username-omitted produces `":" . $password`, password-omitted produces `$username . ":"`, both-present produces `$username . ":" . $password`. Both-omitted returns `undefined` and is filtered before the loop. Verify the string building is correct for each case.
4. **WireMock mapping values**: `basic-auth-pw-omitted` asserts `Basic dGVzdC11c2VybmFtZTo=` (`base64("test-username:")`). `basic-auth` asserts `Basic dGVzdC11c2VybmFtZTp0ZXN0LXBhc3N3b3Jk` (`base64("test-username:test-password")`). Verify both base64 values.
5. **Go fixture `ir-version: v63`**: The Go SDK uses IR v61 natively, so its seed fixture still needs an explicit `ir-version: v63` to test the omit feature. PHP's constraint was removed since PHP already consumes IR v66.

Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14412" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


seed/php-sdk/basic-auth-pw-omitted/src/SeedClient.php
```diff
@@ -0,0 +1,69 @@
...
+     *   baseUrl?: string,
+     *   client?: ClientInterface,
+     *   maxRetries?: int,
+     *   timeout?: float,
+     *   headers?: array<string, string>,
+     * } $options
+     */
+    public function __construct(
+        string $username,
+        string $password,
```
Swimburger: why is password still here?

generators/php/sdk/src/root-client/RootClientGenerator.ts
```diff
@@ -357,22 +357,40 @@ export class RootClientGenerator extends FileGenerator<PhpFile, SdkCustomConfigS
...
+                        if (!usernameOmitted && !passwordOmitted) {
+                            condition = `$${usernameName} !== null && $${passwordName} !== null`;
+                        } else if (usernameOmitted && !passwordOmitted) {
+                            condition = `$${passwordName} !== null`;
+                        } else if (!usernameOmitted && passwordOmitted) {
+                            condition = `$${usernameName} !== null`;
+                        } else {
+                            // Both fields omitted — skip auth header entirely when auth is non-mandatory
+                            continue;
+                        }
```
amckinney: This code is fairly verbose / difficult to read. Can we extract this into a helper function to clear this up?

generators/php/sdk/src/root-client/RootClientGenerator.ts
```diff
@@ -357,22 +357,40 @@ export class RootClientGenerator extends FileGenerator<PhpFile, SdkCustomConfigS
                 );
                 if (basicAuthSchemes.length > 0) {
                     const isAuthOptional = !this.context.ir.sdkConfig.isAuthMandatory;
+                    let isFirstBlock = true;
```
amckinney: Is there a better way to capture this intent? This dangling `boolean` checks are often a code smell.

generators/php/sdk/src/root-client/RootClientGenerator.ts
```diff
@@ -605,8 +623,13 @@ export class RootClientGenerator extends FileGenerator<PhpFile, SdkCustomConfigS
             case "basic": {
                 const username = this.context.getParameterName(scheme.username);
                 const password = this.context.getParameterName(scheme.password);
-                return [
-                    {
+                // When omit is true, the field is completely removed from the end-user API.
+                const schemeRecord = scheme as unknown as Record<string, unknown>;
+                const usernameOmitted = schemeRecord.usernameOmit === true;
+                const passwordOmitted = schemeRecord.passwordOmit === true;
```
amckinney: Is it not sufficient to just check that `if (schemeRecord.usernameOmit)` rather than an explicit assignment to `=== true`?

generators/php/dynamic-snippets/src/EndpointSnippetGenerator.ts
```diff
@@ -321,16 +321,24 @@ export class EndpointSnippetGenerator {
         auth: FernIr.dynamic.BasicAuth;
         values: FernIr.dynamic.BasicAuthValues;
     }): NamedArgument[] {
-        return [
-            {
+        // usernameOmit/passwordOmit may exist in newer IR versions
+        const authRecord = auth as unknown as Record<string, unknown>;
+        const usernameOmitted = authRecord.usernameOmit === true;
+        const passwordOmitted = authRecord.passwordOmit === true;
```
amckinney: Is it not sufficient to just check that `if (authRecord.usernameOmit)` rather than an explicit assignment to `=== true`? The same is true for `passwordOmit`.

fern-support (approved):

PR #14468, opened 2026-04-01
fix(csharp): fix namespace conflict detection in subpackage and WebSocket client generators

Swimburger:
## Description

Fixes C# compilation errors (e.g., CS0118 namespace-vs-type ambiguity) in generated subpackage client files when `explicit-namespaces` is enabled.

Three generators were using `this.registry.allNamespacesOf(namespace)` for the `allNamespaceSegments` parameter, which returns fully-qualified namespace prefix paths (e.g., `{"Contoso", "Contoso.Net", "Contoso.Net.ScimConfiguration"}`). Every other generator in the codebase uses `this.context.getAllNamespaceSegments()` which returns individual segment names (e.g., `{"Contoso.Net", "ScimConfiguration", "System"}`). The `hasProjectNamespaceConflict()` check does `allNamespaceSegments.has(typeName)`, so with the wrong set it would never find a bare segment like `"ScimConfiguration"` — only fully-qualified paths — causing the conflict to go undetected.

Additionally, `SubPackageClientGenerator` and `SubPackageClientInterfaceGenerator` passed `this.namespaces.root` as the `namespace` parameter instead of the actual subpackage namespace (`this.classReference.namespace` / `this.interfaceReference.namespace`).

**Concrete example:** With namespace `Auth0.ManagementApi` and a `ScimConfiguration` subpackage, the root client gets a property `public IScimConfigurationClient ScimConfiguration { get; }` that shadows namespace `Auth0.ManagementApi.ScimConfiguration`. Without this fix, the Writer doesn't detect the conflict — causing CS0118 at compile time.

**Verified against Auth0 Management API spec:** Generated C# SDK with both 2.55.3 (pre-fix) and 2.55.4 (post-fix). Only 2 files differ — `DirectoryProvisioningClient.cs` and `IDirectoryProvisioningClient.cs` — where `Pager<DirectoryProvisioning>` is now correctly qualified as `Pager<Auth0.ManagementApi.DirectoryProvisioning>`, resolving the type-vs-namespace collision. `dotnet build` passes with no namespace-related errors.

## Changes Made

- **`SubPackageClientGenerator.ts`**: Changed `allNamespaceSegments` from `this.registry.allNamespacesOf(...)` → `this.context.getAllNamespaceSegments()`, and `namespace` from `this.namespaces.root` → `this.classReference.namespace`
- **`SubPackageClientInterfaceGenerator.ts`**: Same two changes, using `this.interfaceReference.namespace`
- **`WebsocketClientGenerator.ts`**: Changed `allNamespaceSegments` from `this.registry.allNamespacesOf(...)` → `this.context.getAllNamespaceSegments()` in both `generate()` and `generateInterface()` methods (same bug pattern)
- **New test definition**: Added `ScimConfiguration` subpackage to `csharp-namespace-collision` fixture with types (`ScimConfiguration`, `ScimToken`) and a cross-package type reference (`listUsers` returns `list<root.User>`), which forces the subpackage client to reference a type from a different namespace — exercising `hasProjectNamespaceConflict()` and producing visible `global::Contoso.Net.User` qualifications in the generated output
- **New seed test variants**: Added `explicit-namespaces` and `no-client-namespace-match` variants to the `csharp-namespace-collision` fixture
- **Updated snapshots**: Regenerated IR, dynamic-snippets, and JSON schema snapshots for `csharp-namespace-collision`
- Bumped version to `2.55.4` in `versions.yml`

## Testing

- [x] Seed tests pass for `csharp-namespace-collision` (4/4 variants), `csharp-namespace-conflict` (1/1), `exhaustive` (6/6)
- [x] No snapshot diffs on existing fixtures (backward-compatible)
- [x] Manual `dotnet build` succeeds on `no-client-namespace-match` variant
- [x] Cross-package type reference (`listUsers` → `root.User`) produces `global::Contoso.Net.User` in `ScimConfigurationClient.cs` for the `explicit-namespaces` variant — this would revert to unqualified `User` (causing CS0118) if the fix is reverted
- [x] Verified zero remaining instances of `allNamespaceSegments: this.registry.allNamespacesOf` in the codebase
- [x] IR and dynamic-snippets JSON snapshots updated and passing
- [x] Verified against Auth0 Management API spec — only `DirectoryProvisioningClient` files change, `dotnet build` passes

## ⚠️ Human Review Checklist

- [x] **`getAllNamespaceSegments()` is the correct source**: Confirmed — 25+ call sites across all generators (TypeGenerator, ErrorGenerator, RootClientGenerator, etc.) use `this.context.getAllNamespaceSegments()`. Returns individual segment names, consistent with what `hasProjectNamespaceConflict()` expects.
- [x] **`CSharpFile` namespace parameter is dead code**: Confirmed — the constructor destructures `clazz`, `directory`, `allNamespaceSegments`, `allTypeClassReferences`, `generation`, `fileHeader` but **not** `namespace`. `resolveFileContents()` uses `this.clazz.namespace` directly. The `namespace` param change is a no-op at the `CSharpFile` level. Latent dead code worth a separate cleanup.
- [x] **`irVersion: 65` is correct**: Confirmed — latest IR SDK on main is 65.6.0; all recent C# generator versions use irVersion 65.
- [ ] **WebSocket fix is untested**: No seed fixture exercises WebSocket + `explicit-namespaces` together. The fix is pattern-consistent with every other generator, but there's no test that would catch a regression specifically in the WebSocket code path.
- [ ] **No `dotnet build` in seed CI**: Seed tests use `--skip-scripts`; compilation was only verified manually against Auth0 spec and one seed variant.

Link to Devin session: https://app.devin.ai/sessions/68fd44be23004ed3b975778eb0a7bf14
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14468" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


generators/csharp/sdk/versions.yml
```diff
@@ -1,4 +1,20 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 2.55.4
+  changelogEntry:
+    - summary: |
+        Fix namespace conflict detection in generated subpackage client and
```
fern-support: too much implementation detail info, keep it focused on what it means to SDK generator consumers

fern-support (approved):

PR #14482, opened 2026-04-01
fix(csharp): RFC 3986 spec-compliant encoding with separate path, query key, and query value contexts

jon-fern:
## Description

Refs docusign-demo C# SDK generation — makes `QueryStringBuilder` encoding fully RFC 3986 spec-compliant by distinguishing three encoding contexts (path segments, query keys, query values) instead of using a single shared character set. Also adds percent-encoding for path parameter values, which were previously passed through raw.

## Changes Made

- **Three distinct safe character sets** derived from RFC 3986 ABNF grammar:
  - **Query keys**: `unreserved + (sub-delims \ {& = +}) + : @ / ?` — excludes `&`, `=`, `+`, `#`
  - **Query values**: `unreserved + (sub-delims \ {& +}) + : @ / ?` — excludes `&`, `+`, `#` (but allows `=`)
  - **Path segments** (pchar): `unreserved + sub-delims + : @` — excludes `/` and `?`
- Renamed `SafeQueryChars` → `SafeQueryKeyChars` / `SafeQueryValueChars` / `SafePathChars`
- Renamed `EncodeComponent` → `EncodeWithCharSet` with new `EncodingContext` enum parameter
- Added `IsSafeQueryKeyChar`, `IsSafeQueryValueChar`, `IsSafePathChar` replacing the single `IsSafeQueryChar`
- New public `EncodePathSegment(string)` method for path parameter encoding
- `ValueConvert.ToPathParameterString(string v)` now calls `QueryStringBuilder.EncodePathSegment(v)` instead of `ToString(v)` — path parameters are now percent-encoded
- Added unit tests covering path segment encoding, query key vs value differentiation, `?` passthrough, `+` encoding, and OData `$filter` preservation
- Updated all 161 csharp-sdk seed snapshots (648 files total)
- Version bump: 2.56.1

## Important Review Points

1. **`=` is now unencoded in query _values_ but encoded in query _keys_** — e.g. `?key%3Dwith%3Dequals=value=with=equals`. RFC 3986 allows `=` in query values, but verify this is safe across the server ecosystem.
2. **`?` is now unencoded in both query keys and values** — e.g. `?url=https://example.com/path?query=value`. RFC 3986 allows this, but some naive query-string parsers may split on all `?` characters.
3. **Path parameters are now percent-encoded** — previously `ToPathParameterString(string)` just called `ToString()`. Now it encodes via `EncodePathSegment()`, which encodes `/`, `?`, `#`, `[`, `]`, spaces, etc. This is a behavioral change for any path parameter containing those characters.
4. **`(char)39` instead of `'\''`** — required because `.Template.cs` files pass through Eta's `renderString`, which strips backslashes.
5. **Seed snapshot updates were scripted** — 648 files via find-and-replace. Pattern is mechanical but worth a spot-check.

### ⚠️ Human Review Checklist

- [ ] Confirm the three character sets match RFC 3986 §3.3 (path) and §3.4 (query)
- [ ] Verify that allowing unencoded `=` in query values doesn't break any known server or framework
- [ ] Verify that allowing unencoded `?` in query keys/values doesn't confuse query-string parsers
- [ ] Check that newly-encoded path parameters don't break existing generated SDKs (e.g., path params that previously relied on raw passthrough)
- [ ] Spot-check a few seed snapshots to confirm the scripted update is correct

## Testing

- [x] Unit tests added for `EncodePathSegment` (unreserved, sub-delimiters, colon/at, slash/question encoding, space, empty/null)
- [x] Unit tests added for query key vs value encoding (`=` encoded in key, unencoded in value)
- [x] Unit tests added for `?` passthrough, `+` encoding, OData `$filter` preservation
- [x] Existing tests updated for new encoding behavior
- [x] Seed snapshots updated for all 161 csharp-sdk seeds
- [x] Local lint (`pnpm run check`) passes

Link to Devin session: https://app.devin.ai/sessions/b09769ac33ba4a4085df1b9207921e48
Requested by: @jon-fern

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14482" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


generators/csharp/base/src/asIs/ValueConvert.Template.cs
```diff
@@ -29,7 +29,7 @@ public static class ValueConvert
 
     internal static string ToPathParameterString(ulong v) => ToString(v);
     
-    internal static string ToPathParameterString(string v) => ToString(v);
+    internal static string ToPathParameterString(string v) => QueryStringBuilder.EncodePathSegment(v);
```
Swimburger, line 32: Is this necessary? I want to make sure we're not double-encoding by accident.

Swimburger (approved):

PR #14487, opened 2026-04-02
feat(typescript): propagate Error.cause across all SDK error paths

Swimburger:
## Description

When a custom fetcher throws an error (network failure, DNS issues, custom logic), the generated TypeScript SDK catches it but only preserves `error.message` as a flat string. The original error object — including stack trace, custom properties (`code`, `hostname`, `port`), and any nested error chain — is completely lost.

This PR preserves the original thrown error as `cause` on **all** SDK error classes, following the ES2022 `Error.cause` convention. Every error path where an exception is caught now forwards it as `cause`.

### 1. `Fetcher.ts` (core utility — copied into every generated SDK)
- Added `cause?: unknown` to both the `UnknownError` and `TimeoutError` interfaces
- Attached the original error as `cause` in all catch paths: user abort, timeout (AbortError), `Error` instance, and fallback

### 2. `GeneratedGenericAPISdkErrorImpl.ts` (code generator for `<SdkName>Error` class)
- Added `public readonly cause?: unknown` class property declaration via `getClassProperties()`
- Constructor now accepts optional `cause` parameter in the destructured options
- Conditionally assigns `this.cause = cause` in the constructor body (only when `cause != null`)
- Does **not** pass cause to `super()` — generated SDKs target ES6 where `Error()` only accepts a message string

### 3. `GeneratedTimeoutSdkErrorImpl.ts` (code generator for `<SdkName>TimeoutError` class)
- Added `public readonly cause?: unknown` class property declaration
- Constructor now accepts an optional options object as its second parameter: `constructor(message: string, opts?: { cause?: unknown })` — matching the native ES2022 `Error` convention (`new Error(msg, { cause })`)
- Conditionally assigns `this.cause = opts.cause` in the constructor body when `opts?.cause != null`

### 4. `GeneratedNonStatusCodeErrorHandlerImpl.ts` (code generator for `handleNonStatusCodeError`)
- Forwards `error.cause` when constructing the SDK error for both `"unknown"` and `"timeout"` error reasons
- Timeout case passes `{ cause: error.cause }` as the options object (matching the new constructor signature)

### 5. Supporting interface/type updates
- `GeneratedGenericAPISdkError` interface: added optional `cause` to `build()` and `buildConstructorArguments()` signatures
- `GeneratedTimeoutSdkError` interface: added optional `cause` parameter to `build()` signature
- Commons `Fetcher.ts`: added `cause: "cause"` property mapping to both `UnknownError` and `TimeoutSdkError`

### 6. Test updates
- `NonStatusCodeErrorHandler.test.ts`: Added `cause` to mock `TimeoutSdkError` and `UnknownError` contexts, and mock `build()` args
- Updated snapshots for both `NonStatusCodeErrorHandler` and `GenericSdkErrorGenerators` tests

### 7. Seed snapshots
- ~800+ files updated across all ts-sdk fixtures (4 files per fixture: `Fetcher.ts`, `SeedXxxError.ts`, `SeedXxxTimeoutError.ts`, `handleNonStatusCodeError.ts`)

- [ ] Updated README.md generator (not applicable — no docs changes)

- Fixed inconsistency flagged by Devin Review: changed TimeoutError's cause check from `if (opts?.cause)` (truthiness) to `if (opts?.cause != null)` (proper null check), now consistent with the generic API error class.
- Fixed user abort path: changed from `cause: error instanceof Error ? error : undefined` to `cause: error`, consistent with all other catch paths that pass the caught value directly.

## Human Review Checklist

- [ ] **`cause` property declared on both generated error classes**: Both `<SdkName>Error` and `<SdkName>TimeoutError` declare `public readonly cause?: unknown`. On ES2022+ targets this shadows the native `Error.cause`, but both are typed `unknown` so there's no type conflict. On ES6 targets (which generated SDKs use), native `Error.cause` doesn't exist, making the explicit declaration required.
- [ ] **Both constructors use named/options patterns**: The generic error uses destructured options (`{ cause }`) and the timeout error uses an options object as the second parameter (`opts?: { cause?: unknown }`), matching the native `new Error(msg, { cause })` convention. Neither uses positional `cause` arguments.
- [ ] **Null checks are consistent**: Both error classes use `!= null` checks for cause assignment — the generic error uses `if (cause != null)` and the timeout error uses `if (opts?.cause != null)`.
- [ ] **Conditional assignment**: `this.cause` is only set when cause is present, so errors without a cause retain the default `undefined` from the property declaration.
- [ ] **Backwards compatibility**: `cause` is optional everywhere, so existing consumers are unaffected.
- [ ] **No `super()` change**: `super()` still only receives the message string in both error classes. Cause is assigned in the constructor body after `super()`, which is the correct pattern for ES6 targets.
- [ ] **All caught-exception paths covered**: user abort → `cause: error`, timeout (AbortError) → `cause: error`, Error instance → `cause: error`, non-Error fallback → `cause: error`. All paths now consistently pass the caught value directly. Status-code errors, non-json, and body-is-null don't involve caught exceptions so they don't set `cause`.
- [ ] **Version bump**: `versions.yml` goes from `3.61.0-rc.0` → `3.61.1`, skipping `3.61.0` stable. Verify this is intentional.

## Testing

- [x] Unit tests updated (GenericSdkErrorGenerators + NonStatusCodeErrorHandler snapshots)
- [x] All ts-sdk seed tests pass (203/204; 1 pre-existing expected failure: `inferred-auth-implicit-reference`)
- [x] `pnpm compile` succeeds for all TypeScript generator packages
- [x] `pnpm format` passes

Link to Devin session: https://app.devin.ai/sessions/a5911fc3415248b8a309407b1c4e7e9d
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14487" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


seed/ts-sdk/oauth-client-credentials-mandatory-auth/no-custom-config/src/core/fetcher/Fetcher.ts
```diff
@@ -338,6 +340,7 @@ export async function fetcherImpl<R = unknown>(args: Fetcher.Args): Promise<APIR
                 error: {
                     reason: "unknown",
                     errorMessage: "The user aborted a request",
+                    cause: error instanceof Error ? error : undefined,
```
Swimburger: we don't want to set teh cause when it's not an instance of Error?

fern-support (approved):

PR #14520, opened 2026-04-02
chore(csharp): pin csharpier version via local tool-path instead of global install

Swimburger:
## Description

When running `pnpm seed test --local` for the C# generator, the version of `csharpier` used for formatting was whatever happened to be globally installed on the developer's machine, rather than the pinned version (`1.2.6`) used in the Docker container.

## Changes Made

- **Generator simplified**: `CsharpFormatter` and `CsharpProject` now call `csharpier` directly, assuming it is on PATH. All path-resolution logic (`findDotNetToolPath.ts`) is removed.
- **Local tool install**: `preInstallScript` in `seed/csharp-sdk/seed.yml` installs csharpier `1.2.6` to `generators/csharp/.tools/` using `--tool-path` (non-global, no side effects on the developer's system).
- **PATH injection**: `test.local.env` in seed.yml sets `PATH: "{REPO_ROOT}/generators/csharp/.tools:{PATH}"`. `LocalTestRunner` expands the `{REPO_ROOT}` and `{PATH}` placeholders at runtime before passing env to the generator subprocess.
- **devbox shell**: Same local install + PATH export in `devbox.json` for interactive development.
- **`.gitignore`**: `generators/csharp/.tools/` is excluded from version control.

## Testing

- [x] Unit tests added for `resolveEnv` (placeholder expansion logic)
- [x] `pnpm compile` passes for all affected packages


fern-support (approved):

Swimburger:
Is this still rel

PR #14656, opened 2026-04-05
fix(openapi): preserve schemaless request body examples in new OpenAPI 3.1 importer

fern-support:
## Description

Refs: Follow-up to #14553 which fixed the same issue in the **old** OpenAPI importer (`openapi-ir-parser`). This PR fixes the **new** OpenAPI 3.1 importer (`openapi-to-ir`) which is what the CLI actually uses for modern specs.

Specs like Close's have ~2/3 of endpoints with native OpenAPI `example`/`examples` on the request body but no `schema` defined. The new importer was silently dropping these request bodies and their examples, so docs showed no request body panel and curl snippets had no `-d` body.

## Changes Made

**`RequestBodyConverter.ts`** — new `handleSchemalessJsonContent()` method:
- When a JSON content type has no schema but **does** have native examples, creates an `unknown`-typed reference request body with examples preserved in `v2Examples.userSpecifiedExamples`
- When there are no examples either (e.g. `google.protobuf.Empty` → empty JSON body), returns `undefined` to preserve existing behavior

**`OperationConverter.ts`** — new `synthesizeEndpointExamplesFromNativeRequestBody()` method:
- When `convertExamples()` finds no `x-fern-examples` or `x-code-samples`, extracts native OpenAPI examples from the request body's media type objects
- **Only processes media types that have no schema** — media types WITH schemas already have their examples handled by the normal schema-based pipeline
- Wraps each example into a `V2HttpEndpointExample` with the request body populated, so docs and curl snippets can render them

**`OperationConverter.ts`** — new `extractNativeResponseExample()` method:
- When synthesizing endpoint examples for schemaless request bodies, also extracts the first native example from the operation's successful (2xx) response
- This preserves the 200 response panel in docs (e.g. Close's `POST /lead/` has both a schemaless request body example AND a response example with a `$ref` schema — previously only the request body appeared in the curl snippet while the response panel was empty)

**ETE test timeout fix** (`add.test.ts`, `generators-list.test.ts`):
- Increased timeout from 60s to 180s for three network-dependent tests (`fern add <generator>`, `fern add <generator> --group sdk`, `fern generator list with include`) that were timing out on CI runners when resolving generator versions from external registries

## ⚠️ Items for reviewer attention

1. **Potential duplicate example storage**: Examples are stored in *both* `RequestBodyConverter` (as `v2Examples` on the request body type) *and* `OperationConverter` (as endpoint-level `v2Examples`). Verify downstream consumers (FDR write, `convertPackage.ts`) don't produce duplicate entries.
2. **Empty path parameters**: `synthesizeEndpointExamplesFromNativeRequestBody` passes `{}` for path parameters. For endpoints like `/plants/{plantId}`, the generated example path will contain the raw param name instead of a real value. Is this acceptable?
3. **Single response example shared across all request examples**: `extractNativeResponseExample()` returns only the first 2xx example found. This same response is attached to every synthesized request body example. If the spec has multiple request examples that should pair with different responses, they'll all share one response.
4. **ETE timeout increase**: The 60s → 180s increase fixes the immediate CI failure but the root cause of the slowness (external registry resolution) is not addressed. These tests may still be flaky under heavy CI load.

## Testing
- [x] TypeScript compilation passes (`tsc --noEmit`)
- [x] Biome lint passes
- [x] Existing `openapi-to-ir` tests pass
- [x] All 198 `@fern-api/register` snapshot tests pass (including 4 that previously failed before the `schema != null` guard was added)
- [x] Verified against Close spec: 68 endpoints with schemaless request body examples now preserved
- [ ] Unit tests added for new methods

Link to Devin session: https://app.devin.ai/sessions/041b28e41b704c96ba9c94fd19ba9ceb
Requested by: @cdonel707


dsinghvi:
Fix the ETE test

PR #14666, opened 2026-04-06
fix(go): use *int instead of *float64 for custom pager GetLimit/GetCount

iamnamananand996:
## Description

Fixes the `CustomPaginatedResponse` interface in the Go SDK generator's custom pagination template to use `*int` instead of `*float64` for `GetLimit()` and `GetCount()` return types.

When an OpenAPI spec defines `limit` and `count` fields as `type: integer`, the generated concrete response types correctly produce `GetLimit() *int` and `GetCount() *int` getters. However, the `CustomPaginatedResponse` interface hardcoded `*float64` for these methods, causing a type mismatch that prevents concrete types from satisfying the interface.

Reported via https://github.com/payroc/papi-sdk-go/pull/348/changes#r2822855927

## Changes Made

- Updated `generators/go-v2/base/src/asIs/core/custom_pagination.go_` to change `GetLimit() *float64` → `GetLimit() *int` and `GetCount() *float64` → `GetCount() *int`
- Added version 1.33.2 changelog entry in `generators/go/sdk/versions.yml`
- Rewrote `pagination-custom` test fixture to use a proper custom pager response type (`UsersListResponse`) with integer `limit`/`count`, `has_more`, `links`, and `data` fields
- Regenerated Go SDK seed output for `pagination-custom`, confirming `*int` return types on both the `CustomPaginatedResponse` interface and the concrete `UsersListResponse` type
- Removed `pagination-custom` from `allowedFailures` in `seed/go-sdk/seed.yml` (fixture now passes)

## Testing

- [x] Go SDK seed test passes for `pagination-custom` fixture with `customPagerName: PayrocPager`
- [x] Generated `core/custom_pagination.go` confirms `GetLimit() *int` and `GetCount() *int`
- [x] Generated `users.go` confirms concrete `UsersListResponse.GetLimit() *int` and `GetCount() *int`

## Human Review Checklist

- [ ] **Breaking change risk**: Verify no existing SDK users have specs where `limit`/`count` are `type: number` (which would generate `*float64` getters that no longer satisfy the interface). Pagination counts are semantically integers, so `*int` should be universally correct.
- [ ] **No other float64 references**: Confirm no other generator code constructs or casts these values assuming `float64`.
- [ ] **Fixture rewrite scope**: The `pagination-custom` fixture definition was fully rewritten (old `UsernameCursor`/`UsernamePage` types replaced with `UsersListResponse`/`Link`). Other generators (ts-sdk, java-sdk, python-sdk, php-sdk, ruby-sdk-v2) also reference this fixture — their seed output may need regeneration if CI fails.

Link to Devin session: https://app.devin.ai/sessions/3c113a1beb8c44bea87a422b4100e273


amckinney (approved):
I'm surprised this was `float64` to start.

PR #14717, opened 2026-04-07
feat(cli): present TUI dropdown for --api, --instance, --group flags in CLI v2

iamnamananand996:
## Description

Linear ticket: Closes [FER-9415](https://linear.app/buildwithfern/issue/FER-9415/cli-v2-present-a-dropdown-of-options-for-common-flags)

When users run CLI v2 commands that require `--api`, `--instance`, or `--group` but forget to specify them and multiple options exist, the CLI now presents an interactive dropdown (via inquirer) instead of failing hard. This only applies to TTY environments; non-TTY (e.g. CI) retains the existing error behavior.

**Important:** Batch commands (`api check`, `api merge`, `api split`) are **not** changed — they continue to run on all APIs by default when `--api` is omitted, matching their existing behavior. The TUI prompt only applies to commands that require exactly one selection (e.g. `api compile`).

## Changes Made

- **New shared utility** `packages/cli/cli-v2/src/ui/promptSelect.ts` — reusable `promptSelect<T>()` function that checks `isTTY`, shows an inquirer list prompt in interactive mode, or throws a `CliError` with the original error message in non-interactive mode. Supports an optional `flagHint` callback that annotates each choice with the equivalent CLI flag (e.g. `rest   --api rest`). Includes an empty-choices guard that throws `CliError.Code.InternalError` when the choices array is empty, preventing `Math.max(...[])` → `-Infinity`.
- **`api compile`** — `resolveApi()` now prompts when multiple APIs are found and no `--api` flag is provided.
- **`docs publish`** — `resolveInstanceUrl()` now prompts when multiple docs instances are configured and no `--instance` flag is provided.
- **`sdk generate`** — `getTargets()` now prompts for group selection when multiple SDK groups exist and no `--group`/`--target` is specified. When no groups are defined, falls back to target-based selection. Uses `undefined` as the "all" sentinel to avoid any collision with real group/target names.
- **`sdk generate --target` not-found** — when an explicit `--target` flag doesn't match, the CLI now throws a clear `CliError` listing available targets instead of offering a selection prompt. This better respects the user's explicit intent.
- **Unit tests** — new `getTargets.test.ts`, `resolveApi.test.ts`, and `resolveInstanceUrl.test.ts` covering the selection/error paths.
- **`YamlConfigLoader.ts`** — improved `invalid_union` error handling to prefer surfacing `unrecognized_keys` errors over generic type errors, giving more actionable messages. Also exposes `pathMappings` on the `Success` result for `$ref` resolution tracking.
- **`$ref` path resolution** — `ApiDefinitionConverter`, `DocsConfigConverter`, and `LegacyDocsWorkspaceAdapter` now resolve spec/config paths relative to the correct source file when a config section (e.g. `api`, `apis`, `docs`) was loaded via `$ref`, rather than always resolving relative to `fern.yml`.
- Added changelog entry via the unreleased versioning system (`packages/cli/cli/changes/unreleased/tui-dropdown-prompts.yml`).

## Testing

- [x] Unit tests added/updated — `getTargets.test.ts`, `resolveApi.test.ts`, and `resolveInstanceUrl.test.ts` cover the selection/error paths (non-TTY via CliError assertions)
- [x] Manual testing completed — Biome, lint, compile, and all CI checks pass (30/30)

1. **`ALL_SENTINEL` collision risk** → replaced `"__all__"` string sentinel with `undefined`, eliminating any possibility of collision with real group/target names.
2. **`--target` not-found prompt** → changed to always throw `CliError` with available targets listed, instead of offering a selection dropdown when the user explicitly provided a `--target` flag.
3. **Switched to unreleased versioning system** — removed hardcoded version entry from `versions.yml` and added `packages/cli/cli/changes/unreleased/tui-dropdown-prompts.yml` per repo conventions. Version numbers are now assigned during release.
4. **Empty choices guard** — `promptSelect` now throws `CliError.Code.InternalError` when the choices array is empty, preventing `Math.max(...[])` from evaluating to `-Infinity`.
5. **`CliError` import/constructor updates** — after rebase, updated all imports from `../errors/CliError.js` to `@fern-api/task-context` and added the now-required `code` property to all `CliError` constructor calls.
6. **Restored batch-command behavior** — removed `resolveApiFilter` from `api check`, `api merge`, and `api split`. These commands now match main's behavior: they run on **all** APIs by default when `--api` is omitted and never prompt. The TUI prompt is limited to commands that require exactly one selection (`api compile`, `docs publish`, `sdk generate`).

## Human Review Checklist

- **`LegacyDocsWorkspaceAdapter` semantics change**: `DocsWorkspace.absoluteFilePath` is now set to `dirname(docsFilePath)` (the containing directory) instead of the config file path itself, and `absoluteFilepathToDocsConfig` is set to the actual config file. Verify downstream consumers of `DocsWorkspace.absoluteFilePath` expect a directory, not a file path.
- **`LegacyProjectAdapter` null-check relaxed**: The condition for building `docsWorkspace` changed from `workspace.docs != null && workspace.absoluteFilePath != null` to just `workspace.docs != null`, with `absoluteFilePath` falling back through `workspace.docs.absoluteFilePath ?? workspace.absoluteFilePath ?? this.context.cwd`. Verify this doesn't create unexpected docs workspaces when `absoluteFilePath` is null.
- **`$ref` path resolution in `ApiDefinitionConverter`**: The `findBasePathInMappings` helper resolves spec paths from the `$ref`'d file. Verify this handles nested `$ref` chains and doesn't break workspaces that don't use `$ref` (should fall back to `fernYml.absoluteFilePath`).
- **`undefined` as "all" sentinel in `promptSelect`**: The "all targets/groups" option passes `undefined` as the value. Verify inquirer handles `undefined` values correctly (it should — `undefined` is a valid JS value).
- **`YamlConfigLoader` `invalid_union` change**: Separate concern from the TUI feature — improves union validation error messages by preferring `unrecognized_keys` over generic type errors. The fallback to `invalid_type` is preserved. Verify `unrecognized_keys` preference doesn't suppress useful type errors in edge cases where the unrecognized key message is misleading.
- **Test helpers use type casts**: Test files access private methods via `as unknown as { ... }`. Fragile if method signatures change but will fail loudly.
- **`collectGroups` sorting**: Groups are sorted alphabetically in the dropdown. Confirm this is the desired ordering vs. declaration order in config.

Link to Devin session: https://app.devin.ai/sessions/0bcbc7f8893c4cb3acfb276f6074d50e

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14717" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/cli/cli/versions.yml
```diff
@@ -12,6 +12,17 @@
...
+- version: 4.67.0
+  changelogEntry:
+    - summary: |
+        Present an interactive TUI dropdown when the user omits `--api`,
+        `--instance`, or `--group` flags and multiple options exist.
+        In TTY environments the CLI now prompts for selection instead of
+        failing hard. Non-TTY environments retain the previous error behavior.
+      type: feat
+  createdAt: "2026-04-07"
+  irVersion: 66
```
Swimburger, line 166: We should use the new unreleased versioning system

packages/cli/cli-v2/src/ui/promptSelect.ts
```diff
@@ -0,0 +1,62 @@
...
+     * When provided, each choice label is annotated with the equivalent CLI flag
+     * so the user can see how to skip the prompt next time, e.g. "--api rest".
+     */
+    flagHint?: (value: T) => string | undefined;
+}): Promise<T> {
+    if (!isTTY) {
+        throw new CliError({ message: nonInteractiveError });
+    }
+
+    const maxNameLen = Math.max(...choices.map((c) => c.name.length));
```
Swimburger, line 49: promptSelect needs a guard for empty choices

Right now if choices is an empty array:

Math.max(...[]) evaluates to -Infinity
In TTY mode, inquirer renders an empty list that hangs — the user can't select anything
In non-TTY mode it throws nonInteractiveError, but that message ("Multiple APIs found…") would be misleading since there are none

packages/cli/cli-v2/src/commands/sdk/generate/command.ts
```diff
@@ -155,11 +157,14 @@ export class GenerateCommand {
             });
         }
 
-        // After validation, these are guaranteed to be defined.
-        const api = args.api as string;
-        const target = args.target as string;
-        const org = args.org as string;
-        const output = args.output as string;
+        // All four are guaranteed to be defined after the missingFlags check above.
+        if (args.api == null || args.target == null || args.org == null || args.output == null) {
```
Swimburger, line 161: This is fine, but perhaps we can take advantage of zod or a small validation library. 
Empty strings are still falling through, which may already be converted to null/undefined by yargs.
No need to change anything, just food for thought

Swimburger (approved):

PR #14718, opened 2026-04-07
fix(cli): pass in schema raw name if title not available

Ryan-Amirthan:
## Description
Linear ticket: Closes FER-8940
Union variants (both discriminated and undiscriminated) were displaying as "object" in documentation if they did not have a title (can use the $ref name instead)

## Changes Made
  - **Removed `"schema_identifier"` override source**: This was a broken pattern that passed schema names as overrides but then ignored them
  - **Added displayName fallback chain**: `schema.title ?? discriminant ?? rawSchemaName` ensures TypeReferences always have meaningful display names
  - **Fixed discriminated union variants**: Set displayName on TypeDeclarations (which docs read) instead of only on union metadata. Scoped to title-only so discriminant keys don't leak to global type declarations
  - **Added `else` fallback branches in all converter contexts**: AsyncAPI, OpenRPC, and OpenAPI contexts now consistently fall back to `title ?? rawSchemaName` when no override source is specified, preventing `undefined` displayName on undiscriminated union variants
  - **Consistent `split('/').pop()` for schema name extraction**: Replaced inline regex (`/\/schemas\/([^/]+)/`) and `getRawSchemaNameFromReference` with `reference.split('/').pop()` across all contexts
  - **Hoisted `getDiscriminatorKeyForRef` in undiscriminated union handler**: Avoids redundant double call; restructured as `const mappingEntry = ...; if (mappingEntry != null)` pattern

displayName is needed as a separate field because: name contains multiple casing variants (PascalCase, camelCase, snake_case), typeId includes the full path (e.g., "root.TreeNode"), and displayName can be customized via OpenAPI title

## Human Review Checklist
- [ ] Verify TypeDeclaration displayName is only set when `schemaTitle != null` — discriminant keys should not propagate to global type names
- [ ] Confirm `split('/').pop()` handles all `$ref` formats (not just `#/components/schemas/X`) without returning unexpected names
- [ ] Check that the `else` branch in `convertAsUndiscriminatedUnion` (standard refs with no mapping entry) correctly delegates to `convertReferenceToTypeReference` without an override
- [ ] Verify AsyncAPI v2 message fallback (`messageId ?? name ?? rawName`) matches expected priority

## Testing
  - [x] Unit tests updated (v3-importer-tests: 240 tests, register: 198 tests — all passing)
  - [x] Snapshot files regenerated across v3-importer-tests and register packages
  - [x] Manual testing: Verified union variants now show schema names instead of "object" in a test docs site

Link to Devin session: https://app.devin.ai/sessions/f34d1fca69f34b3cbec03bed76da3b45
Requested by: @Ryan-Amirthan


Ryan-Amirthan:
@claude review

packages/cli/api-importers/openrpc-to-ir/src/1.x/OpenRPCConverterContext3_1.ts
```diff
@@ -44,11 +44,12 @@ export class OpenRPCConverterContext3_1 extends AbstractConverterContext<Openrpc
...
             displayName = displayNameOverride ?? resolvedReference.value.title;
-        } else if (
-            displayNameOverrideSource === "discriminator_key" ||
-            displayNameOverrideSource === "schema_identifier"
-        ) {
+        } else if (displayNameOverrideSource === "discriminator_key") {
             displayName = resolvedReference.value.title ?? displayNameOverride;
+        } else {
+            // No override source - use schema title or fallback to extracted schema name
+            const rawSchemaName = reference.$ref.match(/\/schemas\/([^/]+)/)?.[1] ?? typeId;
```
amckinney: This is brittle -- the user could define schemas in any ref syntax -- it doesn't _need_ to be defined in `/components/schemas`. It can be any valid JSON pointer reference.

If the `typeId` isn't sufficient, I imagine we'd need to use the _basename_ of the JSON pointer reference (i.e. the identifier after the last `/`).

If so, that would look something like the following:

```typescript
const rawSchemaName = reference.$ref.split('/').pop() ?? typeId;
```

packages/cli/api-importers/v3-importer-tests/src/__test__/__snapshots__/v3-sdks/url-reference.json
```diff
@@ -3295,6 +3305,7 @@
                   "name": "Pet",
                   "typeId": "Pet",
                   "inline": false,
+                  "displayName": "a Pet",
```
amckinney, line 3308: Is this what we want? Should `displayName` actually be `Pet` here?

packages/cli/api-importers/v3-importer-tests/src/__test__/__snapshots__/v3-sdks/url-reference.json
```diff
@@ -3103,6 +3110,7 @@
                                           "name": "Tag",
                                           "typeId": "Tag",
                                           "inline": false,
+                                          "displayName": "Pet Tag",
```
amckinney, line 3113: Confirming this is intended -- the `displayName` should be `Pet Tag` instead of `Tag`?

packages/cli/api-importers/v3-importer-tests/src/__test__/__snapshots__/v3-sdks/discriminated-union-mapping.json
```diff
@@ -12,7 +12,8 @@
           "allParts": [],
           "packagePath": []
         },
-        "name": "RequestOne"
+        "name": "RequestOne",
+        "displayName": "A simple request title"
```
amckinney, line 16: Confirming this is what we want?

amckinney (approved):
Assuming the `displayName` is accurate, lgtm

PR #14727, opened 2026-04-07
feat(cli): thread FERN_RUN_ID through CLI startup and Sentry

Swimburger:
## Summary

- Generates a `FERN_RUN_ID` (UUIDv4) at CLI startup if one is not already set in the environment — inherits the value when set upstream (e.g. by `fern-api/actions/setup-cli`) so all steps in a workflow share a single correlation ID
- Tags Sentry errors with `fern_run_id` and `github_run_id` via dedicated helpers in `telemetry/sentryRunId.ts`, called once after `Sentry.init()`

## Changes

- `packages/cli/cli/src/fernRunId.ts` — `getOrCreateFernRunId()` (generate or inherit) and `getFernRunId()` (read-only)
- `packages/cli/cli/src/cli.ts` — calls `getOrCreateFernRunId()` at the top of `runCli()` before any command runs
- `packages/cli/cli/src/telemetry/sentryRunId.ts` — `setSentryFernRunIdTag()`, `setSentryGithubRunIdTag()`, and `setSentryRunIdTags()` (calls both)
- `packages/cli/cli/src/telemetry/SentryClient.ts` — calls `setSentryRunIdTags()` after `Sentry.init()`

## Testing

- [x] Compiles cleanly (`pnpm compile`)
- [x] No new test files needed — `fernRunId.ts` is a thin env-var wrapper; Sentry tag helpers delegate entirely to the Sentry SDK


FedeZara (approved):
Thank you! We might want to include this logic also in CLI v2: a correlation ID is already set up, but we are not reading ENV variables for it, so the logic should slightly change.

PR #14748, opened 2026-04-08
chore(cli): eagerly resolve TelemetryClient distinctId via async factory

FedeZara:
## Description

Part of the CLI error classification effort — see #14749 for full context.

Makes telemetry synchronous in **CLI v2** by eagerly resolving the PostHog `distinctId` at startup via an async factory, so that `Context.telemetry` becomes a synchronous `TelemetryClient`. This is the CLI v2 counterpart to #14747 (CLI v1).

## Why

Same motivation as #14747: the new error system (#14749) introduces a `reportError` function called from `failWithoutThrowing` and top-level catch blocks. This function needs synchronous access to telemetry to capture Sentry exceptions with PostHog correlation. Previously `Context.telemetry` was a `Promise<TelemetryClient>`, which forced all error handlers to be async. Moving the async resolution to startup keeps the hot path simple and avoids changing the `TaskContext` interface.

## Changes Made

- Added `TelemetryClient.create()` async factory that resolves `distinctId` eagerly
- `Context.telemetry` is now a synchronous `TelemetryClient` (no more `Promise<TelemetryClient>`)
- Updated all test helpers to use the new factory (`TelemetryClient.create()` or direct construction)
- Adjusted CLI v2 `withContext` to `await` telemetry creation at startup

## Testing

- [x] Updated TelemetryClient tests for new factory pattern
- [x] Updated test context helpers

PR #14747, opened 2026-04-08
chore(cli): make instrumentPostHogEvent synchronous

FedeZara:
## Description

Part of the CLI error classification effort — see #14749 for full context.

Makes `instrumentPostHogEvent` synchronous in **CLI v1** by removing its `async` qualifier and dropping unnecessary `await` calls at all call sites. This is the CLI v1 counterpart to #14748 (CLI v2).

## Why

The new error system (#14749) needs to call telemetry from `failWithoutThrowing` — a synchronous method on the `TaskContext` interface. If `instrumentPostHogEvent` remains async, every `failWithoutThrowing` call site would need to become async too, rippling changes through the entire `TaskContext` interface. Making telemetry synchronous avoids that: PostHog's `capture()` is fire-and-forget anyway (it buffers locally and flushes in the background), so the `async` was always unnecessary.

## Changes Made

- Changed `instrumentPostHogEvent` from `async` to synchronous in `TaskContext` interface, `CliContext`, `TaskContextImpl`, `TaskContextAdapter`, and test helpers
- Removed `await` from all call sites (21 files)

## Testing

- [x] Existing tests pass (behavioral no-op)

PR #14746, opened 2026-04-08
chore(cli): rename FernCliError to TaskAbortSignal

FedeZara:
## Description

Part of the CLI error classification effort — see #14749 for full context.

Renames `FernCliError` to `TaskAbortSignal` to better reflect its purpose: it is a control-flow signal used to abort a task after an error has already been logged, not an error class itself. This clears the naming space for the `CliError` class introduced in #14749.

## Changes Made

- Renamed `FernCliError` class to `TaskAbortSignal` across all usages
- Deleted `packages/cli/task-context/src/FernCliError.ts`, created `packages/cli/task-context/src/TaskAbortSignal.ts`
- Updated all `import { FernCliError }` → `import { TaskAbortSignal }` and `instanceof FernCliError` → `instanceof TaskAbortSignal` across CLI v1, CLI v2, seed, and test helpers

## Context

The old `FernCliError` was thrown by `failAndThrow()` purely as a sentinel to unwind the call stack — it carried no error metadata. Renaming it to `TaskAbortSignal` makes this intent explicit and avoids confusion with the new `CliError` class that carries typed error codes for Sentry routing.

## Testing

- [x] Existing tests pass (rename is mechanical)

PR #14749, opened 2026-04-08
feat(cli): introduce structured CliError with typed error codes and Sentry routing

FedeZara:
## Description

Introduces a structured `CliError` class in `@fern-api/task-context` with typed error codes and automatic Sentry routing. This replaces the ad-hoc `new Error(...)` pattern throughout the CLI with a system that categorizes errors into user-facing codes (e.g. `CONFIG_ERROR`, `AUTH_ERROR`) vs internal errors that should be reported to Sentry. It also simplifies error handling by removing redundant error classes and unifying the ones we already have.

> **Recommended review approach:** commit-by-commit. Each commit is self-contained and builds on the previous one.

### Prerequisite PRs

- #14746 — rename `FernCliError` to `TaskAbortSignal` (clears the naming space)
- #14747 — make CLI v1 telemetry synchronous (so `failWithoutThrowing` can report errors)
- #14748 — make CLI v2 telemetry synchronous (same, for CLI v2)

### Follow-up PRs

- #14750 — temporarily disable Sentry for unclassified errors (safety net while migrating)
- #14753 — migrate `@fern-api/cli` package (example of a package migration PR)
- ~34 more package migration PRs (one per package, converting `new Error(...)` → `new CliError(...)` and adding error codes to `failAndThrow` and `failWithoutThrowing` calls)
- #14752 — re-enable Sentry for unclassified errors (merge last, after all migrations)

### Two ways to trigger and track errors

There are two paths through which errors are captured and reported:

1. **Throwing `CliError` directly.** Code anywhere in the CLI can `throw new CliError({ message, code })`. The top-level catch handler in each CLI entry point (CLI v1's `runCli` catch in `cli.ts`, CLI v2's `withContext` catch in `withContext.ts`) intercepts it, resolves the error code, and routes it to Sentry and/or PostHog.

2. **Calling `failAndThrow` / `failWithoutThrowing` on the task context.** These methods log the error, resolve the code (explicit override > `CliError.code` > fallback), and report to Sentry if the code is Sentry-reportable. `failAndThrow` then throws a `TaskAbortSignal` to unwind the stack; `failWithoutThrowing` marks the task as failed and returns. This is the preferred path when the caller needs control over what happens after the error — for example, to continue processing other tasks or to clean up resources before aborting. It also adapts naturally to the existing error-handling patterns already used throughout the codebase.

Both paths converge on the same code-resolution logic (`resolveErrorCode`) and Sentry-routing rules (`shouldReportToSentry`), so tracking is consistent regardless of which path is used.

### Default behavior: unclassified errors are internal

Any error that doesn't carry a `CliError.Code` (i.e. a plain `new Error(...)`) is treated as `INTERNAL_ERROR` at the top-level catch boundaries — and therefore reported to Sentry. The assumption is that if nobody explicitly categorized an error as user-facing, it's likely an internal bug.

**Temporary exception:** the follow-up PR #14750 temporarily downgrades this so unclassified errors skip Sentry during the migration period, to avoid noise from the ~34 packages that haven't been migrated yet. Once all migrations land, #14752 re-enables it.

### Error code taxonomy

Errors are classified into 12 typed codes:

| Code | Sentry? | Description |
|------|---------|-------------|
| `INTERNAL_ERROR` | Yes | Unexpected bugs — should be investigated |
| `RESOLUTION_ERROR` | Yes | Type/reference resolution failures (likely IR bugs) |
| `IR_CONVERSION_ERROR` | Yes | IR generation failures |
| `CONTAINER_ERROR` | Yes | Docker container failures |
| `VERSION_ERROR` | Yes | Version parsing/compatibility issues |
| `PARSE_ERROR` | No | Malformed user input (YAML, OpenAPI, etc.) |
| `ENVIRONMENT_ERROR` | No | Missing env vars, wrong Node version, etc. |
| `REFERENCE_ERROR` | No | Dangling references in user config |
| `VALIDATION_ERROR` | No | Schema/rule validation failures |
| `NETWORK_ERROR` | No | HTTP failures, timeouts |
| `AUTH_ERROR` | No | Authentication/authorization issues |
| `CONFIG_ERROR` | No | Invalid generators.yml, fern.config.json, etc. |

Only the first 5 codes are Sentry-reportable — they indicate bugs in Fern itself. The rest are user-actionable and would just create noise in Sentry.

### Simplifying error handling by removing redundant classes

- **Unified `CliError` across CLI v1 and v2.** CLI v2 had its own `CliError` class in `packages/cli/cli-v2/src/errors/CliError.ts`. This PR deletes it and makes both CLIs share the single `CliError` from `@fern-api/task-context`, ensuring consistent error codes and Sentry routing regardless of which CLI entry point is used.

- **Existing error classes now extend `CliError`.** `ValidationError`, `SourcedValidationError`, and `KeyringUnavailableError` now extend `CliError` with their appropriate codes (`VALIDATION_ERROR` and `AUTH_ERROR`). This means `resolveErrorCode()` handles them automatically — no special-casing needed in every catch block.

- **Removed `LoggableFernCliError`.** This was a wrapper that carried a log message alongside an error. With `CliError` now carrying a `message` field (it extends `Error`), the wrapper is redundant. All former `LoggableFernCliError` usages are replaced with `CliError`.

### `reportError` — single error reporting path (CLI v2)

CLI v2's `withContext.ts` previously had separate `shouldReportToSentry` and `extractErrorCode` functions. These are consolidated into a single `reportError(context, error, options?)` function that:
1. Skips `TaskAbortSignal` (already logged)
2. Resolves the error code via `resolveErrorCode()` (explicit override > `CliError.code` > fallback to `INTERNAL_ERROR`)
3. Reports to Sentry if the code is in `SENTRY_REPORTABLE_CODES`
4. Always reports to PostHog with the error code as a property

### Sentry tags

Error codes are now passed as Sentry tags (`errorCode`) on captured exceptions, making it possible to filter and alert on specific error categories in the Sentry dashboard.

## Commits (review in order)

1. **`add shared CliError class`** — introduces `CliError` in `@fern-api/task-context` with the code taxonomy, `shouldReportToSentry`, `resolveErrorCode`, and static factory methods
2. **`unify CLI v2 CliError`** — deletes CLI v2's local `CliError`, switches to shared one, introduces `reportError` in `withContext.ts`
3. **`clean up imports`** — mechanical import reordering for consistency
4. **`make error classes extend CliError`** — `ValidationError`, `SourcedValidationError`, `KeyringUnavailableError` now extend `CliError`
5. **`pass error code as Sentry tag`** — adds `errorCode` tag to `captureException` calls
6. **`remove LoggableFernCliError`** — replaces all usages with `CliError`, deletes the class

## Testing

- [x] Updated test helpers and mocks for new `CliError` / `MockTaskContext` signatures
- [x] Existing CLI v1 and v2 tests pass

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14749" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/cli/cli-v2/src/context/withContext.ts
```diff
@@ -26,29 +26,20 @@ export function withContext<T extends GlobalArgs>(
...
-                context.telemetry.captureException(error);
-            }
-            context.telemetry.sendLifecycleEvent({
-                command: context.info.command,
-                status: "error",
-                durationMs: Date.now() - startTime,
-                errorCode: extractErrorCode(error)
-            });
+            reportError(context, error);
             await context.telemetry.flush();
```
amckinney, line 43: Nice 👍

packages/cli/cli/src/cli-context/CliContext.ts
```diff
@@ -108,14 +119,35 @@ export class CliContext {
...
+
+        this.instrumentPostHogEvent({
+            command: process.argv.join(" "),
+            properties: {
+                failed: true,
+                source: "cli",
+                error,
+                errorCode: code
+            }
+        });
```
amckinney: This supersedes the handling that was previously in `packages/cli/cli/src/cli.ts`, right?

packages/cli/cli/src/cli-context/TaskContextImpl.ts
```diff
@@ -74,15 +82,41 @@ export class TaskContextImpl implements Startable<TaskContext>, Finishable, Task
...
+
+        this.instrumentPostHogEventImpl({
+            command: process.argv.join(" "),
+            properties: {
+                failed: true,
+                source: "task",
+                error,
+                errorCode: code
+            }
+        });
```
amckinney: Do we need to capture `task` events separately? We have the same `command` name, so I don't know how valuable this would be. It looks like we didn't have this before, right?

amckinney (approved):

packages/cli/cli/src/cli-context/TaskContextImpl.ts
FedeZara: I'm sending PostHog events also here because the error that will surface to the CLI main catch clause will be a generic TaskFailureSignal that is not being tracked anymore in PostHog. The behavior will change a little compared to the current, but we'll have better error categorization thanks to this change (before we would capture generic TaskFailureSignal in most of the cases)

FedeZara: Let me know if you agree on the change or you want this to be consistent with the past

packages/cli/cli/src/cli-context/CliContext.ts
FedeZara: Exactly, I moved it here to have a single handling place. The behavior should be the same as before, but I'll check with some testing.

FedeZara: To make it simple, it's now `resolveErrorCode` responsible of categorizing the error and understand what we should do with it.

packages/cli/task-context/src/CliError.ts
```diff
@@ -0,0 +1,111 @@
+export const CliErrorCode = {
```
Swimburger: one thing we often do is create a namespace with the same name as the class, then exported types from it.
So you could do
```ts
import { CliError } from "...";

const code = CliError.Code.ConfigError;
```

Example from Claude on how to do this.

```ts
export class CliError extends Error {
    public readonly code: CliError.Code;

    public constructor(message: string, code: CliError.Code) {
        super(message);
        this.name = "CliError";
        this.code = code;
    }
}

export namespace CliError {
    export type Code = (typeof Code)[keyof typeof Code];
    export const Code = {
        ConfigError: "ConfigError",
        NetworkError: "NetworkError",
        AuthError: "AuthError",
    } as const;
}
```

Swimburger (changes requested):

FedeZara:
* **#14750** <a href="https://app.graphite.com/github/pr/fern-api/fern/14750?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#14752](https://github.com/fern-api/fern/pull/14752) <a href="https://app.graphite.com/github/pr/fern-api/fern/14752?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#14753](https://github.com/fern-api/fern/pull/14753) <a href="https://app.graphite.com/github/pr/fern-api/fern/14753?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#14749** <a href="https://app.graphite.com/github/pr/fern-api/fern/14749?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/fern-api/fern/14749?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14748** <a href="https://app.graphite.com/github/pr/fern-api/fern/14748?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14747** <a href="https://app.graphite.com/github/pr/fern-api/fern/14747?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14746** <a href="https://app.graphite.com/github/pr/fern-api/fern/14746?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Swimburger (approved):

FedeZara:
### Merge activity

* **Apr 14, 6:41 PM UTC**: @FedeZara merged this pull request with [Graphite](https://app.graphite.com/github/pr/fern-api/fern/14749).

Issue #7027, opened 2025-05-07
next-server process causes problems


cwcowell:
### Which Fern component?

Fern CLI

### How urgent is this?

P2 - Medium (Would be helpful)

### What's the issue?

macOS 15.4.1
node 20.12.2

The process `next-server (v15.3.0-canary.1)` starts when you run `fern docs dev`.  It has a few problems:

- it doesn’t always stop when you CTRL-C out of `fern docs dev`, so it's possible to get multiple instances of the process running
- it sometimes gets in a state where it permanently pegs the CPU and slaughters battery, to the extent that it’s not really usable unless the laptop is plugged in (e.g., 50% battery drain in 90 mins).
- it often doesn’t recompile pages as reliably as it does with the `—legacy` option. For example, sometimes I need to reload the page 3 or 4 times (or maybe just wait for awhile), and sometimes I need to restart the server completely, in order to see Markdown changes show up on the locally served pages.

<img width="740" alt="Image" src="https://github.com/user-attachments/assets/e4f3aa1a-3cf7-4005-819b-61182225dddf" />

### Fern CLI & Generator Versions

Fern CLI version 0.60.21

### Workaround

Run `fern docs dev —legacy` instead

### Are you interested in contributing a fix?

No

hkosova:
We're also experiencing this issue.

macOS 15.5
Node.js 20.18.0
Fern CLI 0.62.5 .. 0.65.1

chdeskur:
This should be resolved on Fern CLI v0.63.27 and above. If you shut down the open processes and ensure you are always upgraded above this version, next-server processes should no longer persist. Please open a new issue if you see this behavior confirmed in a version above v0.63.27!

PR #14751, opened 2026-04-08
fix(cli): kill entire process group when stopping fern docs dev

thesandlord:
## Description

Linear ticket: Closes FER-9564 (sub-ticket of FER-9539)

When `fern docs dev` is killed (Ctrl+C or `kill`), the Next.js standalone server's child worker processes (`next-server`) survive as orphaned zombies, consuming CPU and holding the port. Users must `kill -9` them manually. This is a regression of https://github.com/fern-api/fern/issues/7027.

**Root cause (two bugs):**
1. The cleanup function called `serverProcess.kill()` which only sends SIGTERM to the direct child `node` process. Next.js worker processes never receive a signal.
2. The SIGKILL fallback was gated on `!serverProcess.killed`, but execa sets `.killed = true` immediately when `.kill()` is called — not when the process actually exits. So SIGKILL was **never sent**.

## Changes Made

- Spawn the Next.js server with `detached: true` so it becomes a process group leader
- In cleanup, use `process.kill(-pid, "SIGTERM")` to signal the entire process group (server + all workers)
- After 2s, probe the group with signal 0; if still alive, `process.kill(-pid, "SIGKILL")` to force-kill
- Close the file watcher during cleanup to prevent resource leaks
- Removed the broken `!serverProcess.killed` gate that prevented SIGKILL from ever firing
- Replaced redundant `serverProcess.kill()` + `process.kill(-pid)` calls with a single process-group kill path

## Testing

- [ ] Unit tests added/updated — not practical for process lifecycle; requires manual verification
- [ ] Manual testing: run `fern docs dev`, kill with Ctrl+C, verify no orphaned `next-server` processes remain
- [ ] Manual testing: run `fern docs dev`, `kill <pid>` from another terminal, verify cleanup

## Review checklist

- [ ] **`detached: true` changes Ctrl+C behavior**: With `detached`, the child group no longer receives SIGINT directly from the terminal. Instead, our SIGINT handler sends SIGTERM to the group. Verify that Next.js workers shut down cleanly on SIGTERM (an earlier iteration caused 100% CPU during the 2s SIGTERM→SIGKILL window — the SIGKILL fallback should now reliably fire, but worth confirming).
- [ ] **Windows: no direct-child fallback**: The old code called `serverProcess.kill()` as a fallback. The new code only uses `process.kill(-pid, ...)` which is POSIX-only. On Windows, both SIGTERM and SIGKILL to the group will throw (caught silently), and **no kill happens at all**. If Windows support matters, a `serverProcess.kill()` fallback should be added in the catch blocks.
- [ ] **2s SIGKILL delay**: Workers may spin at high CPU for up to 2 seconds between SIGTERM and SIGKILL. Consider whether a shorter timeout (e.g. 500ms) is more appropriate.

Link to Devin session: https://app.devin.ai/sessions/d013e1cb6c3b40e28113067ff849a556
Requested by: @thesandlord

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14751" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/cli/docs-preview/src/runAppPreviewServer.ts
+++ b/packages/cli/docs-preview/src/runAppPreviewServer.ts
@@ -937,43 +937,34 @@ export async function runAppPreviewServer({
         // Close file watcher to prevent resource leaks
         watcher.close();
 
-        if (serverProcess != null && !serverProcess.killed) {
-            const pid = serverProcess.pid;
+        const pid = serverProcess?.pid;
+        if (pid != null) {
             context.logger.debug(`Killing server process tree with PID: ${pid}`);
+
+            // Send SIGTERM to the entire process group (detached: true makes
+            // the server a process group leader, so -pid targets it and all
+            // its child workers).
             try {
-                // Kill the entire process group to ensure all child processes
-                // (including Next.js workers) are terminated.
-                // The negative PID targets the process group.
-                if (pid != null) {
-                    try {
-                        process.kill(-pid, "SIGTERM");
-                    } catch (error) {
-                        context.logger.debug(`Failed to kill process group (may already be dead): ${error}`);
-                    }
-                }
-                serverProcess.kill();
-                setTimeout(() => {
-                    if (serverProcess != null && !serverProcess.killed) {
-                        context.logger.debug(`Force killing server process tree with PID: ${pid}`);
-                        try {
-                            if (pid != null) {
-                                try {
-                                    process.kill(-pid, "SIGKILL");
-                                } catch (error) {
-                                    context.logger.debug(
-                                        `Failed to kill process group (may already be dead): ${error}`
-                                    );
-                                }
-                            }
-                            serverProcess.kill("SIGKILL");
-                        } catch (err) {
-                            context.logger.error(`Failed to force kill server process: ${err}`);
-                        }
-                    }
-                }, 2000);
-            } catch (err) {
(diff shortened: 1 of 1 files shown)


amckinney (approved):

fern-support (approved):

packages/cli/docs-preview/src/runAppPreviewServer.ts
```diff
@@ -408,6 +409,54 @@ function cleanFernDocsCacheSync(bundleRoot: string, context: TaskContext): void
...
+ * Finds and kills any processes still listening on the given port.
+ * This is a best-effort fallback to clean up zombie Next.js worker
+ * processes that may survive after the main server process is killed.
+ */
+function killProcessesOnPort(targetPort: number, context: TaskContext): void {
+    context.logger.debug(`Killing any leftover processes on port ${targetPort}`);
+    const isWindows = process.platform === "win32";
+    const command = isWindows
+        ? `netstat -ano | findstr :${targetPort} | findstr LISTENING`
+        : `lsof -ti tcp:${targetPort}`;
```
amckinney, line 422: Confirming this is POSIX compliant / the same flags will work across both Linux and MacOS?

thesandlord: yup!

PR #14755, opened 2026-04-08
fix(cli): resolve absolute links within version/product context in broken-link checker

Ryan-Amirthan:
## Description
Linear ticket: Closes FER-9584

Fixes broken-link checker incorrectly flagging absolute links in versioned and product-based docs as broken.

## Problem

When you have versioned (or product-based) docs and write an absolute link like `/about` in a page, the Fern docs runtime automatically prefixes it with the current version/product slug (e.g., `/v1/about`). The link checker didn't know about this behavior, so it only checked if the unprefixed `/about` existed, which failed for non-default versions/products.

## Changes Made

- **`check-if-pathname-exists.ts`**: Added `versionSlugs` and `productSlugs` parameters. For absolute links, checks if `<contextSlug>/<pathname>` exists in the slug set, where context slugs include both version and product prefixes.
- **`valid-markdown-link.ts`**: Collects version slugs via `collector.getVersionNodes()` and product slugs via `collector.getProductNodes()` (filtered to internal products only), then passes both to `checkIfPathnameExists`.
- **`versions.yml`**: Added changelog entry for v4.65.3.
- Added ETE test fixtures for both versioned docs and product-based docs, with corresponding test cases.

## Testing

- [x] ETE test added for versioned docs (absolute links resolving within version context)
- [x] ETE test added for product-based docs (absolute links resolving within product context — two products: default "Docs" and non-default "SDKs" with shared pages containing absolute links)
- [x] Manual testing on [ryanstep-config test repo](https://github.com/fern-api/ryanstep-config/pull/39) — false positives gone, true broken links still detected

CleanShot 2026-04-09 at 20 55 34@2x
(error on the right terminal is intended, the 6 duplicate errors are not intended due to the version prefixing)

## ⚠️ Human Review Checklist

- [ ] Consider whether nested product+version scenarios (`<productSlug>/<versionSlug>/<page>`) need additional handling — currently version and product slugs are checked independently, not as combined prefixes
- [ ] Confirm `isInternalProductNode` filter correctly excludes external product nodes (`productLink` type) that lack a `slug` property
- [ ] The context slug matching logic iterates all context slugs for each file slug — verify this doesn't produce false negatives for files that belong to multiple contexts

Link to Devin session: https://app.devin.ai/sessions/19e0cecae19c4e41a500df39d9f158ac
Requested by: @Ryan-Amirthan

### DIFF
--- a/packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
+++ b/packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
@@ -22,7 +22,8 @@ export async function checkIfPathnameExists({
     pageSlugs,
     absoluteFilePathsToSlugs,
     redirects = [],
-    baseUrl
+    baseUrl,
+    versionSlugs = []
 }: {
     pathname: string;
     markdown: boolean;
@@ -39,6 +40,7 @@ export async function checkIfPathnameExists({
         domain: string;
         basePath?: string;
     };
+    versionSlugs?: string[];
 }): Promise<true | string[]> {
     pathname = removeTrailingSlash(pathname);
     const slugs = absoluteFilepath != null ? (absoluteFilePathsToSlugs.get(absoluteFilepath) ?? []) : [];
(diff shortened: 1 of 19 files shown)


packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
```diff
@@ -73,6 +75,22 @@ export async function checkIfPathnameExists({
             return true;
         }
 
+        // For versioned pages, absolute links like `/about` resolve within the current version context.
```
fern-support: Do we also need to worry about product slugs in the same way?

fern-support (approved):

packages/cli/yaml/docs-validator/src/rules/valid-markdown-link/check-if-pathname-exists.ts
Ryan-Amirthan: can you add ETE test for product-based docs specifically since we have some for versions already?

PR #14758, opened 2026-04-08
fix(cli): add log rotation for `fern docs dev` debug logs

thesandlord:
## Description

Closes FER-9568

`fern docs dev` writes debug logs to `~/.fern/logs/` but never cleans them up, causing unbounded disk usage over time. This adds automatic log rotation with a 100 MB directory size cap.

## Changes Made

- Added `enforceLogSizeLimit()` private method in `DebugLogger` that runs fire-and-forget at the end of `initialize()`
- Scans the logs directory, sorts files by mtime (oldest first), and deletes until total size is ≤ 100 MB
- Skips the current session's log file to avoid deleting active logs
- Added optional `consoleLogger` callback parameter to `DebugLogger.initialize()` so rotation status is logged to the console (debug when under cap, info when rotating) — wired to `context.logger` in `runAppPreviewServer.ts`
- All errors are silently caught to avoid breaking the main `fern docs dev` flow
- Added `versions.yml` entry (4.63.2, fix)

## Review Checklist

- [ ] **`file.fullPath === this.logFilePath` comparison**: `this.logFilePath` is an `AbsoluteFilePath` (branded string), while `file.fullPath` is built with `path.join()` (plain string). Works at runtime since `AbsoluteFilePath` is `string & { __AbsoluteFilePath: void }`, but verify paths are normalized consistently.
- [x] Cleanup runs fire-and-forget (`void`) — does not block startup
- [x] Cleanup only runs at startup — a single very long session could still exceed 100 MB. Acceptable tradeoff.

## Testing

- [ ] Unit tests added/updated — **not added**; cleanup logic is defensive (all errors caught) and straightforward
- [x] Lint passes (`biome check`)
- [x] Verified console output shows rotation status with `--log-level=debug`

Link to Devin session: https://app.devin.ai/sessions/0212b2e6da7543248d179fb0d4f0450a
Requested by: @thesandlord

### DIFF
--- a/packages/cli/docs-preview/src/DebugLogger.ts
+++ b/packages/cli/docs-preview/src/DebugLogger.ts
@@ -324,6 +324,19 @@ export class DebugLogger {
                 return;
             }
 
+            const totalSizeMB = Math.round((totalSize / 1024 / 1024) * 100) / 100;
+            await this.writeEntry({
+                timestamp: new Date().toISOString(),
+                source: getCliSource(),
+                level: "info",
+                eventType: "log_rotation",
+                data: {
+                    message: `Rotating logs: total size ${totalSizeMB} MB exceeds ${MAX_LOGS_DIR_SIZE_BYTES / 1024 / 1024} MB cap`,
+                    totalSizeMB,
+                    fileCount: fileInfos.length
+                }
+            });
+
             // Sort oldest first so we delete the oldest logs first
             fileInfos.sort((a, b) => a.mtimeMs - b.mtimeMs);
 
--- a/packages/cli/docs-preview/src/DebugLogger.ts
+++ b/packages/cli/docs-preview/src/DebugLogger.ts
@@ -162,8 +162,8 @@ export class DebugLogger {
         await writeFile(this.logFilePath, header, "utf-8");
         this.initialized = true;
 
-        // Enforce the 100 MB cap on the logs directory
-        await this.enforceLogSizeLimit(logsDir);
+        // Enforce the 100 MB cap on the logs directory (fire-and-forget)
+        void this.enforceLogSizeLimit(logsDir);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/cli/docs-preview/src/DebugLogger.ts
```diff
@@ -158,6 +161,9 @@ export class DebugLogger {
 
         await writeFile(this.logFilePath, header, "utf-8");
         this.initialized = true;
+
+        // Enforce the 100 MB cap on the logs directory
+        await this.enforceLogSizeLimit(logsDir);
```
thesandlord: dont await, run in bg

sbawabe (approved):

packages/cli/docs-preview/src/DebugLogger.ts
```diff
@@ -293,6 +299,54 @@ export class DebugLogger {
         await this.logCliMetric(event);
     }
 
+    /**
+     * Delete oldest log files until total directory size is within the cap.
+     */
+    private async enforceLogSizeLimit(logsDir: AbsoluteFilePath): Promise<void> {
+        try {
```
thesandlord: add a debug log here "rotating logs" or something like that

packages/cli/docs-preview/src/DebugLogger.ts
```diff
@@ -293,6 +299,67 @@ export class DebugLogger {
...
+                    const stats = await stat(fullPath);
+                    fileInfos.push({ name, fullPath, size: stats.size, mtimeMs: stats.mtimeMs });
+                } catch {
+                    // File may have been removed between readdir and stat
+                }
+            }
+
+            let totalSize = fileInfos.reduce((sum, f) => sum + f.size, 0);
+            if (totalSize <= MAX_LOGS_DIR_SIZE_BYTES) {
+                return;
```
thesandlord: add a debug log here saying "does not exceed"

PR #14857, opened 2026-04-09
chore(ci): Add Docker image publishing for fern-cli

pgragg:
## Description

Adds Docker image publishing to the CLI release workflow so that `fernapi/fern` and `fernapi/fern-dev` images are available on Docker Hub. This enables customer GitHub Actions workflows to use `container: image: fernapi/fern:latest` instead of running `npm install -g fern-api` on every run, saving ~10s per workflow execution.

Companion PR: fern-api/fern-platform#9604 (updates customer workflow templates to use the container-based approach).

## Changes Made

- **`packages/cli/cli/Dockerfile`**: New Dockerfile based on `node:22-slim` with git and the Fern CLI pre-installed. Uses build args (`FERN_CLI_PACKAGE`, `FERN_CLI_VERSION`) to support both prod and dev CLI variants. `FERN_CLI_VERSION` is a **required** build arg with no default — the build fails explicitly if not supplied, ensuring hermetic builds. `ENTRYPOINT []` allows GitHub Actions to run arbitrary commands in the container. Placed adjacent to `versions.yml` for consistency with the SDK generator setup.

- **`.github/workflows/publish-cli-docker.yml`**: New reusable `workflow_call` workflow that builds and pushes Docker images to Docker Hub. Accepts `version` (string) and `is_dev` (boolean) inputs. Tags images with both the specific version and `latest`. Uses Docker Buildx with GHA cache.

- **`.github/workflows/publish-cli.yml`**: Hooks the new Docker publish workflow into every CLI release path:
  - `dev` → publishes `fernapi/fern-dev` at the git version
  - `prod` → publishes both `fernapi/fern` and `fernapi/fern-dev` at the release version
  - `manual` → publishes both images at the input version
  - Adds job `outputs` to `dev` and `prod` jobs to pass the resolved version to downstream Docker jobs
  - `prod` job's `get_cli_version` step uses `yq '.[0].version'` for robust version extraction from `versions.yml`

Addressed review feedback:
- Removed `latest` default from `FERN_CLI_VERSION` Dockerfile ARG. The build now fails explicitly if the version is not supplied, ensuring the installed CLI version is always in lock-step with the Docker image tag.
- Replaced fragile `head | grep | awk` version extraction in the `prod` job's `get_cli_version` step with `yq '.[0].version'`. The old `grep "version:"` pattern could match `irVersion:` lines.

## Human Review Checklist

- [ ] **`dev` job refactor**: The version computation was split from the publish step into a separate `get_git_version` step so it can be exposed as a job output. Verify the publish step still correctly consumes the version via `${{ steps.get_git_version.outputs.git_version }}`.
- [ ] **`secrets: inherit`**: All Docker jobs use `secrets: inherit` to pass `FERN_API_DOCKERHUB_PASSWORD` to the reusable workflow. Confirm this secret is available in the repo (it's already used by `seed-dockers.yml`).
- [ ] **Dockerfile `ENTRYPOINT []`**: Required for GitHub Actions `container:` directive to work—without it, Actions can't run arbitrary shell commands.
- [ ] **`yq` usage**: The `prod` job now uses `yq '.[0].version'` to extract the CLI version. `yq` is pre-installed on `ubuntu-latest` runners. Note: the `bulk-update-cli` job still uses the old `grep`/`awk` pattern (tracked in #14860).

## Testing

- [ ] Manual testing not possible for workflow changes, but pushed up an image from local https://hub.docker.com/repository/docker/fernapi/fern-cli-dev/tags
- [ ] Docker Hub credentials (`fernapi` / `FERN_API_DOCKERHUB_PASSWORD`) are consistent with existing usage in `seed-dockers.yml`

Link to Devin session: https://app.devin.ai/sessions/9750ac886908477394299692162b2aa8
Requested by: @pgragg

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14857" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/cli/cli/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:22-slim
+
+ARG FERN_CLI_PACKAGE=fern-api
+ARG FERN_CLI_VERSION=latest
```
amckinney: We'll want this version to be consistent with the docker tag that we assign the image. That way the CLI release is in lock-step with the Docker image release. Otherwise, this docker image will _always_ just pull in the latest image from NPM (i.e. it's not hermetic).

packages/cli/cli/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:22-slim
```
amckinney, line 1: Can we put this in `./packages/cli/cli/Dockerfile` so that its consistent with our SDK generator setup (i.e. adjacent to the `versions.yml` file?

amckinney, line 1: Let's name the image `fernapi/fern` instead of `fernapi/fern-cli` -- the `-cli` is implied/reundant.

packages/cli/cli/Dockerfile
amckinney: We'll likely want to read this version from the environment or as a docker arg in the build command in CI. It should be the same version we use for the CLI release (i.e. the one defined in `versions.yml`).

pgragg: Devin, for safety let's remove the default and explicitly require a version. Fail the Docker image build if it is not supplied.

PR #14863, opened 2026-04-09
fix(csharp): skip literal properties in snippet object initializers when generateLiterals is enabled

Swimburger:
## Description

Refs #19261

When the `generate-literals` config option is enabled (or via the deprecated `experimental-readonly-constants`), the C# generator emits `TypeLiteral` readonly structs for literal properties instead of plain strings. However, the snippet generators (used for mock server tests and code examples) were still emitting string assignments like `Type = "session_settings"` for these properties, causing **CS0029 compilation errors** because `TypeLiteral` has implicit conversion *to* `string` but not *from* `string`.

This was reported by HumeAI whose C# SDK CI is blocked ([HumeAI/hume-dotnet-sdk#53](https://github.com/HumeAI/hume-dotnet-sdk/pull/53)).

## Changes Made

- **`GeneratorContext.isLiteralValue`** (new): Added a boolean convenience method wrapping `getLiteralValue(...) != null` so call sites read more clearly.
- **`ObjectGenerator.doGenerateSnippet`**: When `generateLiterals` is enabled, collect wire values of all literal properties (including `extendedProperties`) from the object declaration via `isLiteralValue`, then filter them out of the object initializer. Also skip them in the "default values for missing required properties" loop.
- **`WrappedRequestGenerator.doGenerateSnippet`**: Same approach for inlined request body properties — collect literal property wire values from the endpoint's request body declaration (including `extendedProperties`) and skip them during example snippet generation.
- **Seed output regenerated** for `literal:readonly-constants` fixture, demonstrating the fix across mock server tests, code snippets, reference docs, and client XML doc examples.

In both cases, omitting the property is correct because the generated field already has `= new()` as its default initializer, which sets the only valid literal value.

### Seed diff highlights

The `literal:readonly-constants` seed output now omits literal properties from object initializers. For example in `SendTest.cs`:
```diff
 new SendLiteralsInlinedRequest
 {
-    Prompt = "You are a helpful assistant",
     Context = "You're super wise",
     Query = "query",
     Temperature = 1.1,
-    Stream = false,
-    AliasedContext = "You're super wise",
     MaybeContext = "You're super wise",
-    ObjectWithLiteral = new ATopLevelLiteral
-    {
-        NestedLiteral = new ANestedLiteral { MyLiteral = "How super cool" },
-    },
+    ObjectWithLiteral = new ATopLevelLiteral { NestedLiteral = new ANestedLiteral() },
 }
```

## Testing

- [x] Manual lint/typecheck passes (`pnpm run check`)
- [x] Seed output regenerated for `literal:readonly-constants` and `websocket:with-websockets` fixtures

### 🔍 Key areas for human review

1. **Recursive all-literal objects**: This PR skips direct literal-typed properties but still includes objects whose subtree is entirely literal (e.g. `ObjectWithLiteral = new ATopLevelLiteral { NestedLiteral = new ANestedLiteral() }`). These are required properties without `= new()` defaults, so the compiler needs them. A follow-up could detect "all-literal" object trees and either skip them or give them default initializers at the generator level.
2. **Completeness**: Are there other snippet generation paths beyond `ObjectGenerator` and `WrappedRequestGenerator` that could emit string assignments for `TypeLiteral` properties?
3. **Seed output correctness**: Verify the seed diff — literal properties (`Prompt`, `Stream`, `AliasedContext`, `Literal1`, `Literal2`, `MyLiteral`, `Context`, `Ending`) are omitted while non-literal properties (`Temperature`, `Query`, `MaybeContext`, `StrProp`) are preserved.

Link to Devin session: https://app.devin.ai/sessions/d14d656ef74644d2b22a573ee1c96caa
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/14863" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/generators/csharp/model/src/object/ObjectGenerator.ts
+++ b/generators/csharp/model/src/object/ObjectGenerator.ts
@@ -183,7 +183,7 @@ export class ObjectGenerator extends FileGenerator<CSharpFile, ModelGeneratorCon
                 ...(this.objectDeclaration.extendedProperties ?? [])
             ];
             for (const prop of allProps) {
-                if (this.context.getLiteralValue(prop.valueType) != null) {
+                if (this.context.isLiteralValue(prop.valueType)) {
                     literalPropertyWireValues.add(getWireValue(prop.name));
                 }
             }
--- a/generators/csharp/model/src/object/ObjectGenerator.ts
+++ b/generators/csharp/model/src/object/ObjectGenerator.ts
@@ -173,19 +173,37 @@ export class ObjectGenerator extends FileGenerator<CSharpFile, ModelGeneratorCon
         exampleObject: ExampleObjectType;
         parseDatetimes: boolean;
     }): ast.CodeBlock {
-        const args = exampleObject.properties.map((exampleProperty) => {
-            const propertyName = this.getPropertyName({
-                className: this.classReference.name,
-                objectProperty: exampleProperty.name
+        // When generateLiterals is enabled, collect wire values of literal properties so we
+        // can skip them in the object initializer. Their default `= new()` already sets the
+        // correct value and assigning a plain string would cause a CS0029 compilation error.
+        const literalPropertyWireValues = new Set<string>();
+        if (this.generation.settings.generateLiterals) {
+            const allProps = [
+                ...this.objectDeclaration.properties,
+                ...(this.objectDeclaration.extendedProperties ?? [])
+            ];
+            for (const prop of allProps) {
+                if (this.context.getLiteralValue(prop.valueType) != null) {
+                    literalPropertyWireValues.add(getWireValue(prop.name));
+                }
+            }
+        }
+
+        const args = exampleObject.properties
+            .filter((exampleProperty) => !literalPropertyWireValues.has(getWireValue(exampleProperty.name)))
+            .map((exampleProperty) => {
+                const propertyName = this.getPropertyName({
+                    className: this.classReference.name,
+                    objectProperty: exampleProperty.name
+                });
+                const assignment = this.exampleGenerator.getSnippetForTypeReference({
+                    exampleTypeReference: exampleProperty.value,
+                    parseDatetimes
+                });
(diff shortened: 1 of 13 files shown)


generators/csharp/model/src/object/ObjectGenerator.ts
```diff
@@ -173,19 +173,37 @@ export class ObjectGenerator extends FileGenerator<CSharpFile, ModelGeneratorCon
...
+        // can skip them in the object initializer. Their default `= new()` already sets the
+        // correct value and assigning a plain string would cause a CS0029 compilation error.
+        const literalPropertyWireValues = new Set<string>();
+        if (this.generation.settings.generateLiterals) {
+            const allProps = [
+                ...this.objectDeclaration.properties,
+                ...(this.objectDeclaration.extendedProperties ?? [])
+            ];
+            for (const prop of allProps) {
+                if (this.context.getLiteralValue(prop.valueType) != null) {
```
Swimburger: should we use a different helper method, isLiteralValue(...)

seed/csharp-sdk/literal/readonly-constants/src/SeedLiteral.Test/Unit/MockServer/Inlined/SendTest.cs
```diff
@@ -55,17 +55,11 @@ public async Task MockServerTest_1()
...
                 Query = "query",
                 Temperature = 1.1,
-                Stream = false,
-                AliasedContext = "You're super wise",
                 MaybeContext = "You're super wise",
-                ObjectWithLiteral = new ATopLevelLiteral
-                {
-                    NestedLiteral = new ANestedLiteral { MyLiteral = "How super cool" },
-                },
+                ObjectWithLiteral = new ATopLevelLiteral { NestedLiteral = new ANestedLiteral() },
```
Swimburger, line 62: why do we even need to provide this literal?

fern-support (approved):

PR #14940, opened 2026-04-13
fix(typescript): propagate license config to publish mode in package.json

iamnamananand996:
## Description

Linear ticket: Closes [FER-9759](https://linear.app/buildwithfern/issue/FER-9759/phonic-sdks-ensure-build-files-reference-generated-license-file)

The `publish` output mode in the TypeScript v2 generator was hardcoding `license: undefined` in the constructed `NpmPackage`, causing the generated `package.json` to omit the `license` field even when the user configured one in `generators.yml`. The `github` output mode already extracted the license correctly — `publish` was the only mode that dropped it.

Additionally, the `github` mode's inline `_visit` logic was refactored to use the shared `licenseFromLicenseConfig` helper for consistency (no behavioral change).

## Changes Made

- **`constructNpmPackage.ts`**: Changed the `publish` output mode from `license: undefined` to `license: licenseFromLicenseConfig(generatorConfig.license)`. Also refactored the `github` mode to use the same `licenseFromLicenseConfig` helper instead of an inline `_visit` call (functionally identical, DRY cleanup).
- **`TestRunner.ts`**: Fixed `extractLicenseInfo` to pass through basic string licenses (e.g. `"MIT"`, `"Apache-2.0"`) — previously it only handled custom object licenses and silently dropped strings.
- **`seed.yml`**: Added `license: MIT` to the `simple-api` bundle fixture to exercise the license config path in seed tests.
- Updated the `simple-api/bundle` seed snapshot — `package.json` now includes `"license": "MIT"`.
- Added unreleased changelog entry for the TypeScript SDK generator.

> **Note:** An earlier revision also added `LICENSE` to the `files` array in `BundledTypescriptProject.ts`, but this was [reverted per review feedback](https://github.com/fern-api/fern/pull/14940#discussion_r3081138809) — npm automatically includes `LICENSE` files regardless of the `files` array.

## Review Checklist

- [ ] Verify the `publish` mode license change is intentional — was `license: undefined` a deliberate choice or an oversight? (The `github` mode already sets it, and `constructNpmPackageFromArgs` also passes it through, suggesting oversight.)
- [ ] The changelog entry (`fix-license-in-build-files.yml`) still references "the LICENSE file in the `files` array" which was reverted — confirm whether the summary should be narrowed to only mention the `license` field propagation fix.

## Testing

- [x] `pnpm run check` passes
- [x] Seed test `simple-api/bundle` passes
- [x] Generated `package.json` now includes `"license": "MIT"` when license config is set

Link to Devin session: https://app.devin.ai/sessions/6bef5e47a63140aab425fbe9f12f26e8

### DIFF
--- a/generators/typescript-v2/browser-compatible-base/src/utils/constructNpmPackage.ts
+++ b/generators/typescript-v2/browser-compatible-base/src/utils/constructNpmPackage.ts
@@ -48,7 +48,7 @@ export function constructNpmPackage({
                     registryUrl: outputMode.registriesV2.npm.registryUrl,
                     token: outputMode.registriesV2.npm.token
                 },
-                license: undefined,
+                license: licenseFromLicenseConfig(generatorConfig.license),
                 repoUrl: undefined
             };
         case "github":
--- a/generators/typescript-v2/browser-compatible-base/src/utils/constructNpmPackage.ts
+++ b/generators/typescript-v2/browser-compatible-base/src/utils/constructNpmPackage.ts
@@ -63,13 +63,7 @@ export function constructNpmPackage({
                 private: isPackagePrivate,
                 publishInfo: undefined,
                 repoUrl: getRepoUrlFromUrl(outputMode.repoUrl),
-                license: generatorConfig.license?._visit({
-                    basic: (basic) => basic.id,
-                    custom: (custom) => `See ${custom.filename}`,
-                    _other: () => {
-                        return undefined;
-                    }
-                })
+                license: licenseFromLicenseConfig(generatorConfig.license)
             };
         default:
             throw new Error(`Encountered unknown output mode: ${outputMode}`);
--- a/generators/typescript/utils/commons/src/typescript-project/BundledTypescriptProject.ts
+++ b/generators/typescript/utils/commons/src/typescript-project/BundledTypescriptProject.ts
@@ -192,7 +192,8 @@ export * from "./${BundledTypescriptProject.TYPES_DIRECTORY}/${folder}.js";
             files: [
                 BundledTypescriptProject.DIST_DIRECTORY,
                 BundledTypescriptProject.TYPES_DIRECTORY,
-                ...this.getAllStubTypeFiles()
+                ...this.getAllStubTypeFiles(),
+                BundledTypescriptProject.LICENSE_FILENAME
             ],
             exports: {
                 ".": this.getExportsForBundle({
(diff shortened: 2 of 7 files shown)


generators/typescript/utils/commons/src/typescript-project/BundledTypescriptProject.ts
```diff
@@ -192,7 +192,8 @@ export * from "./${BundledTypescriptProject.TYPES_DIRECTORY}/${folder}.js";
             files: [
                 BundledTypescriptProject.DIST_DIRECTORY,
                 BundledTypescriptProject.TYPES_DIRECTORY,
-                ...this.getAllStubTypeFiles()
+                ...this.getAllStubTypeFiles(),
+                BundledTypescriptProject.LICENSE_FILENAME
```
Swimburger: license is always included in package.json => files, even if you don't list it
https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files

generators/typescript-v2/browser-compatible-base/src/utils/constructNpmPackage.ts
```diff
@@ -63,13 +63,7 @@ export function constructNpmPackage({
                 private: isPackagePrivate,
                 publishInfo: undefined,
                 repoUrl: getRepoUrlFromUrl(outputMode.repoUrl),
-                license: generatorConfig.license?._visit({
```
Swimburger, line 66: Is generatorConfig.license incorrectly typed? The `licenseFromLicenseConfig` extracts data from it very differently, it doesn't look at basic at all.

Swimburger: Maybe this needs a combination of the two

Swimburger (changes requested):

generators/typescript/utils/commons/src/typescript-project/BundledTypescriptProject.ts
iamnamananand996: Yes, this is the assumption I also have, it will always add `LICENSE` in `files` inside the `package.json`

iamnamananand996: yes please

generators/typescript-v2/browser-compatible-base/src/utils/constructNpmPackage.ts
iamnamananand996: hey @Swimburger, then it that case it is correct.

generators/typescript/utils/commons/src/typescript-project/BundledTypescriptProject.ts
iamnamananand996: I think removing `BundledTypescriptProject.LICENSE_FILENAME` and re-generating the seed `simple-api/bundle` is good enough.

Swimburger (approved):

PR #15026, opened 2026-04-15
feat(typescript): add queryBuilder pattern for query parameter serialization

Swimburger:
## Description

Refactors TypeScript query parameter serialization to use a **fluent `queryBuilder()` pattern**, matching the C# generator architecture. The generator now emits a chained builder (`core.url.queryBuilder().addMany(…).add(…, { style: "comma" }).build()`) so each parameter's serialization format (repeat vs comma) is decided at **code-gen time** rather than runtime via a config map.

OpenAPI query parameters with `explode: false` and `style: form` use `.add("tags", tags, { style: "comma" })` to produce comma-separated values (e.g., `tags=a,b,c`), while all other array parameters use the default repeat format (`tags=a&tags=b&tags=c`).

Each `.add()` call **eagerly serializes** the parameter immediately (matching the C# `QueryStringBuilder` pattern), storing the result in a `Map<string, string>` — no separate format tracking map is needed. Non-comma params are added in bulk via `.addMany()`, then comma-style params override their keys individually.

### Runtime (`core-utilities`)
- **New: `src/core/url/QueryStringBuilder.ts`** — Exports a `queryBuilder()` factory function (class is private). Provides three fluent methods:
  - `.add(key, value, options?)` — eagerly serializes a single parameter via `toQueryString()`, with optional `{ style: "comma" }` for comma-separated arrays
  - `.addMany(params)` — batch-adds all entries from a `Record<string, unknown>` using the default repeat format; null/undefined values are silently skipped
  - `.mergeAdditional(params?)` — overrides existing keys with runtime `requestOptions.queryParams` (last-write-wins via `Map.set`)
  - `.build()` — joins pre-serialized parts with `&`
- **`src/core/url/index.ts`** — Re-exports `queryBuilder`.
- **`src/core/url/qs.ts`** — Added `"comma"` to the `ArrayFormat` type. Comma format encodes each value individually *before* joining with literal commas (so commas within values become `%2C` while separator commas remain literal). Null/undefined values within arrays are skipped entirely (matching `qs` npm library behavior). The API matches the `qs` npm library — no custom extensions.
- **`src/core/fetcher/Fetcher.ts`** — Added `queryString?: string` to `Fetcher.Args`. When provided, the fetcher uses the pre-built string instead of calling `createRequestUrl` for query params.
- **`src/core/fetcher/createRequestUrl.ts`** — Removed the old `arrayFormats` third parameter; the function now only takes `baseUrl` and `queryParameters`.

### Generator (`commons` + `client-class-generator`)
- **`commons/src/core-utilities/UrlUtils.ts`** — Exposes `queryBuilder._invoke()` so generated code can call the factory function.
- **`commons/src/core-utilities/Fetcher.ts`** — Added `queryString` property to generated `Fetcher.Args` interface and `_invoke` emission.
- **`endpoints/utils/GeneratedQueryParams.ts`** — New `getQueryStringExpression(context)` method emits the builder chain. Separates params into non-comma (batched via `.addMany(_queryParams)`) and comma (individual `.add(key, value, { style: "comma" })` overrides). Uses `explode === false` to select comma style.
- **`endpoint-request/Generated{Default,Bytes,FileUpload}EndpointRequest.ts`** — Updated `getFetcherRequestArgs()` return types and implementations to include `queryString` from the builder.
- **`endpoints/default/GeneratedDefaultEndpointImplementation.ts`** — `buildFetcherArgs()` now passes `queryString` when present.

### Wire test example generation (IR pipeline)
- **`ExampleEndpointFactory.ts`** — Scoped query parameter example filtering: accepts **primitives** (as before) and **arrays** (new, for `explode:false` params), but rejects **objects** (restoring pre-PR behavior). This is the minimal change needed to let array examples flow through to wire tests while keeping backwards compatibility for customers with required object query params.
- **`convertParameters.ts`** — Added `getParameterExample()` to extract array examples from OpenAPI specs via `getExampleAsArray()`.
- **`buildEndpointExample.ts`** — Removed array-to-first-element flattening so full arrays flow through to wire tests.
- **`convertExampleEndpointCall.ts`** — When converting Fern definition examples for `allow-multiple` query params with array values, each element is now converted individually and wrapped in a list container (`ExampleContainer.list`).
- **`validateExampleEndpointCallParameters.ts`** — For `allow-multiple` query params with array examples, validates each element individually against the item type instead of rejecting the whole array. This is **strictly more permissive** — it can only make previously-failing `fern check` pass, never cause new failures.
- **`valid-example-endpoint-call.ts`** — Passes `isAllowMultiple` callback for query parameter validation.
- **`test-definitions/fern/apis/query-parameters-openapi/openapi.yml`** — Added `example` values to `tags` (`["ACCESS_GRANTED", "COPY"]`) and `optionalTags` (`["DELETE", "MOVE"]`) parameters so wire tests exercise the comma format with real array data.

### Snapshot updates
- **IR generator test definitions** — Regenerated `query-parameters-openapi.json` and `query-parameters-openapi-as-objects.json`. Endpoints with required object query params now have empty `userSpecifiedExamples` (restored pre-PR behavior due to scoped ExampleEndpointFactory change).
- **IR migration snapshots** — Updated `v57-to-v56` and `v58-to-v57` migration snapshots.
- **ETE test snapshots** — Updated `dynamic.test.ts.snap`, `ir.test.ts.snap`, `validate.test.ts.snap`, `diff.test.ts.snap`, and `dynamic.json`.

### Seed output
- Regenerated all affected `ts-sdk` seed fixtures. Generated code now uses the `addMany` + comma override pattern:
  ```ts
  queryString: core.url
      .queryBuilder()
      .addMany(_queryParams)
      .add("tags", _queryParams.tags, { style: "comma" })
      .mergeAdditional(requestOptions?.queryParams)
      .build()
  ```

### Changelog
- Added unreleased `feat` entry under `generators/typescript/sdk/changes/unreleased/`.
- Added unreleased `fix` entry under `packages/cli/cli/changes/unreleased/` for array example support.

## Human Review Checklist

- [ ] **ExampleEndpointFactory scoping**: The factory now accepts primitives and arrays but rejects objects for query params. This means endpoints with a required object query param (e.g. `user` as `$ref`) will have empty `userSpecifiedExamples`. Verify this is the desired minimal-change approach.
- [ ] **Validator permissiveness**: The `isAllowMultiple` validator change is strictly more permissive — previously-failing array examples now pass validation. Confirm no customer specs rely on the old rejection behavior.
- [ ] **`addMany` + comma override pattern**: Comma-style params are added twice — first via `.addMany(_queryParams)` (repeat format) then overridden via `.add(key, value, { style: "comma" })`. This relies on `Map.set` last-write-wins. Verify this is acceptable vs. filtering comma params out of the `addMany` call.
- [ ] **Builder emitted for all endpoints with query params**: Every endpoint that has query parameters gets a full builder chain, even if none use `explode: false`. Confirm this is acceptable vs. only emitting the builder when at least one comma param exists.
- [ ] **Redundant `queryParameters` computation**: Generated code still computes the `queryParameters` object *and* the `queryString` builder chain. When `queryString` is present, it takes precedence in the fetcher, making the object value unused at runtime. Verify this dual-emit is acceptable or if the legacy `queryParameters` path should be removed.
- [ ] **`mergeAdditional` override semantics**: When `requestOptions.queryParams` contains a key that was already `.add()`-ed, `mergeAdditional` uses `Map.set` (last-write-wins). Confirm this is the desired precedence.
- [ ] **Comma encoding divergence from qs library**: The comma format encodes each value *then* joins with literal commas, producing `a%2Cb,c` for `["a,b", "c"]`. The real `qs` library encodes commas too (`a%2Cb%2Cc`). This divergence is intentional (unambiguous output), but worth confirming.
- [ ] **Null/undefined skip in comma format**: Null/undefined array elements are skipped entirely (`["a", null, "c"]` → `a,c`), matching `qs` library behavior. Verify this is the desired behavior.

## Testing
- [x] Unit tests added for `QueryStringBuilder` (34 tests: `.add()` repeat + comma styles, `.addMany()` batch, chaining, `.mergeAdditional()`, edge cases, end-to-end scenario)
- [x] Unit tests added for comma array format in `qs.ts` (12 tests)
- [x] Existing `qs.ts` and `createRequestUrl.ts` tests passing
- [x] Wire tests exercise comma format with real array examples (`tags: ["ACCESS_GRANTED", "COPY"]`)
- [x] All `ts-sdk` seed fixtures regenerated and passing
- [x] IR generator test definitions regenerated
- [x] IR migration snapshots updated (v57-to-v56, v58-to-v57)
- [x] ETE test snapshots updated (dynamic, IR, validate, diff)
- [x] Validator tests passing (69 tests across 47 files)
- [x] Generator packages compile successfully
- [x] `biome`, `compile`, `depcheck`, `lint` CI checks all green

Link to Devin session: https://app.devin.ai/sessions/5b96d911886f499cbfabba83155f56a6
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/15026" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


generators/typescript/utils/core-utilities/tests/unit/url/qs.test.ts
```diff
@@ -241,11 +241,101 @@ describe("Test qs toQueryString", () => {
...
+                description: "should handle null values in comma format",
+                input: { items: ["a", null, "c"] },
+                options: { arrayFormat: "comma" },
+                expected: "items=a,,c",
+            },
+            {
+                description: "should handle undefined values in comma format",
+                input: { items: ["a", undefined, "c"] },
+                options: { arrayFormat: "comma" },
+                expected: "items=a,,c",
```
Swimburger: does this match qs behavior?

generators/typescript/utils/core-utilities/tests/unit/url/qs.test.ts
```diff
@@ -241,11 +241,101 @@ describe("Test qs toQueryString", () => {
...
+                description: "should handle boolean array values",
+                input: { flags: [true, false, true] },
+                options: { arrayFormat: "comma" },
+                expected: "flags=true,false,true",
+            },
+            {
+                description: "should handle null values in comma format",
+                input: { items: ["a", null, "c"] },
+                options: { arrayFormat: "comma" },
+                expected: "items=a,,c",
```
Swimburger: does this match qs behavior?

seed/ts-sdk/query-parameters-openapi-as-objects/src/Client.ts
```diff
@@ -135,6 +135,28 @@ export class SeedApiClient {
...
+                .add("bytes", _queryParams.bytes)
+                .add("user", _queryParams.user)
+                .add("userList", _queryParams.userList)
+                .add("optionalDeadline", _queryParams.optionalDeadline)
+                .add("keyValue", _queryParams.keyValue)
+                .add("optionalString", _queryParams.optionalString)
+                .add("nestedUser", _queryParams.nestedUser)
+                .add("optionalUser", _queryParams.optionalUser)
+                .add("excludeUser", _queryParams.excludeUser)
+                .add("filter", _queryParams.filter)
```
Swimburger: All the non-comma style queryparams could be added in one method, like `.addMany` or something similar

iamnamananand996 (approved):
Looking at the seed output, looks good

PR #15027, opened 2026-04-15
fix(typescript): include literal custom properties in OAuth token request

Swimburger:
## Description

Fixes a TypeScript compilation error in generated SDKs when the OAuth token endpoint has additional required literal properties beyond `client_id` and `client_secret` (e.g., `grant_type: "client_credentials"` in the Box API).

The generated `OAuthAuthProvider.ts` was only passing `client_id` and `client_secret` to the token endpoint call, but the request type (`PostOAuth2Token`) requires `grant_type` as a required literal field, causing:

```
error TS2345: Argument of type '{ client_id: string; client_secret: string; }' is not assignable to parameter of type 'PostOAuth2Token'.
  Property 'grant_type' is missing in type '{ client_id: string; client_secret: string; }' but required in type 'PostOAuth2Token'.
```

This had two root causes:
1. The **TypeScript SDK generator** did not emit custom properties even when they were present in the IR
2. The **OpenAPI-to-IR converter** hardcoded `customProperties: undefined`, so custom properties were never populated for OpenAPI specs
3. The **Fern definition-to-IR converter** only extracted property names from inline request body `properties`, missing referenced/extended types

### TypeScript SDK Generator (`OAuthAuthProviderGenerator.ts`)
- Added `getCustomPropertyAssignments()` method that iterates over `requestProperties.customProperties` from the IR, resolves literal types, and generates inline property assignments for the token request call
- Updated both the expiration and non-expiration token request code paths to include the custom property assignments
- Uses `JSON.stringify()` for safe string literal serialization in generated code
- **Skips custom properties for `inlinedRequestBody` endpoints** — for those, the generated client method already injects literals into the body via `getLiteralProperties` in `GeneratedDefaultEndpointRequest`, and the generated request type excludes them. Emitting them from `OAuthAuthProvider` would cause TypeScript excess-property errors. Custom properties are only emitted for non-inlined bodies (e.g., form-encoded OpenAPI endpoints where the request type includes literals as required fields).

### OpenAPI-to-IR Converter (`resolveOAuthEndpointReferences.ts`)
- Added `findCustomRequestProperties()` function that extracts all request body properties except `client_id` and `client_secret` from the token endpoint
- Handles both `inlinedRequestBody` and `reference` (named type) body types
- Replaced the hardcoded `customProperties: undefined` with the resolved custom properties

### Fern Definition-to-IR Converter (`convertOAuthTokenEndpoint.ts`)
- Rewrote `getRequestBodyPropertyNames` to use proper `RawSchemas.HttpEndpointSchema` type instead of `{ request?: unknown }` with `as X` assertions
- Uses `isInlineRequestBody()` for type narrowing instead of manual field checks
- Added handling for `extends` in inline request bodies — resolves each extended type via `TypeResolver` and includes their property names (handles `ObjectExtendsSchema` which can be `string | string[]`)
- Added `getPropertyNamesFromType()` helper that resolves a type reference via `TypeResolver` and extracts property names from both `properties` and `base-properties`
- Threaded `TypeResolver` through `convertOAuthClientCredentials`, `convertApiAuth`, and `generateIntermediateRepresentation`

### Changelogs
- Added changelog entries for both the TypeScript SDK generator and the CLI (OpenAPI-to-IR)

## Important Review Notes

- **Only literal custom properties are emitted** — non-literal custom properties (which are regular request parameters) are silently skipped since they are handled as part of the request class constructor.
- **Inlined vs non-inlined request bodies** — The `inlinedRequestBody` check is the key heuristic: `GeneratedDefaultEndpointRequest.getLiteralProperties()` handles literal injection for inlined bodies, so `OAuthAuthProvider` only needs to emit them for non-inlined bodies (e.g., OpenAPI form-encoded endpoints). Reviewers should verify this heuristic is sound for all cases.
- **`scopes` exclusion** — The Fern definition path (`convertOAuthTokenEndpoint.ts`) excludes `scopes` from custom properties based on explicit config. The OpenAPI path does not have this explicit scopes config, so scopes would appear as a custom property if present. Worth verifying this is acceptable.
- **`propertyPath: undefined`** — `findCustomRequestProperties` sets `propertyPath: undefined` when constructing `RequestProperty` values. This matches the existing pattern in `findRequestBodyProperty` but reviewers should confirm no downstream consumer requires a non-null path.
- **`getPropertyNamesFromType` does not recurse into `extends`** — It reads `properties` and `base-properties` from the resolved declaration but does not follow any `extends` chains on the resolved type itself. For OAuth token endpoints this should be fine (they tend to be flat), but worth noting.
- **Other generators** — The C# and Java generators already handle `customProperties` in their code. This change only fixes the TypeScript generator and OpenAPI-to-IR converter.

## Testing

- [x] Existing unit tests pass (538 tests in client-class-generator)
- [x] TypeScript compilation succeeds
- [x] Seed tests `oauth-client-credentials-nested-root` and `oauth-client-credentials-with-variables` pass locally (no snapshot changes — confirms inlined bodies are correctly skipped)
- [x] Verified generated output against Box API config via `pnpm seed run`

Link to Devin session: https://app.devin.ai/sessions/80f907c196a34373bde620359da01ccf
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/15027" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/seed/ts-sdk/oauth-client-credentials-nested-root/no-custom-config/src/auth/OAuthAuthProvider.ts
+++ b/seed/ts-sdk/oauth-client-credentials-nested-root/no-custom-config/src/auth/OAuthAuthProvider.ts
@@ -91,6 +91,8 @@ export class OAuthAuthProvider implements core.AuthProvider {
                 const tokenResponse = await this.authClient.getToken({
                     client_id: clientId,
                     client_secret: clientSecret,
+                    audience: "https://api.example.com",
+                    grant_type: "client_credentials",
                 });
 
                 this.accessToken = tokenResponse.access_token;
--- a/seed/ts-sdk/oauth-client-credentials-nested-root/no-custom-config/src/auth/OAuthAuthProvider.ts
+++ b/seed/ts-sdk/oauth-client-credentials-nested-root/no-custom-config/src/auth/OAuthAuthProvider.ts
@@ -91,8 +91,6 @@ export class OAuthAuthProvider implements core.AuthProvider {
                 const tokenResponse = await this.authClient.getToken({
                     client_id: clientId,
                     client_secret: clientSecret,
-                    audience: "https://api.example.com",
-                    grant_type: "client_credentials",
                 });
 
                 this.accessToken = tokenResponse.access_token;
--- a/generators/typescript/sdk/changes/unreleased/fix-oauth-custom-properties.yml
+++ b/generators/typescript/sdk/changes/unreleased/fix-oauth-custom-properties.yml
@@ -1,0 +1,6 @@
+- summary: |
+    Fix OAuth token request to include literal custom properties (e.g., grant_type)
+    from the token endpoint request. Previously, only client_id and client_secret
+    were passed, causing TypeScript compilation errors for APIs that require
+    additional literal properties like grant_type on the token request.
+  type: fix
--- a/generators/typescript/sdk/client-class-generator/src/auth-provider/OAuthAuthProviderGenerator.ts
+++ b/generators/typescript/sdk/client-class-generator/src/auth-provider/OAuthAuthProviderGenerator.ts
@@ -874,7 +874,7 @@ export class OAuthAuthProviderGenerator implements AuthProviderGenerator {
             if (resolvedType.type === "container" && resolvedType.container.type === "literal") {
                 const propertyName = this.getName(customProperty.property.name, context);
                 const literalValue = resolvedType.container.literal._visit<string>({
-                    string: (val: string) => `"${val}"`,
+                    string: (val: string) => `"${val.replace(/\\/g, "\\\\").replace(/"/g, '\\"')}"`,
                     boolean: (val: boolean) => `${val}`,
                     _other: () => {
                         throw new Error("Encountered non-boolean, non-string literal");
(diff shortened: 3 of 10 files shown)


fern-support (approved):

seed/ts-sdk/oauth-client-credentials-nested-root/no-custom-config/src/auth/OAuthAuthProvider.ts
```diff
@@ -91,6 +91,8 @@ export class OAuthAuthProvider implements core.AuthProvider {
                 const tokenResponse = await this.authClient.getToken({
                     client_id: clientId,
                     client_secret: clientSecret,
+                    audience: "https://api.example.com",
```
Swimburger: this seems wrong, since the .getToken already includes the literals:

```
    private async __getToken(
        request: SeedOauthClientCredentials.auth.GetTokenRequest,
        requestOptions?: AuthClient.RequestOptions,
    ): Promise<core.WithRawResponse<SeedOauthClientCredentials.auth.TokenResponse>> {
        const _headers: core.Fetcher.Args["headers"] = mergeHeaders(this._options?.headers, requestOptions?.headers);
        const _response = await core.fetcher({
            url: core.url.join(
                (await core.Supplier.get(this._options.baseUrl)) ??
                    (await core.Supplier.get(this._options.environment)),
                "/token",
            ),
            method: "POST",
            headers: _headers,
            contentType: "application/json",
            queryParameters: requestOptions?.queryParams,
            requestType: "json",
            body: { ...request, audience: "https://api.example.com", grant_type: "client_credentials" },
            timeoutMs: (requestOptions?.timeoutInSeconds ?? this._options?.timeoutInSeconds ?? 60) * 1000,
            maxRetries: requestOptions?.maxRetries ?? this._options?.maxRetries,
            abortSignal: requestOptions?.abortSignal,
            fetchFn: this._options?.fetch,
            logging: this._options.logging,
        });
```

fern-support (approved):

PR #15028, opened 2026-04-15
fix(seed): ignore changelog/changes directories in affected detection

Swimburger:
## Description

Refs https://github.com/fern-api/fern/pull/15023

Changelog and release metadata files under `generators/*/sdk/changes/` (e.g. `.template.yml`, unreleased changelog entries) were incorrectly triggering seed tests because their paths match `GENERATOR_SOURCE_PATHS` prefixes like `generators/typescript/`. This caused PR #15023 (a release-script-only change) to run the full seed test matrix for every generator.

## Changes Made

- Added `IGNORED_PATH_PATTERNS = [/\/sdk\/changes\//]` in `getAffected.ts` — a regex that specifically targets the `generators/*/sdk/changes/` changelog directory structure, following the same skip pattern as `IGNORED_FILENAMES` for `versions.yml`
- Added a regex check in the generator detection loop to skip files matching these changelog paths
- Added unit tests covering: changelog templates only, changelog entries only, mixed changelog + real source changes, and an end-to-end scenario matching the exact file list from PR #15023
- Added a safety test verifying that a non-changelog path containing `/changes/` (e.g. `generators/python/src/changes/handler.ts`) is **not** incorrectly ignored

## Testing

- [x] Unit tests added (5 new test cases, all 67 tests passing)
- [x] Manual verification: the exact file list from PR #15023 now correctly produces zero affected generators

- Tightened the match from a generic `/changes/` substring to a specific `/sdk/changes/` regex pattern (addresses [review feedback](https://github.com/fern-api/fern/pull/15028#discussion_r3087675518))
- Added test case ensuring hypothetical source files under non-changelog `/changes/` paths still correctly trigger seed tests

## Human Review Checklist

- **Regex specificity**: The pattern `/\/sdk\/changes\//` targets only the known changelog directory layout (`generators/*/sdk/changes/`). A source file at `generators/python/src/changes/handler.ts` would **not** be skipped (verified by test). A file at e.g. `generators/python/sdk/changes_util.ts` would also not match (the regex requires a trailing `/`).
- **Scope**: The filter only applies to the generator source detection loop. It does **not** suppress global infrastructure detection, fixture detection, Dockerfile detection, or seed.yml detection — those remain unaffected.

Link to Devin session: https://app.devin.ai/sessions/ee2ef92363ce43b6917e378ae667bdab
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/15028" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/seed/src/commands/affected/getAffected.ts
+++ b/packages/seed/src/commands/affected/getAffected.ts
@@ -34,6 +34,14 @@ export interface AffectedResult {
 const IGNORED_FILENAMES = ["versions.yml"];
 
 /**
+ * Path segments that indicate changelog / release metadata directories.
+ * Files under paths containing these segments live under generator source
+ * paths but do not affect generated code (e.g.
+ * `generators/typescript/sdk/changes/unreleased/.template.yml`).
+ */
+const IGNORED_PATH_SEGMENTS = ["/changes/"];
+
+/**
  * Paths that, when changed, affect ALL generators and ALL fixtures.
  * These are infrastructure-level changes that feed into IR generation
  * or affect how seed tests execute. Includes:
@@ -214,6 +222,9 @@ export function detectAffected(changedFiles: string[], allGenerators: GeneratorW
         // Skip metadata files that live under generator paths but don't affect codegen
         const basename = file.split("/").pop() ?? "";
         if (IGNORED_FILENAMES.includes(basename)) {
+            continue;
+        }
+        if (IGNORED_PATH_SEGMENTS.some((segment) => file.includes(segment))) {
             continue;
         }
         for (const [generatorName, sourcePaths] of Object.entries(GENERATOR_SOURCE_PATHS)) {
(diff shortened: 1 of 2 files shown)


packages/seed/src/commands/affected/getAffected.ts
```diff
@@ -33,6 +33,14 @@ export interface AffectedResult {
  */
 const IGNORED_FILENAMES = ["versions.yml"];
 
+/**
+ * Path segments that indicate changelog / release metadata directories.
+ * Files under paths containing these segments live under generator source
+ * paths but do not affect generated code (e.g.
+ * `generators/typescript/sdk/changes/unreleased/.template.yml`).
+ */
+const IGNORED_PATH_SEGMENTS = ["/changes/"];
```
Swimburger: is this too aggressive? Could something be in a differnet /changes/ folder and be ignored by accident?

fern-support (approved):

PR #15073, opened 2026-04-15
fix(python): wrap non-string global header values with str()

fern-support:
## Description

Fixes a `mypy` failure in the generated Python SDK when an API declares a non-string global header (e.g. `Request-Timeout: integer`, as used by Vectara).

The generator emits `headers: Dict[str, str]` but previously assigned the raw typed member directly, producing:

```
src/vectara/core/client_wrapper.py:46: error: Incompatible types in assignment
    (expression has type "int", target has type "str")  [assignment]
headers["Request-Timeout"] = self._request_timeout
```

No existing seed fixture declared a non-string global header, so the code path was never exercised.

## Changes Made
- Added `is_type_reference_string` helper in `client_generator/type_utilities.py` that unwraps `Optional`/`nullable`/alias to detect the underlying primitive.
- Added a `needs_str_conversion` flag on `ConstructorParameter` in `core_utilities/client_wrapper_generator.py`, set when the header's Fern IR type is not a string.
- Updated `_get_write_get_headers_body` to wrap the RHS with `str(...)` only when `needs_str_conversion` is True — string-typed headers are emitted unchanged to avoid snapshot churn across existing fixtures.
- Recorded the fix in `generators/python/sdk/versions.yml` as `5.4.0-rc.2`.

Generated code now produces:
```python
if self._request_timeout is not None:
    headers["Request-Timeout"] = str(self._request_timeout)
```

## Testing
Regenerated the Vectara SDK against `/Users/Patrick/configs/vectara-fern-config/fern` using `pnpm seed run --generator python-sdk --skipScripts`, then ran the generated project's CI workflow locally:

- `poetry run mypy .` — Success: no issues found in 906 source files
- `poetry run pytest -rP -n auto .` — 49 passed, 4 skipped
- `poetry install --extras aiohttp && poetry run pytest -rP -n auto -m aiohttp .` — 3 passed

### DIFF
--- a/generators/python/sdk/versions.yml
+++ b/generators/python/sdk/versions.yml
@@ -1,4 +1,16 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 5.4.0-rc.2
+  changelogEntry:
+    - summary: |
+        Fix `mypy` error in generated client wrapper when a non-string global header is
+        declared (e.g. `Request-Timeout: integer`). The generator emits `headers: Dict[str, str]`
+        but previously assigned the raw typed member (e.g. `int`) directly, causing
+        `Incompatible types in assignment`. Non-string global header values are now wrapped
+        with `str(...)` when assigned. String-typed headers are unchanged.
+      type: fix
+  createdAt: "2026-04-15"
+  irVersion: 66
+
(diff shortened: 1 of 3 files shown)


patrickthornton:
Recreating under correct author (auth switched after initial creation).

PR #15143, opened 2026-04-19
fix(cli): fail with clear error when docs.yml is missing from `fern docs dev`

broady:
## Description

`fern docs dev` previously exited silently when no `docs.yml` was present. Now it fails with a clear, actionable error message.

## Changes Made

- Added `failAndThrow` call with `ConfigError` code when `docsWorkspace` is null in `previewDocsWorkspace`
- Removed the silent `return` that previously caused the command to exit without feedback
- Added unreleased changelog entry

## Testing

- [x] Lint and format checks pass (`pnpm lint:biome`, `pnpm format`)
- [x] No existing tests affected (the related test in `generate.test.ts` tests `fern generate --docs` via a separate code path in `generateDocsWorkspace.ts`)

Link to Devin session: https://app.devin.ai/sessions/5e2f8546653e495680c64d2911a609f4
Requested by: @broady

### DIFF
--- a/packages/cli/cli/changes/unreleased/fix-docs-dev-missing-docs-yml.yml
+++ b/packages/cli/cli/changes/unreleased/fix-docs-dev-missing-docs-yml.yml
@@ -1,0 +1,5 @@
+# yaml-language-server: $schema=../../../../../fern-changes-yml.schema.json
+
+- summary: |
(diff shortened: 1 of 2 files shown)


thesandlord (approved):

broady:
@claude please copy the error message and PR description from #15667

PR #15185, opened 2026-04-20
chore(cli): normalize source map paths for Sentry display

FedeZara:
## Summary

Post-process the CLI bundle's source map so that each entry in `sources`
is a clean repo-root-relative path (e.g. `packages/cli/cli/src/cli.ts`)
instead of the esbuild default (`../../src/cli.ts`).

## Why

Sentry resolves stack frames by looking up the entry in `sources[i]`
pointed to by the bundle's debug ID. It then renders that string
verbatim in the issue header. Because esbuild emits `sources` relative
to each output file's directory, every frame currently shows up as
something like:

```
packages/cli/cli/dist/prod/../../src/cli.ts
packages/cli/cli/dist/prod/../../../workspace/loader/src/loadDocsWorkspace.ts
```

The canonical paths Sentry computes on hover (`app:///packages/...`)
are correct, but the displayed filename is noisy and the varying depth
of `..` segments makes it awkward to write one GitHub code mapping that
covers every package pulled into the bundle.

`sourceRoot` doesn't help on its own: per the source-map spec it's a
URL prefix, and Sentry's display concatenates it with each `sources[i]`
literally rather than normalising the result.

## What

A small post-build step in `buildCli` that, after tsup finishes,
rewrites every `.map` file so that:

- each entry in `sources` is resolved against the map's directory, then
  made relative to the repo root (`packages/cli/cli/src/cli.ts`, ...)
- `sourceRoot` is dropped (no longer needed)

Everything else in the map (`mappings`, `names`, `sourcesContent`,
debug-id metadata that sentry-cli injects later) is preserved as-is.

## Impact

- **Runtime / published `fern-api` package**: zero. The `.map` file is
  not included in the npm package (`files: ["cli.cjs"]`); only Sentry
  (and anyone debugging the bundle locally) ever reads it.
- **Sentry**: stack frames now render as `packages/cli/cli/src/cli.ts`
  directly, and a single GitHub code mapping (`app:///` -> empty root)
  resolves every frame regardless of which package it lives in.
- **Existing uploaded maps**: unaffected. Matching is done via
  debug IDs, so old events continue to resolve against their old maps.

Only the CLI is touched in this PR; the equivalent change for
generators will ship separately.

## Test plan

- [x] `pnpm dist:cli:prod` succeeds
- [x] Resulting `cli.cjs.map` has no `sourceRoot`, no `..` segments in
  any `sources` entry, and includes `packages/cli/cli/src/cli.ts`
- [ ] Upload the rebuilt map with `sentry-cli sourcemaps upload` and
  trigger a test error; confirm the Sentry issue header shows
  `packages/cli/cli/src/cli.ts`

### DIFF
--- a/packages/cli/cli/build-utils.mjs
+++ b/packages/cli/cli/build-utils.mjs
@@ -1,5 +1,6 @@
 import { exec } from "child_process";
-import { writeFile } from "fs/promises";
+import { existsSync } from "fs";
+import { readdir, readFile, writeFile } from "fs/promises";
 import path from "path";
 import tsup from "tsup";
 import { fileURLToPath } from "url";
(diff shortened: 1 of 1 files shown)


FedeZara:
Reopening under correctly-named branch 'FedeZara/chore/normalize-sentry-sourcemap-paths'.

PR #15193, opened 2026-04-20
feat(cli): harden Sentry config with additional integrations

FedeZara:
## Summary

Hardens the CLI's Sentry initialisation by opting into four additional integrations and two config flags that were previously missing.

## Changes

- **`onUncaughtExceptionIntegration` / `onUnhandledRejectionIntegration`** — catches errors and promise rejections that escape the CLI's top-level handler (fire-and-forget callbacks, unhandled rejections from background work, etc.). More impactful here than in a generator container because the CLI is long-lived, runs on arbitrary user machines, and has a larger async surface area.
- **`linkedErrorsIntegration`** — follows the `.cause` chain on wrapped errors so both the original and the wrapper appear in Sentry, rather than just the outermost exception.
- **`nodeContextIntegration`** — attaches Node.js version, OS name/version, and architecture to every event. Useful for diagnosing environment-specific failures.
- **`attachStacktrace: true`** — forces a stack trace even when the thrown value is not an `Error` instance (plain string, object, etc.).
- **`sendClientReports: false`** — suppresses the client-report envelope Sentry sends when events are dropped due to rate limiting.

`defaultIntegrations: false` and `tracesSampleRate: 0` are kept as-is — HTTP tracing, local-variable capture, and performance sampling are not needed for a CLI.

### DIFF
--- a/packages/cli/cli/src/telemetry/SentryClient.ts
+++ b/packages/cli/cli/src/telemetry/SentryClient.ts
@@ -20,9 +20,33 @@ export class SentryClient {
                 dsn: sentryDsn,
                 release,
                 environment: sentryEnvironment,
+                // Opt out of every built-in integration (HTTP tracing, local
+                // variables, console breadcrumbs, etc.) — error capture is all
+                // we need.
                 defaultIntegrations: false,
-                integrations: [Sentry.rewriteFramesIntegration()],
-                tracesSampleRate: 0
+                // Rewrite absolute frame paths to repo-root-relative paths so
+                // they align with the source maps uploaded at publish time.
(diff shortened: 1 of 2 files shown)


Swimburger (approved):

FedeZara:
* **#15193** <a href="https://app.graphite.com/github/pr/fern-api/fern/15193?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/fern-api/fern/15193?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #15215, opened 2026-04-21
feat(go): add clientDefault support to Go SDK generator

Swimburger:
## Description

Adds `clientDefault` support (from the `x-fern-default` OpenAPI extension) to the Go SDK generator. When a header, query parameter, or path parameter has a `clientDefault` value, the generated SDK uses that value as a fallback when the caller does not provide one.

Both the **v1** and **v2** Go generators are updated.

### Go v1 generator (`generators/go/internal/generator/sdk.go`)

- **IR types** (`http.go`): Added `ClientDefault *Literal` field to `HttpHeader`, `QueryParameter`, and `PathParameter` structs with getters, setters, and bitmask serialization control.
- **Request-options headers** (`sdk.go:WriteRequestOptionsDefinition`): Headers with `clientDefault` initialize a local var with the default, optionally fall back to an env var, then use the caller-provided value if present.
- **Client constructor headers** (`sdk.go:WriteClient`): After the existing env-var fallback, a second fallback to `clientDefault` is applied for string-typed headers. Note: the prior `continue` after the env block was **removed** so execution falls through to the new clientDefault check.
- **Endpoint headers**: Endpoint-scoped headers with `clientDefault` use a local variable initialized with the default, overridden by the request field if provided.
- **Query parameters**: After normal serialization, a `clientDefault` is added only if the key is not already present (`if _, ok := queryParams[...]; !ok`).
- **Path parameters**: For both wrapped-request and positional styles, a local variable is initialized from the request field (or param), falling back to the clientDefault when the value is the zero string.
- **Helpers**: Added `isStringType()` which guards clientDefault generation to non-optional string primitives, and `pathParameterDefault` struct for path param default tracking.

### Go v2 generator (`generators/go-v2/sdk/src/`)

- **`authUtils.ts`**: Added `isPlainStringType()` helper — returns true only for non-optional, non-nullable `string` primitives. Guards clientDefault generation for path params and headers since the `== ""` zero-value check only works for Go `string` (not `*string`).
- **`ClientGenerator.ts`**: Modified `writeHeaderEnvironmentVariables()` to apply clientDefault fallback for global headers after the env-var fallback. Added `writeClientDefaultConditional()` method using safe `getLiteralValue` (AST-based, properly escaped).
- **`HttpEndpointGenerator.ts`**:
  - **Path parameters**: Before URL encoding, applies clientDefault fallback using local variables (avoids mutating caller's request struct). Uses `getFieldName()` for safe Go identifiers. `endpointURL :=` assignment moved after local var declarations.
  - **Query parameters**: After type-level defaults, adds clientDefault entries for query params that don't already have a type-level default. No type guard — works for both string and boolean literals (matches v1 behavior).
  - **Endpoint headers**: After existing header handling, applies clientDefault fallback using local variables (avoids request struct mutation).

### Other

- **Unreleased changelog**: Added `generators/go/sdk/changes/unreleased/add-client-default-support.yml` (uses the unreleased version mechanism instead of editing `versions.yml` directly).
- **Seed snapshots**: Updated `seed/go-sdk/x-fern-default/` (clientDefault output) and `seed/go-sdk/exhaustive/` (unrelated UUID→Uuid naming regeneration from main).

## Issues found during review & fixed

- **Security: unescaped string interpolation** — Replaced all 3 `getLiteralAsString` callsites (which do no escaping) with safe `getLiteralValue` (AST-based `go.TypeInstantiation.string()` → `escapeGoString`). Affected: `ClientGenerator.ts:writeClientDefaultConditional`, `HttpEndpointGenerator.ts:buildEndpointUrl`, `HttpEndpointGenerator.ts:buildHeaders`.
- **Request struct mutation** — v2 endpoint headers and path parameters now use local variables (`_FieldName := request.FieldName`) instead of mutating the caller's `*Request` struct through the pointer.
- **Invalid Go syntax** — Moved `writer.write("endpointURL := ")` from before clientDefault local var declarations to after them, preventing `endpointURL := _region := request.Region` (double `:=`).
- **Unsafe Go identifiers** — Path param local variable names now use `getFieldName()` instead of `getOriginalName()` to handle names with hyphens/special characters.
- **Silently dropped boolean defaults** — Removed `isPlainStringType` guard for query param clientDefault in v2 (query params use a defaults map, so any literal type works — matches v1 behavior).

## Human Review Checklist

> Items worth a reviewer's attention — not all are bugs, but they carry risk.

- [ ] **`continue` removal in v1 client constructor** (~line 1330): The `continue` after the env-var block was removed so execution falls through to the clientDefault check. When `env != nil` but `clientDefault == nil`, the code enters the env block then skips the clientDefault block (guarded by `clientDefault != nil`). Behavior should be equivalent, but worth verifying.
- [ ] **Space in `options. ` field access** (v1 lines ~1334, 1336): `f.P("options. ", ...)` has a space — this is pre-existing in the codebase. Go's `f.P` concatenates args, so it generates `options. HeaderName`. Confirm this is intentional.
- [ ] **`isPlainStringType` / `isStringType` don't resolve aliases**: If a type is an alias of `string`, clientDefault will be silently skipped. Same limitation in both v1 and v2.
- [ ] **Pre-existing `getLiteralAsString` calls for type-level literals**: Lines 832 and 886 of `HttpEndpointGenerator.ts` still use `getLiteralAsString` for type-level literal values (not clientDefault). These have the same escaping gap but are pre-existing and out of scope for this PR.
- [ ] **`_FieldName` local variable prefix convention**: Path param and header clientDefault logic creates local variables prefixed with `_` (e.g., `_Region`). Verify this doesn't collide with any Go reserved identifiers or existing generated variable names.
- [ ] **Query param clientDefault extracts raw string from Literal**: `queryParameter.clientDefault.string` is passed to `go.TypeInstantiation.string()` which handles escaping. Confirm this is safe for all string contents (quotes, backslashes, newlines).
- [ ] **Exhaustive snapshot churn**: The `seed/go-sdk/exhaustive/` changes are unrelated to clientDefault — they're from main's UUID→Uuid naming convention update and metadata cleanup that got regenerated when running seed tests.

## Testing
- [x] v2 TypeScript compilation (`tsc`) passed via seed test build
- [x] Self-review of full diff completed
- [x] All Devin Review issues resolved (6 findings addressed across 6 commits)
- [x] hex-security-app finding addressed (unescaped `getLiteralAsString` → safe `getLiteralValue`)
- [x] Docker seed tests: `x-fern-default` fixture — generated Go code compiles (`go build`) and passes tests (`go test`) ✓
- [x] Docker seed tests: `exhaustive` fixture (both variants) — compiles and passes tests ✓
- [x] Seed snapshots committed (`x-fern-default` + `exhaustive` regeneration)
- [x] CI fully green: 53/53 checks passing (compile, biome, depcheck, lint, test, test-ete, all 16 go-sdk seed jobs)

Link to Devin session: https://app.devin.ai/sessions/dae61f87717a46d781e579e47b4758e5
Requested by: @Swimburger

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/15215" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


generators/go/sdk/versions.yml
```diff
@@ -1,4 +1,16 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 1.36.0
```
iamnamananand996: Can update this with the new unreleased mechanism.

seed/go-sdk/x-fern-default/client/raw_client.go
```diff
@@ -42,11 +42,20 @@ func (r *RawClient) TestGet(
 		r.baseURL,
 		"",
 	)
+	_Region := request.Region
```
iamnamananand996, line 45: Looks good

iamnamananand996 (approved):

generators/go/sdk/versions.yml
Swimburger: Instead of editing versions.yml, use the unreleased version mechanism. It is documented in CLAUDE.md and other places.

PR #15230, opened 2026-04-22
fix(cli): proper fallback for securityscheme descriptions

Ryan-Amirthan:
## Description
  Linear ticket: Closes FER-9586

  When security schemes are overridden in generators.yml, Fern now preserves descriptions from the OpenAPI spec instead of using generic placeholder text.

## Changes Made
  - Preserve OpenAPI descriptions by injecting them into generators.yml config before conversion (if auth scheme description does not exist)
  - Fix bug where auth scheme descriptions from generators.yml were ignored
  
## Testing

- [ X ] Unit tests added/updated
- [ X ] Manual testing completed
CleanShot 2026-04-21 at 21 39 24@2x


packages/cli/api-importers/openapi-to-ir/src/3.1/OpenAPIConverter.ts
```diff
@@ -110,21 +110,39 @@ export class OpenAPIConverter extends AbstractSpecConverter<OpenAPIConverterCont
     }
 
     private convertSecuritySchemes(): void {
+        const openApiSchemes = this.convertOpenApiSecuritySchemes();
+
         if (this.context.authOverrides) {
-            const overrideAuth = convertApiAuth({
-                rawApiFileSchema: this.context.authOverrides,
+            const descriptions = new Map(openApiSchemes.map((scheme) => [scheme.key, scheme.docs]));
```
jsklan, line 115: Let's pop this if block's content into its own function to make it clearer what we're doing here.

jsklan (approved):

PR #15232, opened 2026-04-22
feat(java): run v2 generator in parallel with v1

lifanzou:
## Description

Performance optimization for the Java SDK generator: run the v2 generator in parallel with v1 instead of sequentially.

## Changes Made

The Java SDK generator runs two sub-generators in sequence:
- **v1** (Java/Gradle): generates core SDK code -- models, clients, error types, HTTP infrastructure
- **v2** (TypeScript/Node.js): generates documentation and test artifacts -- README.md, reference.md, snippets, wire tests

Since v2 reads the generator config and IR independently and its output (docs and tests) does not depend on v1's output, the two generators can safely run concurrently.

This PR starts v2 on a background thread *before* v1 begins parsing the IR, so both generators execute in parallel. After v1 completes, it joins the v2 thread with a 5-minute timeout before reporting success. If v2 fails or times out, the overall generation fails with a clear error message.

**Result**: reduces overall Java SDK generation time by ~18%.

### Summary of code changes

- In `AbstractGeneratorCli.run()`, launch `runV2Generator()` on a named daemon thread (`java-v2-generator`) before calling `getIr()`
- After v1 generation completes, `join()` the v2 thread with a 300-second timeout
- If v2 is still alive after the timeout, throw a `RuntimeException` with a descriptive message
- Any exception from v2 is wrapped in a `RuntimeException` and propagated when the thread is joined

## Testing

- [x] Manual testing completed -- verified parallel execution produces identical output to sequential execution
- [x] Verified v2 failures still propagate correctly and fail the overall generation

### DIFF
--- a/generators/java/generator-utils/src/main/java/com/fern/java/AbstractGeneratorCli.java
+++ b/generators/java/generator-utils/src/main/java/com/fern/java/AbstractGeneratorCli.java
@@ -365,13 +365,15 @@ public abstract class AbstractGeneratorCli<T extends ICustomConfig, K extends ID
 
             // Start v2 generator asynchronously — it reads the config/IR independently
             // and generates docs (README, reference, snippets) which don't depend on v1 output.
-            Thread v2Thread = new Thread(() -> {
-                try {
-                    runV2Generator(generatorExecClient, args);
-                } catch (Exception e) {
-                    throw new RuntimeException("V2 generator failed", e);
-                }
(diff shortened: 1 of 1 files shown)


lifanzou:
Closing: CI benchmark showed +10.6% regression. The parallel v2 optimization improves native execution but doubles memory pressure in Docker's constrained environment.

PR #15233, opened 2026-04-22
chore(internal): use JSON.stringify instead of streaming for IR serialization

lifanzou:
## Description

Replace `json-stream-stringify` with native `JSON.stringify` for writing IR files to disk, improving serialization performance by 3-8x.

## Changes Made

- Replaced `JsonStreamStringify` + `createWriteStream` pipe with `JSON.stringify` + `writeFile` in `streamObjectToFile.ts`
- Switched from `fs.createWriteStream` to `fs/promises.writeFile` for a simpler async API
- Removes the `json-stream-stringify` dependency from the serialization path

### Why

The `streamObjectToFile` utility previously used `json-stream-stringify` piped to a `createWriteStream` to write JSON to disk. This streaming approach was originally used as a precaution for very large files, but in practice Fern IR files are well under V8's ~512MB string limit, so native `JSON.stringify` works without issue and is significantly faster.

The streaming pipeline adds overhead from creating and coordinating the stream without any practical benefit at these file sizes. Switching to `JSON.stringify` + `writeFile` is 3-8x faster and simplifies the code.

## Testing

- [x] Manual testing completed — verified IR serialization produces identical output
- The change is a straightforward swap of serialization method; the function signature and behavior are unchanged

### DIFF
--- a/packages/commons/fs-utils/package.json
+++ b/packages/commons/fs-utils/package.json
@@ -34,7 +34,6 @@
     "@fern-api/core-utils": "workspace:*",
     "@fern-api/path-utils": "workspace:*",
     "glob": "catalog:",
-    "json-stream-stringify": "catalog:",
     "stream-json": "catalog:",
     "tmp-promise": "catalog:"
   },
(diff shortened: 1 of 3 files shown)


lifanzou:
Folded into #15235 — the changes from this PR have been cherry-picked onto that branch. Closing this PR since its file paths (packages/commons/fs-utils) don't trigger the benchmark CI, and #15235 already triggers those checks.

PR #15234, opened 2026-04-22
feat(java): single-pass IR deserialization

lifanzou:
## Summary

- **Previous approach**: `readTree` (allocates full Jackson tree from file) -> `IntegerOverflowProcessor` (walks entire tree) -> extract `casingsConfig` (from tree) -> `treeToValue` (converts full tree to POJO) — two full passes over the data plus a full intermediate tree allocation.
- **New approach**: `readAllBytes` (read file to byte array once) -> `extractCasingConfig` (streaming parse, reads only the top-level `casingsConfig` field then stops) -> `readValue` (deserialize bytes directly to POJO in a single pass) — no intermediate tree allocation.
- The integer overflow processor is no longer called as it is no longer needed for current IR versions.
- This avoids allocating the full Jackson `JsonNode` tree, which is expensive for large IRs (the tree can be several times larger in memory than the raw JSON bytes).
- Reduces IR parsing time significantly for large APIs (e.g., Square).

## Test plan

- [ ] Verify Java SDK generation produces identical output for a representative set of APIs
- [ ] Verify IR files with and without `casingsConfig` are handled correctly
- [ ] Verify v66 compressed name deserialization still works end-to-end

### DIFF
--- a/generators/java/generator-utils/src/main/java/com/fern/java/AbstractGeneratorCli.java
+++ b/generators/java/generator-utils/src/main/java/com/fern/java/AbstractGeneratorCli.java
@@ -105,45 +105,64 @@ public abstract class AbstractGeneratorCli<T extends ICustomConfig, K extends ID
     /**
      * Loads and preprocesses the IR from file, handling integer overflow values and v66 compressed name
      * deserialization.
+     *
+     * Uses a two-phase approach: first extracts casingsConfig via partial tree parse,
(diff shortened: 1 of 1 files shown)


lifanzou:
Closing: CI benchmark showed +15.1% regression. The single-pass IR deserialization (Files.readAllBytes) adds a 57MB allocation that hurts Docker's memory-constrained environment.

PR #15308, opened 2026-04-23
feat(cli): add docs theme upload and list commands

aditya-arolkar-swe:
## What this does

Introduces **global themes** — a way for one repo (the \"control repo\") to define a shared visual identity and have multiple child repos automatically inherit it when they publish their docs.

---

### Step 1 — Control repo: export and upload the theme

A control repo maintains the canonical theme. Two new CLI commands manage this:

**Export the theme from an existing docs site:**
```bash
fern beta docs theme export --name my-theme
```
This reads the branding config from your local `docs.yml` (colors, logo, fonts, CSS, JS, etc.) and produces a `theme.yml` file ready to be uploaded.

**Upload the theme to Fern's cloud:**
```bash
fern beta docs theme upload fern/theme --name my-theme
```
This uploads the theme config and all its file assets (logo images, fonts, custom CSS/JS) to Fern's registry, where they're stored by content hash.

**Confirm the upload:**
```bash
fern beta docs theme list
```
Lists all themes uploaded for your org, with their names and last-updated timestamps.

---

### Step 2 — Child repos: reference the theme

In a child repo's `docs.yml`, add one line:

```yaml
global-theme: my-theme
```

---

### Step 3 — Child repos: publish as normal

```bash
fern generate --docs
```

The CLI automatically:
1. Fetches the named theme from Fern's registry
2. Downloads any file assets (logos, fonts, CSS, JS) referenced by the theme
3. Merges the theme into the local `docs.yml` config — **theme values win** for the fields listed below
4. Publishes the merged config to the docs site

No extra steps needed. The child repo just runs its normal publish command.

---

## What the theme controls vs. what the child repo controls

| Owned by the theme (control repo wins) | Owned by the child repo |
|---|---|
| `logo` (brand logo images and link) | `navigation` (tabs, sections, pages) |
| `favicon` (browser tab icon) | `apis` (API references) |
| `background-image` (page background) | `redirects` |
| `colors` (accent and background colors) | `versions` |
| `typography` (body, heading, and code fonts) | `instances` (domain / URL) |
| `layout` (sidebar width, content width, tab/searchbar placement) | |
| `theme` (light/dark mode default) | |
| `settings` (misc display settings) | |
| `integrations` (analytics and tracking) | |
| `css` (custom stylesheets) | |
| `js` (custom scripts) | |
| `header` (custom header component) | |
| `footer` (custom footer component) | |
| `navbar-links` (top navigation links) | |
| `footer-links` (footer navigation links) | |
| `ai-search` (AI search configuration) | |
| `announcement` (announcement banner) | |

---

## Test plan

- [ ] Upload a theme with `fern beta docs theme upload` against a local FDR instance
- [ ] Confirm it appears in `fern beta docs theme list`
- [ ] Add `global-theme: <name>` to a child repo's `docs.yml`
- [ ] Run `fern generate --docs` and verify theme fields (logo, colors, fonts) appear in the published site
- [ ] Verify child-repo-only fields (navigation, API refs) are preserved unchanged
- [ ] Verify a missing theme name produces a clear error with instructions to upload first


thesandlord (approved):

packages/cli/cli-v2/src/commands/docs/theme/export/command.ts
```diff
@@ -0,0 +1,122 @@
...
+                message: `No docs.yml found at ${docsYmlPath}`,
+                code: CliError.Code.ConfigError
+            });
+        }
+        const docsDir = path.dirname(docsYmlPath);
+        const outDir = args.directory
+            ? path.resolve(process.cwd(), args.directory)
+            : path.resolve(process.cwd(), "fern", "theme");
+
+        context.stdout.info(`Exporting theme-eligible fields from docs.yml → ${outDir}`);
```
amckinney, line 50: Messages like this should be written to stderr (the same applies elsewhere across these commands). That way if this command ever needs to write output that needs to be piped to other commands, the output stream isn't corrupted by the logs.

packages/cli/cli-v2/src/commands/docs/theme/export/command.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export declare namespace ExportThemeCommand {
+    export interface Args extends GlobalArgs {
+        directory?: string;
+    }
+}
+
+export class ExportThemeCommand {
+    public async handle(context: Context, args: ExportThemeCommand.Args): Promise<void> {
+        const docsYmlPath = path.resolve(process.cwd(), "fern", "docs.yml");
```
amckinney, line 36: This works for most cases, but not if the user runs fern nested within the fern directory. You'll need to use the FileFinder logic here, which only supports fern.yml in v2.

We can add support for finding the fern directory in v2, but that would need to be an opt-in option that we remove following the v2 release (I.e. we'll require fern.yml instead of the fern directory).

packages/cli/cli-v2/src/commands/docs/theme/export/command.ts
```diff
@@ -0,0 +1,122 @@
...
+        } catch {
+            throw new CliError({
+                message: `No docs.yml found at ${docsYmlPath}`,
+                code: CliError.Code.ConfigError
+            });
+        }
+        const docsDir = path.dirname(docsYmlPath);
+        const outDir = args.directory
+            ? path.resolve(process.cwd(), args.directory)
+            : path.resolve(process.cwd(), "fern", "theme");
```
amckinney, line 48: Similar to above, you'll want to use the workspace root (i.e. the path to the fern directory) to access the theme once you've resolved it above.

packages/cli/cli-v2/src/commands/docs/theme/export/command.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    private async copyLocalFiles(
+        obj: unknown,
+        sourceDir: string,
+        assetsDir: string,
+        context: Context
+    ): Promise<unknown> {
+        if (obj === null || typeof obj !== "object") {
+            if (typeof obj === "string" && !obj.startsWith("http://") && !obj.startsWith("https://")) {
```
amckinney, line 79: I think there's a helper somewhere where you can access "isRemote" which does similar HTTP scheme checks.

packages/cli/cli-v2/src/commands/docs/theme/export/command.ts
```diff
@@ -0,0 +1,122 @@
...
+                const abs = path.resolve(sourceDir, obj);
+                const dest = path.join(assetsDir, path.basename(abs));
+                try {
+                    await copyFile(abs, dest);
+                    context.stdout.debug(`  Copied ${obj} → assets/${path.basename(abs)}`);
+                    return `./assets/${path.basename(abs)}`;
```
amckinney, line 85: If we only capture the basename is this prone to collisions? foo/example.png and bar/example.png both resolve to example.png

packages/cli/cli-v2/src/commands/docs/theme/export/command.ts
```diff
@@ -0,0 +1,122 @@
...
+export function addExportThemeCommand(cli: Argv<GlobalArgs>): void {
+    const cmd = new ExportThemeCommand();
+    command(
+        cli,
+        "export",
+        "Export theme-eligible fields from docs.yml into a standalone theme directory",
+        (context, args) => cmd.handle(context, args as ExportThemeCommand.Args),
+        (yargs) =>
+            yargs
+                .option("directory", {
```
amckinney, line 114: We should call this --output, similar to other commands.

packages/cli/cli-v2/src/commands/docs/theme/list/command.ts
```diff
@@ -0,0 +1,75 @@
...
+export function addListThemesCommand(cli: Argv<GlobalArgs>): void {
+    const cmd = new ListThemesCommand();
+    command(
+        cli,
+        "list",
+        "List all themes for an org",
+        (context, args) => cmd.handle(context, args as ListThemesCommand.Args),
+        (yargs) =>
+            yargs
+                .option("org", {
```
amckinney, line 69: Let's add a --json flag which writes the output as a JSON array so it's compatible with jq. We can capture multiple properties within each theme like updatedAt, etc. Newline delimited is the correct default behavior.

packages/cli/cli-v2/src/commands/docs/theme/list/command.ts
```diff
@@ -0,0 +1,75 @@
...
+            });
+        }
+
+        const data = (await res.json()) as { themes: Array<{ name: string; updatedAt: string }> };
+        if (data.themes.length === 0) {
+            context.stdout.info(`No themes found for org "${orgId}".`);
+            return;
+        }
+        for (const theme of data.themes) {
+            context.stdout.info(`  ${theme.name}  (updated ${new Date(theme.updatedAt).toLocaleString()})`);
```
amckinney, line 41: For the default behavior I would just list the theme name without any indentation or additional metadata -- the user can specify --json if they need more details. That way this is easy to loop over in a bash array. I imagine we'll eventually have a "get" endpoint so users can fetch details about a specific theme.

packages/cli/cli-v2/src/commands/docs/theme/list/command.ts
```diff
@@ -0,0 +1,75 @@
...
+            code: CliError.Code.ConfigError
+        });
+    }
+}
+
+export function addListThemesCommand(cli: Argv<GlobalArgs>): void {
+    const cmd = new ListThemesCommand();
+    command(
+        cli,
+        "list",
```
amckinney, line 64: The payloads are small enough for now, but I imagine we'll eventually want pagination here so users can specify how many themes they want to fetch and where to start, etc. Not actionable now.

packages/cli/cli-v2/src/commands/docs/theme/upload/command.ts
```diff
@@ -0,0 +1,326 @@
...
+import mime from "mime-types";
+import path from "path";
+import type { Argv } from "yargs";
+import type { Context } from "../../../../context/Context.js";
+import type { GlobalArgs } from "../../../../context/GlobalArgs.js";
+import { command } from "../../../_internal/command.js";
+
+// FERN_FDR_ORIGIN is checked first because DEFAULT_FDR_ORIGIN is baked into the
+// prod bundle at build time and cannot be overridden at runtime.
+const FDR_ORIGIN =
```
amckinney, line 15: If we need this across multiple commands, move it to a common package that they can import from. You should see references in other commands to model this off of.

packages/cli/cli-v2/src/commands/docs/theme/upload/command.ts
```diff
@@ -0,0 +1,326 @@
...
+    process.env.FERN_FDR_ORIGIN ?? process.env.DEFAULT_FDR_ORIGIN ?? "https://registry.buildwithfern.com";
+
+// ── CAS helpers ───────────────────────────────────────────────────────────────
+
+function getGitRoot(): string {
```
amckinney, line 20: All these private functions in this file should be moved to private methods defined within the upload class (like you've done in the other commands). No need to define them elsewhere.

packages/cli/cli-v2/src/commands/docs/theme/upload/command.ts
```diff
@@ -0,0 +1,326 @@
...
+import type { Context } from "../../../../context/Context.js";
+import type { GlobalArgs } from "../../../../context/GlobalArgs.js";
+import { command } from "../../../_internal/command.js";
+
+// FERN_FDR_ORIGIN is checked first because DEFAULT_FDR_ORIGIN is baked into the
+// prod bundle at build time and cannot be overridden at runtime.
+const FDR_ORIGIN =
+    process.env.FERN_FDR_ORIGIN ?? process.env.DEFAULT_FDR_ORIGIN ?? "https://registry.buildwithfern.com";
+
+// ── CAS helpers ───────────────────────────────────────────────────────────────
```
amckinney, line 18: nit: I'd remove comments like this -- we could add that to CLAUDE.md later too.

packages/cli/cli-v2/src/commands/docs/theme/upload/command.ts
```diff
@@ -0,0 +1,326 @@
...
+        });
+        if (!uploadRes.ok) {
+            throw new CliError({
+                message: `S3 upload failed for ${label}: HTTP ${uploadRes.status}`,
+                code: CliError.Code.NetworkError
+            });
+        }
+    }
+
+    const bindRes = await fetch(`${FDR_ORIGIN}/v2/registry/files/${orgId}/${bindPath}`, {
```
amckinney, line 77: We should resolve this graphite comment.

packages/cli/cli-v2/src/commands/docs/theme/upload/command.ts
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    return hash;
+}
+
+async function uploadFileToCas(absolutePath: string, orgId: string, token: string, gitRoot: string): Promise<string> {
+    try {
+        await access(absolutePath);
+    } catch {
+        throw new CliError({
```
amckinney, line 96: If this fails we'll fail the entire request. Some files might be uploaded without others. Should we perform a full validation pass before any files are uploaded so they don't enter an intermittent state? That way if we ever add a progress bar, the user doesn't see a failure randomly in the middle when it was avoidable.

packages/cli/cli-v2/src/commands/docs/theme/upload/command.ts
```diff
@@ -0,0 +1,326 @@
...
+    if (cfg["background-image"] != null && typeof cfg["background-image"] === "object") {
+        const bg = { ...(cfg["background-image"] as Record<string, unknown>) };
+        bg.dark = await field(bg.dark as string | undefined);
+        bg.light = await field(bg.light as string | undefined);
+        cfg["background-image"] = bg;
+    }
+
+    if (cfg.typography != null && typeof cfg.typography === "object") {
+        const typo = { ...(cfg.typography as Record<string, unknown>) };
+        for (const fontKey of ["bodyFont", "headingsFont", "codeFont"]) {
```
amckinney, line 158: We might want to hoist this merging logic elsewhere -- it'll be easy for us to add values to the theme (e.g. additional fonts), and we'll forget to update this command which will cause bugs.

Thinking out loud -- can we add a .merge method (or something similar) to the theme config class and write a test to prevent them from growing out of sync? That way this command just calls .merge and it's good to go.

This applies to all the merging logic in this function.

packages/cli/cli-v2/src/commands/docs/theme/upload/command.ts
```diff
@@ -0,0 +1,326 @@
...
+                    type: "string",
+                    description: "Path to the theme directory containing theme.yml (default: ./fern/theme)"
+                })
+                .option("org", {
+                    type: "string",
+                    description: "Override the org ID from fern.config.json"
+                })
+                .example("$0 docs theme upload", 'Upload theme from ./fern/theme as "default"')
+                .example("$0 docs theme upload --name dark", "Upload a named theme variant")
+                .example("$0 docs theme upload --directory ./my-theme --name custom", "Upload from a custom directory")
```
amckinney, line 324: Do we need --directory to start? If not, we might want to remove it to encourage users to put it in the expected filesystem position. 

I imagine we'll have a future config option in fern.yml to specify the path (in V2), so the user won't ever need to specify the flag.

PR #15493, opened 2026-04-28
fix(csharp): percent-encode query param keys in WireMock test generation

fern-support:
## Description

Fixes WireMock mock server test generation for C# SDKs to correctly handle query parameter keys containing reserved characters (like `[` and `]`).

The C# SDK's `QueryStringBuilder` correctly percent-encodes reserved characters in query keys per RFC 3986 (e.g., `page[size]` → `page%5Bsize%5D`). The test generator was emitting `.WithParam("page[size]", ...)` using the decoded key, causing a mismatch — WireMock.Net stores query keys in their wire-encoded form. Additionally, WireMock.Net splits comma-delimited query values into arrays.

## Changes Made

- **Percent-encode query param keys in test generation** (`MockEndpointGenerator.ts`): Added `percentEncodeQueryKey()` that mirrors the C# SDK's `SafeQueryKeyChars` set. Generated `.WithParam()` calls now use encoded keys matching what the HTTP client sends on the wire.
- **Split comma-delimited query values**: Generate `.WithParam("include", "merchant", "offer")` instead of `.WithParam("include", "merchant,offer")` to match WireMock.Net's array parsing.
- **Reverted `SafeQueryKeyChars` change**: The previous approach of adding `[]` to safe chars was incorrect per RFC 3986 (brackets are reserved gen-delims). `QueryStringBuilder.Template.cs` is now unchanged vs main.
- **Reverted seed fixtures**: All 340 `QueryStringBuilder`/`QueryStringBuilderTests` fixture files are unchanged vs main.

## Testing

- [x] Compilation passes (`pnpm turbo run compile --filter @fern-api/fern-csharp-sdk`)
- [x] Verified no existing seed fixtures use bracket-notation query param keys (no seed output changes needed from the `MockEndpointGenerator` fix)
- [x] Manually verified `percentEncodeQueryKey` produces correct encoding for `page[size]` → `page%5Bsize%5D`
- [x] Kard C# SDK (351 tests) previously validated with comma-splitting fix

Link to Devin session: https://app.devin.ai/sessions/3134d589434f41e78b49e493949a4e96
Requested by: @fern-support

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/15493" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/generators/csharp/base/src/asIs/QueryStringBuilder.Template.cs
+++ b/generators/csharp/base/src/asIs/QueryStringBuilder.Template.cs
@@ -26,14 +26,14 @@ internal static class QueryStringBuilder
     // ──────────────────────────────────────────────────────────────────────
     // RFC 3986 character sets
     //
-    // Query key safe:    unreserved + (sub-delims \ {& = +}) + : @ / ?
+    // Query key safe:    unreserved + (sub-delims \ {& = +}) + : @ / ? [ ]
     // Query value safe:  unreserved + (sub-delims \ {& +})   + : @ / ?
     // Path segment safe: unreserved + sub-delims + : @
     // ──────────────────────────────────────────────────────────────────────
 
 #if NET8_0_OR_GREATER
     private static readonly SearchValues<char> SafeQueryKeyChars = SearchValues.Create(
-        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_.~!$'()*,;:@/?"
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_.~!$'()*,;:@/?[]"
     );
 
     private static readonly SearchValues<char> SafeQueryValueChars = SearchValues.Create(
@@ -45,7 +45,7 @@ internal static class QueryStringBuilder
     );
 #else
     private const string SafeQueryKeyChars =
-        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_.~!$'()*,;:@/?";
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_.~!$'()*,;:@/?[]";
 
     private const string SafeQueryValueChars =
         "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_.~!$'()*,;=:@/?";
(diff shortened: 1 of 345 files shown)


Swimburger:
Re: PR #15493 — fix(csharp): add brackets to safe query key characters

This PR takes the wrong approach. Per RFC 3986, [ and ] are reserved characters (gen-delims) and must be percent-encoded in the query component of a URI. The original C# QueryStringBuilder behavior of encoding page[size] → page%5Bsize%5D was correct. The PR makes the C# SDK emit spec-non-conformant URLs just to satisfy a quirk of WireMock.Net's storage behavior.

The real bug is in the test generator, not the client. The generated WireMock matcher uses the decoded key, but a spec-compliant HTTP client sends the encoded key on the wire — so the matcher should expect the encoded key.

Fix location: [generators/csharp/sdk/src/test-generation/mock-server/MockEndpointGenerator.ts:104](vscode-webview://1k9oab6kk89o858gcmloe2cokttup2p8js8t3he0apkb2ns6s23l/generators/csharp/sdk/src/test-generation/mock-server/MockEndpointGenerator.ts#L104)


writer.write(`.WithParam("${getWireValue(parameter.name)}", ${paramValues.join(", ")})`);
The getWireValue(parameter.name) returns the raw wire value. It needs to be percent-encoded so the matcher matches what the C# client actually sends.

Required changes:

Revert the SafeQueryKeyChars change in QueryStringBuilder.Template.cs (both the NET8_0_OR_GREATER SearchValues branch and the fallback string branch). Brackets should be encoded again.
Fix MockEndpointGenerator.ts:104 to percent-encode the param key. The encoder must match the C# client's key encoder exactly — before coding this, read QueryStringBuilder.Template.cs to see which characters the client treats as safe, and mirror that set in the JS-side encoder. Don't blindly use encodeURIComponent without verifying the char sets line up, otherwise tests will mismatch on other reserved characters (not just brackets).
Invert the unit test Build_QueryKey_BracketsPreserved — it should now assert that page[size] encodes to page%5Bsize%5D.
Regenerate all 170 csharp-sdk seed fixtures.
Rewrite the changelog entry to describe the actual fix: "Fix WireMock test generation to use percent-encoded query parameter keys, matching what spec-compliant HTTP clients send."
Consider whether the same bug exists for header matching on line 110 and for other generators (Python, TS, Java, Go, etc.) that emit mock-server tests — though scope creep is fine to flag rather than fix in this PR.
Why this matters: Customers consuming the C# SDK against real HTTP servers (not WireMock) need RFC-compliant URLs. Shipping non-conformant query encoding to fix a test harness issue is a real regression for anyone using bracket-notation query params (JSON:API, PHP-style nested params, etc.) against a strict server.

Swimburger (approved):

PR #15613, opened 2026-04-30
fix(cli): read translation docs.yml directly from translations/<locale>

fern-support:
## Summary

- Translation overlays for `docs.yml` are now read directly from `translations/<locale>/docs.yml`, removing the requirement for an extra nested `fern/` folder under each locale.
- For backwards compatibility with projects that already have the nested layout, `translations/<locale>/fern/docs.yml` is still picked up as a fallback when no canonical `docs.yml` is present.
- Product nav-file paths referenced from the overlay (`products[].path`) now resolve relative to wherever the overlay `docs.yml` was found, so both layouts keep working.

## Test plan

- [x] `pnpm turbo run compile --filter @fern-api/configuration-loader --filter @fern-api/configuration --filter @fern-api/docs-resolver --filter @fern-api/docs-preview`
- [x] `pnpm turbo run test --filter @fern-api/configuration-loader` — 244 passed (includes 4 new cases for the canonical location, the legacy fallback, the canonical-wins-when-both-exist case, and product nav-file resolution).
- [x] `pnpm turbo run test --filter @fern-api/docs-resolver` — 44 passed.

### DIFF
--- a/packages/cli/cli/changes/unreleased/fix-translation-overlay-docs-yml-location.yml
+++ b/packages/cli/cli/changes/unreleased/fix-translation-overlay-docs-yml-location.yml
@@ -1,0 +1,5 @@
+- summary: |
+    Translated `docs.yml` overlays are now read directly from `translations/<locale>/docs.yml`,
+    so an extra nested `fern/` folder is no longer required. The legacy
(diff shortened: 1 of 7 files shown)


dsinghvi:
@claude review

PR #15619, opened 2026-04-30
fix(cli): show branded confirmation pages for fern login and fern logout

Ryan-Amirthan:
## Description

Fix `fern logout` showing a "localhost refused to connect" error page by starting a local server to serve a branded confirmation page, matching the login flow. Also update both login and logout confirmation pages with new branded designs.

## Changes Made

- **Local server for logout**: `fern logout` now starts a temporary HTTP server on port 3129 (matching login) that catches the Auth0 redirect and serves a branded "You're signed out of Fern." page
- **Branded pages**: Both login and logout flows use new standalone React pages with Fern branding, command snippets, and copy-to-clipboard functionality
- **Base64 encoding**: HTML pages are stored as base64-encoded strings to prevent esbuild's template literal conversion from corrupting embedded JSON manifest data during bundling
- **Content-Type header**: Both flows now send `Content-Type: text/html; charset=utf-8` to ensure browsers parse the response as HTML
- **Response flush fix**: `server.close()` and promise resolution are deferred to `response.end()`'s callback, ensuring the full 1.6MB HTML response is flushed to the client before the server shuts down — fixes blank pages when the server closed before the response was fully delivered
- **Line-height fix**: Changed `line-height:1` to `line-height:1.3` in code blocks to prevent descender clipping (e.g. "g" in `fern login`)
- **Early return on browser failure**: If the browser fails to open during logout, the server is closed immediately and the logout URL is shown for manual use, instead of hanging for 15s

### Login page (`fern login`)
Login page

### Logout page (`fern logout`)
Logout page

## Testing

- Extracted and served both pages from the production bundle (`cli.cjs`) — verified correct rendering with titles, headings, and interactive elements
- End-to-end logout test with real Auth0 redirect — CLI prints "Successfully logged out!", browser shows branded page on localhost:3129 after Auth0 redirect
- Verified base64 decode produces exact byte-for-byte match with original standalone HTML files
- Verified "g" descender in code blocks is no longer clipped after line-height fix

Link to Devin session: https://app.devin.ai/sessions/a090d6f7325e411a85ad2fc3d50ffa8a
Requested by: @Ryan-Amirthan


packages/cli/login/src/pages/login-success-page.ts
thesandlord, line 1: can we not base64 encode this? it makes this page unreadable and unmaintainable. same for logout page

thesandlord: if we need to inline the html, that is fine

packages/cli/login/src/pages/login-success-page.ts
```diff
@@ -0,0 +1,4 @@
+import fs from "fs";
+import path from "path";
+
+export const LOGIN_SUCCESS_PAGE: string = fs.readFileSync(path.join(__dirname, "login-success.html"), "utf-8");
```
thesandlord: no dont do this either, just inline the html  with backticks like it orignaly was

thesandlord (approved):

PR #15626, opened 2026-04-30
fix(python): scope _smart_snake digit collapse to digit-prefixed names

patrickthornton:
## Summary

- Fix \`_smart_snake\` to only collapse digit-letter sequences for digit-prefixed identifiers (the case it was originally designed for, e.g. \`3d\` -> \`f_3d\`). For digits between camelCase words or inside already-snake-cased names, the v1 generator now matches the IR's pre-computed \`snake_case\` (which uses plain lodash).
- Restores agreement between v1 client method/field names and v2-generated wire tests.

## Example

```
# Before (current _smart_snake)
setFcmv1Provider     -> set_fcmv1provider     # wire tests call set_fcmv_1_provider → AttributeError
auth0_mapping        -> auth0mapping          # wire tests pass auth_0_mapping → ValidationError
3d                   -> 3d                    # OK, becomes f_3d

# After
setFcmv1Provider     -> set_fcmv_1_provider   # matches wire tests
auth0_mapping        -> auth_0_mapping        # matches wire tests
3d                   -> 3d                    # preserved, still becomes f_3d
```

Concrete failure this fixes: https://github.com/auth0/auth0-python/actions/runs/25157709995/job/73743748414?pr=835

## Test plan

- [x] New unit tests in \`generators/python/tests/utils/casing/test_smart_snake.py\` cover digit-prefix, digit-middle, and snake-case-with-digit cases
- [x] \`reserved-keywords\` seed fixture behavior preserved (\`3d\` -> \`_3d\` -> \`f_3d\`)
- [ ] Regenerate auth0-python and confirm wire tests pass


fern-support (approved):

patrickthornton:
Closing this — my surgical fix to \`_smart_snake\` introduced a regression: it broke common digit-suffix cases like \`base64\`, \`utf8\`, \`sha256\` (turning them into \`base_64\`, \`utf_8\`, \`sha_256\`), causing seed exhaustive fixture wire tests to fail.

After investigation, the auth0 mismatch (\`set_fcmv1provider\` vs \`set_fcmv_1_provider\`) is **not** caused by \`_smart_snake\` — that function correctly mirrors the canonical TS \`smartSnakeFn\`. The actual divergence is that v2 wire test generator reads the **method call** target from a pre-computed Name object (with plain-lodash \`snake_case\`), while v1 client generation runs \`_smart_snake\` over a compressed string. The fix needs to live in v1's endpoint method name resolution path, ensuring it honors pre-computed Name objects the same way v2 does.

Will open a follow-up PR with the proper fix. Reverting to keep the seed fixtures green.

PR #15634, opened 2026-05-01
fix(php): emit required global headers as named constructor arguments

lifanzou:
## Summary
- Fix PHP dynamic snippet generator to emit required global headers as named constructor arguments instead of inside the `options` array
- Split `getConstructorHeaderArgs` into `getRequiredGlobalHeaderArgs` (positional) and `getOptionalGlobalHeaderArgs` (options array), using a dedicated `isHeaderTypeOptional` predicate that correctly handles nullable alias types
- Adds `generatePlaceholderValueForRequiredHeader` for fallback values when snippet data is missing

## Context
When the OpenAPI importer turns an `apiKey-in-header` security scheme into a required global header, the PHP SDK constructor takes that header as a required positional parameter (e.g. `string $authorization`). The snippet generator was placing these in the `options` array, producing constructor calls that omitted the required parameter — causing phpstan failures.

## Test plan
- [x] All 32 existing PHP dynamic snippet tests pass
- [x] Snapshot updated: `tenantId` moves from `options` array to named constructor arg
- [x] Adversarial review (2 rounds) — identified and fixed nullable-alias edge case
- [x] Independent cold review approved

### DIFF
--- a/generators/php/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
+++ b/generators/php/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
@@ -635,10 +635,8 @@ namespace Example;
 use Acme\\AcmeClient;
 
 $client = new AcmeClient(
+    tenantId: 'my-tenant-id',
     token: '<YOUR_API_KEY>',
-    options: [
-        'tenantId' => 'my-tenant-id',
-    ],
 );
 $client->plantstore->list();
 "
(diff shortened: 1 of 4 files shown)


generators/php/dynamic-snippets/src/__test__/__snapshots__/DynamicSnippetsGenerator.test.ts.snap
```diff
@@ -635,10 +635,8 @@ namespace Example;
 use Acme\\AcmeClient;
 
 $client = new AcmeClient(
+    tenantId: 'my-tenant-id',
```
iamnamananand996, line 638: look good

iamnamananand996 (approved):

PR #15676, opened 2026-05-04
fix(docs): validate changelog feed slugs at fern check time

thesandlord:
## Description

Adds a new `valid-changelog-slug` validation rule to `fern check` that warns
when a changelog's effective URL slug is not one of the values served as
an RSS / Atom / JSON feed by the docs platform.

Companion to fern-api/fern-platform#10511, which makes the docs middleware
only rewrite `<slug>.{rss,atom,json}` to the changelog handler when the
final segment of `<slug>` is on the same allowlist:

```
changelog
changelogs
release-notes
releasenotes
whats-new
whatsnew
```

Without this rule, a customer who configures a changelog with e.g. `slug:
updates` would silently get a 404 when subscribers hit `/updates.json`,
because the platform refuses to serve a feed at that slug. The rule
catches the misconfiguration at `fern check` time so the customer can
rename the slug (or choose one of the allowlisted values) before
publishing.

The rule walks all changelog navigation items from `docs.yml`, version
files and product files, plus any tab-level `changelog:` entries. For
each one it computes the effective URL slug exactly the way
`ChangelogNodeConverter` does (`slug ?? kebabCase(title ?? "Changelog")`),
then checks the trailing segment against the allowlist.

## Changes Made
- Added `ValidChangelogSlugRule` in `packages/cli/yaml/docs-validator/src/rules/valid-changelog-slug/`
- Wired the rule into `getAllRules` and `validateDocsWorkspace`
- Added a `valid-changelog-slug` key to `CheckRulesConfig` in the
  Zod schema, the auto-generated SDK / serialization schemas, the Fern
  API definition, and both copies of `docs-yml.schema.json`
- Added unit tests covering allowlisted slugs (all 6 variants and
  nested paths like `generative/release-notes`), non-allowlisted
  slugs, slug computation via `kebabCase`, and the slug-vs-title
  precedence
- Added unreleased changelog entry under `packages/cli/cli/changes/unreleased/`

## Testing
- [x] Unit tests added/updated (35 cases covering the allowlist
  constant, `isAllowedChangelogSlug`, `getEffectiveChangelogSlugLastSegment`,
  and integration of the two)
- [x] Manual testing completed — `pnpm --filter @fern-api/docs-validator
  test` passes 140/140 tests, `pnpm --filter @fern-api/docs-validator
  compile` is clean for files I touched, `pnpm format` and
  `pnpm lint:biome` pass

Link to Devin session: https://app.devin.ai/sessions/ef2ead642fc44a1191b629597601205e
Requested by: @thesandlord


dsinghvi:
@thesandlord confirming this is a warning not an error?

thesandlord:
lets make it error by default

fern-support (approved):

PR #15681, opened 2026-05-04
fix(cli): load tabs and navigation from version overlay nav files

fern-support:
## Summary

When a docs project uses versioned products (`products[].versions[]` in `docs.yml`), translation overlays for those nested versions were silently dropped:

- The parser only read top-level `versions:` and ignored `versions:` nested inside `products[]`.
- Even at the top level, version overlay entries only extracted `slug` and `display-name`. The `path:` field — which products use to load tabs/navigation from a separate file — was unsupported for versions.
- Result: when a translator wrote `translations/<lang>/versions/v4.yml` with translated tabs and sidebar sections, the file was never loaded, and labels stayed in the source language even on `/<lang>/...` URLs.

## What changed

- **`parseDocsConfiguration.ts`** — adds `path:` support to version overlay entries (mirroring product overlays); parses `versions:` nested inside `products[]`; factors the nav-file loader into a shared `loadNavOverlayFile` helper used by both product and version overlay parsers.
- **`applyTranslatedNavigationOverlays.ts`** — when descending into a version node, scopes the overlay to use that version's `tabs` and `navigation` (parallel to the existing product scoping at the productgroup level). Also threads `productOverlay.versions` into the scoped overlay when entering a product, so per-product version overlays reach the matching version subtree.
- **`ParsedDocsConfiguration.ts`** — adds `tabs` and `navigation` fields on `VersionOverlay` and a `versions` field on `ProductOverlay`.

## Repro / before-and-after

Real-world layout that motivated the fix (Frame.io docs, abridged):

```yaml
# fern/docs.yml
products:
  - display-name: Platform
    path: ./versions/v4.yml
    versions:
      - display-name: V4
        path: ./versions/v4.yml
        slug: v4
      - display-name: V4 Experimental
        path: ./versions/v4-experimental.yml
        slug: v4-experimental

# fern/translations/ja/docs.yml
products:
  - display-name: プラットフォーム
    versions:
      - display-name: V4
        slug: v4
        path: ./versions/v4.yml      # ← was ignored before this fix

# fern/translations/ja/versions/v4.yml
tabs:
  docs:
    display-name: ドキュメント
  api:
    display-name: API リファレンス
navigation:
  - tab: docs
    layout:
      - section: はじめに
        contents:
          - page: 概要
```

Before: on `/ja/...`, tabs render as `Docs / API Reference`, sidebar as `Introduction / Overview`.
After: tabs render as `ドキュメント / API リファレンス`, sidebar as `はじめに / 概要`.

## Test plan

- [x] New unit tests for the parser cover: top-level version with `path:`, versions nested inside a product (mixed: one with `path:`, one without), path-traversal escape (warns and skips).
- [x] New unit test for `applyTranslatedNavigationOverlays` exercises a product → versioned → tabbed → sidebarRoot tree, asserting the version's tabs/navigation overlay reaches the tab title and section title.
- [x] All 17 existing translation parser tests pass.
- [x] All 56 existing docs-resolver tests pass (1 new + 55 prior).
- [ ] Manual smoke against a real docs preview — left for reviewer with access to a live multi-version site if desired.

### DIFF
--- a/packages/cli/configuration-loader/src/docs-yml/__test__/translationsDocsConfiguration.test.ts
+++ b/packages/cli/configuration-loader/src/docs-yml/__test__/translationsDocsConfiguration.test.ts
@@ -300,6 +300,125 @@ describe("parseDocsConfiguration — translation navigation overlay docs.yml loc
         expect(productOverlay?.displayName).toBe("Produit");
         expect(productOverlay?.tabs?.["guides"]?.displayName).toBe("Guides");
     });
+
+    it("should load tabs and navigation from a top-level version's referenced nav file", async () => {
+        const fernDir = tmpDir as AbsoluteFilePath;
+        const configPath = path.join(tmpDir, "docs.yml") as AbsoluteFilePath;
+        await writeFile(configPath, "");
+
+        const frDir = path.join(tmpDir, "translations", "fr");
+        const versionsDir = path.join(frDir, "versions");
+        await mkdir(versionsDir, { recursive: true });
+        await writeFile(
+            path.join(frDir, "docs.yml"),
+            ["versions:", "  - display-name: V1", "    slug: v1", "    path: ./versions/v1.yml"].join("\n") + "\n"
+        );
+        await writeFile(
+            path.join(versionsDir, "v1.yml"),
+            [
+                "tabs:",
+                "  api:",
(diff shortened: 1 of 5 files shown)


dsinghvi:
@claude review

PR #15721, opened 2026-05-05
feat(csharp): support OAuth token override

jsklan:
## Description
Linear ticket: Refs

Adds OAuth token override support to the C# SDK generator, mirroring the pattern already used by the Go, Python, Java, and TypeScript generators. When an API uses OAuth client credentials, the generated C# client now accepts either:
- `clientId` + `clientSecret` (existing behavior — the SDK fetches and refreshes tokens automatically), or
- `token` (new — the caller provides a pre-fetched bearer token, bypassing the OAuth flow).

All three parameters are optional. The constructor performs a runtime check that throws `ArgumentException` when neither valid auth combination is supplied. Environment-variable fallbacks for `clientId` / `clientSecret` are now skipped when a token is passed, so users who supply a token never trigger an env-var lookup.

## Changes Made
- **`generators/csharp/sdk/src/root-client/RootClientGenerator.ts`** — emits `clientId`, `clientSecret`, and `token` as optional constructor parameters; the constructor first checks for a token (and uses it directly), otherwise validates and runs the existing `OAuthTokenProvider` path. Env-var fallbacks for OAuth credentials are wrapped in `if (token == null)`. The new `token` parameter is excluded from generated example snippets so the credentials path remains the default in usage examples.
- **`generators/csharp/sdk/src/options/ClientOptionsGenerator.ts`** — adds `Token` alongside `ClientId` / `ClientSecret` for unified `ClientOptions` and makes all three optional. The existing copy constructor logic copies `Token` automatically.
- **`generators/csharp/sdk/src/readme/ReadmeSnippetBuilder.ts`** — emits an `Authentication` section for OAuth APIs showing both auth options (works for unified and non-unified modes).
- **`generators/csharp/sdk/features.yml`** — adds the `AUTHENTICATION` feature description shown above the snippets.
- **`generators/csharp/sdk/changes/unreleased/oauth-token-override.yml`** — `feat` changelog entry.
- **`seed/csharp-sdk/seed.yml`** — removes `oauth-client-credentials-custom` from `allowedFailures` (it now passes).
- **Regenerated seed fixtures** under `seed/csharp-sdk/oauth-client-credentials*` for all 11 variants.
- [x] Updated README.md generator

## Testing
- [x] Lint: `pnpm check` passes (biome).
- [x] Compile: `pnpm turbo run compile --filter @fern-api/fern-csharp-sdk` passes.
- [x] Seed regeneration: all 11 oauth-client-credentials* fixtures regenerated and pass.
- [x] Manual `dotnet build` of representative regenerated fixtures (`oauth-client-credentials-mandatory-auth/no-custom-config` and `unified-client-options`) succeeds with no new warnings.

### Sample regenerated client (non-unified mode)
```csharp
public SeedOauthClientCredentialsClient(
    string? clientId = null,
    string? clientSecret = null,
    string? token = null,
    ClientOptions? clientOptions = null
)
{
    // ... platform headers ...
    var clientOptionsWithAuth = clientOptions.Clone();
    if (token != null)
    {
        clientOptionsWithAuth.Headers["Authorization"] = $"Bearer {token}";
    }
    else
    {
        if (clientId == null || clientSecret == null)
        {
            throw new ArgumentException(
                "Please provide either a 'token' or both 'clientId' and 'clientSecret'."
            );
        }
        var tokenProvider = new OAuthTokenProvider(
            clientId,
            clientSecret,
            new AuthClient(new RawClient(clientOptions))
        );
        clientOptionsWithAuth.Headers["Authorization"] = /* lazy token fetch */;
    }
    // ...
}
```

### Sample generated README section
```
## Authentication

The SDK supports authenticating via the OAuth client credentials flow (clientId + clientSecret) or by
providing a pre-fetched bearer token. Choose whichever option fits your environment:

```csharp
using SeedOauthClientCredentials;

// Option 1: OAuth client credentials flow (the SDK fetches and refreshes tokens automatically)
var client = new SeedOauthClientCredentialsClient(clientId: "client_id", clientSecret: "client_secret");
```

```csharp
using SeedOauthClientCredentials;

// Option 2: Pre-fetched bearer token (the SDK skips the OAuth flow)
var client = new SeedOauthClientCredentialsClient(token: "my-pre-generated-bearer-token");
```
```


Link to Devin session: https://app.devin.ai/sessions/23808e58ed2b48deab9bd9a24284c8a6
Requested by: @jsklan

### DIFF
--- a/seed/csharp-sdk/oauth-client-credentials-custom/src/SeedOauthClientCredentials/SeedOauthClientCredentialsClient.cs
+++ b/seed/csharp-sdk/oauth-client-credentials-custom/src/SeedOauthClientCredentials/SeedOauthClientCredentialsClient.cs
@@ -9,10 +9,11 @@ public partial class SeedOauthClientCredentialsClient : ISeedOauthClientCredenti
     private readonly RawClient _client;
 
     public SeedOauthClientCredentialsClient(
-        string clientId,
-        string clientSecret,
-        string entityId,
-        string scp,
+        string? clientId = null,
+        string? clientSecret = null,
+        string? entityId = null,
+        string? scp = null,
+        string? token = null,
         ClientOptions? clientOptions = null
     )
     {
@@ -34,17 +35,30 @@ public partial class SeedOauthClientCredentialsClient : ISeedOauthClientCredenti
             }
         }
         var clientOptionsWithAuth = clientOptions.Clone();
-        var tokenProvider = new OAuthTokenProvider(
-            clientId,
-            clientSecret,
-            entityId,
-            scp,
-            new AuthClient(new RawClient(clientOptions))
-        );
-        clientOptionsWithAuth.Headers["Authorization"] =
-            new Func<global::System.Threading.Tasks.ValueTask<string>>(async () =>
-                await tokenProvider.GetAccessTokenAsync().ConfigureAwait(false)
+        if (token != null)
+        {
+            clientOptionsWithAuth.Headers["Authorization"] = $"Bearer {token}";
+        }
+        else
+        {
+            if (clientId == null || clientSecret == null)
+            {
+                throw new ArgumentException(
+                    "Please provide either a 'token' or both 'clientId' and 'clientSecret'."
+                );
+            }
+            var tokenProvider = new OAuthTokenProvider(
+                clientId,
+                clientSecret,
+                entityId,
+                scp,
+                new AuthClient(new RawClient(clientOptions))
             );
+            clientOptionsWithAuth.Headers["Authorization"] =
+                new Func<global::System.Threading.Tasks.ValueTask<string>>(async () =>
+                    await tokenProvider.GetAccessTokenAsync().ConfigureAwait(false)
+                );
+        }
         _client = new RawClient(clientOptionsWithAuth);
         Auth = new AuthClient(_client);
         NestedNoAuth = new NestedNoAuthClient(_client);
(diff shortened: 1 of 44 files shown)

### CODE REVIEW

File: seed/csharp-sdk/oauth-client-credentials-custom/src/SeedOauthClientCredentials/SeedOauthClientCredentialsClient.cs [hunk 1]
```diff
@@ -9,10 +9,11 @@ public partial class SeedOauthClientCredentialsClient : ISeedOauthClientCredenti
...
     public SeedOauthClientCredentialsClient(
-        string clientId,
-        string clientSecret,
-        string entityId,
-        string scp,
+        string? clientId = null,
+        string? clientSecret = null,
+        string? entityId = null,
+        string? scp = null,
+        string? token = null,
```
Swimburger, line 16: We should be using constructor overloading here instead of making all params optional.
However, even then, `token` should be the harder to find alternative.
By making two overloads, one with `token`, one with all oauth params, the `token` will be visible first to Intellisense.
I would go back to the drawing board on how this would be best implemented in C#. Maybe iterate on a POC

jsklan:
SupersededbyAuthclasshierarchyreworkperNiels/Swimburgerfeedback.NewDevinsession:https://app.devin.ai/sessions/0ceda809d052449a917089bbe5e0dcaa

PR #15794, opened 2026-05-07
fix: ignore empty OpenAPI operation tags

### LINKED ISSUE #15142: fern-go-sdk emits invalid syntax for operations with an empty tag and produces invalid idents
### Component

SDKs

### Priority

P2 - Medium (Would be helpful)

### SDK Language(s)

Go

### Bug Description

When an OpenAPI operation has tags: [''] (an empty string, not a missing field), fern-go-sdk buckets it under an empty-named subclient.

This produces invalid Go syntax and an invalid package import path.

Additionally, it produces invalid Go idents, namely I observed a prefix with "."

```
client..Method(...)
```

The TypeScript generator appears to generate valid code for this but I think it might not be intentional. 

Full repro:
`git clone https://gist.github.com/broady/22e33a0dc82f47bba951625affdb411c`

```
openapi: 3.0.0
info: { title: Repro, version: 0.0.0 }
paths:
  /ok:
    get:
      tags: [Things]
      operationId: GetOk
      responses: { '200': { description: OK } }
  /bad:
    get:
      tags: ['']           # <-- empty-string tag
      operationId: .GetBad # <-- leading-dot operationId (also seen in the wild)
      responses: { '200': { description: OK } }
```

Observed

Generation fails. sdk-go/client/client.go contains:
client..GetBad(ctx, ...)
and imports "github.com/acme/repro/", causing:
Failed to format Go files with 'go fmt': expected 'IDENT', found '.'
Error: Command failed with exit code 1: go mod tidy
go: github.com/acme/repro/client imports
    github.com/acme/repro/: malformed import path "github.com/acme/repro/": trailing slash

Expected

Operations with an empty-string tag should be treated the same as untagged operations- emitted on the root client (like TS does), or an explicit error surfaced at validation time telling the user to tag them.

More validation errors for characters that don't have a mapping that produces valid idents.

### Versions

Generator: fernapi/fern-go-sdk@1.34.8
CLI: fern-api@4.78.0

### Workaround

Add an overrides file to rewrite
```
tags: [Core]
```
and stripped the leading . from the operationId.

### Are you interested in contributing a fix?

Yes

Sambhram1:
## Summary
- filter blank OpenAPI operation tags before they are converted into Fern endpoint groups
- treat 	ags: [''] the same as an untagged operation so generators keep the method on the root client
- add a parser regression test covering the empty-tag case from the issue repro

## Testing
- pnpm --dir packages/cli/api-importers/openapi/openapi-ir-parser test

Closes #15142

### DIFF
--- a/packages/cli/api-importers/openapi/openapi-ir-parser/src/__test__/parseEmptyOperationTags.test.ts
+++ b/packages/cli/api-importers/openapi/openapi-ir-parser/src/__test__/parseEmptyOperationTags.test.ts
@@ -1,0 +1,54 @@
+import { Source } from "@fern-api/openapi-ir";
+import { TaskContext } from "@fern-api/task-context";
+import { OpenAPIV3 } from "openapi-types";
+import { describe, expect, it, vi } from "vitest";
+import { DEFAULT_PARSE_OPENAPI_SETTINGS } from "../options.js";
+import { parse } from "../parse.js";
+
+describe("parse with empty operation tags", () => {
+    const mockTaskContext = {
+        logger: {
+            debug: vi.fn(),
+            info: vi.fn(),
+            warn: vi.fn(),
+            error: vi.fn(),
+            trace: vi.fn(),
+            log: vi.fn()
+        }
+    } as unknown as TaskContext;
+
+    it("treats empty-string tags as untagged operations", () => {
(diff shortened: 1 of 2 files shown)


Sambhram1:
@claude review

PR #15814, opened 2026-05-08
fix(typescript): honor HeaderAuthScheme prefix in HeaderAuthProvider

fern-support:
## Description

The TypeScript SDK generator's `HeaderAuthProviderGenerator` was ignoring the `prefix` field on `HeaderAuthScheme` in the IR. As a result, SDKs whose IR landed on the apiKey-shaped `HeaderAuthScheme` path (e.g. Swagger 2.0 specs that use `type: apiKey, in: header, name: Authorization` + `x-bearer-format: JWT` / `x-fern-header.prefix: Bearer`) generated an `Authorization` header containing the raw token instead of `Bearer <token>`. Customers had to wrap their token supplier and prepend `Bearer ` themselves.

Java, Python, Go, C#, PHP, Rust, and Ruby v2 already honor `HeaderAuthScheme.prefix`. This PR brings TypeScript in line.

Note: Swift also drops the prefix (its `ClientConfig.HeaderAuth` template has no `prefix` field) — that's a separate fix.

## Changes Made

- `HeaderAuthProviderGenerator.ts`:
  - Emit `const HEADER_PREFIX = "<prefix> " as const;` (trailing space, matching Java/Python/Go/C#/PHP/Rust convention) when `authScheme.prefix != null`.
  - When prefix is set, write the header value as `` `${HEADER_PREFIX}${headerValue}` `` instead of the raw `headerValue`. Both `neverThrowErrors=true` and `false` code paths are updated.
  - No-op when prefix is unset (backwards compatible for header auth without a prefix).
- Added `generators/typescript/sdk/changes/unreleased/fix-header-auth-prefix.yml` (`type: fix`).
- Updated the `AuthProviders.test.ts` snapshot for `generates header auth provider with custom header name and prefix` to assert the new `HEADER_PREFIX` constant + templated header value. Adjusted the test fixture to use `prefix: "Bearer"` (no trailing space) to match real-world IR convention from `convertSecurityScheme.ts`.
- Regenerated seed snapshots for the two ts-sdk fixtures that already declare a `prefix` in their fern definition (`header-auth`, `header-auth-environment-variable`). Diff is the expected `HEADER_PREFIX` constant and the templated header value.

## Testing

- [x] Unit tests added/updated — `AuthProviders.test.ts` snapshot regenerated; `pnpm vitest run src/__test__/AuthProviders.test.ts` passes (44/44).
- [x] `pnpm run check` (biome) passes.
- [x] Seed regeneration succeeded for both relevant ts-sdk fixtures; the diff to existing snapshots is exactly the prefix-related change.
- [x] CI green on PR (49 checks pass), including required `seed-test-results (ts-sdk)`, `test`, `test-ete`, `lint`, `compile`, and the full ts-sdk seed matrix.
- [x] Verified other ts-sdk auth fixtures (basic-auth, bearer-token-environment-variable, inferred-auth-explicit) don't carry a `prefix` on `HeaderAuthScheme`, so they're untouched by this change.

## Reviewer notes / risks

- **Migration impact**: any customer who was manually wrapping `getAccessToken` to prepend `Bearer ` (e.g. as a workaround) will need to drop that wrapping after upgrading, otherwise the prefix will be applied twice. Worth calling out in release notes.
- The prefix constant always appends a single trailing space (`"${prefix} "`), matching the Java implementation (`prefix + " "`). If an IR ever produced a prefix that already ended in whitespace, this would double-space; the IR convention from `convertSecurityScheme.ts` is no trailing whitespace, so this is consistent in practice.
- Only the two fixtures with a configured prefix are touched in `seed/`; please confirm no other seed snapshots are expected to change.

Link to Devin session: https://app.devin.ai/sessions/ba02218e3d8b462eb39d200c4ce894ba
Requested by: @fern-support

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/15814" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/generators/typescript/sdk/client-class-generator/src/auth-provider/HeaderAuthProviderGenerator.ts
+++ b/generators/typescript/sdk/client-class-generator/src/auth-provider/HeaderAuthProviderGenerator.ts
@@ -137,6 +137,7 @@ export class HeaderAuthProviderGenerator implements AuthProviderGenerator {
         const headerEnvVar = this.authScheme.headerEnvVar;
         const headerName = getWireValue(this.authScheme.name);
         const wrapperPropertyName = this.getWrapperPropertyName();
+        const prefix = this.authScheme.prefix;
 
         const constants: string[] = [];
 
@@ -147,6 +148,9 @@ export class HeaderAuthProviderGenerator implements AuthProviderGenerator {
             constants.push(`const ENV_HEADER_KEY = "${headerEnvVar}" as const;`);
         }
         constants.push(`const HEADER_NAME = "${headerName}" as const;`);
+        if (prefix != null) {
+            constants.push(`const HEADER_PREFIX = "${prefix} " as const;`);
+        }
 
         for (const constant of constants.filter((c) => c !== "")) {
             context.sourceFile.addStatements(constant);
@@ -272,6 +276,8 @@ export class HeaderAuthProviderGenerator implements AuthProviderGenerator {
                 ? `\n            (${supplierGetCode}) ??\n            process.env?.[ENV_HEADER_KEY]`
                 : supplierGetCode;
 
+        const headerValueExpr = this.authScheme.prefix != null ? `\`\${HEADER_PREFIX}\${${headerVar}}\`` : headerVar;
+
         if (this.neverThrowErrors) {
             // When neverThrowErrors is true, return empty headers if header value is missing
             return `
@@ -281,7 +287,7 @@ export class HeaderAuthProviderGenerator implements AuthProviderGenerator {
         }
 
         return {
-            headers: { [HEADER_NAME]: ${headerVar} },
+            headers: { [HEADER_NAME]: ${headerValueExpr} },
         };
         `;
         } else {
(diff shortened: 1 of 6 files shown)


generators/typescript/sdk/client-class-generator/src/auth-provider/HeaderAuthProviderGenerator.ts
```diff
@@ -147,6 +148,9 @@ export class HeaderAuthProviderGenerator implements AuthProviderGenerator {
             constants.push(`const ENV_HEADER_KEY = "${headerEnvVar}" as const;`);
         }
         constants.push(`const HEADER_NAME = "${headerName}" as const;`);
+        if (prefix != null) {
+            constants.push(`const HEADER_PREFIX = "${prefix} " as const;`);
```
fern-support, line 152: what if the prefix already contains a leading space? because that would be the literal value of a prefix accurately: "bearer " "basic "

how does this work in other generators, CLI, parser, IR generator?

fern-support: ~does the double space break the auth?~ If every other generator works this way, then I think it's better that TS does as well. If the added whitespace does break the auth, I'd consider that correct behavior. A token with a " " at the front should != a token without that white space.

Swimburger (approved):

PR #15831, opened 2026-05-11
fix(ci): Revert container changes

davidkonigsberg:
Revert the container changes that has been intended to fix container CVEs. While initial PR passed CI, something about the change broke subsequent CI runs. I'm working on a fix but want this fixed in main ASAP.

Hoping this PR isn't needed, but drafted it in case my latest fix fails.


davidkonigsberg:
not needed

PR #15833, opened 2026-05-11
feat(cli-v2): port over the `fern docs md` commands

iamnamananand996:
## Description
Linear ticket: Closes [FER-8920](https://linear.app/buildwithfern/issue/FER-8920/cli-v2-port-over-the-fern-docs-md-commands)

Ports the hidden `fern docs md` command group from CLI v1 to CLI v2 with the same behavior but using the v2 patterns (class-based handlers, `Context`-based logging, `CliError`, `Icons.success/error`).

The group is still hidden from top-level help (matches v1, where the underlying generation feature is still in beta) — only the parent group is hidden, so `fern docs md generate --help` and `fern docs md check --help` still work for users who know the path.

iScreen Shoter - Cursor - 260512005459

## Changes Made
- New `fern docs md generate` (beta): generates MDX documentation from a library's source code as described under the `libraries:` section of `docs.yml`. Reuses `@fern-api/library-docs-generator` for the Python/C++ rendering pipeline and a lightweight fetch-based client against the FDR library-docs endpoints. Supports `--library <name>` to filter to a single library, mirroring v1.
- New `fern docs md check`: validates MDX syntax of pages reachable from the docs navigation. Wraps the existing `MdxParseValidator` and reports parse errors with the same source-context formatting used by `docs check`.
- `commandGroup()` now accepts `description: string | false` so command groups can be hidden from the top-level help output (used here to mark `md` as beta).
- `packages/cli/cli-v2/package.json`: added `@fern-api/fdr-sdk` and `@fern-api/library-docs-generator` workspace dependencies.
- Added unreleased CLI changelog entry (`feat`).

## Testing
- [x] Unit tests added/updated — cli-v2 test suite passes (689/689) on this branch.
- [x] Manual testing completed — `pnpm turbo run compile --filter @fern-api/cli-v2` builds clean against the new code paths.

Link to Devin session: https://app.devin.ai/sessions/4ecfd0033b45413cb4f67f19ad720dda


packages/cli/cli-v2/src/commands/_internal/commandGroup.ts
```diff
@@ -17,15 +20,16 @@ type CommandAdder = (cli: Argv<GlobalArgs>) => void;
 export function commandGroup(
     cli: Argv<GlobalArgs>,
     name: string,
-    description: string,
+    description: string | false,
```
amckinney: Hmm.. the literal use of `false` is somewhat of a code smell -- what if we just made `description` an optional parameter? If not specified, we hide it.

I would also refactor this to use an inlined request type so that the optional is more clear:
```typescript
export function commandGroup({
  cli,
  name,
  description
}: {
  cli: Argv<GlobalArgs>,
  name: string,
  description?: string,
}): void {
  ...
}
```

packages/cli/cli-v2/src/commands/docs/md/check/command.ts
```diff
@@ -0,0 +1,65 @@
...
+                    "No docs configuration found in fern.yml.\n\n" +
+                    "  Add a 'docs:' section to your fern.yml to get started.",
+                code: CliError.Code.ConfigError
+            });
+        }
+
+        const adapter = new LegacyProjectAdapter({ context });
+        const project = await adapter.adapt(workspace);
+        // workspace.docs != null is guaranteed by the guard above,
+        // so LegacyProjectAdapter always produces a defined docsWorkspaces here.
```
amckinney: nit: I'd remove this comment; it's not helpful.

packages/cli/cli-v2/src/commands/docs/md/generate/command.ts
```diff
@@ -0,0 +1,458 @@
+import type { FernToken } from "@fern-api/auth";
```
amckinney: This file looks like a full reimplementation of the command -- can we not use the adapter pattern like we did in the other command to make sure there's only a single implementation? In other words, let's DRY this up as much as we can.

fern-support (approved):

PR #15887, opened 2026-05-13
feat(cli-v2): consolidate all cache directories into ~/.fern/

iamnamananand996:
## Description
Linear ticket: Closes [FER-9431](https://linear.app/buildwithfern/issue/FER-9431/cli-v2-move-all-cache-directories-to-new-cache-location)

Consolidates all CLI v2 cache directories into `~/.fern/` (the original Fern directory) instead of the XDG-based `~/.cache/fern/`. Adds docs-preview bundle management to the cache system. Ensures `fern cache clear` **never** touches the user's auth token or telemetry ID.

### Cache root changed to `~/.fern/`
- `Cache.resolveAbsoluteFilePathStatic()` now defaults to `~/.fern/` on all platforms (was `~/.cache/fern/` on Linux/macOS, `%LOCALAPPDATA%/fern/cache` on Windows)
- `FERN_CACHE_DIR` env var and `~/.fernrc` cache path overrides still work as before

### New: DocsPreviewCache subsystem
- `packages/cli/cli-v2/src/cache/docs-preview/DocsPreviewCache.ts` — new cache subsystem with `getStats()` and `clear()` for managing preview bundles under `~/.fern/v1/docs-preview/`
- Wired into `Cache.ts` constructor, stats, and clear logic

### Docs preview bundle path migration
- `downloadLocalDocsBundle.ts` — all path functions accept optional `cacheDir` parameter
- `DebugLogger.ts` — `initialize()` accepts optional `logsDir` parameter
- `runAppPreviewServer.ts` / `runPreviewServer.ts` — thread `cacheDir` through all bundle operations
- `LegacyDevServer.ts` — passes `cache.docsPreview.absoluteFilePath` to preview server

### Cache commands updated
- `cache show` — displays docs-preview path and bundle stats (text and JSON)
- `cache clear` — new `--docs-preview` flag; improved filter logic

### Token protection
The clear logic only operates on specific versioned subsystem directories (`v1/ir/`, `v1/logs/`, `v1/docs-preview/`). Root-level files like `token` and `id` are **never** touched. This is documented in the `Cache` class JSDoc.

### Directory structure
```
~/.fern/
├── token                    # Auth token (NOT managed by cache)
├── id                       # Telemetry distinct ID (NOT managed by cache)
├── v1/                      # Cache schema version
│   ├── ir/
│   ├── logs/
│   ├── migrations/
│   └── docs-preview/
│       ├── app-preview/     # Next.js docs bundle
│       └── preview/         # Legacy docs bundle
└── tmp/                     # Atomic write staging
```

## Testing
- [x] Compilation passes for `@fern-api/docs-preview` and `@fern-api/cli-v2`
- [x] Biome lint/format checks pass
- [ ] CI checks

Link to Devin session: https://app.devin.ai/sessions/d489776a5ae24482b99bb9be0d1f4d39

### DIFF
--- a/packages/cli/cli-v2/src/cache/Cache.ts
+++ b/packages/cli/cli-v2/src/cache/Cache.ts
@@ -2,6 +2,7 @@ import { AbsoluteFilePath, join, RelativeFilePath } from "@fern-api/fs-utils";
 import type { Logger } from "@fern-api/logger";
 import os from "os";
 import { FernRcSchemaLoader } from "../config/fern-rc/FernRcSchemaLoader.js";
+import { DocsPreviewCache } from "./docs-preview/index.js";
 import { IrCache } from "./ir/index.js";
 import { LogsCache } from "./logs/index.js";
 
@@ -22,7 +23,11 @@ const CACHE_VERSION = "v1";
  * │   │   │       └── 0a/
  * │   │   │           └── 0a3f9c2e4a7d1b...json
  * │   │   └── v62/
- * │   └── logs/
+ * │   ├── logs/
+ * │   ├── migrations/
+ * │   └── docs-preview/
+ * │       ├── app-preview/     # Next.js docs bundle
+ * │       └── preview/         # Legacy docs bundle
  * └── tmp/                    # Atomic write staging
  * ```
  */
@@ -35,6 +40,8 @@ export declare namespace Cache {
         ir: IrCache.Stats;
         /** Logs statistics */
         logs: LogsCache.Stats;
+        /** Docs preview cache statistics */
+        docsPreview: DocsPreviewCache.Stats;
     }
 
     /** Options for clearing cache entries */
@@ -43,6 +50,8 @@ export declare namespace Cache {
         ir?: boolean;
         /** Clear log files */
         logs?: boolean;
+        /** Clear docs preview bundles */
+        docsPreview?: boolean;
         /** Preview what would be cleared without actually deleting */
         dryRun?: boolean;
     }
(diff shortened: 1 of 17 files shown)


amckinney:
@iamnamananand996 does this change the cache path for the original CLI (including outside of `cli-v2` contexts)? If we merge this as-is, will the old cached artifacts not be respected? That's OK as long as the behavior is _valid_ and the user won't experience a regression.

iamnamananand996:
Also, no migration from `~/.cache/fern/`: Cached bundles at the old XDG location are orphaned. Users will re-download on next `fern docs dev`. this is intentional because cli-v2 is not used by any user right now.

amckinney:
> Also, no migration from ~/.cache/fern/: Cached bundles at the old XDG location are orphaned. Users will re-download on next fern docs dev. this is intentional because cli-v2 is not used by any user right now.

Got it. I imagine we'll need some sort of way to clean up the old cache paths once we officially flip the switch on CLI v2. Otherwise we'll just leave those orphaned artifacts indefinitely.

I wonder if we should drop the `v1` element from the cache path so that the CLI v2 paths are truly the same as the original? That way `fern cache clear` will actually work for both the old and the new artifacts. Thoughts?

iamnamananand996:
> > Also, no migration from ~/.cache/fern/: Cached bundles at the old XDG location are orphaned. Users will re-download on next fern docs dev. this is intentional because cli-v2 is not used by any user right now.
> 
> Got it. I imagine we'll need some sort of way to clean up the old cache paths once we officially flip the switch on CLI v2. Otherwise we'll just leave those orphaned artifacts indefinitely.
> 
> I wonder if we should drop the `v1` element from the cache path so that the CLI v2 paths are truly the same as the original? That way `fern cache clear` will actually work for both the old and the new artifacts. Thoughts?

The subdirectory structure is still different. CLI v1 puts bundles at `~/.fern/app-preview/` directly, while CLI v2 nests them under `~/.fern/docs-preview/app-preview/`. They'd still be separate paths even without the version prefix.

To truly share paths you'd have to make CLI v2 use the exact same layout as CLI v1. That would work for clearing, but we will lose the cache schema versioning — which is the whole point of the v1 segment. One way to avoid collisions with stale data from the old structure is to keep v1 and handle the flip-the-switch moment explicitly. Two reasonable approaches:

- **One-time migration on first CLI v2 run** — detect old `~/.fern/app-preview/` etc. and move/delete them. Runs once, documented in the release notes.

- **Add cleanup to fern cache clear** — when CLI v2's fern cache clear runs, also sweep the known CLI v1 legacy paths. Opt-in via a --legacy flag or just always included.

Option 2 is lower risk since it doesn't run automatically and gives users control. Either way, this needs to be done time, when we officially deprecate `CLI v1`.

amckinney (approved):

PR #15890, opened 2026-05-13
fix(cli): resolve FERN_TOKEN on `fern generate --local` path

jsklan:
## Description
Linear ticket: Closes FER-10551

When `useLocalDocker` is true (i.e. `fern generate --local`), `generateAPIWorkspaces` was skipping the entire auth block, leaving `token` as `undefined`. The Venus client then sent unauthenticated requests on the local path.

The remote path resolves a token via `askToLogin()`, which (1) optionally prompts to log in interactively (TTY only), and (2) calls `getToken()` from `@fern-api/auth` to pick up `FERN_TOKEN` or a saved login token.

This PR mirrors step 2 on the local path — non-interactively — so an existing `FERN_TOKEN` or saved login token is used for Venus requests. If `getToken()` returns `undefined`, generation proceeds without auth (local generation does not strictly require it, so we do not hard-fail).

## Changes Made
- `packages/cli/cli/src/commands/generate/generateAPIWorkspaces.ts`: when `useLocalDocker` is true, set `token = (await getToken()) ?? undefined` instead of leaving it `undefined`. Imported `getToken` from `@fern-api/auth` (already in the same module's `createOrganizationIfDoesNotExist`/`FernToken` import).
- `packages/cli/cli/changes/unreleased/fix-local-docker-token-resolution.yml`: changelog entry (`type: fix`).

## Testing
- [x] Unit tests added/updated — relied on existing tests; no new behavior to mock beyond what is already covered by the `@fern-api/auth` `getToken()` path.
- [x] Manual testing completed — `pnpm turbo run compile --filter @fern-api/cli` and `pnpm turbo run test --filter @fern-api/cli` (565 tests, all passing). `pnpm run check` (biome) clean.

Link to Devin session: https://app.devin.ai/sessions/4ea640d585c2449e8797a44b0c44d9a8
Requested by: @jsklan

### DIFF
--- a/packages/cli/cli/changes/unreleased/fix-local-docker-token-resolution.yml
+++ b/packages/cli/cli/changes/unreleased/fix-local-docker-token-resolution.yml
@@ -1,0 +1,8 @@
+# yaml-language-server: $schema=../../../../../fern-changes-yml.schema.json
+
+- summary: |
+    Resolve the Fern token non-interactively (via `FERN_TOKEN` or a saved login token)
+    when running `fern generate --local`, so Venus requests are authenticated. Auth
+    remains optional for local generation, so the command does not hard-fail if no
+    token is available.
+  type: fix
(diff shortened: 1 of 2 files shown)


jsklan:
Closing in favor of #15891

PR #15951, opened 2026-05-16
fix(cli): CLI-3J handle cross-drive OpenAPI init paths

FedeZara:
## Description
Refs CLI-3J

Prevents `fern init --openapi` from surfacing a Windows cross-drive OpenAPI path as an internal `RelativeFilePath` invariant failure.

## Changes Made
- Copy the OpenAPI spec into the newly created workspace when Windows cannot express the source spec path relative to that workspace.
- Keep existing relative-path behavior when a relative path can be represented.
- Add a regression test for Windows paths on different drives.
- Mark CLI-3J as shipped in the sentry triage ledger.

## Testing

Tested with:
- `pnpm turbo run test --filter @fern-api/init`


packages/cli/init/src/createWorkspace.ts
```diff
@@ -57,16 +58,58 @@ export async function createOpenAPIWorkspace({
     if (!(await doesPathExist(directoryOfWorkspace))) {
         await mkdir(directoryOfWorkspace);
     }
+    const relativeOpenAPIFilePath = getRelativeOpenAPIFilePathForWorkspace({ directoryOfWorkspace, openAPIFilePath });
+    const openAPIFilePathForConfig =
+        relativeOpenAPIFilePath ??
+        (await copyOpenAPIFileIntoWorkspace({
```
Swimburger: do we prompt the user for this? I don't think we can do this automatically without their explicit consent.

FedeZara: Ok let me throw an error then

Swimburger:
I get we probably don't support this today, but I think it should be fine to use an absolute path, even on a different disk.

PR #15983, opened 2026-05-19
fix(generator-cli): restore web-flow signing on cloud-generated commits

tstanmay13:
## Summary

- `pushSignedCommit` was sending a hardcoded `committer` to `octokit.git.createCommit` on every API-created commit, which silently suppressed GitHub's web-flow GPG signature. Cloud-generated PRs have shipped without the "Verified" badge since 0.9.28 (2026-05-06).
- Fix: always send `author` (preserves the stable Fern-bot attribution that 0.9.28 was after) but only send `committer` when a caller passes an explicit `author`. The signer keys off `committer` — omitting it lets GitHub fill it in from the installation and stamp the signature.
- Bumps `@fern-api/generator-cli` to 0.9.32. Catalog pin and CLI changelog will follow in a separate PR after the publish, matching the recent #15971 / #15961 / #15927 pattern.

| Scenario | Author | Committer | Signed |
|---|---|---|---|
| Cloud / App auth, default | `fern-api` bot | omitted → GitHub fills in App identity | ✅ |
| GHE / PAT, default | `fern-api` bot | omitted → PAT owner | ❌ (PATs can't be signed) — attribution still stable |
| Explicit `author: { ... }` | caller-provided | caller-provided (=author) | ❌ (caller opts out for deterministic committer) |

## Why both halves matter

- 0.9.28 (#15893) added explicit `author` + `committer` to fix GHE/PAT attribution showing the PAT-owning service account as the committer. The fix worked for attribution but broke signing for every cloud user.
- A naive revert would re-break GHE attribution. This change keeps the attribution goal by always sending `author` and only suppresses signing when the caller explicitly needs a deterministic committer — defaulting both behaviours to "do the right thing for cloud."

## Test plan

- [x] `pnpm turbo run test --filter @fern-api/generator-cli` — 412/413 passing (1 pre-existing skip), including the two new opt-in tests
- [x] Existing default-path test updated to assert `committer` is *not* sent (with `.not.toHaveProperty` belt-and-suspenders against vitest's loose object equality)
- [x] Rebase-on-conflict retry test updated to assert the same invariant on the second attempt
- [x] New test covers the explicit-author opt-in path (single attempt + rebase retry)
- [ ] After merge: verify a fresh cloud-generated PR ships with the green "Verified" badge

## Reviewer follow-ups (deferred)

- `findExistingUpdatablePR.ts:88-101` uses commit author email as one half of an "is this a Fern-authored commit" check. Pre-existing fragility — surfaces here because the committer email is no longer pinned. Branch-name + commit-message prefix are the resilient legs of that check.
- `PostGenerationPipeline.ts:118-119` still sets `git config user.name/user.email` for the local working copy. Cosmetic since the API commit is what gets pushed; cleanup ticket.

### DIFF
--- a/packages/generator-cli/src/__test__/pushSignedCommit.test.ts
+++ b/packages/generator-cli/src/__test__/pushSignedCommit.test.ts
@@ -97,8 +97,7 @@ describe("pushSignedCommit", () => {
             "local-sha",
             expect.stringMatching(/^refs\/temp\/fern-/)
         );
-        // Default path: `author` is always sent (stable attribution), but `committer` is
-        // omitted so GitHub fills it in from the App installation and signs the commit.
+        // `author` always sent; `committer` omitted so GitHub fills it in and signs.
         expect(octokit.git.createCommit).toHaveBeenCalledWith({
             owner: "acme",
             repo: "acme-sdk",
--- a/packages/generator-cli/src/__test__/pushSignedCommit.test.ts
+++ b/packages/generator-cli/src/__test__/pushSignedCommit.test.ts
@@ -97,15 +97,18 @@ describe("pushSignedCommit", () => {
             "local-sha",
             expect.stringMatching(/^refs\/temp\/fern-/)
         );
+        // Default path: `author` is always sent (stable attribution), but `committer` is
+        // omitted so GitHub fills it in from the App installation and signs the commit.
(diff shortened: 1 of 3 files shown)


tstanmay13:
@claude review once

fern-support (approved):

PR #15990, opened 2026-05-19
feat(cli-v2): add X-Request-Id header to all CLI v2 HTTP requests

iamnamananand996:
## Description
Linear ticket: Closes [FER-8774](https://linear.app/buildwithfern/issue/FER-8774/cli-v2-add-a-x-request-id-header)

Adds `X-Request-Id` header (UUID v4) to all CLI v2 HTTP requests for request tracing. The header is generated once per CLI invocation and passed to every SDK client.

## Changes Made
- Added optional `headers` parameter to `createVenusService` and `createFiddleService` in `@fern-api/core` (matching existing `createFdrService` signature)
- CLI v2 `Context` generates a `requestId` (UUID v4) once per invocation and passes `{ "X-Request-Id": requestId }` to all service clients
- All CLI v2 commands that make HTTP calls pass headers from context
- CLI v1 callers are unaffected — they don't pass `headers`, so the parameter defaults to `undefined`
- `@fern-fern/generators-sdk` doesn't support custom headers, so registry lookups (`sdk list`, `SdkChecker`) aren't instrumented yet

## Testing
- [x] Manual testing completed (compile, lint, all 704 tests pass)

Link to Devin session: https://app.devin.ai/sessions/edba4c237d3e4728875798105d40be9f


fern-support (approved):

packages/cli/cli-v2/src/services/fdr.ts
```diff
@@ -0,0 +1,20 @@
+import { FdrClient } from "@fern-api/fdr-sdk";
```
iamnamananand996: why we haven't added it inside core packages so we can import it like this

`import { createVenusService } from "@fern-api/core";`

iamnamananand996: yes, but we can have it as new fucntion inside core

packages/cli/cli-v2/src/context/Context.ts
```diff
@@ -247,22 +250,34 @@ export class Context {
             return;
         }
 
-        const result = await checkOrganizationMembership({ organization, token });
+        const venus = createVenusService({ token: token.value, headers: this.headers });
```
iamnamananand996: why we didn't used `checkOrganizationMembership` ??

PR #16083, opened 2026-05-26
fix(cli): disable yarn/npm engine checks in generator containers

tstanmay13:
## Update (CLI workaround for in-container installs)

The generator-side `resolutions` injection below is the durable fix for downstream SDK consumers, but it can't help installs running inside the *already-published* TS SDK image — 3.71.3 hasn't shipped yet, so ETE on this PR (and every other open PR) was still failing on the same `mute-stream@4.0.0` engine check.

This commit adds a CLI-side workaround: pass `YARN_IGNORE_ENGINES=true` and `npm_config_engine_strict=false` as container env vars in `packages/cli/generation/local-generation/local-workspace-runner/src/ContainerExecutionEnvironment.ts`, so every container the CLI runs in `--local` mode ignores engine declarations at install time. yarn v1 honors `YARN_IGNORE_ENGINES`; pnpm already defaults to non-strict engines, so the npm-style env is a defensive reaffirmation. The check is a soft hint — generator images bundle their own Node baseline, and transitive deps regularly declare newer `engines.node` ranges that don't actually reflect a runtime requirement (as is the case with `mute-stream@4`). The workaround removes a recurring class of false-positive install breaks, not just today's incident, and the `resolutions` injection still protects end users running `yarn install` on their generated SDK outside the container.

CLI changelog entry added at `packages/cli/cli/changes/unreleased/disable-engine-checks-in-generator-containers.yml`.

---

## Description

`mute-stream@4.0.0` (published 2026-05-08) declared `engines.node: ^22.22.2 || ^24.15.0 || >=26.0.0`. On **2026-05-25 19:47 UTC** `@inquirer/core@11.2.0` bumped its `mute-stream` range from `^3.0.0` to `^4.0.0` — reachable transitively in the package.json that the TS SDK generator writes to a customer's output via `msw` → `@inquirer/confirm` → `@inquirer/core`.

The TS SDK generator runs `yarn install --ignore-scripts --prefer-offline` (or pnpm equivalent) inside its Docker container, which is on Node 20.x for the older published images that ETE fixtures still pin to (e.g. `fernapi/fern-typescript-sdk:3.63.3`). yarn's engine check then trips on `mute-stream@4.0.0` and the install aborts, breaking these 4 ETE tests on every open PR opened since 2026-05-25 ~20:00 UTC:

- `packages/cli/ete-tests/src/tests/generate/fernignore.test.ts` (2 cases)
- `packages/cli/ete-tests/src/tests/generate/generate.test.ts > default api (fern init)`
- `packages/cli/ete-tests/src/tests/generate/output-directory-prompts.test.ts`

## Changes Made

- `generators/typescript/utils/commons/src/typescript-project/PersistedTypescriptProject.ts` — new private helper `pinKnownProblematicTransitiveDeps` that reads the on-disk `package.json` immediately before `install` runs and injects yarn `resolutions` (or pnpm `pnpm.overrides`) for known-problematic transitive deps. Currently pins `mute-stream@^3.0.0`. Idempotent: user-supplied entries for the same key are preserved.
- Called from both `installDependencies` and `generateLockfile`.
- 4 new unit tests covering yarn/pnpm injection, idempotency, and the `runScripts: false` skip path.
- Bumps TS SDK to **3.71.3** with a `fix`-typed changelog entry under `generators/typescript/sdk/changes/3.71.3/`.

## Testing

- [x] Local reproduction: with the `seed/ts-sdk/license/package.json` template, adding `"resolutions": { "mute-stream": "^3.0.0" }` to the package.json before `yarn install --ignore-scripts --prefer-offline` forces yarn to resolve `mute-stream@3.0.0` instead of `^4`. Confirmed via grep of `yarn.lock`.
- [x] `pnpm test` on `@fern-typescript/commons` — 19/19 pass (4 new + 15 existing).
- [x] Biome clean on both touched files.
- [x] TypeScript compile clean on the commons package.

After this lands and 3.71.3 publishes via autorelease, downstream ETE fixtures pinning `fernapi/fern-typescript-sdk` to a pre-3.71.3 version will still fail until they're bumped — a separate follow-up.

### DIFF
--- a/generators/typescript/sdk/changes/3.71.3/pin-mute-stream-transitive.yml
+++ b/generators/typescript/sdk/changes/3.71.3/pin-mute-stream-transitive.yml
@@ -1,0 +1,14 @@
+# yaml-language-server: $schema=../../../../../fern-changes-yml.schema.json
+
+- summary: |
+    Pin transitive `mute-stream` to `^3.0.0` via yarn `resolutions` /
+    pnpm `overrides` before `install` runs. `mute-stream@4.0.0` declares
+    `engines.node: ^22.22.2 || ^24.15.0 || >=26.0.0`, and
+    `@inquirer/core@11.2.0` (2026-05-25) bumped its `mute-stream` range
+    from `^3.0.0` to `^4.0.0` — reachable transitively via
+    `msw` → `@inquirer/confirm` → `@inquirer/core` in the generated
+    SDK's `package.json`. On generator images running Node 20.x the
+    engine check tripped, aborting `yarn install` / `pnpm install` and
+    breaking generation. The pin is idempotent (user-supplied
+    resolutions/overrides for the same key are preserved).
+  type: fix
--- a/generators/typescript/sdk/versions.yml
+++ b/generators/typescript/sdk/versions.yml
@@ -1,4 +1,20 @@
 # yaml-language-server: $schema=../../../fern-versions-yml.schema.json
+- version: 3.71.3
+  changelogEntry:
+    - summary: |
+        Pin transitive `mute-stream` to `^3.0.0` via yarn `resolutions` /
+        pnpm `overrides` before `install` runs. `mute-stream@4.0.0` declares
+        `engines.node: ^22.22.2 || ^24.15.0 || >=26.0.0`, and
+        `@inquirer/core@11.2.0` (2026-05-25) bumped its `mute-stream` range
+        from `^3.0.0` to `^4.0.0` — reachable transitively via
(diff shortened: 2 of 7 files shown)


patrickthornton:
Closing in favor of #16095, which addresses the root cause (Node 20 in the published 2.3.2 fallback image) by bumping the `fern init` preset to `fern-typescript-sdk@3.71.2` (which runs on the Node 24 image already used by all current generator containers). Thanks for taking a swing at this!

PR #16097, opened 2026-05-26
feat(docs): content-addressed doc publishing

broady:
## Description

**Part 1 of 2** — Docs deployment ledger with dual-write support: content-addressed blob storage, multi-locale translation publishing, stable file paths for deployment-level dedup, and MIME type detection.

Part 2 ([#16460](https://github.com/fern-api/fern/pull/16460), draft) adds `integrations` + `favicon` mapping once `@fern-api/fdr-sdk` publishes `LedgerConfig` with those fields.

### Core ledger architecture
- `mapDocsConfigToLedgerConfig`: Maps classic `DocsConfig` (FileId-based) → ledger-native `LedgerConfig` (path-based) with `ImageRef`, `PathOrUrl`, typography/JS file resolution
- `publishDocsLedger` / `publishDocsLedgerPreview`: Ledger register → upload → finish flow with structured git provenance
- `publishDocs.ts`: Dual-write integration — `deployMode` env var controls `"legacy"` / `"dual"` / `"ledger"` paths; `readAndHashFile` helper; `canonicalizeUploadedFilesForResolver` for FileId ↔ sanitized path mapping
- `docsDeployMode.ts`: Parse + validate `FERN_DOCS_DEPLOY_MODE` with `console.warn` on unrecognized values (S-1)
- `buildLedgerInput`: Assemble locale entries from `DocsDefinition` + API definitions

### Dual-write semantics
In `dual` mode, legacy and ledger run **in parallel** via `Promise.allSettled`:
```
Promise.allSettled([runLegacyPublish(), runLedgerPublish()])
```

**Failure semantics — any failure is fatal (non-zero exit):**
The two writes are not atomic and which store serves the live site depends on rollout state (edge config / docs-bundle), so a partial success cannot be reported as a successful publish.
- **Both fail** → `failAndThrow`: "both publish paths failed. Your docs were not updated."
- **Legacy fails, ledger succeeds** → `failAndThrow`: "one of two publish paths failed. Your docs may or may not have been updated."
- **Ledger fails, legacy succeeds** → failure is **deferred**: the legacy publish still completes fully (translations, AI chat), then the command exits non-zero with the same "may or may not have been updated" message.
- **Ledger fails in `ledger` mode** → `failAndThrow` — no legacy fallback.

### Audit fixes (MF-1, MF-2, S-1)
- **MF-1**: `translatedDefinition` remapped via `mapDocsDefinitionToLegacyFileIds` before V2 translations POST
- **MF-2**: `canonicalizeUploadedFilesForResolver` gated behind `deployMode !== "legacy"` — pure legacy is byte-identical to pre-PR behavior
- **S-1**: Warn on unrecognized deploy mode values

### Code quality
- Consolidated changelog entries → single `docs-ledger.yml`
- Named types: `LedgerMetadata`, `LedgerJsFile`, `PreviewRegisterInput`
- Exhaustive `mapColorsV3` switch with `assertNever`
- Exhaustive `normalizeRepoUrlToHttps` switch with `assertNever`
- `asyncPool(4)` replacing unbounded `Promise.all` in translation building
- Removed dead `buildSnippetsConfigForSdk` (reverted wiring in 89cba44)
- Removed spurious `fix-changelog-overview-detection.yml` (change was reverted)
- Migrated `visitDiscriminatedUnion` import from `@fern-api/ui-core-utils` → `@fern-api/core-utils`

### Deferred to [#16460](https://github.com/fern-api/fern/pull/16460)
- `integrations` mapping (intercom) — `LedgerConfig` in current SDK lacks field
- `favicon` path resolution — same SDK blocker

## Testing
- [x] Unit tests added/updated (`buildLedgerInput.test.ts`, `docsDeployMode.test.ts`, `mapDocsDefinitionToLegacyFileIds.test.ts`, `normalizeRepoUrl.test.ts`)
- [x] `pnpm check` (biome lint) passes locally
- [x] CI passing

Link to Devin session: https://app.devin.ai/sessions/19242359a2c54f51877acef2bfb1863f
Requested by: @broady

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/16097" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


broady:
blocked on #16099 and #16098

ctondreau:
## 🔍 Multi-Agent Audit — Docs Deployment Ledger

A 6-role review team (architect, senior eng, QA, security, release, product owner) audited this PR, with every critical/major finding independently verified to refute false positives (**58 confirmed, 1 refuted**).

**Verdict: Request changes.** The ledger architecture itself is sound — content-addressed blobs (full SHA-256), stable sanitized-path file tokens, register→upload→finish with legacy-authoritative dual-write ordering, and checksum-bound presigned PUTs all verified clean. The blockers are a default-path regression, a mechanically unmergeable state, and untested orchestration.

---

### 🔴 Must Fix Before Merge

#### MF-1 (CRITICAL) — V2 translations register sends path tokens the legacy store cannot resolve (default-mode regression)
**File:** `publishDocs.ts` (translation block)
`canonicalizeUploadedFilesForResolver` runs in **all** modes, so the resolver substitutes `file:<sanitizedPath>` tokens everywhere. The base publish compensates via `mapDocsDefinitionToLegacyFileIds` before `finishDocsRegister` — but `translatedDefinition` is built from the **un-remapped** `docsDefinition` and POSTed raw to `/v2/registry/docs/translations/register`. FDR's legacy store keys files by UUID (`APIV1Write.FileId(uuidv4())`), so every in-page image/file ref on translated pages breaks — silently (the per-locale catch only warns). Affects default `legacy` mode, **no flag required**.
**Fix:** Apply `mapDocsDefinitionToLegacyFileIds({ docsDefinition: translatedDefinition, pathToFileId: legacyFilePathToId })` before the translations POST. Also audit `updateAiChatFromDocsDefinition`, which receives the un-remapped definition.

#### MF-2 (MAJOR) — Default legacy mode is not behavior-preserving
**Files:** `publishDocs.ts`, `mapDocsDefinitionToLegacyFileIds.ts`
Even with the env var unset, every publish now flows fileId→sanitizedPath→UUID. Verified hazards: (a) `FILE_REF_PATTERN /file:([^\s"'<>)}\]]+)/g` cannot round-trip filenames containing spaces/parens/quotes (the sanitizer only rewrites `../`), so common filenames like `image (1).png` that worked with UUIDs now ship broken `file:` tokens; (b) the exact-match branch (`pathToFileId.get(value)`) rewrites **any** string field coincidentally equal to a sanitized path, including inside code fences; (c) `remapValue` deep-clones the entire multi-MB DocsDefinition every publish — the benchmark bot measured **+10.1s (+4.7%)** on the default path. There is no opt-out.
**Fix:** Gate `canonicalizeUploadedFilesForResolver` + `mapDocsDefinitionToLegacyFileIds` behind `deployMode !== "legacy"` so the default path is byte-identical to pre-PR behavior. This also eliminates MF-1's blast radius in pure legacy mode.

#### MF-3 (MAJOR) — In-memory blobs are PUT with `application/octet-stream` instead of their declared Content-Type
**File:** `publishDocsLedger.ts` (`uploadMissingBlobs`)
`contentType` is only inferred for lazily-read file blobs; page blobs (declared `text/markdown`) and apiManifest/jsFiles blobs (`application/json`) are uploaded as octet-stream while FDR presigns with the declared type. Verifiers split on the consequence: one concluded Content-Type is a signed header → real S3 returns 403 `SignatureDoesNotMatch` and every ledger publish with a new page blob hard-fails; the other concluded `@aws-sdk/s3-request-presigner` marks content-type unsignable → uploads succeed but CAS objects carry wrong metadata (and the in-code comment "The presigned URL enforces this value in its signature" is false). **Either outcome is wrong, the local S3 emulator masks both.**
**Fix:** Thread a hash→contentType map (from BlobRefs/fileManifest) into `uploadMissingBlobs` and send the declared type on every PUT. Fix the comment. **Validate one real publish against actual AWS S3 before promoting any environment to dual/ledger.**

#### MF-4 (MAJOR) — Multi-locale ledger preview is guaranteed to fail at finish
**File:** `publishDocsLedgerPreview.ts`
`previewRegisterWithLocales` sends `locales[]`, but the pinned server contract (`LedgerPreviewRegisterInputSchema`) has no such field — zod strips it, the server presigns base-locale blobs only, then `finish({ ..., locales })` HeadObject-verifies every locale and throws `Missing uploaded blob(s) in CAS`. Any translated preview fails (fatal in ledger mode, silently warn-swallowed in dual). The preview/init `deploymentHash` also differs from the finish-side hash.
**Fix:** Either extend the server contract + handler to accept `locales[]` (sequence with the SDK publish in MF-7), or have preview finish send only the base locale until the server supports multi-locale preview.

#### MF-5 (MAJOR) — Translation parity is broken
**Files:** `buildTranslatedDocsDefinition.ts`, `publishDocs.ts`, `publishDocsLedger.ts`, `publishDocsLedgerPreview.ts`
(a) `publishDocsViaLedger` runs **before** the translated-API registration block, so ledger apiManifests contain only base-language definitions; (b) `buildTranslatedDocsDefinition` omits `applyTranslatedApiTitlesToNavTree` + `updateApiDefinitionIdInTree` that the inline V2 block performs — the comment "both stores receive identical translated content" is false; (c) the helper is a ~150-line near-verbatim copy of the inline V2 pipeline (which was NOT refactored to call it), so dual mode builds every locale twice and the copies have already drifted. Also, the translated-API registration block isn't gated on `deployMode !== "ledger"`, so ledger-only mode performs legacy registry writes nothing references.
**Fix:** Move translated-API registration ahead of the ledger publish; pass `translatedApiIdsByLocale` into `buildTranslatedDocsDefinition`; make the V2 inline block call the shared helper; gate the translated-API registration on mode.

#### MF-6 (MAJOR) — Undisclosed customer-visible scope
**Files:** `runRemoteGenerationForGenerator.ts`, `packages/cli/docs-resolver/src/ChangelogNodeConverter.ts`
(a) `buildSnippetsConfigForSdk` replaces the all-undefined snippetsConfig for **every** `fern generate` run — a real FDR-registration behavior change, mentioned nowhere in the title/description, with no changelog entry. (b) The changelog fallback promotes the **first** non-date file in a changelog directory to overview page — changes navigation/slugs for existing sites with stray files, order-dependent, untested.
**Fix:** Split both into their own PRs (preferred — they currently make the ledger feature non-revertable in isolation), or document them, add the missing snippets changelog entry, and add deterministic selection + tests for the changelog fallback.

#### MF-7 (MAJOR) — PR is unmergeable and depends on an unpublished SDK
**Files:** `pnpm-workspace.yaml`, `pnpm-lock.yaml`, `mapDocsConfigToLedgerConfig.ts`
`mergeable: CONFLICTING`. The pinned `@fern-api/fdr-sdk@1.2.16-b5078a0a01` verifiably lacks `LedgerConfig.integrations` (zero occurrences in the published tarball's `contract.d.ts`), so compile is red by design.
**Fix:** Sequence — publish the fdr-sdk containing `integrations` (and the `locales[]` preview shape per MF-4, from #16098/#16099) → bump catalog/override/lockfile → rebase onto main → merge only on fully green CI. **Do not merge on the promise of a later SDK publish.**

---

### 🟡 Should Fix

| # | Item | File | Fix |
|---|------|------|-----|
| S-1 | Invalid `FERN_DOCS_DEPLOY_MODE` silently falls back to legacy; the comment promises a warning that doesn't exist (worst failure mode for a migration flag — typo → operator believes dual is running). Flagged by 3 roles. | `docsDeployMode.ts` | Return `{ mode, invalidRawValue }` and warn visibly, or hard-fail on unrecognized non-empty values. Pin in the existing test. |
| S-2 | `uploadMissingBlobs` warn-and-skips unsourceable blobs, then calls `finish()`. The server's HeadObject verification backstops this, but the client should fail fast with the missing hashes. | `publishDocsLedger.ts` | Throw on skip (non-fatal in dual, correctly fatal in ledger). |
| S-3 | Three mutually contradictory FileId-lifecycle comments (buildLedgerInput dedup caveat, mapDocsConfigToLedgerConfig "identity function" claim, nonexistent "finishTranslation endpoint" reference). | `publishDocsLedger.ts`, `mapDocsConfigToLedgerConfig.ts`, `publishDocs.ts` | One accurate description; document the FDR content-addressed-apiDefinitionId version floor. |
| S-4 | `resolveFileIdToPath` never returns undefined despite its documented contract — unmapped UUIDs leak into LedgerConfig as fake paths; the `mapJsFiles` null guard is dead code. | `mapDocsConfigToLedgerConfig.ts` | Return undefined on miss; keep identity passthrough only for ledger mode. |
| S-5 | SVG (unmeasured) logos silently dropped from ledger config — legacy site shows logo, ledger doesn't; defeats parity checks. | `mapDocsConfigToLedgerConfig.ts` | Warn when a configured logo/og-image is dropped; consider dimensionless ImageRefs. |
| S-6 | `normalizeRepoUrlToHttps` hardcodes github.com (wrong on GHE), doesn't strip userinfo from passthrough URLs, mangles ssh-form remotes. | `normalizeRepoUrl.ts` | Plumb `GITHUB_SERVER_URL`; `new URL()` + clear credentials in passthrough; reject `git@host:` forms. |
| S-7 | Wire `defaultLocale` is always `"en"` regardless of docs.yml `defaultLanguage`. | `publishDocsLedger.ts` | Seed base locale from `docsConfig.defaultLanguage ?? "en"`. |
| S-8 | Preview register bypasses the typed oRPC client (raw fetch, blind cast, no retry) — contract drift won't be caught at compile time. | `publishDocsLedgerPreview.ts` | Land `locales[]` previewRegister in the fdr-sdk contract and call through `createDocsLedgerClient`. |
| S-9 | Dual-write divergence is observable only as a CLI warn line — no telemetry, exit-code, or server marker. "Is the ledger ready for promotion?" is unanswerable. A ledger-only publish silently strands the legacy store readers still serve. | `publishDocs.ts` / FDR follow-up | Emit a structured event on dual-write failure; have FDR record/flag ledger-only publishes. Document rollback. |
| S-10 | Changelog hygiene: description claims 4→1 consolidation but 3 ymls ship; `ledger-favicon.yml` describes a sub-feature of the same feature; wire-format jargon in user-facing text. Plus the undescribed monorepo-wide ui-core-utils 0.x→1.x override bump. | `changes/unreleased/`, `pnpm-workspace.yaml` | Fold favicon into docs-ledger.yml; trim jargon; note the ui-core-utils bump. |
| S-11 | `previewId` sanitized on the preview path but passed raw to `publishDocsViaLedger` — can split one logical preview across two server identities. | `publishDocs.ts` | Sanitize at a single choke point. |
| S-12 | Git provenance (private repo slug/branch/SHA) now durably persisted with no opt-out. | `publishDocs.ts` | Document; consider honoring a telemetry opt-out; keep git fields org-gated server-side. |

---

### 🧪 Test Gaps

All five shipped test files cover only pure helpers at the happy path; everything risky lives in untested orchestration.

1. **Dual-write failure semantics:** nothing tests that ledger failure is non-fatal in dual mode and fatal in ledger mode, that ledger mode skips `startDocsRegister`, or that `canonicalizeUploadedFilesForResolver` populates the three id/path maps both publish paths depend on. A one-line regression in the catch block either breaks all publishes during rollout or silently drops the legacy write.
2. **Ledger orchestration:** no test that register and finish receive deep-equal inputs (the server recomputes the deployment hash on finish), that only `missingContent` hashes upload, that translation blobs merge into the pool, or that preview finish forwards `domain/basepath/previewId` verbatim.
3. **Path sanitization & special filenames:** zero coverage for spaces (demonstrably breaks `FILE_REF_PATTERN` round-trip on the DEFAULT path — see MF-2), unicode NFC/NFD, case divergence, or sanitization collisions (last-write-wins manifest clobber).
4. **Dead dedup test:** `Object.values(blobs)` on a Map is always `[]` — the "deduplicates pages with identical content" test asserts nothing. Replace with `expect([...blobs.values()].filter(b => b.toString() === markdown)).toHaveLength(1)`.
5. **`uploadBlobWithRetry` matrix** (429/5xx backoff, 412, 403/400 fail-fast, retry exhaustion) and the lazy-read MIME branch (`.PNG`, extensionless, `.svg`) — never executes in any test.
6. **Translation pipeline:** `buildTranslatedDocsDefinition` (frontmatter strip, locale fallback, editThisPageUrl rewrite, page-level vs locale-level failure semantics) — pin the asymmetry before a refactor inverts it.
7. **Round-trip property test:** canonicalize → `mapDocsDefinitionToLegacyFileIds` must be byte-identical to the pre-canonicalization legacy definition.

---

> **Note on the one refuted finding:** "finish() after skipped blobs persists a corrupt deployment" is **false** — the FDR finish handler HeadObject-verifies every referenced blob before persisting anything and throws `PRECONDITION_FAILED`. The server backstop is solid; the client-side fail-fast (S-2) is purely an ergonomics improvement.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — multi-agent SCRUM review (29 agents, findings adversarially verified)</sub>

ctondreau:
## 🧪 Proposed E2E / Integration Tests

Follow-up to the audit above. We researched both test harnesses — the CLI's Vitest suites in this repo and FDR's `docsLedger` integration suite in `fern-platform` — so these proposals fit existing patterns rather than inventing new infrastructure. Each maps to a finding (MF-_n_ / S-_n_) and names the exact harness, seam, and a sketch.

There are **three layers** worth adding to, in priority order:

1. **CLI hermetic integration tests** (this repo, `remote-workspace-runner`) — fast, no network, catch the orchestration + mode-branching bugs that today have zero coverage.
2. **CLI E2E ledger publish** (`packages/cli/ete-tests`) — extends the existing real-CLI test; needs a reachable FDR.
3. **FDR server integration tests** (`fern-platform`, `servers/fdr/src/__test__/local/docsLedger.test.ts`) — exercises register→upload→finish against docker-compose + s3mock.

---

### Layer 1 — CLI hermetic integration tests (highest value, lowest cost)

These live alongside the PR's existing unit tests in `packages/cli/generation/remote-generation/remote-workspace-runner/src/__test__/`. The seams already exist: `publishDocsViaLedger` does `await import("@fern-api/fdr-sdk/orpc-client")` for `createDocsLedgerClient` (mock with `vi.mock`), and `uploadMissingBlobs` uses global `fetch` (mock with `vi.spyOn(globalThis, "fetch")`). No msw/nock needed — matches the repo's plain-Vitest + DI house style.

| Test file | Targets | What it asserts |
|-----------|---------|-----------------|
| `publishDocsViaLedger.test.ts` (new) | MF-5, gaps #1/#2 | register→upload→finish ordering; `register` and `finish` receive **deep-equal** inputs (server recomputes the deployment hash on finish — drift = silent corruption); `uploadMissingBlobs` uploads **only** the hashes in `missingContent`, not the whole blob set; translation blobs merge into the pool. |
| `uploadMissingBlobs.test.ts` (new) | MF-3, S-2, gap #5 | **Every** PUT carries the blob's declared Content-Type (`text/markdown` for pages, `application/json` for manifests) — this is the unit-level guard for MF-3 since s3mock can't catch it (see Layer 3); retry matrix (429/5xx backoff, 412, 403/400 fail-fast, exhaustion) via stubbed `fetch` + fake timers; **fail-fast on skipped blob** (S-2) instead of proceeding to `finish()`. |
| `mapDocsConfigToLedgerConfig.test.ts` (new) | S-4, S-5, S-7 | pure, mirrors the style of the existing `mapDocsDefinitionToLegacyFileIds.test.ts`: `resolveFileIdToPath` returns `undefined` on a lookup miss (not a fake path — S-4); a configured SVG/dimensionless logo is **not silently dropped** (S-5); `defaultLocale` seeds from `docsConfig.defaultLanguage`, not hardcoded `"en"` (S-7); exhaustive `mapColorsV3` + `assertNever`. |
| `buildTranslatedDocsDefinition.test.ts` (new) | MF-5, gap #6 | use `createTempFixture` (`packages/cli/ete-tests/src/utils/createTempFixture.ts`) for the real `readFile` calls; assert the output applies `applyTranslatedApiTitlesToNavTree` + `updateApiDefinitionIdInTree` so it is **byte-identical** to what the inline V2 block produces (the "both stores receive identical content" claim); page-level failure is lenient, locale-level is fatal — pin the asymmetry before a refactor inverts it. |

**Two property/regression tests that pin the confirmed bugs directly:**

```ts
// canonicalize-roundtrip.test.ts  → MF-1 / MF-2 / gap #7
// In legacy mode the canonicalize → mapDocsDefinitionToLegacyFileIds round-trip
// MUST be byte-identical to the pre-canonicalization legacy definition.
it("round-trips filenames with spaces/parens without corrupting file refs", () => {
  const def = makeDocsDefinition({ files: { "image (1).png": fileId } });
  const canon = canonicalizeUploadedFilesForResolver(def, /* maps */);
  const back  = mapDocsDefinitionToLegacyFileIds({ docsDefinition: canon, pathToFileId });
  expect(back).toEqual(def);            // currently FAILS — FILE_REF_PATTERN drops at "image ("
});

// translations-remap.test.ts  → MF-1 (the critical default-mode regression)
it("translated definition is remapped to legacy FileIds before the V2 translations POST", () => {
  // Build a translatedDefinition with a file: ref, run the publish path with a spy on the
  // translations register call, assert the posted body contains UUID FileIds, not file:<path> tokens.
});
```

The dead test from the audit (gap #4) should be fixed in the same pass — `Object.values(blobs)` on a `Map` is always `[]`, so the existing "deduplicates pages with identical content" assertion is vacuous:

```ts
expect([...blobs.values()].filter((b) => b.toString() === markdown)).toHaveLength(1);
```

---

### Layer 2 — CLI E2E ledger publish (`packages/cli/ete-tests`)

The existing `src/tests/generate/generate.test.ts` already drives the **real compiled CLI** (`runFernCli(["generate", "--docs", ...])`) against the `fixtures/docs/` site and asserts on the published-URL stdout. Extend it:

```ts
it("generate docs in ledger mode publishes via the ledger flow", async ({ signal }) => {
  const { stdout } = await runFernCli(["generate", "--docs", "--no-prompt"], {
    cwd: join(fixturesDir, RelativeFilePath.of("docs")),
    env: { FERN_DOCS_DEPLOY_MODE: "ledger", FERN_FDR_ORIGIN: "http://localhost:8080", FERN_TOKEN: "dummy" },
    includeAuthToken: false, reject: false, signal,
  });
  expect(stdout).toContain("Started.");           // and assert on the [ledger] log lines
}, 300_000);
```

Add a sibling case with a **translated** fixture (`locales` in `docs.yml`) to exercise the multi-locale path end-to-end, and a `FERN_DOCS_DEPLOY_MODE: "dual"` case asserting a ledger failure does **not** fail the legacy publish (the dual-write non-fatal contract).

> ⚠️ **Caveat:** the ete suite runs against the **real dev FDR** in CI (`FERN_TOKEN=secrets.FERN_ORG_TOKEN_DEV`); there is no hermetic mock-FDR container for the CLI today. So a CI ledger E2E either depends on dev FDR supporting the new contract (gated behind the SDK publish in MF-7) or needs a small mock-FDR harness added to `ete-tests/docker-compose.yml`. The Layer-1 tests are the reliable, hermetic coverage; treat Layer 2 as the smoke test once the SDK lands.

---

### Layer 3 — FDR server integration tests (`fern-platform`)

`servers/fdr/src/__test__/local/docsLedger.test.ts` already has the full `register → uploadBlob → finish → read → archive` cycle running against docker-compose (Postgres + adobe/s3mock), runnable with `pnpm --filter=@fern-platform/fdr test:local`. Add server-side guards for the findings:

- **MF-4 (preview rejects/handles `locales[]`):** ✅ fully testable here. Add to the preview suite — assert `previewRegister` with a `locales[]` field either presigns blobs for **all** locales or the contract rejects it, so `finish` doesn't later throw `PRECONDITION_FAILED: Missing uploaded blob(s) in CAS`. This is the server half of the MF-4 fix and pins the contract decision.
- **MF-1 (translated files keyed by path, retrievable per-locale):** ✅ testable. Multi-locale `register`/`finish`, then read `fileMetadata({ domain, basepath, locale, filePath })` for a **non-default** locale and assert the in-page file ref resolves. Guards the server contract the CLI fix in MF-1 depends on.
- **MF-3 (Content-Type):** ⚠️ **NOT catchable here.** Confirmed: adobe/s3mock does not validate the signed `Content-Type`, so a mismatched PUT returns 200 against the mock even though real S3 returns 403 `SignatureDoesNotMatch`. A green test would give false confidence. Options: (a) keep MF-3 coverage at the CLI unit level (Layer 1 `uploadMissingBlobs.test.ts`) where we assert the declared type is sent, **plus** (b) one manual/staging publish against real AWS S3 before promoting any env to dual/ledger. Don't rely on the local suite for this one.

---

### Suggested sequencing

1. **Now (this PR):** Layer 1 — all four new test files + the two property tests + the dead-test fix. Hermetic, no infra, and they fail-red on MF-1/MF-2/MF-3/MF-5 today, which is exactly what you want gating the fixes.
2. **With the SDK publish (MF-7):** Layer 3 MF-1 and MF-4 server tests in `fern-platform`, since they exercise the new contract.
3. **Smoke, post-SDK:** Layer 2 ete cases once dev FDR speaks the ledger contract.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — test proposals researched against both repos' actual Vitest harnesses.</sub>

ctondreau:
## 🔧 Requested fixes — two behavior changes that affect customers who do *not* opt into dual/ledger

Both of these run regardless of `FERN_DOCS_DEPLOY_MODE` (neither checks `deployMode`), so they ship to every customer on their next publish/generate. Neither is mentioned in the PR title/description. They should be made intentional + safe before merge.

---

### (a) SDK install snippets are now registered for every `fern generate`
**File:** `packages/cli/generation/remote-generation/remote-workspace-runner/src/runRemoteGenerationForGenerator.ts:216`

The diff changes the API-registration payload from an all-`undefined` snippets config to a populated one:

```diff
-        snippetsConfig: {
-            typescriptSdk: undefined,
-            pythonSdk: undefined,
-            javaSdk: undefined,
-            ...
-        },
+        snippetsConfig: buildSnippetsConfigForSdk({
+            language: generatorInvocation.language,
+            packageName,
+            version: resolvedVersion
+        }),
```

This is in the IR→FDR path (`convertIrToFdrApi`), not docs-specific and not mode-gated, so **every SDK generate now registers package + version snippet metadata** where it previously registered nothing. The result is customer-visible: API-reference pages will start rendering install snippets (`npm i <pkg>@<version>`, etc.) that weren't there before.

This is likely a *desirable* change — the issue is that it's silent, bundled into a ledger PR, and would be reverted along with the ledger if the ledger is rolled back. **Gating it behind docs deploy mode would be wrong** (it's correct behavior; don't tie it to the flag). Instead:

**Requested fix:**
1. **Split this into its own PR** (preferred) so it's independently revertable and reviewable, **or** keep it here with an explicit changelog entry under `packages/cli/cli/changes/unreleased/` and a line in the PR description.
2. **Verify the version fallback can't emit a wrong install command.** `buildSnippetsConfigForSdk` calls `resolveVersionFallback(version)`. Confirm that when `version` is `undefined`/unpublished it does **not** resolve to a misleading string (e.g. a literal `"latest"` or `"0.0.0"`) that would render an install snippet pointing at a version the customer never published. Add a test case for the `version == null` path asserting the snippet is omitted rather than emitted with a bogus version.

---

### (b) Changelog overview page is now auto-selected from the first non-reserved file
**File:** `packages/cli/docs-resolver/src/ChangelogNodeConverter.ts:74–78`

Previously only files named `summary` / `index` / `overview` became a changelog overview page. The new fallback promotes the **first non-date file** in iteration order when no reserved-name file exists:

```ts
// Fall back to any non-date file as the overview when no reserved-name file is found.
if (overviewPagePath == null && fallbackOverviewPath != null) {
    overviewPagePath = fallbackOverviewPath;   // fallbackOverviewPath = first non-date, non-reserved file
}
```

Two problems for existing sites (all modes):
- **Backward-incompatible & silent:** a site with a stray non-date file in its changelog dir (e.g. `release-notes.mdx`, a `_template.mdx`, a `README.mdx`) that previously had *no* overview page will suddenly get one — changing navigation and slugs with no opt-in and no warning.
- **Non-deterministic:** selection depends on `this.changelogFiles` iteration order (`fallbackOverviewPath == null` → first wins). If that array isn't stably ordered, which file becomes the overview can vary between runs/machines.

**Requested fix (pick one of these, in order of safety):**

1. **Require an explicit signal instead of guessing** (safest — no surprise for existing sites). Only promote a non-reserved file when it opts in via frontmatter, e.g. `overview: true` or an explicit `slug` override:
   ```ts
   } else if (fallbackOverviewPath == null && this.markdownToFrontmatter.get(absoluteFilepath)?.overview === true) {
       fallbackOverviewPath = absoluteFilepath;
   }
   ```
2. **If you keep the implicit fallback, make it deterministic and loud:**
   - Sort candidates before picking (e.g. lexicographic on relative path) so selection is stable regardless of `changelogFiles` order — don't rely on first-seen.
   - `context.logger.warn(...)` when the fallback fires, naming the file it promoted, so customers notice the nav change.
3. **Either way, add tests** in the docs-resolver suite covering:
   - reserved-name file present → it wins over any non-reserved file (precedence);
   - no reserved name, multiple non-date files → deterministic, documented choice;
   - all files are dated → no overview page (today's behavior preserved);
   - the chosen file's frontmatter `slug` override is honored.

This logic lives in `packages/cli/docs-resolver`, which runs for **every** docs build (preview + prod), so a regression here hits all customers — it warrants explicit tests and, ideally, the opt-in approach.

---

**Summary:** (a) split out or document + guard the version fallback; (b) prefer an explicit-opt-in overview marker, otherwise make the fallback deterministic + warned, and add tests either way. Both should stop being implicit side effects of a ledger PR.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

ctondreau:
## 🐞 Reproduced: MF-2 special-filename breakage still occurs in **dual** mode

Followed up on MF-2 with a real end-to-end publish (built CLI from this branch → local FDR + s3mock). Good news: the `deployMode !== "legacy"` gating means **default/legacy publishes are safe**. But the underlying `FILE_REF_PATTERN` defect was never fixed, so **dual mode still corrupts file references for filenames containing spaces/parens/quotes** — and dual mode keeps the legacy store as the served source until read-cutover, so this hits real rendered pages during migration.

### Reproduction
Fixture page referencing two images — a normal one and a spaced one:
```md
![fern](fern.png)
![spaced](<image (1).png>)
```

**Legacy mode** (`FERN_DOCS_DEPLOY_MODE` unset) — both resolve ✅
```
![fern](file:997b7044-…)              → UUID in files map ✓
![spaced](<file:a74ab9be-…>)          → UUID in files map ✓
```

**Dual mode** (`FERN_DOCS_DEPLOY_MODE=dual`) — the spaced one breaks ❌. From the **served** `v1/fdr.json`:
```
![fern](file:aeffa3e0-…)              → UUID, RESOLVES ✓
![spaced](<file:image (1).png>)       → NOT remapped, DANGLING ✗
```
The file *was* uploaded (files map has UUID `f19c3e2c` → path `image (1).png`), but the page markdown still says `file:image (1).png` instead of `file:f19c3e2c-…`, so the renderer can't resolve it. The normal filename immediately above remaps correctly.

### Root cause
`mapDocsDefinitionToLegacyFileIds.ts`:
```ts
const FILE_REF_PATTERN = /file:([^\s"'<>)}\]]+)/g;
```
The character class excludes whitespace, `"`, `'`, `<`, `>`, `)`, `}`, `]`. For `file:image (1).png` it captures only `file:image` (stops at the space); `image` isn't a key in `pathToFileId`, so the `?? path` fallback leaves the token unremapped. Any sanitized path containing one of those characters can never be matched, because the sanitizer only rewrites `../` (it preserves spaces/parens). `image (1).png` is a very common filename (Finder/Windows duplicate-copy convention).

### Scope
| Mode | Spaced/paren filenames | Why |
|---|---|---|
| legacy (default) | ✅ safe | canonicalize gated off; resolver assigns UUIDs directly |
| ledger-only | ✅ safe | path-native refs; manifest keyed by sanitized path |
| **dual** | ❌ **broken in the legacy store** | legacy remap runs but the regex can't match the path; ledger half is fine, legacy half (still served pre-cutover) is dangling |

### Suggested fix
The remap can't be a regex over arbitrary path characters. Options:
1. **Match the full bracketed/parenthesized markdown image target**, not a greedy char-class — e.g. handle `](<...>)` and `](...)` forms and remap the captured path verbatim, allowing spaces/parens inside.
2. **Remap by exact path lookup against `pathToFileId` keys** rather than a character-class scan: for each known sanitized path, replace `file:<thatPath>` occurrences (longest-key-first) so paths with special chars are handled by data, not regex.
3. At minimum, widen the class to allow the characters that real sanitized paths contain, and add a round-trip test.

### Test to add
The `canonicalize-roundtrip` test from the [test-proposal comment](https://github.com/fern-api/fern/pull/16097#issuecomment-4674081039), with a dual-mode case asserting `image (1).png` round-trips to a resolvable UUID. It fails today.

<sub>Verified against a real publish: CLI built from `feat/docs_ledger` → local FDR (`fdr:mocked`) + adobe/s3mock, inspecting the served `v1/fdr.json`. 🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

ctondreau:
## 🧪 Dual-write: remaining test coverage needed before merge

I validated the dual-write happy path end-to-end against a local FDR (`fdr:mocked` + s3mock) with the CLI built from this branch — both stores written, correct ordering (legacy `finishDocsRegister` before ledger), base + **translated** file refs resolve to UUIDs (MF-1 confirmed fixed), and cross-mode CAS deployment dedup works. One defect found: MF-2 special filenames still break in the legacy half (separate comment).

These dual-mode behaviors are **not yet covered** and should be tested before merge. Listed highest-risk first.

### A. Dual-write resilience — ledger failure must be non-fatal (HIGHEST PRIORITY)
This is the safety property the entire migration strategy rests on: in `dual` mode, a ledger failure must **not** fail the publish, because legacy already succeeded.
- **How:** publish in `dual` mode with the ledger pointed at an unreachable/erroring endpoint (e.g. wrong CAS bucket, killed ledger handler, or inject a throw in `publishDocsViaLedger`).
- **Expect:** legacy publish completes, exit 0, and a `[ledger] Dual-write failed (non-fatal): …` warn. The published legacy site must be fully intact.
- **Also assert the inverse:** in `ledger`-only mode the same failure **is** fatal (non-zero exit).
- **Why it matters:** a one-line regression in the catch block either silently drops the legacy write or turns every rollout publish into a hard failure. Today this path has zero coverage.

### B. MF-3 — Content-Type on the ledger CAS PUTs (needs real S3)
Cannot be validated locally — adobe/s3mock does **not** verify the signed `Content-Type`, so wrong-MIME PUTs pass against the mock but real S3 may 403 (`SignatureDoesNotMatch`).
- **How:** one `dual` publish against a real bucket (`fdr-dev-content-addressable-storage`) with a new page blob (`text/markdown`) and a JSON manifest blob.
- **Expect:** no `403 SignatureDoesNotMatch`, and stored CAS objects carry the declared Content-Type (not `application/octet-stream`).
- **Gate:** do this before promoting any environment to dual/ledger.

### C. MF-5 — byte-parity between legacy and ledger translated content
Both stores receive the `es` locale, but their translated content hasn't been byte-compared.
- **How:** `dual` publish with a translated locale; fetch the legacy translation blob (`v1/translations/<locale>.json`) and the ledger's stored translated content; diff the page markdown, nav titles, and API-reference titles.
- **Expect:** identical (the PR comment claims "both stores receive identical translated content"). Specifically verify the ledger applies `applyTranslatedApiTitlesToNavTree` + `updateApiDefinitionIdInTree` (MF-5 flagged these as missing from `buildTranslatedDocsDefinition`).

### D. API reference + SDK snippets in dual mode
My fixture had the API-reference nav stripped, so the dual-write of **API definitions** and the always-on `buildSnippetsConfigForSdk` change (behavior-(a)) went untested.
- **How:** publish a docs site with a real API reference in `dual` mode.
- **Expect:** API definitions register to both stores; API-reference install snippets render with package+version; the ledger apiManifest contains the API. Confirm translated-API registration ordering relative to the ledger publish (MF-5: ledger currently publishes before translated-API registration).

### E. Custom-domain fan-out in dual mode
Single-domain only so far. The translation register fans out to `[fernUrl, ...customDomains]`.
- **How:** `dual` publish with one or more `customDomains` configured.
- **Expect:** base docs + each translation blob written under every domain's S3 prefix; ownership checks pass; no write to unowned hostnames.

---

**Suggested:** A and C are quick local tests (`fdr:mocked` + s3mock); D needs an API fixture; B needs dev S3 creds; E needs multi-domain setup. A is the blocker — it's the contract dual-write depends on and is trivially testable. Most of these map directly to the missing tests in the [test-proposal comment](https://github.com/fern-api/fern/pull/16097#issuecomment-4674081039) (Layer 1 dual-write failure semantics, Layer 3 MF-1/MF-3/MF-4).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

thesandlord (approved):

PR #16132, opened 2026-05-28
fix(python): propagate Params suffix to typeddict alias inner types

patrickthornton:
Closes FER-10868

## Summary

When `use_typeddict_requests: true` is set on the Python SDK generator, type aliases whose body resolves to a named type — directly (`Alias = T`) or wrapped in `List[T]` / `Dict[K, V]` / `Set[T]` — emit the inner element as the Pydantic model `T` rather than the TypedDict variant `TParams`. Request parameters typed as such aliases therefore reject dict-literal arguments at type-check time even though the runtime accepts them.

Surfaced by Auth0 (Slack thread `1776961586.870679`) across four aliases in their Management API generation.

## Example

Before — `src/auth0/management/requests/update_enabled_client_connections_request_content.py`:

```python
from ..types.update_enabled_client_connections_request_content_item import (
    UpdateEnabledClientConnectionsRequestContentItem,
)

UpdateEnabledClientConnectionsRequestContentParams = typing.List[UpdateEnabledClientConnectionsRequestContentItem]
```

And the generated docstring example fails pyright:

```python
client.connections.clients.update(
    id="id",
    request=[{"client_id": "client_id", "status": True}],
    # Argument of type "list[dict[str, str | bool]]" cannot be assigned to
    # parameter "request" of type "UpdateEnabledClientConnectionsRequestContentParams"
)
```

After:

```python
from .update_enabled_client_connections_request_content_item import (
    UpdateEnabledClientConnectionsRequestContentItemParams,
)

UpdateEnabledClientConnectionsRequestContentParams = typing.Sequence[
    UpdateEnabledClientConnectionsRequestContentItemParams
]
```

The same example now type-checks cleanly because the dict literal matches the TypedDict variant.

## Root cause

`AbstractAliasGenerator.__init__` resolved the alias body via `get_type_hint_for_type_reference(self._alias.alias_of)` with the default `in_endpoint=False`. That value propagated through `TypeReferenceToTypeHintConverter`'s `list_`/`map_`/`set_` arms into `_get_type_hint_for_named`, where `TypeDeclarationReferencer.get_class_name` short-circuits the `Params` suffix unless `as_request=True`.

The peer `TypeddictUndiscriminatedUnionGenerator.generate()` already passes `in_endpoint=True` when resolving union members — which is why undiscriminated-union aliases worked correctly under `use_typeddict_requests` and list/map/set aliases did not.

## Fix

Override `_type_hint` in `TypedDictAliasGenerator.__init__` to re-resolve the alias body with `in_endpoint=True`. Localized to the TypedDict variant; the Pydantic alias generator keeps its current behavior.

## Side effects

1. **`List` → `Sequence`.** Under `in_endpoint=True`, `typing.List[T]` swaps to `typing.Sequence[T]` — the same input-side swap already applied to endpoint parameters. Existing callers passing lists continue to work.
2. **Seed fixture drift.** `file-upload/use_typeddict_requests/src/seed/service/requests/{my_alias_object,my_collection_alias_object}.py` previously encoded the bug; updated to the corrected output.

## Test plan

- [x] Auth0 spec regenerated locally — all 4 affected aliases now reference `…Params` variants
- [x] Pyright sweep over 856 docstring examples: `reportArgumentType` errors dropped from 8 → 0; all 8 mapped 1:1 to the 4 aliases (sync + async pair each)
- [x] `pnpm seed test --generator python-sdk --fixture file-upload --skip-scripts` — 3/3 pass; expected fixture diff verified
- [x] `pnpm seed test --generator python-sdk --fixture literal --skip-scripts` — 2/2 pass; no drift in the other `use_typeddict_requests` fixture
- [ ] CI green on python-sdk seed suite


patrickthornton:
Superseded by #16133 — branch was renamed from `pt/...` to `patrick/python/...` for naming-convention consistency, which inadvertently closed this PR. Same commit, same diff.

PR #16140, opened 2026-05-31
fix(cli): emit custom scalars and all named types in GraphQL converter

lifanzou:
## Description
Linear ticket: Refs FER-10878

The GraphQL converter silently dropped custom scalars. In `collectTypeDefinitions`, the loop hit a `continue` for any non-built-in `GraphQLScalarType`, so custom scalars (e.g. `DateTime`, `URL`, `EmailAddress`) never made it into the `types` map. They only appeared inlined as `scalar` primitives at each reference site, which left them without a stable id the frontend can use as an href anchor.

This fix emits every custom scalar as a named **alias** type (mapping to the appropriate base primitive via the existing `convertScalarTypeDefinition` / `getBaseTypeForCustomScalar` logic) with a stable, namespace-aware id, and updates references to custom scalars to point at that id instead of inlining the scalar shape.

Note: this PR intentionally does **not** add navigation/sidebar nodes for types or touch field-level arguments — those are tracked separately.

## Changes Made
- **`GraphQLConverter.collectTypeDefinitions`**: replaced the `continue` for custom scalars with emission of a named alias `TypeDefinition` (stable id, description preserved). Built-in scalars are still skipped (they map to inline primitives).
- **`GraphQLConverter.convertScalarType`**: custom-scalar references now return `{ type: "id", value: <typeId> }` so they anchor to the emitted named type. Built-in scalars are unchanged (inline primitives).
- **Tests**: added assertion-based unit tests proving (1) every custom scalar in the `basic` fixture lands in `types` as an alias, (2) built-in scalars are not emitted, (3) references use the stable id, and (4) ids are namespaced. Updated the `basic` and `account-schema` snapshots accordingly.
- **Changelog**: added `fix-graphql-custom-scalars.yml` (type `fix`).

## Testing
- [x] Unit tests added/updated (`GraphQLConverter.test.ts`, 6 passing)
- [x] Snapshots updated (`basic.json`, `account-schema.json`)
- [x] Package compiles (`tsc`) and lint/format pass (`biome`)


fern-support:
@claude review

rishabh-fern (approved):
LGTM!

PR #16153, opened 2026-06-01
fix(typescript): fix anySignal() race condition between aborted check and addEventListener       

### LINKED ISSUE #16151: fix(typescript-sdk): anySignal() has a race condition between aborted-check and addEventListener
### Component

SDKs

### Priority

P1 - High (Strongly needed)

### SDK Language(s)

TypeScript

### Bug Description

The generated `anySignal()` helper in `src/core/fetcher/signals.ts` has a race condition that silently drops abort events, breaking timeouts and user-initiated cancellation.

### Steps to reproduce

1. Use the generated `anySignal()` from the TypeScript SDK.
2. Create two `AbortController` instances.
3. Proxy one signal so that the real `.abort()` fires inside `addEventListener` (race window between `signal.aborted` check and listener registration).

```ts
const ctrlA = new AbortController()
const ctrlB = new AbortController()

const originalAddEventListener = ctrlA.signal.addEventListener.bind(ctrlA.signal)
const originalRemoveEventListener = ctrlA.signal.removeEventListener.bind(ctrlA.signal)

let abortedAccessCount = 0
const proxy = new Proxy(ctrlA.signal, {
  get(target, prop, receiver) {
    if (prop === "aborted") {
      abortedAccessCount++
      if (abortedAccessCount === 1) return false // lie on first check
      return Reflect.get(target, prop, receiver)
    }
    if (prop === "addEventListener") {
      return (...args: Parameters<typeof originalAddEventListener>) => {
        if (abortedAccessCount >= 1 && args[0] === "abort") {
          ctrlA.abort("too-late") // abort fires in the race window
        }
        return originalAddEventListener(...args)
      }
    }
    if (prop === "removeEventListener") {
      return originalRemoveEventListener
    }
    return Reflect.get(target, prop, receiver)
  },
})

const combined = anySignal(proxy, ctrlB.signal)
```

### Expected behavior

`combined.aborted` should be `true` and `combined.reason` should be `"too-late"`. The abort event from the source signal should always propagate to the combined signal.

### Actual behavior

`combined.aborted` is `false`. The abort event was dispatched during the race window (between `signal.aborted` check and `addEventListener` call) and is silently lost. The combined signal stays permanently unaborted.

### Environment

- OS: macOS 15 (Darwin 25.5.0)
- Node.js: 22.x

### Versions

Fern CLI version: 0.83.0
SDK Generator: `fernapi/fern-typescript-node-sdk` v3.64.1

### Workaround

Consumer project can patch `signals.ts` locally, but the fix is overwritten on every SDK regeneration. The fix must live in the Fern generator so it persists across regenerations.

## Are you interested in contributing a fix?

Yes

## Context

Discovered in [`latitude-dev/latitude-llm`](https://github.com/latitude-dev/latitude-llm/pull/3336) where a local patch was applied. Since `signals.ts` is fully generated by Fern, the fix must live in the generator.

- Upstream PR with local fix and full test suite: https://github.com/latitude-dev/latitude-llm/pull/3336

### Are you interested in contributing a fix?

Yes

### Current `anySignal()` flow

```mermaid
flowchart TD
    A[Start: for each signal in signals] --> B{signal.aborted?}
    B -->|true| C[controller.abort]
    B -->|false| D["addEventListener('abort', handler)"]
    C --> Z[return controller.signal]
    D --> E

    subgraph RACE_WINDOW["RACE WINDOW"]
        direction LR
        D -.->|"signal aborts here"| W[/"abort event dispatched<br/>(no listeners yet!)"/]
        W -.->|"event LOST"| E
    end

    E[continue to next signal]
    E --> A
```

### Race condition sequence

```mermaid
sequenceDiagram
    participant anySignal
    participant SourceSignal

    anySignal->>SourceSignal: signal.aborted?
    SourceSignal-->>anySignal: false

    Note over SourceSignal: controller.abort("timeout") fires<br/>"abort" event dispatched<br/>BUT no listeners registered!

    anySignal->>SourceSignal: addEventListener("abort", handler)
    Note over anySignal,SourceSignal: ❌ Listener added TOO LATE<br/>Event already dispatched<br/>Handler never fires

    Note over anySignal: ❌ combined signal never aborts<br/>operation hangs indefinitely
```

**Why this happens:** Per the WHATWG spec, when `abort()` is called, the "abort" event is dispatched **once** to any registered listeners. If no listener is registered at dispatch time, the event is dropped. Adding a listener **later** on an already-aborted signal does NOT retroactively invoke the callback in Node.js (confirmed with `node -e` on v22.16.0).

## Steps to Reproduce

```typescript
// Triggering the race is timing-dependent in production, but we can
// prove the gap exists with a controlled proxy:

const ctrl = new AbortController()
let accessCount = 0

const proxy = new Proxy(ctrl.signal, {
  get(target, prop) {
    if (prop === "aborted") {
      accessCount++
      // First access: lie, say not aborted (simulate pre-check state)
      if (accessCount === 1) return false
      // Subsequent: return truth
      return Reflect.get(target, prop)
    }
    if (prop === "addEventListener") {
      return (...args: Parameters<typeof ctrl.signal.addEventListener>) => {
        // Abort between check and listener registration
        ctrl.abort("too-late")
        return ctrl.signal.addEventListener(...args)
      }
    }
    return Reflect.get(target, prop)
  },
})

const combined = anySignal(proxy)
// combined.aborted === false   ← BUG: should be true
// combined.reason === undefined
```

See `packages/sdk/typescript/src/core/fetcher/signals.test.ts` for the complete regression test (committed in the linked PR).

## Proposed Fix

Re-check `signal.aborted` immediately after `addEventListener` to close the race window:

```diff
 for (const signal of signals) {
     if (signal.aborted) {
         controller.abort((signal as any)?.reason);
-        break;
+        return controller.signal;
     }

     signal.addEventListener("abort", () => controller.abort((signal as any)?.reason), {
         signal: controller.signal,
     });
+
+    // Re-check after adding listener: the signal may have aborted
+    // between the initial `signal.aborted` check and the
+    // `addEventListener` call. If it did, the abort event was
+    // already dispatched and the listener will never fire.
+    if (signal.aborted) {
+        controller.abort((signal as any)?.reason);
+        return controller.signal;
+    }
 }
```

### Fixed flow

```mermaid
flowchart TD
    A[Start: for each signal in signals] --> B{signal.aborted?}
    B -->|true| C[controller.abort]
    B -->|false| D["addEventListener('abort', handler)"]
    C --> Z[return controller.signal]

    D --> F{signal.aborted?<br/><br/>RE-CHECK}
    F -->|true| G["signal aborted during<br/>registration window"]
    G --> H[controller.abort]
    H --> Z
    F -->|false| I[continue to next signal]
    I --> A
```

faisalaffan:
## Description
Closes #16151

Fix race condition in generated `anySignal()` where abort events are silently dropped when a source signal aborts between the `signal.aborted` check and `addEventListener("abort", ...)` registration.

## Changes Made
- Re-check `signal.aborted` immediately after `addEventListener` to catch signals that aborted during the registration window
- Replace `break` with `return controller.signal` to avoid useless iteration after combined signal is already aborted
- Add unit tests: race condition reproducer via Proxy, abort reason forwarding

## Testing
- [x] Unit tests added/updated — 2 new test cases including the exact race condition scenario from the issue
- [x] Seed test verified: `pnpm seed test --generator ts-sdk --fixture bytes-upload --skip-scripts` — 1/1 passed

## Root Cause
In Node.js, adding an "abort" listener to an already-aborted `AbortSignal` does NOT fire the callback synchronously. The gap between `signal.aborted` check and `addEventListener` creates a window where aborts are silently dropped.

## Downstream Fix Reference
Patch applied in `latitude-dev/latitude-llm` (PR https://github.com/latitude-dev/latitude-llm/pull/3336) — same fix with full test suite.


faisalaffan:
hi @Swimburger @amckinney can you check this ?

Swimburger:
Thank you for reporting and submitting this PR. I went ahead and created a PR with other changes necessary to make CI pass and added a version entry. The change has been published as version 3.71.5.

Issue #15956, opened 2026-05-17
Runnable endpoint component does not respect auth x-extensions


Devon-White:
### Component

Docs

### Priority

P3 - Low (Nice to have)

### Bug Description

Consider the following in our OpenAPI spec:

```yaml
securitySchemes:
    SignalWireBasicAuth:
      type: http
      scheme: Basic
      description: |-
        SignalWire Basic Authentication using Project ID and API Token.

        The client sends HTTP requests with the Authorization header containing
        the word Basic followed by a space and a base64-encoded string of project_id:token.
        The project ID will be used as the username and the API token as the password.

        Example:
        ```
        Authorization: Basic base64(project_id:token)
        ```
      x-fern-basic:
        username:
          name: project_id
          env: SIGNALWIRE_PROJECT_ID
        password:
          name: api_token
          env: SIGNALWIRE_API_TOKEN
```

As seen here, the security schema section correctly replaces `username` with `project_id` and `password` with `api_token` using the `x-fern-basic` extension.

As seen here, it properly replaces the labels in the auth fields: https://signalwire.com/docs/apis/rest/subscribers/tokens/create-subscriber-token?explorer=true

However if I reference this endpoint with the `<RunnableEndpoint />` component, it will retain the default labels for the security schema: `username` & `password`

### Are you interested in contributing a fix?

No

PR #16185, opened 2026-06-02
fix(docs): preserve x-fern-basic labels for runnable endpoints

shubhamsinnh:
﻿## Summary

Adds regression coverage for `x-fern-basic` auth labels on runnable endpoint API definitions.

The existing OpenAPI fixture uses custom basic auth field names:

- `project_id` for the username field
- `api_token` for the password field

This change verifies that those labels are preserved through both:

- OpenAPI/IR to FDR conversion in `@fern-api/register`
- IR to Docs API definition conversion in `@fern-api/docs-resolver`

It also checks that the endpoint `multiAuth` entry points back to the same basic auth scheme, so callers rendering runnable endpoints can resolve the custom labels from `authSchemes`.

## Notes

The runnable endpoint component itself does not appear to live in this package. If the UI still renders `username` / `password`, the remaining fix is likely in the renderer resolving endpoint auth scheme ids into `apiDefinition.authSchemes`.

## Testing

- `tsc --noEmit -p packages/cli/register/tsconfig.json`
- `tsc --noEmit -p packages/cli/docs-resolver/tsconfig.json`
- `git diff --check HEAD`
- `vitest --run src/ir-to-fdr-converter/__test__/openapi-from-flag.test.ts --testNamePattern "x-fern-basic" --config vitest.config.ts --reporter verbose`
- `vitest --run src/__test__/x-fern-basic-auth.test.ts --config vitest.config.ts --reporter verbose`

The Vitest runs used a local Node 20 preload for `navigator`; the repo's DevBox setup pins Node 22.

### DIFF
--- a/packages/cli/docs-resolver/src/__test__/x-fern-basic-auth.test.ts
+++ b/packages/cli/docs-resolver/src/__test__/x-fern-basic-auth.test.ts
@@ -1,0 +1,56 @@
+import assert from "node:assert";
+
+import { OSSWorkspace } from "@fern-api/lazy-fern-workspace";
+import { createMockTaskContext } from "@fern-api/task-context";
+import { loadAPIWorkspace } from "@fern-api/workspace-loader";
+import { AbsoluteFilePath, resolve } from "@fern-api/fs-utils";
+
+import { convertIrToApiDefinition } from "../utils/convertIrToApiDefinition.js";
+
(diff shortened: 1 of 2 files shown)


shubhamsinnh:
@amckinney could you sanity check if this is the right layer for #15956?

PR #16224, opened 2026-06-03
fix(python): respect optional response-property types and oauth request-properties mapping

iamnamananand996:
## Description
Fixes two allowed-failure fixtures in the Python SDK generator by addressing the generator root causes (not the generated output).

**1. `response-property`** — When a `response: { property: ... }` endpoint resolves to an optional body (e.g. a named alias to `optional<...>` such as `OptionalWithDocs`), the raw client emits `HttpResponse(response=_response, data=None)` for empty responses, but the generic was `HttpResponse[str]`, so mypy flagged `data=None`. The generator used `response_type.is_optional` on the body's type hint, which is `False` for a named alias even though the alias resolves to optional. Now it uses `resolved_schema_is_optional_or_unknown(response.response_body_type)`, so the property type is wrapped in `Optional` whenever the underlying body resolves to optional.

**2. `oauth-client-credentials-openapi`** — The generated OAuth token provider hard-coded `client_id`/`client_secret` kwargs when calling the token endpoint, but the endpoint params come from the auth scheme's `request-properties` mapping (e.g. `username`/`password` for the OpenAPI spec). The provider now resolves the actual generated parameter name from the `request-properties` mapping for `client_id`, `client_secret`, and `refresh_token`.

`oauth-client-credentials-custom` is **left in `allowedFailures`** — see note below.

## Changes Made
- `endpoint_function_generator.py`: `_get_nested_json_response_type` wraps the property type in `Optional` based on the resolved schema, not the rendered type hint.
- `oauth_token_provider_generator.py`: token/refresh invocations use parameter names resolved from the `request-properties` mapping instead of hard-coded `client_id`/`client_secret`/`refresh_token`.
- Regenerated `response-property`, `oauth-client-credentials-openapi`, and `oauth-client-credentials-custom` snapshots (all other oauth/response-property fixtures regenerate with no diff).
- Removed `response-property` and `oauth-client-credentials-openapi` from `seed/python-sdk/seed.yml` `allowedFailures`.
- Added changelog entries under `generators/python/sdk/changes/unreleased/`.

### Note on `oauth-client-credentials-custom`
The mapping fix is applied here too (now calls `get_token_with_client_credentials(cid=..., csr=...)` instead of `client_id=`/`client_secret=`), which removes the original 4 "unexpected keyword argument" mypy errors. However, this fixture's token endpoint additionally requires `scp` (the `scopes` mapping) and `entity_id` (a non-literal `customProperties` entry), which have **no value source** in the OAuth token provider abstraction. mypy was previously masking these as soon as the unexpected-kwarg errors are resolved, so the fixture now fails with `Missing named argument "scp"`/`"entity_id"` instead. This is a deeper feature gap (the token provider would need to source/forward scopes and arbitrary custom properties) and is also unsupported in the TypeScript and Java generators (Go only "passes" because Go does not enforce required struct fields). Left in `allowedFailures` pending a decision on how to model custom/scope property values.

## Testing
- [x] Full `poetry run mypy .` passes on regenerated `response-property` (52 files) and `oauth-client-credentials-openapi` (50 files).
- [x] `mypy` + `pre-commit run` pass on the changed generator source.
- [x] All oauth + response-property fixtures regenerate via seed; only the three intended snapshots change.

Link to Devin session: https://app.devin.ai/sessions/aa201fe3b32348b59c79ab37d86a80b3

### DIFF
--- a/generators/python/src/fern_python/generators/sdk/client_generator/oauth_token_provider_generator.py
+++ b/generators/python/src/fern_python/generators/sdk/client_generator/oauth_token_provider_generator.py
@@ -533,6 +533,7 @@ class OAuthTokenProviderGenerator:
         self, client_credentials: ir_types.OAuthClientCredentials
     ) -> AST.FunctionInvocation:
         # TODO(amckinney): Support non-in-lined request types.
+
         # The token endpoint's parameter names come from the configured request-properties
         # mapping (e.g. client-id -> $request.cid), which may differ from "client_id"/"client_secret".
         token_request_properties = client_credentials.token_endpoint.request_properties
--- a/generators/python/src/fern_python/generators/sdk/client_generator/oauth_token_provider_generator.py
+++ b/generators/python/src/fern_python/generators/sdk/client_generator/oauth_token_provider_generator.py
@@ -533,13 +533,16 @@ class OAuthTokenProviderGenerator:
         self, client_credentials: ir_types.OAuthClientCredentials
     ) -> AST.FunctionInvocation:
         # TODO(amckinney): Support non-in-lined request types.
+        # The token endpoint's parameter names come from the configured request-properties
+        # mapping (e.g. client-id -> $request.cid), which may differ from "client_id"/"client_secret".
+        token_request_properties = client_credentials.token_endpoint.request_properties
         kwargs = [
             (
-                "client_id",
+                self._get_request_property_parameter_name(token_request_properties.client_id),
                 AST.Expression(f"self.{self._get_client_id_member_name()}"),
             ),
             (
-                "client_secret",
+                self._get_request_property_parameter_name(token_request_properties.client_secret),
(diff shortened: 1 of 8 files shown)


generators/python/src/fern_python/generators/sdk/client_generator/oauth_token_provider_generator.py
```diff
@@ -533,13 +533,16 @@ def _get_refresh_function_invocation(
         self, client_credentials: ir_types.OAuthClientCredentials
     ) -> AST.FunctionInvocation:
         # TODO(amckinney): Support non-in-lined request types.
```
iamnamananand996, line 535: can we remove this TODO now ?

# TODO(amckinney): Support non-in-lined request types.

or still it is needed ??

fern-support (approved):

PR #16336, opened 2026-06-08
fix(swift): emit .rawValue for nullable enum query parameters [FER-11120]

adidavid014:
## Description

When a query parameter is a **string-backed enum that is nullable** (`Nullable<Enum>`, optionally also optional), the Swift generator emits `.string($0)` instead of `.string($0.rawValue)`, causing `swift build` to fail with type errors. This blocks the Whop Swift SDK (580 errors across 33 clients).

### Where `Nullable<Enum>` comes from
Fern models **optional** (may be absent) and **nullable** (may be explicitly `null`) as distinct concepts. A spec that declares a query param both optional and nullable produces `Optional<Nullable<Enum>>`:

```yaml
# minimal repro
paths:
  /ads:
    get:
      parameters:
        - name: direction
          in: query
          required: false                 # optional
          schema:
            oneOf:
              - $ref: "#/components/schemas/Direction"
              - type: "null"              # nullable
components:
  schemas:
    Direction: { type: string, enum: [asc, desc] }
```

→ generated as `direction: Nullable<Direction>?`. Before this fix that emitted `.string($0)` (a `Direction`, not a `String`) → compile error. This pattern is common in machine-generated specs (Whop marks ~91% of its query params nullable).

## Root Cause

`resolvesToAnEnumWithRawValues()` requires a symbol-type `TypeReference`, but nullable-wrapped enums have variant type `"nullable"` — calling `nonOptional()` on `Optional<Nullable<Direction>>` yields `Nullable<Direction>`, not the underlying `Direction` symbol. Two code paths were affected:

1. **Optional + Nullable** (`Nullable<Enum>?`): `resolvesToAnEnumWithRawValues(swiftType.nonOptional())` — `nonOptional()` peels the `Optional` but leaves `Nullable`, so the check returns false.
2. **Non-optional + Nullable** (`Nullable<Enum>`): The `.map` closure body never checked for enums at all.

## Changes Made

In `EndpointMethodGenerator.ts` query param generation:

- **Optional+Nullable path**: `resolvesToAnEnumWithRawValues(swiftType.nonOptional())` → `resolvesToAnEnumWithRawValues(swiftType.nonOptional().nonNullable())`
- **Non-optional+Nullable path**: Added the missing `resolvesToAnEnumWithRawValues(swiftType.nonNullable())` check with `.rawValue` emission

Before → After in generated Swift:
```diff
-"role": role.wrappedValue.map { .string($0) }
+"role": role.wrappedValue.map { .string($0.rawValue) }
-"secondaryRole": secondaryRole?.wrappedValue.map { .string($0) }
+"secondaryRole": secondaryRole?.wrappedValue.map { .string($0.rawValue) }
```

### Scope — other call sites audited
`resolvesToAnEnumWithRawValues` has four call sites in this file; the other two are correct and intentionally unchanged:
- the plain **required enum** path (no `Optional`/`Nullable` wrapper to peel) — already emitted `.rawValue`;
- `inferQueryParamCaseName`, which recurses through both `optional` and `nullable` before the enum check, so the case name (`.string`) is correct for `Nullable<Enum>`.

Query-param serialization is the only place enum values are hand-mapped to a `QueryParameter` case; path/header params and request bodies serialize via `Codable` (which uses `rawValue` automatically) and are unaffected.

## Testing

- [x] Seed test `nullable-optional` (both `no-custom-config` and `nullable-as-optional` variants) — confirms `.rawValue` is now emitted for all three nullable enum patterns
- [x] Seed test `enum` — confirms no regression for non-nullable enum query params
- [x] `pnpm lint:biome` and `pnpm format` pass
- [x] **End-to-end compile:** regenerated the Whop Swift SDK (242-op spec, 60 nullable-enum query params across 33 clients) with this change → `swift build` → **0 errors (was 580)**. This also clears the ~108 errors that surfaced as the misleading `the compiler is unable to type-check this expression in reasonable time` — those were the same bug (the type mismatch inside a large `[String: QueryParameter]` literal exhausting the solver), not a separate issue.

Link to Devin session: https://app.devin.ai/sessions/06593815f5714564ac04684293e57d04
Requested by: @adidavid014

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/16336" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


adidavid014:
@claude review

kafkas (approved):

PR #16517, opened 2026-06-12
feat(generator-cli): replay --import-history and --fernignore-action migrate now error with migration guidance (train 8/9)

tstanmay13:
## Merge train: 8 of 9 (generator-cli 3/4) — merge after train PR 7

Part of the replay hardening train from the 2026-06-11 coingecko incident (Linear: FER-11258). Consumer half of engine train PR 3 (engine ADR-0005: bootstrap is clean-start only). Independently compile- and runtime-safe against the pinned `@fern-api/replay` 0.18.0, so it does not need to wait for the engine publish.

## Summary

`--import-history` imported overlapping historical patches that duplicate content on every regeneration (compile-breaking). The feature is removed in the engine; this PR gives the generator-cli flags the agreed parse-and-error treatment for one release:

- `replay bootstrap --import-history` and `replay init --import-history` still parse but exit 1 with migration guidance: reset customized generator-owned files to their as-generated state → `replay bootstrap` with no flags → re-apply customizations as one commit (with the duplicate-content rationale)
- `--fernignore-action migrate` gets the same treatment (standalone it was a silent no-op — its action only applied under import-history)
- `importHistory` plumbing removed from `replay-init.ts` / `replay-submit-init.ts`; the auto-bootstrap call site keeps `importHistory: false` against the pinned engine and drops it at the pin bump (coordination item on FER-11258)
- Error message constants are duplicated from the engine deliberately (the engine doesn't export them, keeping its api-surface golden byte-stable) — one-release drift risk, recorded in engine ADR-0005

## Test plan

- New `replay-removed-flags.test.ts` 5/5; red verified on origin/main (flags exited 0 there)
- Live wiring verified against the freshly compiled `lib/cli.js` + pinned engine 0.18.0: both flags print the full message and exit 1; no-flags clean start exits 0 and writes a lockfile
- Full generator-cli lane uncached: 421 passed / 1 skipped; forced uncached tsc compile green

## Review

Adversarially reviewed: **approve, zero required changes**. Next-release follow-up (recorded in engine ADR-0005): delete the parse stubs.

Linear: FER-11258

### DIFF
--- a/packages/generator-cli/src/__test__/replay-removed-flags.test.ts
+++ b/packages/generator-cli/src/__test__/replay-removed-flags.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from "vitest";
+import {
+    getRemovedReplayFlagError,
+    IMPORT_HISTORY_REMOVED_MESSAGE,
+    MIGRATE_FERNIGNORE_REMOVED_MESSAGE
+} from "../replay/removed-flags";
+
+describe("getRemovedReplayFlagError", () => {
+    it("rejects --import-history with migration guidance", () => {
+        const error = getRemovedReplayFlagError({ importHistory: true });
+        expect(error).toBe(IMPORT_HISTORY_REMOVED_MESSAGE);
+        expect(error).toContain("--import-history has been removed");
+        expect(error).toContain("duplicate content");
+        // The three-step clean-start migration sequence.
+        expect(error).toContain("as-generated");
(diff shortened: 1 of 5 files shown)


tstanmay13:
Superseded by #16522, which combines all four generator-cli consumer changes into one linear PR (rebuilt on current main and re-validated as a unit: compile 13/13, 451 tests passing). Closing in favor of the single PR. Branch left intact for reference.

PR #16708, opened 2026-06-25
fix(typescript): guard against null/undefined in file upload core utility

cadesark:
## Description

Fixes a regression (v3.12.x → v3.68.0) where passing `null` or `undefined` to an optional multipart file upload field crashes the SDK with `TypeError: Cannot use 'in' operator to search for 'path' in undefined`.

## Changes Made

**`appendFile` throws on null/undefined** (in `FormDataWrapper`):
```typescript
public async appendFile(key: string, value: Uploadable): Promise<void> {
    if (value == null) {
        throw new Error(
            `File upload for "${key}" received ${value === null ? "null" : "undefined"}. The generated code should not call appendFile with null/undefined — check that optional file fields are guarded before this call.`
        );
    }
    // ...
}
```

This is safe because the generated code already guards optional file fields:
```typescript
if (request.maybe_file != null) {
    await _body.appendFile("maybe_file", request.maybe_file);
}
```

**Defensive throw in `getFileWithMetadata`** (internal function — if null somehow bypasses the guard):
```typescript
if (file == null) {
    throw new Error(`Expected file to be a Blob, Buffer, ReadableStream, ...`);
}
```

**`toMultipartDataPart`** — clean signature (no `| undefined` return, no internal null guard). It assumes callers have already validated.

- Uses `Error` (not `TypeError`) for consistency with line 86 in the same file
- Unit tests added for null/undefined paths in `toBinaryUploadRequest`
- All 35 seed `file.ts`, 12 seed `FormDataWrapper.ts`, and 33 seed test files updated

## Testing

- [x] Unit tests added for null/undefined input
- [x] All seed test outputs updated and passing (50/50 CI checks green)

Link to Devin session: https://app.devin.ai/sessions/4ae2868a605043aeab6adfcd30d22593
Requested by: @cadesark


generators/typescript/utils/core-utilities/src/core/file/file.ts
```diff
@@ -22,7 +22,11 @@ export async function toBinaryUploadRequest(
 
 export async function toMultipartDataPart(
     file: Uploadable,
-): Promise<{ data: Uploadable.FileLike; filename?: string; contentType?: string }> {
+): Promise<{ data: Uploadable.FileLike; filename?: string; contentType?: string } | undefined> {
+    if (file == null) {
+        console.warn(`File upload received ${file === null ? "null" : "undefined"}, skipping.`);
+        return undefined;
```
Swimburger: since `file` is not null | undefined, I still think we should throw an error here.

generators/typescript/utils/core-utilities/src/core/file/file.ts
```diff
@@ -37,6 +41,11 @@ async function getFileWithMetadata(
     file: Uploadable,
     { noSniffFileSize }: { noSniffFileSize?: boolean } = {},
 ): Promise<Uploadable.WithMetadata> {
+    if (file == null) {
+        throw new TypeError(
```
Swimburger: I think we should use a different error type, not `TypeError`. Can you figure out what is most idiomatic in this case?

generators/typescript/utils/core-utilities/src/core/form-data-utils/FormDataWrapper.template.ts
```diff
@@ -79,7 +79,11 @@ export class Node18FormData implements CrossPlatformFormData {
     }
 
     public async appendFile(key: string, value: Uploadable): Promise<void> {
-        const { data, filename } = await toMultipartDataPart(value);
+        const part = await toMultipartDataPart(value);
+        if (part == null) {
+            return;
+        }
```
Swimburger: shouldn't be necessary, as `value` should never be null'ish at this point, and `part` should never be null'ish at this point.

Swimburger (changes requested):
Looking at this code, I think the check for null'ish should be higher up, and enforced so null'ish doesn't ever reach the code that you're changing here.

generators/typescript/utils/core-utilities/src/core/form-data-utils/FormDataWrapper.template.ts
```diff
@@ -79,6 +79,10 @@ export class Node18FormData implements CrossPlatformFormData {
     }
 
     public async appendFile(key: string, value: Uploadable): Promise<void> {
+        if (value == null) {
+            console.warn(`File upload for "${key}" received ${value === null ? "null" : "undefined"}, skipping.`);
+            return;
+        }
```
Swimburger, line 86: We should throw here. appendFile should not be allowed to be called with null.

generators/typescript/utils/core-utilities/src/core/form-data-utils/FormDataWrapper.template.ts
```diff
@@ -181,6 +189,10 @@ export class WebFormData implements CrossPlatformFormData {
     }
 
     public async appendFile(key: string, value: Uploadable): Promise<void> {
+        if (value == null) {
```
Swimburger, line 194: We should throw here. appendFile should not be allowed to be called with null.

generators/typescript/utils/core-utilities/src/core/form-data-utils/FormDataWrapper.template.ts
```diff
@@ -221,6 +233,10 @@ export class FormDataWrapper {
...
+            console.warn(`File upload for "${key}" received ${value === null ? "null" : "undefined"}, skipping.`);
```
Swimburger: We should throw here. appendFile should not be allowed to be called with null.

Swimburger (changes requested):

Swimburger (approved):

PR #16919, opened 2026-07-07
fix(go): decompress gzip responses that net/http skips when the spec sets an explicit Accept-Encoding header

iamnamananand996:
### Why is this needed?
Fern generators never add `Accept-Encoding` themselves — but when an API definition/OpenAPI spec declares `Accept-Encoding: gzip` as a request header, the generated Go SDK sets it explicitly on every request. The request goes through fine and the server (correctly) responds with a gzip-compressed body (`Content-Encoding: gzip`).

The problem: Go's `net/http` transport only performs transparent gzip decompression **when it added the `Accept-Encoding` header itself**. If the caller sets it explicitly, the transport intentionally skips decompression (see the `Transport.DisableCompression` docs: *"if the user explicitly requested gzip it is not automatically uncompressed"*). The SDK then tries to JSON-deserialize raw gzip bytes and fails at runtime with a parse error.

### How this PR fixes it
The generated SDK core now decompresses the body itself whenever the **response** has `Content-Encoding: gzip` and the transport did not already decode it (`resp.Uncompressed == false`). Decompression is driven purely by the response header, so:
- uncompressed responses pass through untouched
- responses already decoded by the transport (the default no-explicit-header path) are not double-decompressed

## Changes Made
- `caller.go_`: new `decompressResponse(resp)` helper — wraps `resp.Body` in a `gzip.Reader` when `Content-Encoding: gzip` is present and not already handled by the transport; strips `Content-Encoding`/`Content-Length` headers and sets `resp.Uncompressed = true`. Empty bodies (`io.EOF` from `gzip.NewReader`) are passed through untouched.
- `caller.go_`: `gzipReadCloser` closes both the gzip reader and the underlying body, and caps total decompressed bytes at `maxDecompressedBodySize` (1 GiB) to guard against gzip decompression bombs (per security review).
- `streamer.go_`: applies the same decompression to streaming responses before handing off to the stream reader.
- Changelog entry under `generators/go/sdk/changes/unreleased/`.

Part of a series of per-language gzip fixes: Go (this PR), Rust #16920, C# #16921, Ruby #16922, Java #16923.

## Testing
- [x] Unit tests added/updated — `TestCallWithGzipResponse` in `caller_test.go_` (httptest server responds with gzip body when the request sets `Accept-Encoding: gzip` explicitly)
- [x] Manual testing completed — rendered templates into the `seed/go-sdk/alias` fixture module; `go build ./...`, `go vet`, and the `./internal` test suite pass

Link to Devin session: https://app.devin.ai/sessions/90aef2978d1448fd85ad153356d5625f

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/16919" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/generators/go-v2/base/src/asIs/internal/caller.go_
+++ b/generators/go-v2/base/src/asIs/internal/caller.go_
@@ -2,6 +2,7 @@ package internal
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"errors"
@@ -107,6 +108,10 @@ func (c *Caller) Call(ctx context.Context, params *CallParams) (*CallResponse, e
 
 	// Close the response body after we're done.
 	defer func() { _ = resp.Body.Close() }()
+
+	if err := decompressResponse(resp); err != nil {
+		return nil, err
+	}
 
 	// Check if the call was cancelled before we return the error
 	// associated with the call and/or unmarshal the response data.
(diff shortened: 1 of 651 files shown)


broady:
Are you sure this is needed for Go? net/http gunzips this by default. I'm reading the PR description and it's still unclear why this change is needed. Should we not be sending Accept-Encoding: gzip? Why is their API sending a zipped response when the request is missing Accept-Encoding?

broady:
We don't need to modify the http.Request - should look like 
```
defer func() { _ = resp.Body.Close() }()
body := resp.Body
if !body.Uncompressed {
  body = &gunzipReader{body}
}
```

broady:
We don't need to wrap in a ReadCloser - returning a Reader should be sufficient

broady:
what is with the wrapped reader? is this some kind of hand-rolled io.LimitedReader?

i don't think the justification of a zip bomb makes sense - these are trusted payloads

please justify this

generators/go-v2/base/src/asIs/internal/caller.go_
```diff
@@ -263,19 +269,43 @@ func newFormURLEncodedRequestBody(request interface{}, bodyProperties map[string
...
+		return resp.Body, nil
+	}
+	if !strings.EqualFold(resp.Header.Get("Content-Encoding"), "gzip") {
+		return resp.Body, nil
+	}
+	gzipReader, err := gzip.NewReader(resp.Body)
+	if err != nil {
+		if err == io.EOF {
+			// The response body is empty, so there is nothing to decompress.
+			return resp.Body, nil
```
broady: This doesn't look correct - gzip.NewReader actually reads from resp.Body - returning resp.Body here is invalid.

broady (commented):
check

broady (approved):

PR #16921, opened 2026-07-07
fix(csharp): enable automatic gzip/deflate decompression on the default HttpClient

iamnamananand996:
### Why is this needed?
Fern generators never add `Accept-Encoding` themselves — but when an API definition/OpenAPI spec declares `Accept-Encoding: gzip` as a request header, the generated C# SDK sets it explicitly on every request. The server then (correctly) responds with a gzip-compressed body (`Content-Encoding: gzip`).

The problem: .NET's `HttpClientHandler.AutomaticDecompression` defaults to `None`, and the generated core never configures it — so `HttpClient` hands back raw gzip bytes and JSON deserialization fails at runtime with a parse error.

### How this PR fixes it
The default `HttpClient` created by the generated core now comes from a new `DefaultHttpClientFactory` that enables .NET's built-in automatic decompression:

```csharp
var handler = new HttpClientHandler
{
#if NET5_0_OR_GREATER
    AutomaticDecompression = DecompressionMethods.All,
#else
    AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
#endif
};
return new HttpClient(handler);
```

The handler decompresses based on the **response's** `Content-Encoding`, so explicit spec-defined `Accept-Encoding` request headers work and uncompressed responses are untouched. No `Accept-Encoding` headers are added manually. User-supplied `HttpClient`s via `ClientOptions` are intentionally out of scope and are not modified.

## Changes Made
- New core as-is file `DefaultHttpClientFactory.Template.cs` (registered in `AsIs.ts` and `SdkGeneratorContext`)
- `BaseOptionsGenerator.getHttpClientField`: default initializer changed from `new HttpClient()` to `DefaultHttpClientFactory.Create()`
- New as-is test template `test/RawClientTests/GzipResponseTests.Template.cs` exercising the default client path: gzip response decoded with empty `Content.Headers.ContentEncoding` after handler decompression, plus a test asserting identity (uncompressed) responses are unchanged
- Changelog entry under `generators/csharp/sdk/changes/unreleased/`
- [ ] Updated README.md generator (not applicable)

Part of a series of per-language gzip fixes: Go #16919, Rust #16920, C# (this PR), Ruby #16922, Java #16923.

## Testing
- [x] Unit tests added/updated — `GzipResponseTests` (WireMock serves a gzip-compressed body with `Content-Encoding: gzip`; asserts decoded payload and empty `ContentEncoding`; identity response test)
[...]

PR #16922, opened 2026-07-07
fix(ruby): restore Net::HTTP transparent gzip decoding when the spec sets an explicit Accept-Encoding header

iamnamananand996:
### Why is this needed?
Fern generators never add `Accept-Encoding` themselves — but when an API definition/OpenAPI spec declares `Accept-Encoding: gzip` as a request header, the generated Ruby SDK sets it explicitly on every request. The server then (correctly) responds with a gzip-compressed body (`Content-Encoding: gzip`).

The problem: `Net::HTTP` transparently inflates gzip/deflate response bodies only when it sets its own default `Accept-Encoding` header — `Net::HTTPGenericRequest#[]=` flips `@decode_content` to `false` the moment the caller assigns `Accept-Encoding` explicitly. The SDK then tries to JSON-parse raw gzip bytes and fails at runtime.

### How this PR fixes it
In the generated `RawClient#build_http_request`: when the merged request headers include `Accept-Encoding`, the request is extended with a `DecodeContent` module whose `decode_content` returns `true`, restoring Net::HTTP's built-in transparent decoding. Decoding is still driven by the **response's** `Content-Encoding: gzip`/`deflate` header, so uncompressed responses are untouched.

## Changes Made
- `raw_client.Template.rb`: re-enable `decode_content` via `request.extend(DecodeContent)` when an `Accept-Encoding` header is present
- `test_raw_client.Template.rb`: unit test with a local TCP server returning a gzip body with `Content-Encoding: gzip`; asserts the request sent `Accept-Encoding: gzip` and the body is decoded
- Changelog entry under `generators/ruby-v2/sdk/changes/unreleased/`

Part of a series of per-language gzip fixes: Go #16919, Rust #16920, C# #16921, Ruby (this PR), Java #16923.

## Testing
- [x] Unit tests added/updated (as-is test template)
- [x] Manual testing completed — verified with a standalone Ruby repro against a local gzip server: without the fix the body is 51 raw gzip bytes; with `DecodeContent` extended, the body decodes to the original JSON

Link to Devin session: https://app.devin.ai/sessions/90aef2978d1448fd85ad153356d5625f

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/16922" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


davidrojas-pm (approved):
one nit on the test template: the `TCPServer` + thread have no `ensure` cleanup, so if the client ever fails to connect `server.accept` blocks and the test hangs instead of failing fast. propagates to all the generated tests. fine to punt.

zamn (approved):

fern-support (approved):

PR #15641, opened 2026-05-01
feat(typescript): add unified `auth` option to BaseClientOptions

Swimburger:
## Description

When `alwaysSendAuth: true` is configured, the generated `BasicAuthProvider` (or other auth providers) throws if credentials aren't provided — even when users handle auth through custom fetchers. This adds a unified `auth` option to `BaseClientOptions` that lets users supply custom auth handling, bypassing the built-in provider.

## Changes Made

- Added `AuthOption` type as an undiscriminated union of:
  1. A function `(arg?: { endpointMetadata? }) => Promise<AuthRequest>` — for fully custom auth
  2. An `AuthProvider` object (has `getAuthRequest` method) — for custom provider instances
  3. The existing auth credential options (e.g. `{ username, password }`) — backwards compatible
- Added optional `auth?: AuthOption` field to generated `BaseClientOptions` type
- Updated `normalizeClientOptionsWithAuth` to detect which variant was passed:
  - If function → wraps in `{ getAuthRequest: fn }` adapter
  - If `AuthProvider` object → uses directly
  - Otherwise → merges as credential options via `Object.assign`, falls through to default provider creation
- Moved `isAuthProvider` type guard to `core/auth/AuthProvider.ts` (colocated with `AuthProvider` interface) instead of generating it inline in each SDK
- Updated test mock context to include `AuthRequest`, `EndpointMetadata`, and `isAuthProvider` references
- Regenerated **all** auth-related seed fixtures (213 fixtures, all pass)
- Added unreleased changelog entry
- [ ] Updated README.md generator (not applicable — generated code only)

**Backwards compatible**: existing top-level credential fields continue to work unchanged. The `auth` option is purely additive.

### Auth schemes covered

All 5 auth scheme types + both multi-auth modes regenerated and verified:
- **bearer**: simple-api, imdb, exhaustive, pagination, trace, etc.
- **basic**: basic-auth, basic-auth-pw-omitted, basic-auth-environment-variables
- **header**: header-auth, header-auth-environment-variable
- **inferred** (OAuth): inferred-auth-explicit, inferred-auth-implicit (+ api-key, no-expiry, reference variants)
- **oauth**: oauth-client-credentials (+ custom, default, environment-variables, mandatory-auth, nested-root, openapi, reference, with-variables)
- **ANY** (multiple schemes): any-auth (3 variants)
- **ENDPOINT_SECURITY** (routing): endpoint-security-auth

### Example usage (custom fetcher pattern)
```typescript
new MavenAGIClient({
[...]

PR #16970, opened 2026-07-08
test(typescript): pin exact typing of multi-scheme (any) auth options (FER-11540)

adidavid014:
## The bug (FER-11540)

Under a multi-scheme `auth: any` config (Payabli: `any: [BearerAuth (oauth client-credentials), ApiKeyAuth]`), the generated TS client nests each scheme's credentials under a wrapper key, e.g. `new PayabliClient({ apiKeyAuth: { apiKey } })`.

A caller migrating from the old single-scheme (flat) shape would write `new PayabliClient({ apiKey: "..." })`. The report was that this **compiled and ran with no error** — the field was silently ignored, no auth header was sent, and the API returned a live 401. Silent no-ops are a nasty footgun for anyone migrating from single-scheme (flat) to multi-scheme (nested) auth.

## Investigation & finding

I reproduced and traced the type construction:
- `BaseClientTypeGenerator` emits `BaseClientOptions = { ...base } & AnyAuthProvider.AuthOptions<[...]>`.
- `AnyAuthProviderGenerator.writeOptions()` defines `AuthOptions<TAuthProviders> = AtLeastOneOf<TAuthProviders>` (a union of intersections; each scheme's credentials nested under a wrapper key like `apiKeyAuth`/`bearerAuth`).

I then type-checked the exact scenario against:
1. A faithful reconstruction of the emitted type (both Payabli's 2-scheme topology and the seed `any-auth` 5-scheme fixture), and
2. **Payabli's own real generated `.d.ts`** in their preview SDK.

Result in every case on current `main`: the flat/unknown shape is **already a compile-time type error**, and the nested shape type-checks cleanly:
- `{ apiKey: "..." }` (inline literal) -> `TS2353: Object literal may only specify known properties, and 'apiKey' does not exist in type 'BaseClientOptions'` (or `TS2559` when the wrapper key collides with the flat name).
- `{ token: "..." }`, `{ nonsense: 123 }` -> type errors.
- `{ apiKeyAuth: { apiKey: "..." } }`, `{ bearerAuth: { clientId, clientSecret } }` -> compile cleanly.

The desired end-state from the ticket — **option (b): the flat/unknown option shape is a compile-time type error, not a silent no-op** — is therefore already the behavior. It was delivered by the auth-provider refactor / unified `auth` option work (#15641, landed 2026-05-01), which made the emitted options type exact via `AtLeastOneOf`. The customer's originally-observed failure came from an older generated SDK that predated that work (the ticket cites reading 3.75.1 *source*, but the runtime 401 was from a stale build).

## Why option (b) and not (a)

Exact typing was already in place and is the cleaner fix: it turns the mistake into an immediate compile error rather than best-effort routing of an ambiguous flat field. Accepting flat fields (option a) would re-introduce ambiguity across schemes (e.g. which provider owns a bare `token`?) and mask the migration boundary. Option (b) keeps existing *nested* multi-scheme code working while making the flat mistake impossible to hit silently.

## What this PR changes

Because the fix already shipped, the highest-value, durable contribution is a **regression guard** so the silent no-op cannot quietly return. Added type-level tests to `AuthProviders.test.ts` that reconstruct the emitted `AnyAuthProvider.AuthOptions` type (the `AtLeastOneOf` / `UnionToIntersection` utilities as emitted by `AnyAuthProviderGenerator.writeOptions()`) for an `any: [OAuth, ApiKey]` config and type-check real usages via ts-morph `getPreEmitDiagnostics()`:

- accepts the nested per-scheme wrapper shape (`apiKeyAuth: { apiKey }`)
- accepts the nested oauth client-credentials shape (`bearerAuth: { clientId, clientSecret }`)
- rejects the flat single-scheme `apiKey` shape (inline literal)
- rejects the flat bearer `token` shape
- rejects an entirely unknown option key

Note on the widened-variable path (`const o = { apiKey }; new Client(o)`): TypeScript deliberately skips excess-property checking for non-literals, so an unknown key on a widened variable is accepted structurally regardless of how the type is authored. That is standard TS semantics and not something the generator can (or should) tighten, so it is intentionally not asserted here. The ticket's documented reproduction (an inline object literal) is fully covered.

## Verification

- Full generator unit suite: **557 passed** (was 552; +5 new tests), no regressions.
- New tests fail as expected if the type is loosened, pass with the current exact typing.
- **No generated-output change**: only a generator test file is touched; seed fixtures are unchanged (regenerating produces no diff). No `versions.yml` bump, since there is no user-facing behavior change to release.

Related OAuth generator fixes for the same customer situation: Rust #16909, C# #16928, Go #16945, Java #16947.

Closes FER-11540

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/16970" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/generators/typescript/sdk/client-class-generator/src/__test__/AuthProviders.test.ts
+++ b/generators/typescript/sdk/client-class-generator/src/__test__/AuthProviders.test.ts
@@ -738,4 +738,104 @@ describe("HeaderAuthProviderGenerator", () => {
             expect(context.sourceFile.getFullText()).toMatchSnapshot();
         });
     });
+
+    // ──────────────────────────────────────────────────────────────────────────
+    // Regression: FER-11540
+    //
+    // Under multi-scheme `auth: any` (e.g. OAuth client-credentials + ApiKey), the
+    // generated `BaseClientOptions` nests each scheme's credentials under a wrapper
+    // key (e.g. `apiKeyAuth: { apiKey }`). Passing the OLD flat single-scheme shape
+    // (`{ apiKey: "..." }`) must NOT silently type-check: previously it compiled,
+    // sent no auth header, and produced a live 401 at runtime.
+    //
+    // These tests pin the exact-object typing of the emitted `AnyAuthProvider.AuthOptions`
+    // type so a regression that re-opens the silent no-op is caught at build time.
+    // We reconstruct the type as emitted by `AnyAuthProviderGenerator.writeOptions()`
+    // (the `AtLeastOneOf` / `UnionToIntersection` utilities) plus representative
+    // per-scheme `AuthOptions`, then type-check real usages.
+    // ──────────────────────────────────────────────────────────────────────────
+    describe("multi-scheme (any) auth options typing [FER-11540]", () => {
+        // Mirrors AnyAuthProviderGenerator.writeOptions() + BaseClientTypeGenerator output
+        // for `auth: any: [OAuth (client-credentials), ApiKey]` with wrapper keys
+        // `bearerAuth` (oauth) and `apiKeyAuth` (api key).
+        const AUTH_OPTIONS_SOURCE = `
+type Supplier<T> = T | Promise<T> | (() => T | Promise<T>);
+
+type UnionToIntersection<U> = (U extends any ? (x: U) => void : never) extends (x: infer I) => void ? I : never;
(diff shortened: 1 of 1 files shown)


Swimburger:
Are there any actual changes in the generated SDKs?

Swimburger (approved):

PR #17111, opened 2026-07-16
fix(csharp): scope nested-type namespace shadow global:: to the enclosing type

willkendall01:
Stacked on top of #17103.

## Problem
#17103 fixes the CS0426 build break caused by a generated nested type whose name equals the SDK's root namespace segment (e.g. the `auth0` union variant → nested `Auth0` struct in `ConnectionResponseContent`, namespace `Auth0.ManagementApi`). It does so by tracking that nested type as a **global** type-name conflict, which forces a `global::` qualifier onto **every** reference beginning with that segment across the whole SDK — even though the shadow is only visible inside the type that declares it.

On the Auth0 Management SDK that emitted **13,070** `global::Auth0` references; only ~1,438 are actually needed.

## Change
A nested type only shadows the namespace root within its enclosing type's lexical scope, so this qualifies references only where the shadow is actually visible:

- **`name-registry.ts`** — record the *enclosing types* a nested shadow is visible in (`nestedTypeShadows`) instead of tracking the name globally; add `hasNestedTypeShadowInScope(name, scopeStack)`.
- **`Writer.ts`** — track the current enclosing-type scope (`typeScopeStack` + `pushTypeScope`/`popTypeScope`).
- **`Class.ts`** — push/pop the enclosing type's FQN around the class body.
- **`ClassReference.ts`** — `shouldGlobal` now also fires when the reference's root segment is shadowed **and** we're writing inside that scope.
- Tests + changelog updated to the scoped API.

Still on-by-default, no flag. It remains a strict no-op for any SDK without a nested type shadowing its root namespace segment.

## Verification
- `global::Auth0` references on the Auth0 SDK: **13,070 → 1,438** (−89%), confined to exactly the two types that declare a nested `Auth0`.
- `dotnet build` of the regenerated SDK: **`Build succeeded`, 0 errors** across netstandard2.0 / net462 / net8.0 / net9.0 incl. the mock-server test project — no CS0426/CS0118/CS0234, with `experimental-fully-qualified-namespaces` off.
- Generator unit tests pass: codegen 630, model 9, sdk 67.

> Not yet run: the seed snapshot suite — if a committed fixture contains this exact collision its snapshot would legitimately shrink (fewer `global::`).

### DIFF
--- a/generators/csharp/codegen/src/ast/core/Writer.ts
+++ b/generators/csharp/codegen/src/ast/core/Writer.ts
@@ -41,6 +41,11 @@ export class Writer extends AbstractWriter {
     /* Whether or not to skip adding global:: qualifier to System namespaces */
     public readonly skipGlobalQualifier: boolean;
 
+    /* Fully qualified names of the enclosing types currently being written, outermost first.
+       Used to detect nested-type namespace shadows that are only visible within a type's body
+       (see NameRegistry.hasNestedTypeShadowInScope). */
+    public readonly typeScopeStack: string[] = [];
+
     constructor({
         namespace,
         allNamespaceSegments,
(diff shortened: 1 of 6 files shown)


willkendall01:
Folded directly into #17103 (commit 39d2c8aea71) instead of stacking a separate PR.

PR #17146, opened 2026-07-21
feat(cli-generator): generate wire tests on CLI generation

rishabh-fern:
## What

Adds an opt-in `generateWireTests` config flag to the CLI generator. When enabled, generation emits a mock-driven integration suite alongside the CLI:

- **`wiremock/wire-test-cases.json`** — one declarative case per endpoint example, built from the shared `@fern-api/mock-utils` engine the SDK wire tests use. Each case is *naming-independent*: it carries `{method, path, params, body, response}`, not a CLI command chain.
- **`tests/wire_test.rs`** — a generic harness. Per case it starts an in-process `wiremock::MockServer`, resolves the CLI command chain by loading the same baked OpenAPI spec the binary runs on (via `fern_cli_sdk::openapi::load_openapi_spec`, so it never reproduces the CLI's command-naming rules), drives the compiled binary (`--base-url` + `--params` + `--json`), and asserts the request hit the mock exactly once with the right method+path and that stdout rendered the mocked response body.

No docker, no `RUN_WIRE_TESTS` gate: `wiremock` is already a `[dev-dependencies]` entry in `sdk/Cargo.toml`, so the emitted `ci.yml` `cargo test` job runs the suite automatically on the generated repo's PR.

Defaults to **off** so existing generations are unaffected.

Also fixes a non-hermetic SDK unit test (`auth/login.rs` `paint_returns_plain_string_when_stderr_is_not_a_tty`) that assumed `cargo test` has no TTY on stderr — false in an interactive shell. It now forces the `NO_COLOR` path explicitly and is TTY-independent.

## Testing

- `tsc` + 217 TS unit tests pass (new coverage for the manifest builder, harness renderer, config validation, and skip path).
- New seed fixture `query-parameters-openapi:with-wire-tests`; `cargo test` on the generated output passes (`wire_search ... ok`), verified under plain `cargo test` (CI parity).
- No IR changes. No breaking changes — opt-in, default off; no `sdk/Cargo.toml` dependency changes (so `cli-seed` image / `Cargo.lock` need no rebuild).

## Notes

- Custom commands (`custom.rs`) are **not** covered — they aren't in the IR/spec, so no case is derived. The underlying request/auth/render plumbing they build on *is* exercised by the spec-command wire tests.
- Multi-spec namespaced command disambiguation in the harness is best-effort (documented inline).

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17146" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


generators/cli/CLAUDE.md
```diff
@@ -169,6 +170,55 @@ docker run --rm \
...
+signal, not a harness bug.
+
+No docker, no `RUN_WIRE_TESTS` gate — `wiremock` is already a `[dev-dependencies]`
+entry in `sdk/Cargo.toml`, so `seed`'s `cargo test --all-features` compiles and
+runs the suite automatically. The `query-parameters-openapi:with-wire-tests`
+seed variant exercises this end to end.
+
+```bash
+# Regenerate the wire-test seed fixture and run the emitted suite locally.
+pnpm seed test --generator cli --fixture query-parameters-openapi --skip-scripts
```
iamnamananand996, line 218: okay

iamnamananand996 (approved):

generators/cli/src/customConfig.ts
```diff
@@ -56,6 +56,19 @@ export interface FernCliCustomConfig {
...
+     * the CLI: `wiremock/wire-test-cases.json` (one case per endpoint
+     * example) and `tests/wire_test.rs` (a generic harness that stands up an
+     * in-process mock server, drives the compiled binary, and asserts the
+     * request/response). `seed`'s `cargo test --all-features` runs these
+     * automatically — no docker, no network.
+     *
+     * Defaults to `false` so existing generations are unaffected until a
+     * consumer opts in.
+     */
+    generateWireTests?: boolean;
```
iamnamananand996, line 71: looks good

PR #17290, opened 2026-07-29
feat(typescript): add opt-in appInfo client option appended to User-Agent

cadesark:
## Description

Adds an opt-in TypeScript SDK generator config, `allowUserAgentAppInfo` (default `false`). When enabled, generated clients accept an optional `appInfo` client option whose product token is appended to the `User-Agent` header the SDK would otherwise send, following [RFC 9110 §10.1.5](https://www.rfc-editor.org/rfc/rfc9110#name-user-agent).

```ts
const client = new AcmeClient({
    token: "...",
    appInfo: { name: "partner-app", version: "3.1.0", comment: "+https://partner.example" },
});
```

**Additive and default-off**, so existing generated output is byte-for-byte identical until a user opts in. No IR change. TypeScript only.

## Gated design (reworked)

The shared `core/runtime` utility (`runtime.ts`) — which is copied verbatim into every SDK — is **not modified**. This keeps all existing (flag-off) fixtures byte-identical to `main` and mirrors how `includePlatformHeaders` already works (core-util files are always shipped identically; flags only change the generated client code).

All `appInfo` logic is emitted into the generated client, only when `allowUserAgentAppInfo` is enabled:

- **`BaseClientContextImpl.ts`** — flag-gated: adds the optional `appInfo?: { name: string; version?: string; comment?: string }` property to `BaseClientOptions`.
- **`BaseClientTypeGenerator.ts`** — when the flag is on, emits a small self-contained `appendAppInfoToUserAgent` helper (a module-level function) into the generated `BaseClient.ts`, and wraps the computed User-Agent expression with `appendAppInfoToUserAgent(<ua>, options?.appInfo)` for **all three** branches: `includePlatformHeaders` → `core.getUserAgent(...)`; the `user-agent` template config (static string); and the default `{package}/{version}` (static string). The helper is emitted only when the flag is on and a User-Agent header is actually produced (so `omitFernHeaders` still suppresses everything). When the flag is **off**, exactly today's code is emitted — zero `appInfo` references.
- The emitted helper token-encodes `name`/`version` (percent-encodes every non-RFC-7230 `tchar`, incl. spaces / control chars / CR / LF) and escapes the `comment` delimiters (`(`, `)`, `\`) and control chars, so caller-supplied values cannot inject header content. It formats `{name}/{version} ({comment})`, dropping `/version` and ` (comment)` when blank, and returns the User-Agent unchanged when `appInfo`/`name` is absent.
[...]

PR #17300, opened 2026-07-30
feat(ruby): add opt-in appInfo client option appended to User-Agent

cadesark:
## Summary

Ports the opt-in `appInfo` User-Agent feature (already shipped for TypeScript in #17290 and PHP in #17299) to the **Ruby SDK generator (ruby-v2)**.

Adds a new SDK generator config `allowUserAgentAppInfo` (boolean, default `false`). When enabled, the generated client accepts an optional `app_info` client keyword (`{ name:, version?:, comment?: }`) whose product token is appended to the `User-Agent` header for **all three** branches: the structured platform value (`includePlatformHeaders`), the configured `user-agent` template value, and the default `{package}/{version}` — producing an RFC 9110 §10.1.5 User-Agent.

## Gated design (default-off byte-identical)

- The sanitize/append logic lives in a self-contained `RawClient.append_app_info` helper emitted into `raw_client.rb`, **gated on a new `allowUserAgentAppInfo` Eta template variable** in `raw_client.Template.rb` (threaded through `RubyProject.ts`). With the flag off, `raw_client.rb` is a **literal empty diff**.
- The `app_info` client keyword and the User-Agent wrapping are gated in `RootClientGenerator.ts` (`emitAppInfoOption()` / `wrapUserAgentWithAppInfo()`), so `client.rb` is byte-identical when the flag is off.
- Suppressed entirely by `omitFernHeaders` (no User-Agent is sent in that mode); still overridable by an explicit `User-Agent`. **No IR change.**

## Before / After User-Agent (from the new `allow-user-agent-app-info` fixture)

Default `{package}/{version}` branch, base value `fern_examples/0.0.1`:

```
# before (no app_info passed / flag off)
User-Agent: fern_examples/0.0.1

# after (client.new(..., app_info: { name: "partner-app", version: "3.1.0", comment: "+https://partner.example" }))
User-Agent: fern_examples/0.0.1 partner-app/3.1.0 (+https://partner.example)
```

Generated wiring:
```ruby
"User-Agent" => Seed::Internal::Http::RawClient.append_app_info("fern_examples/0.0.1", app_info),
```

## Sanitization (trim-before-encode; avoids the whitespace-junk-token bug)

Each value is **trimmed first, blank treated as absent (segment dropped), then encoded**:
- `name`/`version`: every non-RFC-7230 `tchar` percent-encoded (spaces, control chars, CR/LF).
- `comment`: `(`, `)`, `\` and control chars (incl. CR/LF) escaped so it cannot terminate the comment group or inject header content.

## Empty-diff confirmation

Regenerated the `examples` fixtures with `--local`. `git status` shows changes **only** under the new `allow-user-agent-app-info` fixture; every pre-existing fixture (incl. `include-platform-headers`) is byte-identical (metadata churn reverted).

## Rubocop-clean confirmation

`--skip-scripts` does not run the generated SDK's rubocop, so I ran it the way CI does — Docker `ruby:3.3-slim` + `gem install rubocop -v '~> 1.21' rubocop-minitest` + the SDK's `.rubocop.yml`: **118 files inspected, no offenses detected** (initially caught 2 `Style/RedundantFreeze` on the frozen regex literals; fixed by dropping `.freeze`).

## Tests

New `generators/ruby-v2/base/src/__test__/appendAppInfoToUserAgent.test.ts` renders the emitted Ruby helper and **executes it** (local ruby 3.x, else Docker `ruby:3.3-slim`) — 18 tests, all passing:
- null / empty / whitespace-only name → User-Agent unchanged
- version/comment omission when absent or blank
- trims surrounding whitespace rather than encoding it (regression guard for the TS bug)
- CRLF injection via name / version / comment (asserts `%0D%0A`, no raw CR/LF)
- paren + backslash injection in comment (`a)b(c\d` → `a%29b%28c%5Cd`)
- space token-encoding in name; printable URL comment kept human-readable

Also ran the full ruby-v2 TS unit suites (ast/base/sdk) — all green.

## Changelog

`generators/ruby-v2/sdk/changes/unreleased/allow-user-agent-app-info.yml` (`type: feat`). `versions.yml` untouched.

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17300" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


iamnamananand996 (approved):

generators/ruby-v2/ast/src/custom-config/BaseRubyCustomConfigSchema.ts
```diff
@@ -36,6 +36,10 @@ export const BaseRubyCustomConfigSchema = z.object({
     // Opt-in: emit the X-Fern-Runtime, X-Fern-Runtime-Version, and X-Fern-Platform
     // observability headers. Disabled by default so existing output is unchanged.
     includePlatformHeaders: z.boolean().optional(),
+    // Opt-in: expose an optional `app_info` client keyword whose product token is
+    // appended to the User-Agent header (RFC 9110). Disabled by default so existing
+    // output is byte-identical.
+    allowUserAgentAppInfo: z.boolean().optional(),
```
iamnamananand996, line 42: fine

PR #17301, opened 2026-07-30
feat(java): add opt-in appInfo client option appended to User-Agent

cadesark:
## What

Ports the opt-in `appInfo` User-Agent feature (TS PR #17290) to the **Java SDK generator**. Adds a new SDK generator config `allowUserAgentAppInfo` (default `false`). When enabled, the generated client's builder exposes an `appInfo(String name, String version, String comment)` option whose product token is appended to whatever `User-Agent` the SDK would otherwise send, per RFC 9110 §10.1.5.

## Gated design (where it's emitted)

The User-Agent is baked into the `ClientOptions` headers map when `ClientOptions.Builder.build()` constructs the client. When (and only when) `allowUserAgentAppInfo` is on **and** a `User-Agent` header is actually written:

- The builder gets a private `appInfo` field + an `appInfo(name, version, comment)` method that stores the **sanitized product token** (computed by the emitted static `appInfoProductToken`).
- The token is threaded as a new `ClientOptions` constructor parameter (positioned right after `retryJitterFactor`) and passed by `build()`.
- Every User-Agent value expression in the constructor is wrapped with `appendAppInfo(<base>, appInfo)`. This covers **all branches**: the structured `getUserAgent()` value (`includePlatformHeaders`), the `runtime-version` `{coordinate}/ + getSdkVersion()` jar-manifest path, and the baked-in literal (which also covers the configured `user-agent` template value).
- Self-contained sanitizing helpers (`appendAppInfo`, `appInfoProductToken`, `encodeToken`, `encodeComment`, `appendPercentEncoded`) are emitted into the generated `ClientOptions` — the shared always-shipped core-utilities are never touched.

When the flag is off, none of the above is emitted, so existing output is byte-identical. Still overridable by an explicit `User-Agent` and suppressed by `omit-fern-headers`. No IR change.

Composes with `runtime-version`: the manifest-rebuilt value (`prefix + getSdkVersion()`) is the `<base>` passed to `appendAppInfo`, so the app token survives that path too.

## Sanitization

Each of name/version/comment is **trimmed first, treated as absent when blank, then encoded** (avoids the whitespace-junk-token bug). `name`/`version` percent-encode every non-RFC-7230 `tchar` (UTF-8 bytes); `comment` escapes `(`, `)`, `\` and control chars (0x00-0x1F, 0x7F incl. CR/LF). Blank name → User-Agent unchanged; blank version/comment → that segment dropped.

## Before / after User-Agent (new `imdb/allow-user-agent-app-info` fixture)

- Base (`includePlatformHeaders`): `com.fern.imdb/0.0.1 (linux; x86_64) Java/<ver>`
- With `.appInfo("partner-app", "3.1.0", "+https://partner.example")`: `com.fern.imdb/0.0.1 (linux; x86_64) Java/<ver> partner-app/3.1.0 (+https://partner.example)`

## Verification

- **Empty-diff gate:** regenerated the flag-off sibling `imdb/include-platform-headers` — all SDK source (incl. `ClientOptions.java`) byte-identical; the only deltas were snippet-file formatting caused by the on-host `spotlessApply` post-step being unavailable (no local JDK), not by the code change. That regeneration was discarded; only the new flag-on fixture is committed.
- **New-fixture full build:** `./gradlew compileJava` (incl. `sample-app`) + `spotlessApply` + `spotlessJavaCheck` all pass on the new fixture (via Docker `eclipse-temurin:17-jdk`).
- **Generator unit tests:** `:generator-utils:test :sdk:test` pass. New `AppInfoUserAgentTest` compiles the **exact emitted helper MethodSpecs** with the in-process `JavaCompiler` and exercises them via reflection (mirrors `WebhookBodyHashTest`), covering: null/empty/whitespace-only name → unchanged; name-only / name+version / name+version+comment; blank version/comment dropped; whitespace trimmed (not encoded); space encoding; **CRLF injection via name/version/comment** (→ `%0D%0A`); **paren/backslash escaping** in comment; printable comment (URL) kept readable.
- Java generator compiles (`:sdk:compileJava`); spotless applied to generator sources.

## Files

- Config: `JavaSdkCustomConfig.java`, `JavaSdkDownloadFilesCustomConfig.java`, `Cli.java`
- Codegen: `ClientOptionsGenerator.java`
- Test: `AppInfoUserAgentTest.java`
- Seed: `seed/java-sdk/seed.yml` + `seed/java-sdk/imdb/allow-user-agent-app-info/`
- Changelog: `generators/java/sdk/changes/unreleased/allow-user-agent-app-info.yml` (`type: feat`)

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17301" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


generators/java/sdk/src/main/java/com/fern/java/client/JavaSdkCustomConfig.java
```diff
@@ -145,6 +145,20 @@ default Boolean runtimeVersion() {
...
+     * If true, the generated client exposes an optional {@code appInfo(name, version, comment)} builder option whose
+     * product token is appended to the {@code User-Agent} header the SDK would otherwise send (following RFC 9110
+     * §10.1.5), producing e.g. {@code {sdk}/{version} ... partner-app/3.1.0 (+https://partner.example)}.
+     * Caller-supplied values are sanitized (name/version percent-encoded to RFC 7230 {@code tchar}s; comment delimiters
+     * and control characters escaped). Opt-in and disabled by default so existing generated output is unchanged. The
+     * header is still overridable by an explicit {@code User-Agent} and suppressed by {@link #omitFernHeaders()}.
+     */
+    @Value.Default
+    @JsonProperty("allowUserAgentAppInfo")
+    default Boolean allowUserAgentAppInfo() {
```
iamnamananand996, line 158: fine

iamnamananand996 (approved):

PR #17308, opened 2026-07-31
fix(python): honor retries in HttpClient.stream and AsyncHttpClient.stream

ckarnell:
`stream()` in the Python generator's shared `http_client` accepts `retries` and never reads it. The `request()` method next to it reads both `retries` and `max_retries` and recurses on a retryable status or a connection error. So `max_retries` in `request_options` is silently ignored on every streaming call, in both the sync and async clients.

Reproduction is in the new test. A mock transport returning 503, 503, 200 sees one request through `stream()` and three through `request()`.

The fix mirrors `request()`, with one deliberate limit. It retries only before the response is handed to the caller. Once the body has been yielded the caller may already have read part of it, and replaying the request would produce a truncated or duplicated stream. So this covers connection errors and a retryable status on the initial response, not a failure part way through a stream.

I used `ExitStack` instead of driving the httpx context manager by hand. My first attempt called `__enter__()` and then re-entered the same object with `with`, which raises, because contextlib managers are single use. The tests caught that.

The 5 new tests fail without the change and pass with it, and the two negative controls (`max_retries=0`, and the ordinary success path) pass either way, so they are not just detecting the patch. `tests/utils/test_http_client.py` plus `test_http_sse.py` give 157 passed. `ruff check`, `ruff format --check` and `mypy` are clean.

I did not check the generated SDK snapshots, or whether any downstream SDK relies on streaming never retrying. Worth separating too, this is retry, not the resumable stream reconnect that #15886 added to the Go generator.

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17308" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/generators/python/core_utilities/shared/http_client.py
+++ b/generators/python/core_utilities/shared/http_client.py
@@ -4,7 +4,7 @@ import re
 import socket
 import time
 import typing
-from contextlib import asynccontextmanager, contextmanager
+from contextlib import AsyncExitStack, ExitStack, asynccontextmanager, contextmanager
 from random import random
 
 import httpx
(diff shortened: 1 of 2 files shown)


ckarnell:
Two red checks here look unrelated to this change, so here is what I checked.

`test-ete` fails on a CLI snapshot in `packages/cli/ete-tests/src/tests/validate/validate.test.ts`, expecting "Detected empty API definition" and getting "Found 0 errors and 1 warning". This PR only touches `generators/python/core_utilities/shared/http_client.py` and adds one Python test file. The last `test-ete` runs I can see that passed were on 2026-07-30 around 20:40Z, and four CLI commits landed after that (#17307, the OAuth login flows PR, and two release commits), so this run is the first against that state.

`benchmark-report` fails with a 403 posting its PR comment, which is the read-only token on a fork.

All 15 `python-sdk` seed jobs pass, along with `benchmark (python-sdk)`.

Happy to rebase if that helps, but I did not want to touch a CLI snapshot from a Python generator PR.

ckarnell:
Thanks, the non-idempotent body point was right and I have pushed a fix.

`content` can be an iterator and `files` can hold open handles. A retry re-sends the same objects, so replaying one would put an empty or truncated body on the wire instead of failing loudly. `stream()` had no retry path before this PR, so that exposure is mine, not inherited. A request whose body cannot be replayed is now returned as-is, with two tests: an iterator body is not replayed, and a bytes body still is, so the guard cannot pass by disabling retries generally.

Worth flagging separately: `request()` has the same replay behaviour today, since it recurses with the same `content` and `files`. I have not touched it here because changing existing retry behaviour is a bigger call than adding a guard to a path I just created.

On the seed snapshots, I do not think that one holds. There is no `seed/python-sdk/` directory in the tree, and all 36 files under `seed/` matching `http_client` are `.rs`. Happy to regenerate if I am reading the layout wrong.

On asymmetric coverage, fair. Sync has five tests and async has one. The async path is the same code with awaits, and I would rather add cases if you want them than guess at which ones earn their keep.

164 tests pass across `test_http_client.py`, `test_http_sse.py` and the new file. `ruff check`, `ruff format --check` and `mypy` are clean.

PR #17444, opened 2026-08-17
feat(cli-generator): authenticate Homebrew/Scoop publishing with a GitHub App

rishabh-fern:
## Summary

Adds `distribution.githubApp`, letting a generated CLI's Homebrew and Scoop publish jobs mint a **GitHub App installation token** at release time instead of reading a long-lived PAT out of a secret. The token is scoped to one repository and expires in an hour, and revoking it means uninstalling an App rather than rotating a credential attached to a person who may have left the org.

```yaml
config:
  binaryName: acme-cli
  distribution:
    githubApp:
      appIdSecret: PUBLISH_APP_ID
      privateKeySecret: PUBLISH_APP_PRIVATE_KEY
    homebrew:
      tap: acme/homebrew-tap
    scoop:
      bucket: acme/scoop-bucket
```

### Why one shared block, not per channel

One App normally serves both channels. An App can be installed on several accounts, and each job passes its own tap/bucket `owner`, so a tap and a bucket under *different* orgs still need only one App. Duplicating the same two secret names under `homebrew` and `scoop` would only invite the drift where one is renamed and the other is not.

Per-channel `githubApp` is deliberately **not** shipped: adding it later is additive, removing it later would need a generator migration.

### Resolution order

Most-specific-first, per channel:

1. the channel's own `tokenEnvironmentVariable`
2. `distribution.githubApp`
3. the channel's default PAT secret name

Rung 1 above rung 2 makes migrating one channel at a time expressible (shared App, `scoop.tokenEnvironmentVariable` still pinned). It also means two auth mechanisms for one push cannot be expressed, so there is no mutual-exclusivity error to raise.

### Preflight job

`preflight-distribution` is emitted only when a channel uses an App. It declares no `needs`, so a bad secret fails in ~20s instead of after the build matrix and after `gh release create` has already published the release — where the failure is `appId is required`, naming neither the secret nor the fix.

It catches the two failure modes that don't describe themselves:

- An Actions **variable** is not readable as `secrets.<NAME>`, and an undefined secret resolves to the empty string rather than erroring.
- A PEM collapsed onto one line still carries its BEGIN header, so the only signal is the missing line breaks (`secretOrPrivateKey must be an asymmetric key`).

It checks the App identifier for emptiness only — `create-github-app-token` accepts the numeric App ID *or* the Client ID (`Iv23…`), so a shape assertion would reject valid config. Neither secret value is ever echoed. Only the App channels depend on it, so a bad secret costs that channel's publish but still ships archives, installers and `curl | bash`.

### Two load-bearing details

- **Job-level `env` cannot read `steps.*`.** So in the App branch `GITHUB_TOKEN` moves from the job `env` down to the commit step; left up there it renders as the empty string. Worth knowing: nothing in that job actually reads it — the `git push` authenticates off the `http.extraheader` that `actions/checkout` writes because of `persist-credentials: true`. It's inherited cargo-dist decoration, kept so the PAT branch stays byte-identical.
- **Both variants come from one template**, not two copies of the job. Authoring a second copy is how the `announcement_is_prerelease` guard gets dropped from one branch — and a tap has no prerelease channel, so a lost guard means an RC formula becomes what every `brew upgrade` installs, with a green workflow and no error anywhere. There is a test asserting the guard's presence in the App branch specifically.

## Backwards compatibility

Additive and opt-in. **Every pre-existing fixture's `release.yml` regenerates with zero diff** — that's a file comparison, not an argument, and it holds because the App branch interpolates into the same template rather than forking it. No generator migration needed.

Deliberate no-ops, to save a reviewer the walk:

- `patchDistWorkspace.ts` — cargo-dist never reads the push token; we render `release.yml` ourselves.
- `emitReadme.ts` — the install stanzas reference taps and buckets, never secrets.
- `emitPublishWorkflow.ts` (`ci.yml`), `resolveOutputConfig.ts`, and the IR — untouched.

`constructScoopJobYaml` now takes `ScoopJobYamlArgs`; nothing outside `generators/cli/` imports it.

## Testing

- 413 unit tests pass; biome clean. Two named tests cover the failures nothing else would catch: the App branch retains the prerelease guard, and Scoop's token scopes to `scoop-bucket` — asserted with the *same* App on both channels, which is the realistic config and the one where a tap/bucket mixup is invisible on inspection.
- New seed fixture `github-distribution-app`, so the committed `release.yml` shows the shape a consumer's pipeline actually takes.
- Verified the emitted document parses as YAML end to end: preflight has no `needs`, the Homebrew job env is `PLAN`/`GITHUB_USER`/`GITHUB_EMAIL` only, both publish jobs wait on preflight.

Two defects found and fixed during self-review:

1. A **PAT channel was gated on the other channel's App preflight**. With a shared App and `homebrew.tokenEnvironmentVariable` still pinned, a malformed App secret would skip the Homebrew publish — which never touches the App — collapsing the "a bad secret costs one channel" property the job exists to provide. The `needs` entry is now gated on each channel's own resolved auth.
2. `set -u` with a possibly-unset env var in the very job whose purpose is legible diagnostics; values are now normalized through `${VAR:-}` first.

## Reviewer note

`github-distribution-app` is generated against the current generator, whereas its sibling folders predate `rust 0.45.0`'s `custom-reqwest-client`. So `diff -rq github-distribution github-distribution-app` shows `release.yml` (this change) **plus three unrelated drift files**. I kept that regeneration out of this branch — it belongs in its own `chore(seed)` PR.

## Not in this PR

- Docs (`fern-api/docs`: `products/cli-generator/configuration.mdx`, `publishing.mdx`) — follow-up PR.
- End-to-end release validation. Every assertion here is on emitted YAML; nothing proves an installation token actually pushes cross-repo. That needs a real App installed on real tap/bucket repos, then `v0.0.1-rc.1` (expect both channels **skipped**) followed by `v0.0.1` (expect formula + manifest committed). The feature is inert until a consumer sets `githubApp`, so this gates recommending it, not merging it.

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17444" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


rishabh-fern:
## End-to-end validation: complete ✅

The item this PR listed as outstanding — *"nothing proves an installation token actually pushes cross-repo"* — is now closed. Validated on real GitHub Actions against real tap and bucket repos, using a locally built image of **exactly the code in this PR** (working tree clean against `691280b`, verified the bundled `cli.cjs` contains the final-round markers).

**Setup:** a GitHub App with only `Contents: read & write`, installed on the tap and bucket **but not** on the CLI repo — so the App's sole capability is the cross-repo push, and everything in-repo still runs on the built-in `GITHUB_TOKEN`.

| Repo | Role |
|---|---|
| `rishabh-fern/elevenlabs-cli-dist-test` | generated CLI + `release.yml` |
| `rishabh-fern/homebrew-tap` | tap |
| `rishabh-fern/scoop-bucket` | bucket |

### Round 1 — `v0.0.1-rc.1` (prerelease guard)

All 14 jobs green. Release created and marked prerelease with 21 assets.

| Job | Result |
|---|---|
| `preflight-distribution` | ✅ success — **completed alongside `plan`, not after the build matrix** |
| `publish-homebrew-formula` | ⏭ **skipped** |
| `publish-scoop` | ⏭ **skipped** |
| tap / bucket commits | **1 each — untouched** |

A fully successful release that deliberately published nothing to either channel. This is the failure mode worth caring about: a tap has no channels, so brew serves whatever sits at `HEAD` — a lost guard means an RC formula becomes what every `brew upgrade` installs, with a green workflow and no error anywhere.

### Round 2 — `v0.0.1` (first real publish)

All 14 jobs green, **including both publish jobs**. First execution of `create-github-app-token` and the cross-repo push.

```
homebrew-tap   Formula/elevenlabs.rb    commit "elevenlabs-cli 0.0.1"  by github-actions[bot]
scoop-bucket   bucket/elevenlabs.json   commit "elevenlabs 0.0.1"      by github-actions[bot]
```

Each job minted a token scoped to **its own** repo (`repositories: homebrew-tap` vs `repositories: scoop-bucket`) — the tap/bucket mixup that the single-derivation-site helper exists to prevent.

Formula correctness, which is where a wrong `Cargo.toml` `repository` silently 404s every install:

```ruby
desc "CLI for the ElevenLabs API Documentation"
homepage "https://github.com/rishabh-fern/elevenlabs-cli-dist-test"
url "https://github.com/rishabh-fern/elevenlabs-cli-dist-test/releases/download/v0.0.1/elevenlabs-cli-aarch64-apple-darwin.tar.gz"
sha256 "4617f22546cc39b11260bdbd3da1d2965729f50b6dd91d4941e179fff87ccae2"
```

Verified independently: downloaded that asset, `shasum -a 256` matched the formula byte-for-byte, extracted, ran `--version` → `elevenlabs 0.0.1`. Both the darwin tarball and the Windows zip return `200`.

### Install

```
$ brew install rishabh-fern/tap/elevenlabs
🍺  /opt/homebrew/Cellar/elevenlabs/0.0.1: 6 files, 14.2MB, built in 1 second
$ /opt/homebrew/bin/elevenlabs --version
elevenlabs 0.0.1
```

`brew info` parses the generated Ruby cleanly. The `homebrew-` prefix strip works — `brew install rishabh-fern/tap/...` resolves to the `homebrew-tap` repo.

### Round 3 — `v0.0.2` (upgrade path)

The first release only proved the App can write to an *empty* tap; steady state is appending to a populated one. All 14 jobs green, and both repos show a correct sequence rather than a clobber:

```
homebrew-tap                          scoop-bucket
  elevenlabs-cli 0.0.2  ← HEAD          elevenlabs 0.0.2  ← HEAD
  elevenlabs-cli 0.0.1                  elevenlabs 0.0.1
  Initial commit                        Initial commit
```

Formula and manifest both rewritten in place to `0.0.2` with URLs repointed. `brew update && brew upgrade elevenlabs` → `0.0.2`.

### Summary

| Behavior | Status |
|---|---|
| Preflight validates App credentials, reports in ~20s with no `needs` | ✅ |
| Prerelease guard skips both channels while still publishing the release | ✅ |
| App token minted per channel, scoped to its own repo | ✅ |
| Cross-repo push to tap and bucket | ✅ |
| Formula URLs point at the consumer's repo; checksums match; archives reachable | ✅ |
| `brew install` → working binary | ✅ |
| Second release appends to populated downstream repos | ✅ |
| `brew upgrade` picks up the new version | ✅ |

Not exercised live: re-running a completed tag for idempotency (covered by unit assertions), and the negative preflight cases (secret-stored-as-variable, single-line PEM) — both covered by unit tests, and neither can write to a downstream repo by construction.

### Note for reviewers on the seed diff

`seed/cli/query-parameters-openapi/github-distribution-app/` is generated against the current generator, while its sibling folders predate `rust 0.45.0`'s `custom-reqwest-client`. A `diff -rq github-distribution github-distribution-app` therefore shows `release.yml` (this change) **plus three unrelated files**, and regenerating locally will show ~17 modified files in the older folders. That drift is pre-existing on `main` and deliberately excluded here; refreshing those fixtures belongs in its own `chore(seed)` PR.

iamnamananand996 (approved):

PR #17454, opened 2026-08-17
feat(swift): invocation-only dynamic snippets + clientImport

cadesark:
## What

Populates the optional `clientImport` field on `InvocationSnippetResponse` for the Swift dynamic-snippets invocation generator, mirroring the TypeScript port.

## Details

Swift imports whole modules (`import <Module>`) rather than importing the client by symbol (as TS/PHP/C#/Java do). The client type and every SDK type it references resolve through a single module import. So `clientImport` is the `import <Module>` statement the scaffold already emits via `generateImportModuleStatement()`, rendered without a trailing newline for embedding in a docs template.

This is the import a docs template needs to construct the root client. It is distinct from the always-empty per-symbol `imports` field (a bare Swift invocation references no per-symbol import, since the Swift AST has no per-node import mechanism).

## Testing

- `pnpm turbo run compile --filter @fern-api/swift-dynamic-snippets` — passes (10/10 tasks)
- Package vitest — 39/39 pass, including a new assertion that `clientImport === "import Acme"` (and does not end in a newline)

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17454" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/generators/swift/dynamic-snippets/src/EndpointSnippetGenerator.ts
+++ b/generators/swift/dynamic-snippets/src/EndpointSnippetGenerator.ts
@@ -1,4 +1,10 @@
-import { AbstractAstNode, Options, Scope, Severity } from "@fern-api/browser-compatible-base-generator";
+import {
+    AbstractAstNode,
+    InvocationSnippetResponse,
+    Options,
+    Scope,
+    Severity
+} from "@fern-api/browser-compatible-base-generator";
 import { assertNever } from "@fern-api/core-utils";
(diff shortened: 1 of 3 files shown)


cadesark:
Closing: Swift does not have docs snippet support, so the invocation/clientImport snippet generation added here would never render (no dynamic IR is produced for Swift). Can revive if Swift snippet support is added later.

PR #17494, opened 2026-08-20
fix(cli): stabilize global theme asset paths

ctondreau:
## Description

Global-theme stitching downloaded theme assets into a random per-process temp dir (`tmp.dir({ prefix: "fern-theme-" })`), so every publish of the same theme produced a *different* asset path:

```
/tmp/fern-theme--2123-kY46I5owUYrT/theme_asset_1.png   # publish A
/tmp/fern-theme--2100-Nxv5JiQhE3RZ/theme_asset_1.png   # publish B, same theme
```

Publish relativizes that absolute path, so the config ends up naming `_dot_dot_/…/tmp/fern-theme--2123-kY46I5owUYrT/theme_asset_1.png` — a brand-new file key and S3 object on every publish, and a logo path that is not reproducible across publishes. Verified against a customer's ledger deployments: four publishes in one afternoon, four distinct `fern-theme--<pid>-<rand>` directories.

Consequences: asset churn in the bucket on every publish, and any consumer holding a config from one publish alongside a file map from another (e.g. the dashboard visual editor's cached loader state) names a key the file map doesn't have — the logo `src` then falls back to the raw relative path, 404s, and renders as clipped alt text.

The directory is now derived from the theme's identity instead of the process:

```ts
getGlobalThemeAssetDirectoryPath(organization, themeName)
// => `${os.tmpdir()}/fern-theme-${sha256(`${organization}\0${themeName}`).slice(0, 16)}`
```

Asset filenames are unchanged, so the full path is byte-identical across publishes of the same theme. Since the directory is now reused rather than process-private, assets and `stitched-docs.yml.json` are written to a unique temp file and atomically renamed into place, so a concurrent publish can never read a torn file. The exit-time cleanup (and the log line advertising it) is gone — deleting a shared, reused directory would defeat the stability and could race another publish.

A predictable path in a world-writable tmpdir also gives up the 0700-random-name protection `tmp.dir()` provided, so `ensureGlobalThemeAssetDirectory` creates the directory at mode `0o700` and, before anything is written, `lstat`s it and refuses to proceed unless it is a real directory owned by the current uid (tightening the mode back to `0700` when we own a loosened directory). A pre-created or symlinked path fails the publish loudly via `failAndThrow` instead of silently redirecting our writes.

## Changes Made
- `stitchGlobalTheme.ts`: deterministic, reused theme asset directory; atomic writes; no exit cleanup
- `ensureGlobalThemeAssetDirectory`: 0700 creation + ownership/symlink guard
- Unit tests for the directory-name helper (determinism, org/theme separation, unsafe inputs) and the symlink rejection
- CLI changelog entry (`type: fix`)

## Testing
- [x] Unit tests added/updated — `pnpm turbo run test --filter @fern-api/docs-resolver` (112 tests pass); `compile`, `lint:biome`, `format:fix` clean
- [x] Manual testing completed — dev CLI built from this branch, DEV publishes with before/after control and a dashboard-editor check; see the [verification comment](https://github.com/fern-api/fern/pull/17494#issuecomment-5362362817)

Link to Devin session: https://app.devin.ai/sessions/21f651485fc34eac869d118d0898128e
Requested by: @ctondreau

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17494" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/cli/cli/changes/unreleased/fix-stable-global-theme-asset-path.yml
+++ b/packages/cli/cli/changes/unreleased/fix-stable-global-theme-asset-path.yml
@@ -1,0 +1,3 @@
+- summary: |
+    Make global theme asset paths stable across publishes of the same organization and theme.
+  type: fix
--- a/packages/cli/docs-resolver/src/__test__/stitchGlobalTheme.test.ts
+++ b/packages/cli/docs-resolver/src/__test__/stitchGlobalTheme.test.ts
@@ -2,10 +2,12 @@ import { AbsoluteFilePath } from "@fern-api/fs-utils";
 import { createMockTaskContext } from "@fern-api/task-context";
 
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import path from "path";
 
 import {
     deepMergeGlobalWins,
     filenameFromUrl,
+    getGlobalThemeAssetDirectoryPath,
     isPresignedUrl,
     isRemoteUrl,
     mergeThemeOverride,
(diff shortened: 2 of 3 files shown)


emjoseph:
Devin,
Do old files get cleaner or will /tmp continue grow without cleeanup?
What happens if /tmp/fern-theme-[hash] already exists on shared host? Will publishes fail ownership guard?

emjoseph (approved):

PR #17505, opened 2026-08-21
fix(cli): send runtime bundles to sdk-gen-api (FSDK-1116)

zamn:
## Why this PR exists

The first sdk-gen-api CLI integration sent Fern source specifications plus high-level generator invocation metadata, then expected sdk-gen-api to translate that metadata into `SdkConfigIrV1`. That is not sufficient for a wrapped Fern generator runtime: the Fern runtime listener executes a generator from its native `{ config, ir }` input, and the IR must already be migrated to the exact schema version expected by the requested Fern generator ID and version.

This PR moves ownership of that Fern-specific preparation back to the Fern CLI. sdk-gen-api remains responsible for authentication, build state, idempotency, orchestration, artifact status, and dispatch, but it treats each runtime bundle as opaque bytes.

## Behavior before and after

| Scenario | Before this PR | After this PR |
|---|---|---|
| `FERN_USE_SDK_GEN_API` unset or false | Existing Fiddle remote-generation path | Unchanged Fiddle path |
| Local/on-prem generation | Existing local runner | Unchanged |
| Unsupported/non-first-party generator | Existing Fiddle path | Unchanged Fiddle path |
| Supported first-party SDK generator with flag enabled | Protocol v1 request; sdk-gen-api attempted Fern-to-Postman IR translation | Protocol v2 request with one native Fern runtime bundle per target |
| Multi-generator group | One sdk-gen-api build with target metadata | One sdk-gen-api build with shared sources and one ordered bundle per target |
| Artifact download/local output | Existing Fern download and `.fernignore` handling | Same existing download and `.fernignore` handling |

There is no intended change to the SDK configuration or IR semantics supplied to the generator. The new bundle is built with Fern’s existing generator configuration and IR migration helpers, including the exact requested generator identity/version and any explicit IR-version override. When the feature flag is enabled, execution moves to the image selected by the deployed generator-version index; that runtime/environment change is the purpose of this route.

## Exact data flow

1. Fern resolves the generator invocation, SDK version, sources, audiences, output mode, and enriched IR as it does today.
2. Eligibility is checked before selecting sdk-gen-api:
   - recognized first-party SDK generator;
   - concrete SDK version (not `AUTO`);
   - source archive available;
   - generator language matches the mapped language;
   - no whitelabel configuration;
   - replay and full-project generation remain unsupported.
3. Fern resolves the generator IR schema version from FDR or the explicit override and migrates the enriched IR for the exact generator ID/version.
4. Fern builds the native `GeneratorConfig` with:
[...]

PR #17526, opened 2026-08-25
feat(cli): add generator compatibility validation

zamn:
## Summary

- add exact-version generator/configuration compatibility validation to the remote workspace runner
- keep the first-party alias and cutover policy private behind a narrow internal API
- preserve exhaustive cutover, alias, diagnostic, and SemVer coverage
- avoid a separately published dependency

## Architecture

The compatibility policy is colocated with the remote workspace runner and is not exported as a workspace or published package.

## Verification

- 106 focused compatibility tests
- 365 remote workspace runner tests
- dependency-aware TypeScript compile across 48 packages
- Biome
- CSpell
- frozen pnpm install
- `git diff --check`


zamn:
Folded into #17505.

PR #17676, opened 2026-09-08
feat(cli): run and configure Postman's on-prem SDK adapter on local generation

ATechAdventurer:
Everything local generation needs to run Postman's on-prem adapter in an air-gapped customer environment: how the image is *run*, and how it is *configured*.

## Running the image

Three levers, all routed through one `resolveGeneratorImage()` so every caller agrees on which image runs. It returns a digest reference untouched rather than appending `:version`, which would produce the invalid `repo@sha256:...:1.2.3`.

| Lever | Effect | Unset |
|---|---|---|
| `FERN_GENERATOR_NETWORK` | Container network mode. `none` = no network, which air-gapped requires. | Networking unchanged |
| `name: fern-python-sdk@sha256:...` | Runs that exact image. The digest rides the image reference; the generator **name** stays clean, because IR version resolution and the cutover check below both match it exactly. | Tag-based, as before |
| `USE_FERN_RC` | Runs the adapter's moving `rc` tag instead of the workspace's version. Only for invocations already resolving to the adapter; a digest pin still wins. `FERN_RC_NAMESPACE` overrides the namespace, default `fernenterprise`. | Workspace version runs |

Env vars rather than `generators.yml` keys: these are properties of *who is running*, not of the workspace, so the same file serves a customer on a release version and us on a pre-release. First-class config fields are the natural follow-up. `FERN_GENERATOR_NETWORK` is passed through rather than allowlisted — anyone who can set it can already run the container directly.

## Configuring the image

The adapter is published under the Fern generator names it replaces, at the version its language carries in the cutover matrix, so a customer keeps their `generators.yml` entry and changes only the version. The name is identical whichever image it is, so detection is a per-name version cutover rather than a name allowlist — and the name also identifies the language, so nothing has to be declared.

| Workspace | Generator | Result |
|---|---|---|
| No `sdk-config.yml` | At/above cutover | Error naming the generator, with `fern sdk migrate` |
| Has `sdk-config.yml` | At/above cutover | SDK Config IR built from the migrated config |
| Has `sdk-config.yml` | Below cutover | Fern's own `GeneratorConfig`, unchanged |

**Nothing is migrated at runtime.** An unmigrated workspace is refused, not translated from `generators.yml` — that file cannot express what these generators are configured with, and translating it would produce an SDK shaped by a configuration the customer never wrote. Generators below the cutover never reach any of this.

Two fields are imposed rather than carried, because the adapter rejects anything else and a local run can only mean one thing. **`sourceOrigin` is always `fern`** — it is the only input to `fernMode`, the flag selecting `PetsClient` over `PetsService`, so deriving it from the document would let a config silently reshape the SDK. **`output` is always files delivery at the container mount** — a `sdk-config.yml` written for publishing may say `zip` or `github`, both legal externally and both fatal to the container; the configured delivery is warned about and recorded in `compatibility.unsupportedFields`.

Source selection stays with the Fern workspace: `source.specs` is built from the workspace's pre-processed manifest, and only `apiImportSettings` is read off the document. A post-migration edit to `source.specs` is therefore **inert** on local generation. Separately, group audiences from `generators.yml` filter the manifest while `api.audiences` does not. Having `fern sdk migrate` own source selection end to end is the real fix and is not in this PR.

Failures the adapter would hit as `CONFIG_INVALID` are moved onto the host instead, naming the specs involved: spec types it cannot consume (`graphql`, `protobuf`, `openrpc`), and multi-spec workspaces, where it reads `source.specs[0]` and would otherwise produce an SDK covering one spec and exit zero.

## Open question: which namespace publishes the adapter

**This needs a decision before merge.** With `USE_FERN_RC` unset and no `image.registry`, an adapter-version invocation resolves to `fernapi/fern-python-sdk:6.0.0` — Fern's own namespace — while being handed SDK Config IR. So "change only the version" holds only if either Postman publishes adapter releases *into* `fernapi` at the cutover versions, or the customer also sets `image.registry`.

Two other behaviors collapse into the same question. Each cutover currently sits one major above Fern's release (`typescript 3.94 → 4.0`, `go 1.58 → 2.0`, `csharp 2.83 → 3.0`, …), so no existing pin can cross it — but if Fern ever ships its own `fern-go-sdk 2.0.0`, that customer gets SDK Config IR from a generator that cannot read it. And `latest`, which `fern sdk generate --local` defaults to, is treated as Fern's generator, which is safe only while the adapter is not published into `fernapi` under a moving tag. If the answer is "yes, `fernapi`", all three want an explicit registry check rather than a version comparison.

## Verification

265 tests in `local-workspace-runner`, 112 in `configuration-loader`'s `generators-yml` suite; `turbo run compile` clean across the dependency graph; Biome clean.

Ran the built CLI against real workspaces with a stub container capturing the document it was handed — digest-pinned image, `FERN_GENERATOR_NETWORK=none`, and a migrated `sdk-config.yml` in one run:

```
docker run --user root --network none ... --rm \
  localhost:5051/fern-typescript-sdk@sha256:0127e13e... /fern/config.json
```

The container received `sdk-config-ir/v1` with `sourceOrigin: fern`, the TypeScript target selected out of a two-target document, `packageManager` under `generation.language.typescript`, and output at the mount. That run is why these belong together: the digest-pinned reference still resolved the generator *name* to `fernapi/fern-typescript-sdk`, which is what made the cutover check fire. Also verified: unmigrated workspace → the migration error; `zip` delivery → generated to the mount with the warning; below-cutover generator in a workspace that *has* `sdk-config.yml` → Fern's `GeneratorConfig`, untouched.


ATechAdventurer:
Added `USE_FERN_RC` in `2d6750e`.

`USE_FERN_RC=true` makes an on-prem adapter invocation run `fernenterprise/fern-<language>-sdk:rc` instead of the version the workspace asks for. Postman publishes one moving `rc` tag per language pointing at that language's newest pre-release, so this needs no version alongside it.

An environment variable rather than a `generators.yml` key because it is a property of who is running, not of the workspace: the same configuration then serves a customer on a release version and us on a pre-release, with no file change to remember to undo.

Deliberately narrow:

- **Only invocations that already resolve to the adapter.** A generator below the cutover is Fern's own, and Fern publishes no `rc` tag, so substituting one there would run an image that does not exist.
- **A pinned digest still wins.** Pinning an exact artifact is deliberate, and a switch meaning "something newer" should not quietly un-pin it.
- **The log stays truthful.** `ContainerExecutionEnvironment` logs the reference it is handed, so a run under this switch reports the image it actually ran.

10 cases in `resolveGeneratorImage.test.ts` cover it, including that `fern-typescript-node-sdk` resolves to `fernenterprise/fern-typescript-sdk:rc` — both TypeScript generator names map to one language, so the reference has to come from the language rather than the name. 236 tests pass in the package.

Worth knowing: `kotlin` and `cli` are unreachable through this switch, because Fern publishes no generator of those names for the adapter to be invoked as. They are published under `fernenterprise` and can only be run by naming the image directly.

packages/cli/configuration-loader/src/generators-yml/convertGeneratorsConfiguration.ts
```diff
@@ -672,11 +703,15 @@ function getGeneratorNameAndImage(
...
+        //
+        // A digest is split off before normalization: it belongs to the image reference, not to the
+        // generator identity, and IR version resolution must still see a clean name.
+        const { name: imageNameWithoutDigest, digest } = splitImageDigest(generator.image.name, context);
+        const correctedImageName = correctIncorrectDockerOrgWithWarning(imageNameWithoutDigest, context);
         const normalizedImageName = addDefaultDockerOrgIfNotPresent(correctedImageName);
         return {
             normalizedName: normalizedImageName,
-            containerImage: `${generator.image.registry}/${correctedImageName}`
+            containerImage: `${generator.image.registry}/${correctedImageName}${digest != null ? `@${digest}` : ""}`
```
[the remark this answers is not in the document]
ATechAdventurer: Fixed in 83596c5d1c2 — correct diagnosis, thank you.

`splitImageDigest` now runs on both branches. The default path sets `containerImage` to `${normalizedName}@${digest}` when a digest is present and leaves it `undefined` otherwise, so `normalizedName` stays clean for the cutover maps and IR version resolution while the digest still reaches the runtime.

Worth noting the failure was the first half of what you describe, not an invalid reference: `resolveGeneratorImage` short-circuits on `@sha256:` and returned the digest untouched, so the right image ran. What broke is exactly your point about exact map matching — the polluted name missed `ONPREM_ADAPTER`, so the pinned adapter image was handed Fern's `GeneratorConfig` and no raw-spec mount, then failed `CONFIG_INVALID` inside the container.

Four tests in `convertGeneratorsConfiguration.test.ts` cover it through the public API rather than the helper: digest split on the default path, undigested invocation still deriving its image from the name, the custom-registry path unchanged, and a malformed digest rejected on the default path too.

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
```diff
@@ -51,12 +51,177 @@ export const TYPE_RELOCATIONS_FILENAME = ".fern-type-relocations.json";
...
+    const rightParts = right.split(".");
+    for (let index = 0; index < Math.max(leftParts.length, rightParts.length); index++) {
+        const l = Number.parseInt(leftParts[index] ?? "0", 10) || 0;
+        const r = Number.parseInt(rightParts[index] ?? "0", 10) || 0;
+        if (l !== r) {
+            return l < r ? -1 : 1;
+        }
+    }
+    return 0;
+}
```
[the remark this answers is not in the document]
ATechAdventurer: Taken, fixed in 83596c5d1c2.

`compareVersions` is gone; `isOnPremAdapter` now uses `semver.coerce` + `semver.gte`. `coerce` also gives the prerelease behavior the old `split("-")` was hand-rolling — `4.0.0-rc1` coerces to `4.0.0`.

Your point about `|| 0` was the more serious half, and it turned out to be worse than a typo swallow: it also covered `latest`. See the reply on the prerelease thread — unparseable versions are now rejected rather than silently classified as `0.0.0`.

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
```diff
@@ -51,12 +51,177 @@ export const TYPE_RELOCATIONS_FILENAME = ".fern-type-relocations.json";
...
+ * cutover — an rc of the adapter is still the adapter.
+ */
+export function isOnPremAdapter(generatorName: string, version: string): boolean {
+    const cutover = ONPREM_ADAPTER_CUTOVER.get(generatorName);
+    if (cutover == null) {
+        return false;
+    }
+    const release = version.split("-")[0] ?? version;
+    return compareVersions(release, cutover) >= 0;
+}
```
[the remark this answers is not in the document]
ATechAdventurer: Taken, fixed in 83596c5d1c2 — `ONPREM_ADAPTER_CUTOVER` and `ONPREM_ADAPTER_LANGUAGE` are now one `ONPREM_ADAPTER` map of `name -> { language, cutover }`, so the invariant is structural. `onPremAdapterLanguage` reads `.language` off the same entry, and the redundant `language != null` guard in `resolveGeneratorImage` is now genuinely just a lookup.

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
```diff
@@ -51,12 +51,177 @@ export const TYPE_RELOCATIONS_FILENAME = ".fern-type-relocations.json";
...
+ * A prerelease is treated as its release version, so `4.0.0-rc1` is on the adapter side of the
+ * cutover — an rc of the adapter is still the adapter.
+ */
+export function isOnPremAdapter(generatorName: string, version: string): boolean {
+    const cutover = ONPREM_ADAPTER_CUTOVER.get(generatorName);
+    if (cutover == null) {
+        return false;
+    }
+    const release = version.split("-")[0] ?? version;
+    return compareVersions(release, cutover) >= 0;
```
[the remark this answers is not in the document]
ATechAdventurer: Confirmed and fixed in 83596c5d1c2. I probed it before changing anything:

```
isOnPremAdapter("fernapi/fern-python-sdk", "latest") -> false
resolveGeneratorImage(...)                           -> fernapi/fern-python-sdk:latest
```

So routing and the image could disagree exactly as you describe. I had previously reasoned about this as a safe default, which was wrong — it is safe only while `latest` never points at an adapter release, and that is an assumption about someone else's publish process.

Rather than resolve the tag to a concrete version, `isOnPremAdapter` now throws `UnresolvableGeneratorVersionError` when a name that has a cutover carries a version `semver.coerce` cannot parse. Answering `false` is itself a routing decision — it hands over a Fern `GeneratorConfig` — and resolving the tag would mean a registry round-trip on a path that is meant to work air-gapped. The error names the generator and the tag and asks for a concrete version.

Names with no cutover are untouched: `isOnPremAdapter("fernapi/fern-postman", "latest")` is still `false`, since nothing about them is routed on version. Partial versions still work, since they are unambiguous once coerced — `6` and `v6.1` are on the adapter side, `5` is not.

packages/cli/generation/local-generation/local-workspace-runner/src/postman/buildSdkConfigIrFromSdkConfig.ts
```diff
@@ -0,0 +1,185 @@
...
+            sdkName: target.sdkName ?? sdkConfig.sdkName,
+            sdkVersion: target.sdkVersion ?? sdkConfig.sdkVersion,
+            organization,
+            ...(target.generatorVersion != null ? { generatorVersion: target.generatorVersion } : {}),
+            ...(sdkConfig.apiVersion != null ? { apiVersion: sdkConfig.apiVersion } : {})
+        },
+        api: sdkConfig.api,
+        client: { ...sdkConfig.client, ...target.client },
+        package: { ...sdkConfig.package, ...target.package },
+        docs: target.docs ?? sdkConfig.docs,
```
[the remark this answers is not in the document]
ATechAdventurer: Good catch, applied in 83596c5d1c2 — `docs: { ...sdkConfig.docs, ...target.docs }`.

This was straightforwardly inconsistent with the two lines directly above it, which already spread-merge `client` and `package`. A target setting any single `docs` key dropped every root-level one.

Covered by a test that fails against the old `??` and passes against the merge: root sets `includeApiReference` and `referenceBaseUrl`, the TypeScript target overrides only `includeApiReference`, and both survive.

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
```diff
@@ -51,12 +51,177 @@ export const TYPE_RELOCATIONS_FILENAME = ".fern-type-relocations.json";
...
+ * Opts a local generation run into a specific container network mode, `none` being the useful value.
+ *
+ * An environment variable rather than a generators.yml key so the switch exists without a schema
+ * change; a first-class config field or CLI flag is the natural follow-up if that is preferred.
+ */
+export const GENERATOR_NETWORK_ENV_VAR = "FERN_GENERATOR_NETWORK";
+
+export function getConfiguredGeneratorNetwork(env: NodeJS.ProcessEnv = process.env): string | undefined {
+    const value = env[GENERATOR_NETWORK_ENV_VAR]?.trim();
+    return value !== undefined && value !== "" ? value : undefined;
```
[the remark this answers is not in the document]
ATechAdventurer: Not changing this one, and I think the framing overstates it.

`FERN_GENERATOR_NETWORK` is read from the environment of whoever is already invoking the CLI, and the CLI already shells out to `docker run` on their behalf. Anyone who can set that variable can run the container directly with any flags they like, so an allowlist here moves nothing across a trust boundary — it just makes the supported value harder to discover.

The variable exists to make a run *more* isolated than the default, not less. `none` is the supported value and the only one the air-gapped story depends on; the passthrough is deliberate so the switch did not need a schema change to exist. I have said so explicitly in the PR description rather than encoding an allowlist, and a first-class `generators.yml` field is the natural follow-up if this should become constrained configuration rather than an operator escape hatch.

packages/cli/configuration-loader/src/generators-yml/convertGeneratorsConfiguration.ts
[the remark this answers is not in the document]
ATechAdventurer: Fixed in 83596c5d1c2 — the missing split on the default path is real, and this is now applied on both branches.

One correction to the mechanism, since it matters if anyone acts on this later. `resolveGeneratorImage` does not append a version to a digest-bearing reference: it short-circuits on `@sha256:` and returns it unchanged, precisely so `repo@sha256:...:1.2.3` cannot be constructed. I probed the old code to be sure:

```
resolveGeneratorImage({containerImage: undefined,
                       name: "fernapi/fern-python-sdk@sha256:aaa…",
                       version: "6.0.0"})
  -> fernapi/fern-python-sdk@sha256:aaa…   (not "…:6.0.0")
isOnPremAdapter(...)  -> false
```

So the correct image ran. The actual bug is the other half of your comment — the polluted name missed the exact-match cutover lookups, so a pinned adapter got Fern's `GeneratorConfig` and no raw-spec mount. Same fix either way.

packages/cli/generation/local-generation/local-workspace-runner/src/postman/buildSdkConfigIrFromSdkConfig.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    const { warnings, unsupportedFields } = describeImposedOutput({ sdkConfig, target });
+
+    const candidate: SdkConfigIrV1Input = {
+        schemaVersion: SDK_CONFIG_IR_V1_SCHEMA_VERSION,
+        source: {
+            specs: specs.specs,
+            ...(sdkConfig.source.apiImportSettings != null
+                ? { apiImportSettings: sdkConfig.source.apiImportSettings }
+                : {})
```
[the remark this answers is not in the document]
ATechAdventurer: The observation is accurate — `source.specs` is built entirely from the Fern workspace manifest in `collectOnPremSourceSpecs`, and only `source.apiImportSettings` is read off the loaded document. I am keeping the behavior, but the PR was wrong to leave it implicit, so I have written it into the description.

Source selection deliberately stays with the Fern workspace on the local route. The container is handed *pre-processed* specs — bundled with `$ref`s resolved, overrides merged, overlays applied — mounted at container paths, and `specUrl` points at those mounts. `sdk-config.yml`'s `source.specs` describes sources for the hosted route, where the adapter fetches them itself. Honoring them locally would mean resolving and bundling paths that the Fern workspace does not know about, on a path that has to work air-gapped.

So the accurate statement is narrower than "the SDK can target the wrong API": the specs that get bundled and mounted are always the workspace's, and the adapter always generates from what is mounted. What is genuinely true is that a post-migration edit to `source.specs` is **inert** rather than wrong — it changes nothing, silently.

The audience half is the sharper point and I will take it as a follow-up. Group audiences from `generators.yml` filter the manifest while `api.audiences` in the document does not, so two audience selections can coexist with only one taking effect.

I considered warning on divergence here and decided against it in this PR: comparing a document's relative `source.specs` paths against container mount paths is fuzzy enough that a false warning on every ordinary run is the likely outcome. The non-guessy version is for `fern sdk migrate` to own source selection end to end, which is a larger change than this PR should carry.

ATechAdventurer:
Review round 1 addressed in \`83596c5d1c2\`, with a reply on each thread.

Three real bugs, all with tests that fail against the old code:

| Finding | Status |
|---|---|
| Digest not split on the default \`generators.yml\` path | Fixed — split on both branches, 4 tests through the public API |
| \`latest\` / unparseable versions silently routed to Fern | Fixed — \`semver.coerce\`/\`gte\`, moving tags now rejected |
| Target \`docs\` replaced root instead of merging | Fixed — matches \`client\`/\`package\` |
| Parallel cutover/language maps could drift | Fixed — one \`ONPREM_ADAPTER\` map |
| Prerelease at a cutover major | Documented as an invariant between two release processes |
| \`sdk-config.yml\` source selection inert locally | Behavior kept, now stated in the description; audience divergence noted as follow-up |
| \`FERN_GENERATOR_NETWORK\` accepts any mode | Declined, reasoning in-thread and in the description |

One correction for the record: both bots reported the digest bug, and the mechanism in the Devin version is wrong — \`resolveGeneratorImage\` does not build \`repo@sha256:...:1.2.3\`, it short-circuits on \`@sha256:\`. The real failure was the cutover lookup missing on the polluted name. Same fix.

264 tests pass in \`local-workspace-runner\`, 112 in \`configuration-loader\`'s \`generators-yml\` suite; \`turbo run compile\` clean across the dependency graph; Biome clean.

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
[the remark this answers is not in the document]
ATechAdventurer: Correcting my previous reply — rejecting moving tags was wrong, and `test-ete` caught it in `24c70e21b4e`.

`fern sdk generate --target go --local` defaults to `version: latest` for `fernapi/fern-go-sdk`. So `latest` is not an edge case on this path, it is *the* path, and erroring on it broke a first-party command. You said as much in your comment — "a supported and widely used generator version value" — and I underweighted it. My unit tests never caught it because they only ever asked about concrete versions.

Where it landed: a version `semver.coerce` cannot read means **Fern's own generator**, which is the pre-existing behavior, but now as a stated decision rather than a `parseInt` accident. The reasoning is that selecting the adapter is an explicit act — the whole design is "keep your generators.yml entry and change only the version" — so a workspace that names no version has not opted in.

That is safe only while the adapter is not published into `fernapi` under a moving tag. Which is not independently true: it collapses into the namespace question already open on this PR. If the adapter does ship into `fernapi`, `latest` moves under us and a version comparison cannot save it — that needs a registry check. I have written that where the decision is made rather than leaving it implicit.

The half of your comment that did survive is the comparator: `semver.coerce`/`gte` replaced the hand-rolled parser, so `6` and `v6.1` now read correctly instead of going through `|| 0`.

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
```diff
@@ -51,12 +52,167 @@ export const TYPE_RELOCATIONS_FILENAME = ".fern-type-relocations.json";
...
+ * generator of the same name.
+ *
+ * A prerelease sorts below its release under semver, so the cutover is compared against the release
+ * version: `4.0.0-rc1` is on the adapter side, because an rc of the adapter is still the adapter.
+ * This holds only while Fern publishes no prerelease at or above a cutover major — see the note on
+ * {@link ONPREM_ADAPTER}.
+ *
+ * A version semver cannot read at all — `latest` above all, which is what `fern sdk generate`
+ * defaults to — is Fern's own generator. Selecting the adapter is an explicit act: the whole cutover
+ * design is "keep your generators.yml entry and change only the version", so a workspace that names
```
zamn: This comment reads a bit confusing since I think this method is basically saying whether or not we should be using the new sdk config file?

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
```diff
@@ -51,12 +52,167 @@ export const TYPE_RELOCATIONS_FILENAME = ".fern-type-relocations.json";
...
+ */
+export const USE_FERN_RC_ENV_VAR = "USE_FERN_RC";
+
+export function usesOnPremAdapterPrerelease(env: NodeJS.ProcessEnv = process.env): boolean {
+    const value = env[USE_FERN_RC_ENV_VAR]?.trim().toLowerCase();
+    return value === "true" || value === "1";
+}
+
+/** Namespace Postman publishes the on-prem adapter under. */
+const ONPREM_ADAPTER_NAMESPACE = "fernenterprise";
```
zamn: might be worth making this a env var we can override w/ this as default

packages/cli/generation/local-generation/local-workspace-runner/src/runGenerator.ts
```diff
@@ -275,23 +279,46 @@ export async function writeFilesToDiskAndRunGenerator({
...
+            outputPath: paths.outputDirectory,
+            rawSpecsManifest
+        });
+        if (!built.success) {
+            throw new CliError({ message: built.message, code: CliError.Code.ConfigError });
+        }
+        for (const warning of built.warnings) {
+            context.logger.warn(warning);
+        }
+        await writeFile(absolutePathToWriteConfigJson, JSON.stringify(built.sdkConfigIr, undefined, 4));
```
zamn, line 328: I'm a bit confused by this. We are migrating their fern config at runtime? This is something we wanted to have done manually because it is hostile to the user experience. Also can we re-use the migrate command code that already exists?

Ideally we would just error, tell the user to migrate, give them the command (this should all exist in the code already). Then we just load that config and carry on once they have migrated.

packages/cli/generation/local-generation/local-workspace-runner/src/runGenerator.ts
```diff
@@ -275,23 +279,46 @@ export async function writeFilesToDiskAndRunGenerator({
...
+    if (generatorWantsSdkConfigIr(generatorInvocation.name, generatorInvocation.version)) {
+        const built = await resolveSdkConfigIr({
+            generatorInvocation,
+            absolutePathToFernConfig,
+            organization,
+            outputPath: paths.outputDirectory,
+            rawSpecsManifest
+        });
+        if (!built.success) {
+            throw new CliError({ message: built.message, code: CliError.Code.ConfigError });
```
zamn, line 323: ok I guess since we are flagging on here we actually are going to error if migration is required? Still confused why we are writing a file below?

packages/cli/configuration-loader/src/generators-yml/convertGeneratorsConfiguration.ts
```diff
@@ -660,6 +660,37 @@ async function convertGroup({
     };
 }
 
+/**
+ * Separates an optional `@sha256:...` digest from an image name.
+ *
+ * Pinning by digest is how an immutable deployment is expressed: a tag can be repointed, a digest
```
zamn, line 666: I am a bit fuzzy on this digest stuff. Yes it is good security practice but we can easily control whether or not we allow image tags to be overridden. That is a dockerhub setting we can just enable. It is going to be very hard to correlate which image digest correlates to what core release version, ie "what generator versions live on this image?". How do we plan on debugging that question?

zamn (commented):
There is a lot of redundancies here I think that we can simplify. I am not sure where the digest idea came from. It isnt necessarily bad but I think its overkill for our use case.

packages/cli/generation/local-generation/local-workspace-runner/src/runGenerator.ts
ATechAdventurer: No migration happens at runtime — and the fact that the code reads that way is a fair hit, so I have changed it in `d7290be6c76`.

What it actually does is what you describe as the ideal: if `sdk-config.yml` is absent, `resolveSdkConfigIr` returns a failure and the CLI errors with the generator named and the command to run.

```
fern-python-sdk@6.0.0 is configured by sdk-config.yml, which this workspace does not have.
Migrate your generator configuration:

    fern sdk migrate

That writes sdk-config.yml and archives your generators.yml as generators.archived.yml.
Your other generators keep running from generators.yml either way.
```

It never reads `generators.yml` and translates it. An earlier revision of this branch did have a `generators.yml` → IR translator, and it was deliberately removed for exactly the reason you give — it would produce an SDK shaped by a configuration the customer never wrote. There is a note about that in the PR description.

On reusing the migrate command: the error points at `fern sdk migrate` and that command is unchanged, so the migration logic has one implementation. This path only ever *loads* the file that command produces.

See the reply on your follow-up for the file write, which is the part that genuinely looked wrong.

packages/cli/generation/local-generation/local-workspace-runner/src/runGenerator.ts
ATechAdventurer: Right on both counts, and the file write was genuinely confusing. Fixed in `d7290be6c76`.

The file is `config.json` in the per-run temp dir — `/tmp/fern-xxxx/config.json` — which is the file every generator run writes and mounts as the container's sole argument. It is not in the user's workspace and nothing about their configuration is rewritten.

What made it look wrong is that it was written **twice**: line 254 wrote Fern's `GeneratorConfig`, then line 318 overwrote the same path with SDK Config IR for adapter invocations. Reading top to bottom, the second write looks like it is clobbering the user's config.

It is now decided once, up front:

```ts
const wantsSdkConfigIr = generatorWantsSdkConfigIr(generatorInvocation.name, generatorInvocation.version);
if (!wantsSdkConfigIr) {
    await writeFile(absolutePathToWriteConfigJson, JSON.stringify(await GeneratorConfig.jsonOrThrow(config), undefined, 4));
}
```

so the file is written exactly once whichever generator is running, and the comment at the SDK Config IR branch now says outright that it builds from the existing `sdk-config.yml` and does not migrate.

And yes — the error fires before any of this. An unmigrated workspace never reaches the write.

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
ATechAdventurer: Fair — the doc led with version semantics and buried what the function is for. Rewritten in `d7290be6c76` to open with the decision it drives:

> Whether this invocation is Postman's on-prem adapter rather than Fern's own generator of the same name — which is what decides that the container is configured from `sdk-config.yml` and handed SDK Config IR, instead of from `generators.yml` and handed a Fern `GeneratorConfig`.
>
> `false` for every generator Fern ships today.

Your reading is right, with one nuance worth keeping in view: it is not only about which config file, it also decides whether raw specs get mounted, since the adapter generates from the spec rather than from the Fern IR. Both hang off this one predicate, which is why it is phrased as "is this the adapter" rather than "use the new config".

The version semantics are still there, but demoted to two bullets below the summary, each labelled as a deliberate decision rather than a parsing detail.

packages/cli/generation/local-generation/local-workspace-runner/src/constants.ts
ATechAdventurer: Good call, added in `d7290be6c76` as `FERN_RC_NAMESPACE`, defaulting to `fernenterprise`:

```
USE_FERN_RC=true FERN_RC_NAMESPACE=acme-staging  ->  acme-staging/fern-python-sdk:rc
USE_FERN_RC=true                                 ->  fernenterprise/fern-python-sdk:rc
```

It only applies when `USE_FERN_RC` is on, so it cannot redirect a customer's release-version run, and a blank or whitespace value falls back to the default rather than producing `/fern-python-sdk:rc`. Three tests cover those cases.

Same reasoning as `USE_FERN_RC` itself for making it an env var: it is a property of who is running, not of the workspace, so it does not belong in `generators.yml`.

packages/cli/configuration-loader/src/generators-yml/convertGeneratorsConfiguration.ts
ATechAdventurer: The debuggability objection is the right one, and I have made the log answer it in `d7290be6c76` rather than argue the point.

A digest-pinned run previously logged only the reference, which identifies the artifact but says nothing about which release it is. It now carries the declared version too:

```
Executing generator fernapi/fern-python-sdk using container image:
  ghcr.io/acme/fern-python-sdk@sha256:0127e13e… (generators.yml version 6.0.0)
```

The suffix appears only when the reference does not already end in that version, so an ordinary tag-based run logs exactly what it did before.

Two things that make this less load-bearing than it might look:

**The version never leaves `generators.yml`.** A digest pin is `name: ...@sha256:...` *plus* `version: 6.0.0`, and the version is still what drives IR version resolution and the adapter cutover check — the digest is carried only on the image reference. So "what generator version is this" is answered by the file, and now by the log line as well.

**This is opt-in and additive.** Nothing requires a digest; tag-based references behave exactly as before. If Docker Hub's immutable-tags setting covers the threat for our own images, that is a perfectly good answer for those images — but it does not cover a customer mirroring into their own registry, which is the air-gapped case this branch is aimed at. The cost of supporting it is `splitImageDigest` plus a short-circuit in `resolveGeneratorImage`.

If you would still rather not carry it, it is separable from the rest of this PR — it came in from #17509 and could go back out without touching the `sdk-config.yml` work. Happy to drop it if you think the correlation problem outweighs the mirroring case.

ATechAdventurer:
@zamn thanks — five good ones, four turned into changes in `d7290be6c76`.

**The important correction: nothing is migrated at runtime.** That was the right thing to be alarmed about, and the code was reading that way for a real reason. `config.json` in the per-run temp dir was being written twice — Fern's `GeneratorConfig` first, then overwritten with SDK Config IR for adapter invocations — which top-to-bottom looks like the CLI rewriting the user's configuration. It never was; that file is the container's argument, not anything in the workspace. It is now decided once up front and written exactly once.

The behavior you asked for is what was already there: an unmigrated workspace **errors** with the generator named and `fern sdk migrate` to run. An earlier revision of this branch did have a `generators.yml` → IR translator and it was deliberately deleted, for the reason you give.

| Comment | Outcome |
|---|---|
| Runtime migration? | No — errors and points at `fern sdk migrate`; write path clarified |
| Why write a file below? | Double write removed; written once either way |
| `isOnPremAdapter` doc confusing | Rewritten to lead with the decision it drives |
| Namespace as env var | Added `FERN_RC_NAMESPACE`, defaults to `fernenterprise` |
| Digest ↔ generator version correlation | Log now reports the `generators.yml` version on digest-pinned runs |

On the digest question more broadly — immutable tags on Docker Hub do cover our own images, but not a customer mirroring into their own registry, which is the air-gapped case here. That said, digest pinning arrived from #17509 and is separable from the `sdk-config.yml` work; if you would rather not carry it, say so and I will lift it out.

265 tests pass in `local-workspace-runner`; `tsc` and Biome clean.

zamn (approved):

PR #17678, opened 2026-09-08
fix(rust): generate the auto-paging method alongside the single-page one

iamnamananand996:
A paginated endpoint has generated no pager since #9781, which disabled the wiring in `SubClientGenerator`. It was disabled for a good reason: turning it back on replaced the endpoint's own method — `list_users` stopped returning `ListUsersResponse` and started returning `AsyncPaginator<serde_json::Value>`, losing the typed response for everyone who only wanted one page.

**Emit both.** The single-page method is generated for every endpoint exactly as today, and a paginated endpoint additionally gets `<endpoint>_paginated` returning the paginator. Nothing that compiles today changes shape — the four affected seed fixtures show additions only, and no method was removed.

## Two bugs that kept the pager from compiling

Turning the wiring on surfaced why it could not simply be re-enabled — the generated pager did not build:

- ``new RegExp(`\\bname\\b`)`` was written with a doubled escape, so the regex source was a literal backslash followed by `bname` and the rewrite meant to point the page loader at its captured copies **never matched a single name**. The loader kept referring to the borrowed parameter, and every reference parameter escaped the method body (`E0521`, 7 errors on `seed/rust-sdk/pagination`).
- The capture itself used `.clone()`, which on a `&str` clones the *reference*, so taking a copy did not end the borrow either. String references now take `to_string()`.

Two smaller cleanups so the generated crates build warning-free: the loader for custom/uri/path pagination takes `_cursor_value` (it has no cursor to thread through yet), and an endpoint whose every query parameter *is* a pagination parameter has nothing left for the base query string, so its unread `request` parameter is underscored — it stays in the signature, because the paginated and single-page methods should be called the same way.

## Verification

| check | result |
|---|---|
| `seed test --generator rust-sdk` | 148/148 |
| seed churn | 4 fixtures, 6 files, additions only |
| `cargo build --all-features` on `pagination` | clean (1 pre-existing `unused import` in an unmodified `mod.rs`) |
| `cargo test` on `pagination` | 98 passed |
| `pagination-custom` / `pagination-uri-path` | build and test clean, 0 warnings |
| `exhaustive` | builds; its 3 wire-test errors are pre-existing and unrelated (`Animal::Dog` union) |
| generator unit tests | 42 passed |

## Not in this PR

The README pagination snippet stays disabled. It calls the endpoint with no arguments, so it was already emitting code that does not compile — a separate fix.

The pager still yields `serde_json::Value` rather than the endpoint's item type, starts from `None` instead of a caller-supplied cursor, and advances an offset pager by 1 rather than by the page size. Those are behaviour fixes; this PR is the wiring and the borrow.

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17678" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/generators/rust/sdk/src/SdkCustomConfig.ts
+++ b/generators/rust/sdk/src/SdkCustomConfig.ts
@@ -9,7 +9,14 @@ export const SdkCustomConfigSchema = BaseRustCustomConfigSchema.extend({
      * - Model/type generation is skipped (types come from a co-generated types crate)
      * - The `RequestExecutor` trait and `HttpClient::with_executor()` are the primary API surface
      */
-    cliEmbedded: z.boolean().optional().default(false)
+    cliEmbedded: z.boolean().optional().default(false),
+    /**
+     * How the `step` attribute of `x-fern-pagination` is read on an offset-paginated endpoint.
+     * `"item-index"` (the default, matching the TypeScript and Python generators) treats the offset
+     * as an item index and advances it by the number of items the page returned. `"page-index"`
+     * treats it as a page number and advances it by one.
+     */
+    offsetSemantics: z.enum(["item-index", "page-index"]).optional().default("item-index")
 });
 
 export type SdkCustomConfigSchema = z.infer<typeof SdkCustomConfigSchema>;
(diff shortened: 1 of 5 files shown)


iamnamananand996:
Superseded by #17680, which is the same two commits pushed to a branch on this repo rather than my fork. Fork PRs do not receive repository secrets, so `FERN_TOKEN` was absent and four of the five `test-ete` failures were `Authentication required. Please run 'fern login'` from `fern generate` — not the change. (The fifth, the `validate > docs` snapshot, is in `packages/cli`, which this diff does not touch at all.) Reopening from an upstream branch so CI is meaningful.

PR #17684, opened 2026-09-08
fix(rust): send basic auth credentials on the client-wide auth path

iamnamananand996:
> **This is a shipped correctness bug, not only a conformance row.** An API whose only security scheme is basic auth generates a Rust client that sends **no credential at all**.

## What happens today

`apply_auth_headers` has two paths:

- **Per-endpoint auth routing** resolves basic auth correctly, and always has.
- **The client-wide (ALL/ANY) path** applies an API key and a bearer token, and has no basic branch.

So on the default path, `username` and `password` sit unread on `ClientConfig`, and the only `Authorization` header in the generated client is:

```rust
if let Some(token) = token {
    let auth_value = format!("Bearer {}", token);
    headers.insert("Authorization", ...);
}
```

gated on `config.token` — which a basic-auth API never sets.

`seed/rust-sdk/basic-auth`, whose **only** scheme is basic, shipped in exactly that state. `grep -c Basic src/core/http_client.rs` returned **0**.

## The fix

The flat path now applies `Authorization: Basic <base64(user:pass)>`, using the same expression the per-endpoint path already used. It is placed **before** the bearer branch, so a client configured with both still ends up with the bearer token — that is the existing precedence, not a new one.

## Two dependency gates had the same shape

Both were written as "endpoint-security **and** has basic scheme", and both had to move:

| gate | file |
|---|---|
| the `base64` crate | `AbstractRustGeneratorContext.ts:151` |
| `use base64::Engine;` in `http_client` | `RustProject.ts:194` |

Without the second, the generator compiles fine and emits an SDK that **does not build** (`no method named encode found for GeneralPurpose` — the `Engine` trait isn't in scope). That is how it was caught: by generating and building, not by a generator test.

## Verification

| check | result |
|---|---|
| `seed test --generator rust-sdk` | 148/148 |
| seed churn | 4 fixtures, all basic-auth ones |
| `cargo build --all-features` | clean on all four |
| `cargo test` | 86 passed on each |

`basic-auth`, `any-auth`, `basic-auth-environment-variables` and `basic-auth-pw-omitted` — `pw-omitted` included deliberately, since the `(Some, Some)` guard means an omitted password still sends nothing, matching the per-endpoint path's existing behaviour rather than inventing a new one.

Closes two rows on the Rust behavioural conformance harness, both reporting `authorization_was_absent`.

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17684" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/generators/rust/base/src/context/AbstractRustGeneratorContext.ts
+++ b/generators/rust/base/src/context/AbstractRustGeneratorContext.ts
@@ -146,9 +146,10 @@ export abstract class AbstractRustGeneratorContext<
             this.dependencyManager.add("uuid", { version: "1.0", features: ["serde"] });
         }
 
-        // Conditionally include base64 when base64 types are used, or when per-endpoint
-        // auth routing needs it to encode basic auth credentials.
-        if (this.usesBase64() || (this.isEndpointSecurity() && this.hasBasicAuthScheme())) {
+        // Conditionally include base64 when base64 types are used, or when a basic auth scheme
+        // has to be encoded. Both auth paths need it: per-endpoint routing and the flat
+        // client-wide application.
+        if (this.usesBase64() || this.hasBasicAuthScheme()) {
             this.dependencyManager.add("base64", "0.22");
         }
 
--- a/generators/rust/base/src/project/RustProject.ts
+++ b/generators/rust/base/src/project/RustProject.ts
@@ -189,9 +189,9 @@ export class RustProject extends AbstractProject<AbstractRustGeneratorContext<Ba
             content = content.replace(/\{\{SERDE_ERROR_IMPORT\}\}/g, "");
         }
 
-        // Conditionally include base64 import in http_client (base64 method, or basic
-        // auth encoding in per-endpoint auth routing).
-        if (this.context.usesBase64() || (this.context.isEndpointSecurity() && this.context.hasBasicAuthScheme())) {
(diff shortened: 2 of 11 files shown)


fern-support (approved):

iamnamananand996:
**Correction to this PR's description.** I claimed it closes two rows on the Rust conformance harness, both reporting `authorization_was_absent`. Measured against the released `0.47.5`: **it does not**, and the reason is worth recording.

The conformance `auth` fixture declares `bearerAuth` **first**, then `basicAuth`. `buildAuthSchemes.ts:24` keeps only the first `securityScheme`, so `ir.auth.schemes` contains bearer alone — and this fix is gated on `hasBasicAuthScheme()`, which reads that list. The generated client for that fixture therefore contains **zero** occurrences of `Basic`, even at 0.47.5.

Those two rows are blocked on fern-api/fern#17688, not on this PR.

**The fix itself is correct and still worth having**, which the seed proves: `seed/rust-sdk/basic-auth`, whose *only* scheme is basic auth, went from **0** occurrences of `Basic` to 2 and now sends `Authorization: Basic <base64>`. That was the shipped correctness bug — an API whose only scheme is basic generated a client that sent no credential at all. It is fixed.

What I got wrong was the conformance attribution: for a *multi-scheme* spec, this fix is necessary but not sufficient, because the scheme never survives import.

PR #17759, opened 2026-09-15
feat(cli): make sdk-config authoritative for generation

bears4barrett:
## Description
Makes sdk-config.yml authoritative when generating SDKs with fern generate --sdk-config.

## Changes
- Builds API workspaces directly from SDK Config sources and targets.
- Avoids loading or validating generators.yml when --sdk-config is provided.
- Supports OpenAPI, AsyncAPI, and GraphQL sources.
- Propagates SDK Config package and output settings through sdk-gen-api.
- Rejects --group with --sdk-config, since SDK Config targets are authoritative.
- Updates fern sdk migrate to omit generatorVersion, allowing targets to use the latest supported generator.
- Removes the obsolete logic that matched SDK Config targets against generators.yml groups.

## Backwards compatibility
Generation without --sdk-config continues through the existing generators.yml path unchanged.

Hand-authored generatorVersion values remain supported as exact pins. When omitted, the CLI resolves the latest supported generator during generation.

## Testing
- Added SDK Config workspace construction coverage.
- Added project-loading coverage without generators.yml.
- Added remote request package and output propagation coverage.
- Updated migration unit and end-to-end tests to verify generatorVersion is omitted.

---

<a href="https://app.devin.ai/review/fern-api/fern/pull/17759" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


packages/cli/cli/src/commands/generate/createSdkConfigWorkspace.ts
```diff
@@ -0,0 +1,233 @@
...
+    const response = await fetch(spec.url, { signal: AbortSignal.timeout(30_000) });
+    if (!response.ok) {
+        return context.failAndThrow(
+            `Could not download SDK Config source '${spec.id}' from ${spec.url}: ${response.status} ${response.statusText}`,
+            undefined,
+            { code: CliError.Code.NetworkError }
+        );
+    }
+    const directory = await mkdtemp(path.join(tmpdir(), "fern-sdk-config-source-"));
+    const absolutePath = path.join(directory, `${sanitizeFilename(spec.id)}.json`);
```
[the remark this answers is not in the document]
bears4barrett: Updated remote source downloads to preserve recognized URL extensions. Extensionless URLs use Content-Type, falling back to .yml or .graphql. Added YAML regression coverage.

packages/cli/cli/src/commands/generate/createFernSourceArchiveResolver.ts
```diff
@@ -53,17 +52,7 @@ export function createFernSourceArchiveResolver({
...
-                    const configTarget = sdkConfigV1?.targets.find(
-                        (target) => target.language === request.sdkGenApiRoute?.language
-                    );
-                    validateSdkConfigImportSettings(specs, {
-                        clientPathParameterStyle:
-                            configTarget?.clientPathParameterStyle ?? sdkConfigV1?.clientPathParameterStyle
-                    });
-                }
-                return specs;
+                return await workspace.getAllSpecsForGenerator(request.generatorInvocation.apiOverride?.specs);
```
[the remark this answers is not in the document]
bears4barrett: Restored validateSdkConfigImportSettings for each SDK Config target using its effective path-parameter style. Added coverage confirming validation runs.

packages/cli/cli/src/commands/generate/createSdkConfigWorkspace.ts
```diff
@@ -0,0 +1,233 @@
...
+        name,
+        version,
+        config: {},
+        outputMode: FernFiddle.remoteGen.OutputMode.downloadFiles({}),
+        automation: { generate: true, preview: true, upgrade: true, verify: true },
+        containerImage: undefined,
+        irVersionOverride: undefined,
+        absolutePathToLocalOutput:
+            output?.delivery === "files"
+                ? AbsoluteFilePath.of(path.resolve(configDirectory, output.path ?? `generated/${language}`))
```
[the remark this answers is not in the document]
bears4barrett: Added a named default and documentation explaining that omitted file paths resolve to generated/<language> beside sdk-config.yml. Behavior is unchanged.

packages/cli/cli/src/commands/generate/createSdkConfigWorkspace.ts
```diff
@@ -0,0 +1,270 @@
...
+import { AbsoluteFilePath } from "@fern-api/fs-utils";
+import { bundleRemoteOpenAPI, OSSWorkspace } from "@fern-api/lazy-fern-workspace";
+import { CliError, TaskContext } from "@fern-api/task-context";
+import { FernFiddle } from "@fern-fern/fiddle-sdk";
+import type { SdkConfigV1, SdkConfigV1SourceSpec } from "@postman/sdk-config/sdk-config/v1";
+
+const SDK_CONFIG_GROUP = "sdk-config";
+const DEFAULT_LOCAL_OUTPUT_DIRECTORY = "generated";
+
+const GENERATOR_BY_LANGUAGE: Record<string, string> = {
```
zamn, line 15: i feel like we had defined this elsewhere in the codebase that maybe we could reuse but i could be wrong.

packages/cli/ete-tests/src/tests/sdk-migrate/fixtures/sdk-config.yml
```diff
@@ -12,9 +12,7 @@ targets:
     output:
       delivery: files
       path: ./generated/typescript
-    generatorVersion: 3.63.3
```
zamn, line 15: might be worth having two versions of this file since we definitely should still support version pinning in it

zamn (approved):

packages/cli/cli/src/commands/generate/createSdkConfigWorkspace.ts
bears4barrett: will check

bears4barrett: I checked the existing mappings. The closest is CLI v2’s LANGUAGE_TO_DOCKER_IMAGE, but it has different coverage and canonical choices—most notably no Kotlin/CLI/MCP and the legacy Ruby generator. The sdk-gen compatibility map is generator-to-language and cannot be safely inverted because of aliases. I’m keeping this bridge-specific canonical map here for now.

